PR #15, opened 2022-02-08
merge coreth-0.8.5-rc.2

ceyonur:
Merges Coreth Tag 0.8.5-rc.2

Resolves: #14


ceyonur:
bumped avalanchego to 1.7.5, resolved #17

plugin/evm/vm.go
```diff
@@ -565,6 +626,39 @@ func (vm *VM) CreateStaticHandlers() (map[string]*commonEng.HTTPHandler, error)
  ******************************************************************************
  */
 
+// conflicts returns an error if [inputs] conflicts with any of the atomic inputs contained in [ancestor]
+// or any of its ancestor blocks going back to the last accepted block in its ancestry. If [ancestor] is
+// accepted, then nil will be returned immediately.
+// If the ancestry of [ancestor] cannot be fetched, then [errRejectedParent] may be returned.
+func (vm *VM) conflicts(inputs ids.Set, ancestor *Block) error {
```
patrick-ogrady, line 633: This is not needed (related to atomic UTXOs).

plugin/evm/message/message.go
```diff
@@ -14,7 +16,7 @@ import (
 
 const (
 	// EthMsgSoftCapSize is the ideal size of encoded transaction bytes we send in
-	// any [EthTxs] message. We do not limit inbound messages to
+	// any [Txs] message. We do not limit inbound messages to
 	// this size, however. Max inbound message size is enforced by the codec
 	// (512KB).
 	EthMsgSoftCapSize = common.StorageSize(64 * units.KiB)
```
patrick-ogrady, line 22: nit: `TxsMsgSoftCapSize`

plugin/evm/gossiper_eth_gossiping_test.go
```diff
@@ -0,0 +1,375 @@
+// (c) 2019-2021, Ava Labs, Inc. All rights reserved.
```
patrick-ogrady, line 1: nit: change file name to tx_gossiping_test

plugin/evm/block_builder.go
```diff
@@ -80,15 +80,15 @@ type blockBuilder struct {
 	// isSC is a boolean indicating if SubnetEVM is activated. This prevents us from
```
patrick-ogrady, line 80: nit: `isSE`

consensus/dummy/test_consensus.go
```diff
@@ -0,0 +1,53 @@
+// (c) 2019-2020, Ava Labs, Inc. All rights reserved.
```
aaronbuchwald, line 1: We shouldn't need to migrate this in. We just pass in non-nil callbacks to the tests in `core/test_blockchain.go` to catch any errors that could come from mishandling extra state changes. Since those don't exist in subnet-evm, I think we can remove this file and pass in the normal dummy engine with no callbacks.

plugin/evm/block_builder_test.go
```diff
@@ -83,7 +83,7 @@ func TestBlockBuilderStopsTimer(t *testing.T) {
 	if builder.buildStatus != mayBuild {
 		t.Fatalf("expected build status to be %d but got %d", dontBuild, builder.buildStatus)
 	}
-	if !builder.isSC {
+	if !builder.isSE {
```
aaronbuchwald, line 86: should we update the test fatal message to be `isSE` as well?

plugin/evm/block_builder.go
```diff
@@ -52,8 +52,8 @@ type blockBuilder struct {
 	ctx         *snow.Context
 	chainConfig *params.ChainConfig
 
-	chain   *subnetEVM.ETHChain
-	network Network
+	chain    *subnetEVM.ETHChain
+	gossiper Gossiper
```
aaronbuchwald, line 56: `isSE` seems to be a parallel of `isAP4` in coreth, can we get rid of it and the surrounding code considering it has always been active and there's no reason for anyone to launch without it?

plugin/evm/block_builder.go
aaronbuchwald: It seems like there are a few other places as well that aren't in the diff which haven't been updated. A quick search for SC in `plugin/evm/` should find all of them that need to be replaced.

plugin/evm/vm.go
```diff
@@ -343,18 +366,51 @@ func (vm *VM) Initialize(
...
+	if err := vm.chain.RemoveRejectedBlocks(0, lastAcceptedHeight); err != nil {
+		return err
+	}
+	heightBytes := make([]byte, 8)
+	binary.PutUvarint(heightBytes, lastAcceptedHeight)
+	if err := vm.db.Put(pruneRejectedBlocksKey, heightBytes); err != nil {
+		return err
+	}
+	return vm.db.Commit()
+}
```
aaronbuchwald, line 401: This is not needed for `subnet-evm` since it existed after we starting deleting blocks from disk on reject.

plugin/evm/vm.go
```diff
@@ -79,9 +86,10 @@ const (
 
 var (
 	// Set last accepted key to be longer than the keys used to store accepted block IDs.
-	lastAcceptedKey = []byte("last_accepted_key")
-	acceptedPrefix  = []byte("snowman_accepted")
-	ethDBPrefix     = []byte("ethdb")
+	lastAcceptedKey        = []byte("last_accepted_key")
+	acceptedPrefix         = []byte("snowman_accepted")
+	ethDBPrefix            = []byte("ethdb")
+	pruneRejectedBlocksKey = []byte("pruned_rejected_blocks")
```
aaronbuchwald, line 92: We can delete this key along with `pruneChain` since it's not needed for subnet-evm.

plugin/evm/vm.go
```diff
@@ -343,18 +366,51 @@ func (vm *VM) Initialize(
...
-	vm.bootstrapped = false
-	return nil
+func (vm *VM) initGossipHandling() {
+	if vm.chainConfig.SubnetEVMTimestamp != nil {
+		vm.gossiper = vm.newPushGossiper()
+		vm.Network.SetGossipHandler(NewGossipHandler(vm))
+	} else {
+		vm.gossiper = &noopGossiper{}
+		vm.Network.SetGossipHandler(message.NoopMempoolGossipHandler{})
+	}
```
aaronbuchwald, line 376: we shouldn't need this either since gossiping has existed for the full history of subnet-evm. Is it possible to remove `SubnetEVMTimestamp` or does it serve another purpose as well?

plugin/evm/vm_test.go
```diff
@@ -79,19 +79,30 @@ func init() {
 	genesisBalance = new(big.Int).Mul(big.NewInt(testMinGasPrice), big.NewInt(21000*1000))
 }
 
-// BuildGenesisTest returns the genesis bytes for Subnet EVM VM to be used in testing
-func BuildGenesisTestWithAlloc(t *testing.T, genesisJSON string) []byte {
-	ss := CreateStaticService()
-
+func fundGenesis(genesisJSON string) (string, error) {
```
aaronbuchwald, line 82: Since we are calling this everywhere to fund the two addresses, can we instead just modify the balance on the genesis string and use that?

consensus/dummy/test_consensus.go
ceyonur: I added this specifically for https://github.com/ava-labs/subnet-evm/blob/98737f6a870ffb3785b9acb5553eabee95ed787d/core/blockchain_test.go#L256. In this test we need to set the state so that pruner can work. I don't know if there is any other way but seems like this test breaks without setting the state in callbacks.

plugin/evm/vm.go
ceyonur: done

plugin/evm/message/message.go
ceyonur: done

plugin/evm/gossiper_eth_gossiping_test.go
ceyonur: done

plugin/evm/block_builder.go
ceyonur: done

core/test_blockchain.go
```diff
@@ -28,50 +28,50 @@ type ChainTest struct {
...
-	},
-	{
-		"SetPreferenceRewind",
-		TestSetPreferenceRewind,
-	},
-	{
-		"BuildOnVariousStages",
-		TestBuildOnVariousStages,
-	},
+	// {
```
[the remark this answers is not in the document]
ceyonur: I was debugging one of these. nice catch!

plugin/evm/block_builder_test.go
ceyonur: done

plugin/evm/vm.go
ceyonur: done

plugin/evm/vm.go
ceyonur: done

plugin/evm/vm.go
ceyonur: We still treat SubnetEVMTimestamp as a hardfork because of legacy tests. I tried to remove it couple of times but some tests expect a pre-subnetevm genesis/fork.

plugin/evm/vm_test.go
ceyonur: done

PR #16, opened 2022-02-09
add allowfeerecipients in genesis & add coinbase to chain config

ceyonur:
This PR adds a mechanism to redirect fees to addresses instead of burning them by sending to a blackhole address

Adds`allowRedirectFees` bool flag to genesis chain configs.
Adds a configurable `coinbase` address to chain configs. 
Block builders can specify their `coinbase` addresses to collect fees if chain activates `allowRedirectFees` in the genesis.


ceyonur:
resolves #16

PR #21, opened 2022-02-19
[AV-1297] Add framework for stateful precompiles

aaronbuchwald:
This PR adds the framework for stateful precompiles. This framework is borrowed from coreth and enables us to start building more powerful types of precompiles within the subnet-evm.


ceyonur (approved):
lgtm!

Issue #13, opened 2022-02-07
Add option to specify list of "whitelist" deployers in genesis file


patrick-ogrady:
Some subnet creators want to limit the addresses that can deploy contracts to a pre-defined list.

patrick-ogrady:
Shipped with #31

PR #22, opened 2022-02-19
Allow list precompile

aaronbuchwald:
closes: #13 

This PR adds a stateful precompile that is able to dictate what addresses are allowed to deploy contracts to the chain.

Addresses can be assigned two different privilege levels:
- Deployer can deploy new contracts to the chain
- Admin can deploy new contracts to the chain and modify the permissions of the set of admin and deployer addresses


ceyonur (commented):
I think we also need to add `allowList` to gen genesis. Pushed a fix to this PR.

core/vm/allow_list.sol
```diff
@@ -0,0 +1,28 @@
...
+    
+    function setDeployer(address addr) public {
+        _invokeAllowList(addr, 1);
+    }
+
+    function setAdmin(address addr) public {
+        _invokeAllowList(addr, 2);
+    }
+    
+    function _invokeAllowList(address addr, uint256 role) internal {
```
patrick-ogrady, line 23: Do we want to add getters here?

core/vm/evm.go
```diff
@@ -450,6 +452,12 @@ func (evm *EVM) create(caller ContractRef, codeAndHash *codeAndHash, gas uint64,
 	if evm.StateDB.GetNonce(address) != 0 || (contractHash != (common.Hash{}) && contractHash != emptyCodeHash) {
 		return nil, common.Address{}, 0, ErrContractAddressCollision
 	}
+	// Check that [caller] has the right to deploy contracts
+	modifyStatus := getAllowListStatus(evm.StateDB, caller.Address())
```
patrick-ogrady, line 456: can we extend this to the origin as well? Otherwise, you'll need to give factories deployer privilege which seems sub-optimal.

core/vm/contracts_stateful.go
```diff
@@ -7,6 +7,10 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 )
 
+var (
+	AllowListPrecompileAddress = common.HexToAddress("0x0200000000000000000000000000000000000000")
```
patrick-ogrady, line 11: Any reason for 0x02....0? I know coreth goes 0x010...n?

core/vm/allow_list.go
```diff
@@ -0,0 +1,147 @@
...
+
+// SetAllowListStatus sets the permissions of [address] to [status]
+func SetAllowListStatus(stateDB StateDB, address common.Address, status ModifyStatus) error {
+	if !status.Valid() {
+		return fmt.Errorf("invalid status used as input for allow list precompile: %v", status)
+	}
+	// Generate the state key for [address]
+	addressKey := createAddressKey(address)
+	// Update the type of [status] for the call to SetState
+	role := common.Hash(status)
```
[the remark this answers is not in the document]
patrick-ogrady: I don't think this is actually performing another hash, it is just casting it so it works with `SetState`?

core/vm/evm.go
patrick-ogrady: I'm looking around rn

core/vm/evm.go
```diff
@@ -450,6 +452,12 @@ func (evm *EVM) create(caller ContractRef, codeAndHash *codeAndHash, gas uint64,
...
+	if !modifyStatus.HasDeployPrivileges() {
```
patrick-ogrady, line 457: I think we only need to check this if there are any admins? Otherwise, this prevents the use of `subnet-evm` for chains with permissionless deploys.

core/vm/evm.go
patrick-ogrady: Figured out a way 🚀 ...one sec

patrick-ogrady: https://github.com/ava-labs/subnet-evm/pull/23

core/vm/allow_list.go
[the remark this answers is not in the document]
aaronbuchwald: `common.Hash(status)` casts it to a hash and `common.BytesToHash(input)` just sets the bytes value stored in `common.Hash` to the last 32 bytes of `input` (cropping from the left side as necessary).

core/vm/allow_list.go
```diff
@@ -0,0 +1,147 @@
+// (c) 2019-2020, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package vm
```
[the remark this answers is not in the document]
aaronbuchwald: Yup moving all of this into a `precompile` directory.

core/genesis.go
```diff
@@ -87,6 +88,9 @@ type Genesis struct {
 	GasUsed    uint64      `json:"gasUsed"`
 	ParentHash common.Hash `json:"parentHash"`
 	BaseFee    *big.Int    `json:"baseFeePerGas"`
+
+	// Addresses for the original set of admins over the contract allow list precompile
+	ContractAllowListAdmin []common.Address `json:"allowListAdmin"`
```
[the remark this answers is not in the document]
aaronbuchwald: Yup

core/vm/allow_list.sol
aaronbuchwald: That's a good idea, this will require an additional precompile to read the values stored in the allow list address space.

core/vm/contracts_stateful.go
aaronbuchwald: Wanted to make sure that we do not collide with the precompiles in coreth so that we leave the option open to combine them in the future.

core/vm/evm.go
aaronbuchwald: Yup, updating this to be based off of `IsAllowListEnabled`.

ceyonur:
shall we abandon this PR in favor of #25 ?

aaronbuchwald:
Closing in favor of #31

PR #25, opened 2022-02-22
Allow list precompile chain config

aaronbuchwald:
This PR switches to using the chain config to specify when the allow list should be activated and how it should be configured. This also attempts to define a clearly and easily enabling stateful precompiles at a given block timestamp.


precompile/utils.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+// CreateAddressKey converts [address] into a [common.Hash] value to be used as a storage slot key
+func CreateAddressKey(address common.Address) common.Hash {
+	hashBytes := make([]byte, common.HashLength)
+	copy(hashBytes, address[:])
+	return common.BytesToHash(hashBytes)
+}
+
+// isForked returns whether a fork scheduled at block s is active at the given head block.
```
patrick-ogrady, line 19: can you specify if this is time or height?

patrick-ogrady: (from above, I see it is used with timestamp but imagine it could be used with either)

precompile/stateful_precompile_config.go
```diff
@@ -0,0 +1,47 @@
...
+	// 1) 0 indicates that the precompile should be enabled from genesis.
+	// 2) n indicates that the precompile should be enabled in the first block with timestamp >= [n].
+	// 3) nil indicates that the precompile is never enabled.
+	Timestamp() *big.Int
+	// Configure is called on the first block where the stateful precompile should be enabled.
+	// This allows the stateful precompile to configure its own state via [StateDB] as necessary.
+	// This function must be deterministic since it will impact the EVM state. If a change to the
+	// config causes a change to the state modifications made in Configure, then it cannot be safely
+	// made to the config after the network upgrade has gone into effect.
+
```
patrick-ogrady, line 22: put `//` on this line so it is clear the blocks of text are connected

core/state_processor.go
```diff
@@ -74,13 +75,17 @@ func (p *StateProcessor) Process(block *types.Block, parent *types.Header, state
 		blockNumber = block.Number()
 		allLogs     []*types.Log
 		gp          = new(GasPool).AddGas(block.GasLimit())
+		timestamp   = new(big.Int).SetUint64(header.Time)
 	)
 
+	// Configure the allow list if it goes into effect during the transition from [parent] to [block].
+	precompile.CheckConfigure(new(big.Int).SetUint64(parent.Time), timestamp, p.config.AllowListConfig, statedb)
```
patrick-ogrady, line 82: nit: is there some config array to iterate over here so the addition of more precompiles don't require changing code structure?

patrick-ogrady:
Can you add something to the README about how this works?

params/config.go
```diff
@@ -341,6 +356,17 @@ func (c *ChainConfig) checkCompatible(newcfg *ChainConfig, headHeight *big.Int,
...
 	}
 
+	var allowListTimestamp, newAllowListTimestamp *big.Int
+	if c.AllowListConfig != nil {
+		allowListTimestamp = c.AllowListConfig.Timestamp()
+	}
+	if newcfg.AllowListConfig != nil {
+		newAllowListTimestamp = newcfg.AllowListConfig.Timestamp()
+	}
+	if isForkIncompatible(allowListTimestamp, newAllowListTimestamp, headTimestamp) {
```
patrick-ogrady, line 366: ❤️

patrick-ogrady: I think we should instruct everyone to structure their modifications as hard forks so it is easy to discern changes.

precompile/allow_list.go
```diff
@@ -0,0 +1,165 @@
...
+// precompile
+func PackModifyAllowList(address common.Address, status AllowListRole) ([]byte, error) {
+	// If [allow], set the last byte to 1 to indicate that [address] should be added to the whitelist
+	if !status.Valid() {
+		return nil, fmt.Errorf("unexpected status: %d", status)
+	}
+
+	input := make([]byte, modifyAllowListInputLength)
+	copy(input, address[:])
+	copy(input[20:], status[:])
```
patrick-ogrady, line 96: nit: use common.AddressLength

precompile/allow_list.go
```diff
@@ -0,0 +1,165 @@
...
+	if !status.Valid() {
+		return common.Address{}, common.Hash{}, fmt.Errorf("invalid status used as input for allow list precompile: %v", status)
+	}
+
+	return address, statusHash, nil
+}
+
+// setAllowListStatus sets the permissions of [address] to [status]
+// assumes [status] has already been verified as valid.
+func setAllowListStatus(stateDB StateDB, address common.Address, status common.Hash) error {
```
patrick-ogrady, line 120: nit: why return an error?

precompile/params.go
```diff
@@ -0,0 +1,16 @@
...
+package precompile
+
+import "github.com/ethereum/go-ethereum/common"
+
+// Gas costs for stateful precompiles
+const (
+	ModifyAllowListGasCost = 20_000
+)
+
+// Designated addresses of stateful precompiles
```
patrick-ogrady, line 13: nit: NOTE not to use same addresses for different precompiles

precompile/stateful_precompile_config.go
```diff
@@ -0,0 +1,47 @@
...
+func CheckConfigure(parentTimestamp *big.Int, currentTimestamp *big.Int, config StatefulPrecompileConfig, state StateDB) {
+	// If the stateful precompile is nil, skip configuring it
+	if config == nil {
+		return
+	}
+	forkTimestamp := config.Timestamp()
+	isParentForked := isForked(parentTimestamp, forkTimestamp)
+	isCurrentBlockForked := isForked(currentTimestamp, forkTimestamp)
+	// If the network upgrade goes into effect within this transition, configure the stateful precompile
+	if !isParentForked && isCurrentBlockForked {
```
patrick-ogrady, line 44: Won't this not get triggered in genesis as parentTimestamp == currentTimetsamp?

core/vm/contracts_stateful.go
```diff
@@ -4,34 +4,40 @@
...
+
+// RunStatefulPrecompiledContract confirms runs [precompile] with the specified parameters.
+func RunStatefulPrecompiledContract(precompile precompile.StatefulPrecompiledContract, accessibleState precompile.PrecompileAccessibleState, caller common.Address, addr common.Address, input []byte, suppliedGas uint64, readOnly bool) (ret []byte, remainingGas uint64, err error) {
+	// Confirm that there is sufficient gas to execute [precompile]
+	// Note: we perform this check here to so that we can return ErrOutOfGas if [precompile] will run out of gas.
+	requiredGas := precompile.RequiredGas(input)
+	if requiredGas > suppliedGas {
+		return nil, 0, ErrOutOfGas
+	}
+
```
ceyonur, line 41: should we deduce `requiredGas` here instead of in the `Run`?

params/config.go
```diff
@@ -198,11 +201,20 @@ func (c *ChainConfig) IsIstanbul(num *big.Int) bool {
 	return isForked(c.IstanbulBlock, num)
 }
 
-// IsSubnetEVM returns whether num is either equal to the SubnetEVM fork block timestamp or greater.
+// IsSubnetEVM returns whether [blockTimestamp] is either equal to the SubnetEVM fork block timestamp or greater.
 func (c *ChainConfig) IsSubnetEVM(blockTimestamp *big.Int) bool {
 	return isForked(c.SubnetEVMTimestamp, blockTimestamp)
 }
 
+// IsSubnetEVM returns whether [blockTimestamp] is either equal to the AllowList fork block timestamp or greater.
```
ceyonur, line 209:
```suggestion
// IsAllowList returns whether [blockTimestamp] is either equal to the AllowList fork block timestamp or greater.
```

ceyonur (commented):
some nits, overall looks very good. I LOVE that single file encapsulating all logic.

As we talked in slack + and @patrick-ogrady's suggestion: we might be able to take this one step further where precompiles register themselves without any code change in chain configs (and maybe even in genesis?) for upgrades. But this can probably be done in a separate PR.

Nice work!

core/vm/contracts_stateful.go
patrick-ogrady: Nah, I think it is good to do inside `Run` to be consistent

core/state_processor.go
aaronbuchwald: Not yet, this is exactly what I was thinking of: adding a function to ChainConfig that returns a slice of these that can be easily iterated over.

precompile/allow_list.go
aaronbuchwald: Good call, this is unnecessary.

precompile/stateful_precompile_config.go
aaronbuchwald: Using nil as the parent timestamp within the genesis.

aaronbuchwald:
Need to change `AllowListConfig` from being an interface back to the actual struct so that it can be unmarshalled correctly. This has the unfortunate repercussion that nil checks don't work on it as expected due to go weirdness.

core/vm/evm.go
```diff
@@ -65,6 +66,12 @@ func (evm *EVM) precompile(addr common.Address) (StatefulPrecompiledContract, bo
 	default:
 		precompiles = PrecompiledContractsHomestead
 	}
+
+	// Enable the optional stateful precompiles
+	if evm.chainRules.IsAllowListEnabled {
+		precompiles[precompile.ModifyAllowListAddress] = precompile.ModifyAllowListPrecompile
+		precompiles[precompile.ReadAllowListAddress] = precompile.ReadAllowListPrecompile
```
patrick-ogrady, line 73: Why have a separate precompile for reading? Is this to avoid more complex input parsing? (will continue reading)

miner/worker.go
```diff
@@ -161,6 +161,8 @@ func (w *worker) commitNewWork() (*types.Block, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed to create new current environment: %w", err)
 	}
+	// Configure any stateful precompiles that should go into effect during this block.
```
patrick-ogrady, line 164: nit \n

miner/worker.go
```diff
@@ -161,6 +161,8 @@ func (w *worker) commitNewWork() (*types.Block, error) {
...
+	w.chainConfig.CheckConfigurePrecompiles(new(big.Int).SetUint64(parent.Time()), bigTimestamp, env.state)
```
patrick-ogrady, line 165: good catch!

params/config.go
```diff
@@ -113,6 +115,8 @@ type ChainConfig struct {
 
 	FeeConfig          *FeeConfig `json:"feeConfig,omitempty"`
 	AllowFeeRecipients bool       `json:"allowFeeRecipients,omitempty"` // Allows fees to be collected by block builders.
+
+	AllowListConfig precompile.AllowListConfig `json:"allowListConfig"` // Config for the allow list precompile
```
patrick-ogrady, line 119: nit: `,omitempty`

params/config.go
```diff
@@ -438,5 +449,27 @@ func (c *ChainConfig) AvalancheRules(blockNum, blockTimestamp *big.Int) Rules {
...
 
 	rules.IsSubnetEVM = c.IsSubnetEVM(blockTimestamp)
+	rules.IsAllowListEnabled = c.IsAllowList(blockTimestamp)
 	return rules
 }
+
+// enabledStatefulPrecompiles returns a list of stateful precompile configs in the order that they are enabled
+// by block timestamp.
+func (c *ChainConfig) enabledStatefulPrecompiles() []precompile.StatefulPrecompileConfig {
+	statefulPrecompileConfigs := make([]precompile.StatefulPrecompileConfig, 0)
```
patrick-ogrady, line 459: make 0? just do `[]precompile.StatefulPrecompileConfig{}` or `make([]precompile.StatefulPrecompileConfig)`

precompile/allow_list.go
```diff
@@ -0,0 +1,195 @@
...
+
+// modifyAllowListPrecompile implements StatefulPrecompiledContract and can be used as a thread-safe singleton.
+// Provides designated admins to modify the contract deployers allow list.
+type modifyAllowListPrecompile struct{}
+
+// Run verifies that [callerAddr] has the correct permissions to modify the allow list and if so updates the the allow list
+// as requested by the arguments encoded in [input].
+func (mal *modifyAllowListPrecompile) Run(evm PrecompileAccessibleState, callerAddr common.Address, addr common.Address, input []byte, suppliedGas uint64, readOnly bool) (ret []byte, remainingGas uint64, err error) {
+	// Note: this should never happen since the required gas should be verified before calling Run.
+	if suppliedGas < ModifyAllowListGasCost {
```
patrick-ogrady, line 139: If we handle this error, why do we enforce the invariant in the caller?

precompile/allow_list.sol
```diff
@@ -0,0 +1,44 @@
+// (c) 2022-2023, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+// SPDX-License-Identifier: MIT
+
+pragma solidity >=0.8.0;
+
+library AllowList {
+    address constant modifyAllowListAddr = 0x0200000000000000000000000000000000000000;
+    address constant readAllowListAddr = 0x0200000000000000000000000000000000000000;
```
patrick-ogrady, line 10: nit: should be 0x....1

core/vm/evm.go
patrick-ogrady: Seems like we could instead use the first byte as a function selector? like 0x0 == modify and 0x1 == read

I think that is a much more scalable pattern than splitting.

precompile/stateful_precompile_config.go
```diff
@@ -0,0 +1,49 @@
...
+	Configure(StateDB)
+}
+
+// CheckConfigure checks if [config] is activated by the transition from block at [parentTimestamp] to [currentTimestamp].
+// If it does, then it calls Configure on [config] to make the necessary state update to enable the StatefulPrecompile.
+// Note: this function is called within genesis to configure the starting state if it [config] specifies that it should be
+// configured at genesis, or happens during block processing to update the state before processing the given block.
+// TODO: add ability to call Configure at different timestamps, so that developers can easily re-configure by updating the
+// stateful precompile config.
+func CheckConfigure(parentTimestamp *big.Int, currentTimestamp *big.Int, config StatefulPrecompileConfig, state StateDB) {
```
patrick-ogrady, line 37: Please note that genesis activations treat `parentTimestamp` as `nil`

precompile/params.go
```diff
@@ -0,0 +1,25 @@
...
+	ModifyAllowListGasCost = 20_000
+	ReadAllowListGasCost   = 5_000
+)
+
+// Designated addresses of stateful precompiles
+// Note: it is important that none of these addresses conflict with each other or any other precompiles
+// in core/vm/contracts.go.
+// We start at 0x0200000000000000000000000000000000000000 and will increment by 1 from here to reduce
+// the risk of conflicts.
+// For forks of subnet-evm, users should start at 0x0300000000000000000000000000000000000000 to ensure
```
patrick-ogrady, line 19: ❤️ ❤️ ❤️

core/vm/evm.go
```diff
@@ -226,7 +238,7 @@ func (evm *EVM) Call(caller ContractRef, addr common.Address, input []byte, gas
 	}
 
 	if isPrecompile {
-		ret, gas, err = p.Run(evm, caller, addr, input, gas, evm.interpreter.readOnly)
+		ret, gas, err = RunStatefulPrecompiledContract(p, evm, caller.Address(), addr, input, gas, evm.interpreter.readOnly)
```
patrick-ogrady, line 241: nit: why call `.Address()` here instead of passing in caller context?

core/vm/evm.go
```diff
@@ -32,13 +32,32 @@ import (
...
+	"github.com/ava-labs/subnet-evm/constants"
 	"github.com/ava-labs/subnet-evm/params"
+	"github.com/ava-labs/subnet-evm/precompile"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/holiman/uint256"
 )
 
+var prohibitedAddresses = map[common.Address]struct{}{
```
patrick-ogrady, line 44: much cleaner than array structure ❤️

aaronbuchwald:
Closing in favor of #31

PR #32, opened 2022-03-01
[proposal] automatically set marker + use code instead of nonce to be more canonical

patrick-ogrady:


params/config.go
```diff
@@ -374,12 +394,27 @@ func (c *ChainConfig) checkCompatible(newcfg *ChainConfig, headHeight *big.Int,
...
+		if isForkIncompatible(c.AllowListConfig.Timestamp(), newcfg.AllowListConfig.Timestamp(), headTimestamp) {
+			return newCompatError("AllowList fork block timestamp", c.AllowListConfig.Timestamp(), newcfg.AllowListConfig.Timestamp())
+		}
+	} else if !match {
+		return newCompatError("AllowList", currentTime, currentTime)
+	}
 
-	if isForkIncompatible(c.AllowListConfig.Timestamp(), newcfg.AllowListConfig.Timestamp(), headTimestamp) {
-		return newCompatError("AllowList fork block timestamp", c.AllowListConfig.Timestamp(), newcfg.AllowListConfig.Timestamp())
+	// Check compatibility of misc, non-timed configs ([currentTime] returns on
```
patrick-ogrady, line 411: Not sure if we want to check these here 🤷

params/config.go
```diff
@@ -339,6 +344,21 @@ func (c *ChainConfig) CheckConfigForkOrder() error {
...
+func isNil(a interface{}) bool {
+	return a == nil || reflect.ValueOf(a).IsNil()
+}
+
+func matchNil(a, b interface{}) (match, ni bool) {
+	switch {
+	case isNil(a) && isNil(b):
+		return true, true
+	case !isNil(a) && !isNil(b):
+		return true, false
```
ceyonur, line 356: shouldn't we check if their values are same here? like `return a==b, false`

params/config.go
```diff
@@ -374,12 +394,27 @@ func (c *ChainConfig) checkCompatible(newcfg *ChainConfig, headHeight *big.Int,
...
 		return newCompatError("SubnetEVM fork block timestamp", c.SubnetEVMTimestamp, newcfg.SubnetEVMTimestamp)
 	}
+	currentTime := big.NewInt(time.Now().Unix())
+	if match, ni := matchNil(c.AllowListConfig, newcfg.AllowListConfig); match && !ni {
+		if isForkIncompatible(c.AllowListConfig.Timestamp(), newcfg.AllowListConfig.Timestamp(), headTimestamp) {
+			return newCompatError("AllowList fork block timestamp", c.AllowListConfig.Timestamp(), newcfg.AllowListConfig.Timestamp())
+		}
+	} else if !match {
+		return newCompatError("AllowList", currentTime, currentTime)
+	}
```
ceyonur, line 409: should we make a specialized function for that so it can be a example for other precompile configs?

params/config.go
```diff
@@ -483,8 +518,8 @@ func (c *ChainConfig) AvalancheRules(blockNum, blockTimestamp *big.Int) Rules {
 func (c *ChainConfig) enabledStatefulPrecompiles() []precompile.StatefulPrecompileConfig {
 	statefulPrecompileConfigs := make([]precompile.StatefulPrecompileConfig, 0)
 
-	if c.AllowListConfig.Timestamp() != nil {
```
ceyonur, line 486: why we ommitted Timestamp nil check here?

params/config.go
ceyonur: I'm preparing a PR for this.

ceyonur (approved):
It looks good, I think it's best to merge this ASAP and review it in the base branch.

params/config.go
```diff
@@ -234,10 +236,13 @@ func (c *ChainConfig) IsSubnetEVM(blockTimestamp *big.Int) bool {
...
 // IsAllowList returns whether [blockTimestamp] is either equal to the AllowList fork block timestamp or greater.
 func (c *ChainConfig) IsAllowList(blockTimestamp *big.Int) bool {
+	if c.AllowListConfig == nil {
+		return false
+	}
 	return utils.IsForked(c.AllowListConfig.Timestamp(), blockTimestamp)
 }
 
-// IsIstanbul returns whether num is either equal to the Istanbul fork block or greater.
+// GetFeeConfig returns the *FeeConfig if it exists.
```
aaronbuchwald, line 245: Can we update the comment to reflect that it returns the default config otherwise?

params/config.go
aaronbuchwald: I think it would make sense to add a check here, but don't think using matchNil is necessary since they are the same type. Also, I think we would want to enforce a more thorough check since any change here will result in a rule set change and there's no block timestamp currently in this config to dictate when such a change would go into effect.

params/config.go
aaronbuchwald: I believe this is omitted as it's assuming that we are replacing the meaning of a nil timestamp with a nil config entirely.

PR #34, opened 2022-03-01
Set code precompile check

ceyonur:
Adds generic methods for precompiles to reduce any duplicated code for adding new precompiles.

### DIFF
--- a/core/vm/evm.go
+++ b/core/vm/evm.go
@@ -86,7 +86,7 @@ func (evm *EVM) precompile(addr common.Address) (precompile.StatefulPrecompiledC
 	}
 
 	// Enable the optional stateful precompiles
-	if evm.chainRules.IsPrecompileEnabled(precompile.AllowListName) {
+	if evm.chainRules.IsPrecompileEnabled(precompile.AllowListAddress) {
 		precompiles[precompile.AllowListAddress] = precompile.AllowListPrecompile
 	}
 	p, ok := precompiles[addr]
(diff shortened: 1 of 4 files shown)


core/vm/evm.go
```diff
@@ -86,7 +86,7 @@ func (evm *EVM) precompile(addr common.Address) (precompile.StatefulPrecompiledC
 	}
 
 	// Enable the optional stateful precompiles
-	if evm.chainRules.IsAllowListEnabled {
+	if evm.chainRules.IsPrecompileEnabled(precompile.AllowListName) {
```
aaronbuchwald, line 89: We should be able to return a map of the addresses to the stateful precompile singletons to clean this up further and remove the need for adding `Name()` to the interface.

params/config.go
```diff
@@ -509,10 +518,22 @@ func (c *ChainConfig) AvalancheRules(blockNum, blockTimestamp *big.Int) Rules {
 	rules := c.rules(blockNum)
 
 	rules.IsSubnetEVM = c.IsSubnetEVM(blockTimestamp)
-	rules.IsAllowListEnabled = c.IsAllowList(blockTimestamp)
+
+	precompileMap := make(map[string]bool, len(c.enabledStatefulPrecompiles()))
+	for _, enabledPrecompile := range c.enabledStatefulPrecompiles() {
+		precompileMap[enabledPrecompile.Name()] = c.IsPrecompileActivated(enabledPrecompile, blockTimestamp)
+	}
```
aaronbuchwald, line 525: `enabledStatefulPrecompiles()` should only return non-nil configs that need to be enabled. This is so that we can do as few checks as possible for them when processing blocks. We can also order them in the future and cache it, so that we don't do this each time and can stop as soon as we hit a timestamp that doesn't need to be activated instead of iterating through all of them every time.

core/vm/evm.go
ceyonur: I changed it, hope this is what you meant.

params/config.go
ceyonur: changed bool to struct. `enabledStatefulPrecompiles()` already returns non-nil configs. My intention was to reduce the need LOC to add a precompile. With this we can tell developers to add their precompile to `enabledStatefulPrecompiles` with proper code and be done with it. Otherwise they need to add precompile to `Rule` struct + `enabledStatefulPrecompiles()` + `AvalancheRules` + another function for `IsEnabled`.

aaronbuchwald:
Closing in favor of taking a slightly different approach in how we manage the precompiles via evm rules.

PR #36, opened 2022-03-01
Remove genesis contract from test genesis json configs

aaronbuchwald:
This PR removes the deprecated genesis contract from the genesis JSON used in the vm tests.


ceyonur (approved):
LGreenTM

PR #41, opened 2022-03-04
Ci improvements

aaronbuchwald:
This PR marks a few identified flaky tests as flaky so we can skip over them until we can fix them.

Additionally, this adds a coverage.sh script from coreth and modifies the build_test.sh script to take in arbitrary arguments. This is a change from coreth that's used to pass in different arguments when running on different OSs so that we do not need to run with race detection enabled on multiple GH Actions and to optimize the runtime.

### DIFF
--- a/.github/workflows/lint-and-tests.yml
+++ b/.github/workflows/lint-and-tests.yml
@@ -29,5 +29,7 @@ jobs:
         shell: bash
       - run: ./scripts/build.sh ./build/subnetevm
         shell: bash
-      - run: ./scripts/build_test.sh
+      - run: ./scripts/build_test.sh -race
+        shell: bash
+      - run: ./scripts/coverage.sh
         shell: bash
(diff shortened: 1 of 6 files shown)


.github/workflows/lint-and-tests.yml
```diff
@@ -29,5 +29,7 @@ jobs:
         shell: bash
       - run: ./scripts/build.sh ./build/subnetevm
         shell: bash
-      - run: ./scripts/build_test.sh
+      - run: ./scripts/build_test.sh -race
```
ceyonur, line 32: 👍

ceyonur (approved):

plugin/evm/tx_gossiping_test.go
```diff
@@ -77,6 +77,7 @@ func getValidTxs(key *ecdsa.PrivateKey, count int, gasPrice *big.Int) []*types.T
 // to ease up UT, which target only VM behaviors in response to subnet-evm mempool
 // signals
 func TestMempoolTxsAddedTxsGossipedAfterActivation(t *testing.T) {
+	t.Skip("FLAKY")
```
hexfusion, line 80: optional: Just a thought but maybe we could attach an issue to the skip for tracking.

t.Skip("FLAKY") // AV-1234

patrick-ogrady: Nah, we'll keep this private for now.

PR #45, opened 2022-03-07
AV-1296: add minter precompile

ceyonur:
- [x] Add minter precompile
- [x] Add Test


precompile/contract_native_minter.go
```diff
@@ -0,0 +1,133 @@
...
+		callerStatus := getAllowListStatus(evm.GetStateDB(), precompileAddr, callerAddr)
+		if !callerStatus.IsEnabled() {
+			return nil, remainingGas, fmt.Errorf("%w: %s", ErrCannotMint, callerAddr)
+		}
+
+		if !evm.GetStateDB().Exist(to) {
+			if remainingGas < CallNewAccountGas {
+				return nil, 0, fmt.Errorf("%w (%d) < (%d)", ErrExceedsGasAllowance, CallNewAccountGas, suppliedGas)
+			}
+			remainingGas -= CallNewAccountGas
```
patrick-ogrady, line 111: This makes our `RequiredGas` invariant weird because it could surpass the amount returned?

precompile/params.go
```diff
@@ -21,8 +26,10 @@ const (
 // in the future.
 var (
 	ContractDeployerAllowListAddress = common.HexToAddress("0x0200000000000000000000000000000000000000")
+	ContractNativeMinterAddress      = common.HexToAddress("0x0300000000000000000000000000000000000000")
```
patrick-ogrady, line 29: I think we should do `0x020...1`

precompile/params.go
```diff
@@ -9,6 +9,11 @@ import "github.com/ethereum/go-ethereum/common"
 const (
 	ModifyAllowListGasCost = 20_000
 	ReadAllowListGasCost   = 5_000
+
+	MintGasCost = 20_000
+	BurnGasCost = 20_000
+
+	CallNewAccountGas = 25000 // Paid for CALL when the destination address didn't exist prior.
```
patrick-ogrady, line 16: This seems excessive given that a send to any address in ETH (even if it creates a new account is 21k).

precompile/params.go
```diff
@@ -9,6 +9,11 @@ import "github.com/ethereum/go-ethereum/common"
 const (
 	ModifyAllowListGasCost = 20_000
 	ReadAllowListGasCost   = 5_000
+
+	MintGasCost = 20_000
+	BurnGasCost = 20_000
```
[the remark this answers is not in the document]
ceyonur: right, this was a leftover from my first attempt.

precompile/params.go
ceyonur: this is same as https://github.com/ava-labs/subnet-evm/blob/master/params/protocol_params.go#L41

precompile/contract_native_minter.go
ceyonur: ah, true! however the `RequiredGas` function of this precompile requires evm state (for `state.Exist(addrs)`) should we change `RequiredGas` structure to take state. I think we need to re-implement what we already have in `createMint` to determine required gas. Alternatively we can omit `RequiredGas` function altogether and let gas checks run in the execute functions. @aaronbuchwald

patrick-ogrady:
image

precompile/contract_native_minter.go
```diff
@@ -0,0 +1,129 @@
...
+// contract deployer allow list.
+// assumes [role] has already been verified as valid.
+func SetContractNativeMinterStatus(stateDB StateDB, address common.Address, role AllowListRole) {
+	setAllowListRole(stateDB, ContractNativeMinterAddress, address, role)
+}
+
+// PackMintInput packs [address] and [amount] into the appropriate arguments for minting operation.
+func PackMintInput(address common.Address, amount *big.Int) ([]byte, error) {
+	// function selector (4 bytes) + input(hash for address + hash for amount)
+	input := make([]byte, 0, selectorLen+inputLen)
```
aaronbuchwald, line 62: can we change the name of `inputLen` to be specific to the function it's used for?

precompile/contract_native_minter.go
```diff
@@ -0,0 +1,129 @@
...
+		return common.Address{}, nil, fmt.Errorf("invalid input length for minting: %d", len(input))
+	}
+	to := common.BytesToAddress(input[:common.AddressLength])
+	assetAmount := new(big.Int).SetBytes(input[common.AddressLength : common.AddressLength+common.HashLength])
+	return to, assetAmount, nil
+}
+
+// createMint // TODO: add comment
+func createMint(precompileAddr common.Address) RunStatefulPrecompileFunc {
+	return func(evm PrecompileAccessibleState, callerAddr, addr common.Address, input []byte, suppliedGas uint64, readOnly bool) (ret []byte, remainingGas uint64, err error) {
```
aaronbuchwald, line 81: Not sure if we want to follow the same pattern as for the allow list here to create a minting execution function. I did this with the allow list since it might be re-used for different types of allow lists at different addresses, as well as to use it for writing different roles, but here there is only one native asset to be minted, and I would think that we will only ever deploy it at a single address.

precompile/contract_native_minter.go
aaronbuchwald: I'd be in favor of adding state into `RequiredGas`. The reason I separated them was so that we could return `ErrOutOfGas` if there was not sufficient gas to cover the operation.

Alternatively, I think the best solution would be to move all of the errors into a package that could be used by both the EVM and the precompile package, so that we can return `ErrOutOfGas` from the precompile directly, and then I think it would make sense to turn the stateful precompiles into a single execution function and get rid of the `RequiredGas` function entirely.

precompile/params.go
```diff
@@ -9,6 +9,10 @@ import "github.com/ethereum/go-ethereum/common"
 const (
 	ModifyAllowListGasCost = 20_000
 	ReadAllowListGasCost   = 5_000
+
+	MintGasCost = 20_000
+
+	CallNewAccountGas = 25_000 // Paid for CALL when the destination address didn't exist prior.
```
aaronbuchwald, line 15: Can we use the parameter in the `params` package instead of copying it in here?

precompile/contract_native_minter.go
```diff
@@ -0,0 +1,129 @@
...
+		}
+
+		evm.GetStateDB().AddBalance(to, amount)
+		// Return an empty output and the remaining gas
+		return []byte{}, remainingGas, nil
+	}
+}
+
+// createNativeMinterPrecompile returns a StatefulPrecompiledContract with R/W control of an allow list at [precompileAddr]
+func createNativeMinterPrecompile(precompileAddr common.Address) StatefulPrecompiledContract {
```
aaronbuchwald, line 118: nit: update comment to include minting functionality

precompile/contract_native_minter.go
```diff
@@ -0,0 +1,129 @@
...
+			return nil, remainingGas, err
+		}
+
+		// Verify that the caller is in the allow list and therefore has the right to modify it
+		callerStatus := getAllowListStatus(evm.GetStateDB(), precompileAddr, callerAddr)
+		if !callerStatus.IsEnabled() {
+			return nil, remainingGas, fmt.Errorf("%w: %s", ErrCannotMint, callerAddr)
+		}
+
+		if !evm.GetStateDB().Exist(to) {
```
aaronbuchwald, line 103: nit: call `GetStateDB()` once and use the returned `stateDB` throughout.

precompile/contract_native_minter.go
```diff
@@ -0,0 +1,129 @@
...
+)
+
+var (
+	_ StatefulPrecompileConfig = &ContractNativeMinterConfig{}
+	// Singleton StatefulPrecompiledContract for W/R access to the contract deployer allow list.
+	ContractNativeMinterPrecompile StatefulPrecompiledContract = createNativeMinterPrecompile(ContractNativeMinterAddress)
+
+	mintSignature = CalculateFunctionSelector("mint(address,uint256)") // address, amount
+
+	ErrCannotMint = errors.New("non-enabled cannot mint")
```
aaronbuchwald, line 21: nit: move to new errors file

aaronbuchwald (changes requested):

precompile/params.go
ceyonur: cyclic import :/

ceyonur: I wonder why chain configs and params are in the same package...

precompile/contract_native_minter.go
ceyonur: my intention was to encapsulate generic errors in `errors` file and keep precompile-specific errors in their own files.

precompile/params.go
```diff
@@ -9,6 +9,10 @@ import "github.com/ethereum/go-ethereum/common"
 const (
 	ModifyAllowListGasCost = 20_000
 	ReadAllowListGasCost   = 5_000
+
+	MintGasCost = 20_000
```
patrick-ogrady, line 13: Only real nit on this PR is to not make minting dynamic gas. Let's use 30_000 and call it a day for all ops.

precompile/contract_native_minter.go
```diff
@@ -0,0 +1,129 @@
...
+		}
+
+		// Verify that the caller is in the allow list and therefore has the right to modify it
+		callerStatus := getAllowListStatus(evm.GetStateDB(), precompileAddr, callerAddr)
+		if !callerStatus.IsEnabled() {
+			return nil, remainingGas, fmt.Errorf("%w: %s", ErrCannotMint, callerAddr)
+		}
+
+		if !evm.GetStateDB().Exist(to) {
+			if remainingGas < CallNewAccountGas {
```
patrick-ogrady, line 104: can use `deductGas` here

precompile/contract_native_minter.go
ceyonur: removed `RequiredGas`. We can return  `ErrOutOfGas` without moving it into a separate package. Returned it in the helper function.

precompile/params.go
ceyonur: removed this

precompile/contract_native_minter.go
ceyonur: done

precompile/contract_native_minter.go
ceyonur: done

precompile/params.go
ceyonur: removed this completely

precompile/params.go
ceyonur: done

precompile/contract_native_minter.go
```diff
@@ -0,0 +1,129 @@
...
+	to := common.BytesToAddress(input[:common.AddressLength])
+	assetAmount := new(big.Int).SetBytes(input[common.AddressLength : common.AddressLength+common.HashLength])
+	return to, assetAmount, nil
+}
+
+// createMint // TODO: add comment
+func createMint(precompileAddr common.Address) RunStatefulPrecompileFunc {
+	return func(evm PrecompileAccessibleState, callerAddr, addr common.Address, input []byte, suppliedGas uint64, readOnly bool) (ret []byte, remainingGas uint64, err error) {
+		// Note: this should never happen since the required gas should be verified before calling Run.
+		if suppliedGas < MintGasCost {
```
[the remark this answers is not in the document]
ceyonur: done

core/stateful_precompile_test.go
```diff
@@ -281,3 +283,300 @@ func TestContractDeployerAllowListRun(t *testing.T) {
...
+				if err != nil {
+					panic(err)
+				}
+				return input
+			},
+			suppliedGas: precompile.ModifyAllowListGasCost,
+			readOnly:    true,
+			setupState: func(state *state.StateDB) {
+				precompile.SetContractNativeMinterStatus(state, adminAddr, precompile.AllowListAdmin)
+				precompile.SetContractNativeMinterStatus(state, noRoleAddr, precompile.AllowListEnabled)
```
aaronbuchwald, line 368: Can we add a separate address for the enabled status? We shouldn't use an address called `noRoleAddr` as an allow listed address.

precompile/contract_native_minter.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	to := common.BytesToAddress(input[:common.HashLength])
+	assetAmount := new(big.Int).SetBytes(input[common.HashLength : common.HashLength+common.HashLength])
+	return to, assetAmount, nil
+}
+
+// createMint checks if the caller is permissioned for minting operation.
+// The execution function parses the [input] into native token amount and receiver address.
+func createNativeMintToken(accessibleState PrecompileAccessibleState, caller common.Address, addr common.Address, input []byte, suppliedGas uint64, readOnly bool) (ret []byte, remainingGas uint64, err error) {
+	if remainingGas, err = deductGas(suppliedGas, MintGasCost); err != nil {
```
patrick-ogrady, line 84: ❤️ ❤️

precompile/utils.go
```diff
@@ -23,8 +24,10 @@ func CalculateFunctionSelector(functionSignature string) []byte {
 	return hash[:4]
 }
 
-// createConstantRequiredGasFunc returns a required gas function that always returns [requiredGas]
-// on any input.
-func createConstantRequiredGasFunc(requiredGas uint64) func([]byte) uint64 {
-	return func(b []byte) uint64 { return requiredGas }
+// deductGas checks if [suppliedGas] is sufficient against [requiredGas] and deducts [requiredGas] from [suppliedGas].
+func deductGas(suppliedGas uint64, requiredGas uint64) (uint64, error) {
```
patrick-ogrady, line 28: Should we allow the user to also provide a string to wrap the error as well?

PR #49, opened 2022-03-10
refactor(Dockerfile):  Set default values. Remove unused argument

adutchak:
1. Setting some arg defaults. We need this to simplify the requirements for clients and avoid introduction of extra config files.
2. I could not find in the repo where `CURRENT_BRANCH` argument is used, therefore removing it.

### DIFF
--- a/scripts/constants.sh
+++ b/scripts/constants.sh
@@ -6,9 +6,6 @@ GOPATH="$(go env GOPATH)"
 # Avalabs docker hub
 dockerhub_repo="avaplatform/avalanchego"
 
-# Current branch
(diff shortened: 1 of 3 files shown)


scripts/constants.sh
```diff
@@ -6,9 +6,6 @@ GOPATH="$(go env GOPATH)"
 # Avalabs docker hub
 dockerhub_repo="avaplatform/avalanchego"
 
-# Current branch
```
patrick-ogrady, line 9: Can we keep this? This is very useful when running these builds

adutchak: Of course, I just searched everywhere in a repo and could not find where it is actually used.

PR #51, opened 2022-03-10
Add docs on genesis allocation

aaronbuchwald:
This PR adds documentation for defining genesis allocations to the README.


README.md
```diff
@@ -41,7 +41,83 @@ The Subnet EVM is compatible with almost all Ethereum tooling, including [Remix,
...
+    "8db97C7cEcE249c2b98bDC0226Cc4C2A57BF52FC": {
+      "balance": "0x52B7D2DCC80CD2E4000000"
+    },
+    "Ab5801a7D398351b8bE11C439e05C5B3259aeC9B": {
+      "balance": "0xa796504b1cb5a7c0000"
+    }
+  },
+```
+
+The keys in the allocation are the hex addresses to provide initial funds with the normal `0x` prefix stripped from the string. The balances are defined as hex strings with the `0x` prefix included and denominated in Wei.
```
gyuho, line 61: Can we add a reference for hexadecimal https://en.wikipedia.org/wiki/Hexadecimal,

> defined as a hexadecimal number in string with the `0x` prefix.

Also, explain 1 AVAX in C/subnet-evm chain is denominated by 10^18 units of nano avax?

gyuho: I found this link helpful https://www.rapidtables.com/convert/number/hex-to-decimal.html but not sure if we want to include that, or there's any better link/tool

patrick-ogrady: Good comments, will add some small tweaks.

PR #52, opened 2022-03-11
Nits

aaronbuchwald:
This PR:

1) Does a cleanup of the minter precompile test suite
2) Moves VM errors into their own package so they can be shared across the EVM and precompiles in their own package
3) Reverts the change that makes it so only explicitly allowed addresses can mint (makes it so that admins are enabled as well)


core/stateful_precompile_test.go
```diff
@@ -315,19 +314,29 @@ func TestContractNativeMinterRun(t *testing.T) {
...
+			expectedRes: []byte{},
 			assertState: func(t *testing.T, state *state.StateDB) {
-				res := precompile.GetContractNativeMinterStatus(state, adminAddr)
-				assert.Equal(t, precompile.AllowListAdmin, res)
+				res := precompile.GetContractNativeMinterStatus(state, allowAddr)
+				assert.Equal(t, precompile.AllowListEnabled, res)
 
-				res = precompile.GetContractNativeMinterStatus(state, noRoleAddr)
-				assert.Equal(t, precompile.AllowListAdmin, res)
+				assert.Equal(t, common.Big1, state.GetBalance(allowAddr), "expected increased admin funds")
```
ceyonur, line 336:
```suggestion
				assert.Equal(t, common.Big1, state.GetBalance(allowAddr), "expected increased funds")
```
this is not the admin address

precompile/allow_list.go
```diff
@@ -83,7 +84,7 @@ func (s AllowListRole) IsAdmin() bool {
 // IsEnabled returns true if [s] indicates that it has permission to access the resource.
 func (s AllowListRole) IsEnabled() bool {
 	switch s {
-	case AllowListEnabled:
+	case AllowListAdmin, AllowListEnabled:
```
ceyonur, line 87: I feel like this is kind-of misleading since both of role name and function contains `Enabled` however the function returns true for `Admin` as well. Plus do we really want admins to be able to mint tokens?

core/stateful_precompile_test.go
```diff
@@ -339,19 +348,15 @@ func TestContractNativeMinterRun(t *testing.T) {
...
 			suppliedGas: precompile.MintGasCost,
 			readOnly:    false,
-			setupState: func(state *state.StateDB) {
-				precompile.SetContractNativeMinterStatus(state, adminAddr, precompile.AllowListAdmin)
-			},
-			expectedErr: precompile.ErrCannotMint.Error(),
+			expectedRes: []byte{},
 			assertState: func(t *testing.T, state *state.StateDB) {
 				res := precompile.GetContractNativeMinterStatus(state, adminAddr)
 				assert.Equal(t, precompile.AllowListAdmin, res)
```
ceyonur, line 354: should we also remove setupStates for the `TestContractDeployerAllowListRun` tests (above)?

core/stateful_precompile_test.go
```diff
@@ -363,31 +368,23 @@ func TestContractNativeMinterRun(t *testing.T) {
...
-			setupState: func(state *state.StateDB) {
-				precompile.SetContractNativeMinterStatus(state, adminAddr, precompile.AllowListAdmin)
-				precompile.SetContractNativeMinterStatus(state, noRoleAddr, precompile.AllowListEnabled)
-			},
-			expectedErr: precompile.ErrWriteProtection.Error(),
+			expectedErr: vmerrs.ErrWriteProtection.Error(),
 		},
-		"mint with insufficient gas": {
-			caller:         noRoleAddr,
+		"readOnly mint with allow role fails": {
```
ceyonur, line 373: this does not test minting, should we rename this to something like
```suggestion
		"readOnly modify allow list with allow role in mint contract fails": {
```

core/stateful_precompile_test.go
```diff
@@ -363,31 +368,23 @@ func TestContractNativeMinterRun(t *testing.T) {
...
-				precompile.SetContractDeployerAllowListStatus(state, adminAddr, precompile.AllowListAdmin)
-				precompile.SetContractNativeMinterStatus(state, noRoleAddr, precompile.AllowListEnabled)
-			},
-			expectedErr: vm.ErrOutOfGas.Error(),
+			suppliedGas: precompile.ModifyAllowListGasCost,
+			readOnly:    true,
+			expectedErr: vmerrs.ErrWriteProtection.Error(),
 		},
-		"add role with insufficient gas": {
+		"readOnly mint with admin role fails": {
```
ceyonur, line 387:
```suggestion
		"readOnly modify allow list with admin role in mint contract fails": {
```

core/stateful_precompile_test.go
```diff
@@ -397,158 +394,90 @@ func TestContractNativeMinterRun(t *testing.T) {
...
-			assertState: func(t *testing.T, state *state.StateDB) {
-				res := precompile.GetContractNativeMinterStatus(state, noRoleAddr)
-				assert.Equal(t, precompile.AllowListEnabled, res)
-
-				finalBalance := state.GetBalance(noRoleAddr)
-				assert.Equal(t, common.Big1, finalBalance)
-			},
+			expectedErr: vmerrs.ErrOutOfGas.Error(),
 		},
-		"mint big": {
```
ceyonur, line 481: why have we removed this test-case? I'd prefer to test mint capabilities here instead of allowList capabilities. (or preferably both of them).

precompile/allow_list.go
```diff
@@ -153,7 +154,7 @@ func createAllowListRoleSetter(precompileAddr common.Address, role AllowListRole
 		modifyAddress := common.BytesToAddress(input)
 
 		if readOnly {
-			return nil, remainingGas, ErrWriteProtection
+			return nil, remainingGas, vmerrs.ErrWriteProtection
```
ceyonur, line 157: ooh seems I imported `go-ethereum` vm package, instead of subnet-evm. that's why it worked haha

ceyonur: but certainly not a bad idea? :P

vmerrs/vmerrs.go
```diff
@@ -0,0 +1,49 @@
...
+//
+// The go-ethereum library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+package vmerrs
```
ceyonur, line 27: should we rename this to `error` as it's actually not in the vm package? or maybe at least vmerrors?

ceyonur: alternatively we can try to import VM package from geth for errors? it won't be a clean way but still works, right?

precompile/allow_list.go
aaronbuchwald: Admin is a strictly higher and should be considered enabled as well. If we don't consider it enabled, then the UX for an admin will be that they have to create a new address, mark it as enabled, and then mint funds from that address. It's much simpler if we just treat Admin as a strictly higher privilege level.

precompile/allow_list.go
aaronbuchwald: I think we should migrate to a separate package. We should not rely on any geth packages where we also have the same package present in our code.

vmerrs/vmerrs.go
aaronbuchwald: Currently this is specific to vm errors, I'd be alright with renaming it to vmerrors, but prefer the shortened vmerrs tbh.

Same comment on importing the VM package as above.

core/stateful_precompile_test.go
aaronbuchwald: Added it back in. I didn't think this was really different than the other minting test case.

ceyonur (approved):
LGTM 👍

PR #54, opened 2022-03-12
Set precompile code

aaronbuchwald:
This PR sets the code of stateful precompiles when they are configured to 0x00. This ensures that the precompile will pass the check to ensure that there is a non-zero length contract at the called address when Solidity makes a call to another contract.

### DIFF
--- a/precompile/stateful_precompile_config.go
+++ b/precompile/stateful_precompile_config.go
@@ -52,7 +52,7 @@ func CheckConfigure(parentTimestamp *big.Int, currentTimestamp *big.Int, config 
 		// Set the code of the precompile's address to a non-zero length byte slice to ensure that the precompile
 		// can be called from within Solidity contracts. Solidity adds a check before invoking a contract to ensure
 		// that it does not attempt to invoke a non-existent contract.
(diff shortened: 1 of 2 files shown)


precompile/stateful_precompile_config.go
```diff
@@ -49,6 +49,10 @@ func CheckConfigure(parentTimestamp *big.Int, currentTimestamp *big.Int, config
 		// Set the nonce of the precompile's address (as is done when a contract is created) to ensure
 		// that it is marked as non-empty and will not be cleaned up when the statedb is finalized.
 		state.SetNonce(config.Address(), 1)
+		// Set the code of the precompile's address to a non-zero length byte slice to ensure that the precompile
+		// can be called from within Solidity contracts. Solidity adds a check before invoking a contract to ensure
+		// that it does not attempt to invoke a non-existent contract.
+		state.SetCode(config.Address(), []byte{0x00})
```
patrick-ogrady, line 55: Would it be better to do 0x1 here?

aaronbuchwald: Updated

PR #57, opened 2022-03-13
Simple Load Simulator

patrick-ogrady:
image


patrick-ogrady:
TODO:
* Add license headers
* Populate README

cmd/simulator/metrics/metrics.go
```diff
@@ -0,0 +1,80 @@
...
+
+	block, err := client.BlockByNumber(ctx, new(big.Int).SetUint64(blockNumber))
+	if err != nil {
+		return fmt.Errorf("failed to get block at number %d: %w", blockNumber, err)
+	}
+
+	startTime := block.Time()
+	currentTime := startTime
+	totalTxs := 0
+	timeTxs := map[uint64]int{}
```
aaronbuchwald, line 34:
```suggestion
	timeTxs := make(map[uint64]int)
```

cmd/simulator/metrics/metrics.go
```diff
@@ -0,0 +1,80 @@
...
+		if err != nil {
+			log.Printf("failed to get block number: %s", err)
+			time.Sleep(2 * time.Second)
+			continue
+		}
+
+		if blockNumber > newBlockNumber {
+			time.Sleep(2 * time.Second)
+			continue
+		}
```
aaronbuchwald, line 46: Should this be >=? I don't think we should ever regress backwards (assuming we are running with a normal config that only fetches accepted blocks).

cmd/simulator/metrics/metrics.go
```diff
@@ -0,0 +1,80 @@
...
+
+		var block *types.Block
+		for i := blockNumber; i <= newBlockNumber; i++ {
+			for ctx.Err() == nil {
+				block, err = client.BlockByNumber(ctx, new(big.Int).SetUint64(i))
+				if err != nil {
+					log.Printf("failed to get block at number %d: %s", blockNumber, err)
+					time.Sleep(2 * time.Second)
```
aaronbuchwald, line 54: can we move this standard sleep time to a variable at the top of the file?

cmd/simulator/metrics/metrics.go
```diff
@@ -0,0 +1,80 @@
...
+					continue
+				}
+				log.Printf("[block created] index: %d base fee: %d block gas cost: %d block txs: %d\n", i, block.BaseFee().Div(block.BaseFee(), big.NewInt(params.GWei)), block.BlockGasCost(), len(block.Body().Transactions))
+				timeTxs[block.Time()] += len(block.Transactions())
+				totalTxs += len(block.Transactions())
+				break
+			}
+		}
+		blockNumber = newBlockNumber + 1
```
aaronbuchwald, line 66: Ah I see this is how we can end up with `blockNumber < newBlockNumber` in the above loop. Can we update it to treat it as `lastBlockNumber` and then `continue` when `newBlockNumber <= lastBlockNumber` ?

cmd/simulator/worker/worker.go
```diff
@@ -0,0 +1,316 @@
...
+
+	requestAmount = new(big.Int).Mul(maxTransferCost, big.NewInt(100))
+	minFunderBalance = new(big.Int).Add(maxTransferCost, requestAmount)
+}
+
+func createWorkers(ctx context.Context, keys []*key.Key, endpoints []string, desiredWorkers int) (*worker, []*worker, error) {
+	var master *worker
+	var workers []*worker
+	for _, k := range keys {
+		worker, err := newWorker(k, endpoints[rand.Intn(len(endpoints))])
```
aaronbuchwald, line 61: Should we spread the load evenly among the endpoints instead of randomizing it?

Could be simply `endpoints[i%len(endpoints)]`

cmd/simulator/worker/worker.go
```diff
@@ -0,0 +1,316 @@
...
+		}
+		if err := worker.fetchNonce(ctx); err != nil {
+			return nil, nil, fmt.Errorf("could not get nonce: %w", err)
+		}
+		log.Printf("loaded worker %s (balance=%s nonce=%d)\n", worker.k.Address.Hex(), worker.balance.String(), worker.nonce)
+
+		switch {
+		case master == nil:
+			master = worker
+		case new(big.Int).Sub(master.balance, worker.balance).Sign() < 0:
```
aaronbuchwald, line 76:
```suggestion
		master.balance.Cmp(worker.balance) < 0:
```

aaronbuchwald: Could we also add a comment explaining that we swap the worker for the master if it has a higher balance so that the master worker ends up with the highest balance?

cmd/simulator/worker/worker.go
```diff
@@ -0,0 +1,316 @@
...
+		}
+	}
+
+	var err error
+	if master == nil {
+		master, err = newWorker(nil, endpoints[rand.Intn(len(endpoints))])
+		if err != nil {
+			return nil, nil, fmt.Errorf("unable to create master: %w", err)
+		}
+	}
```
aaronbuchwald, line 90: Should we return an error here or will this require that the user of the load simulator sends funds to the generated key? If the latter, can we add a comment here explaining why this isn't a fatal error.

cmd/simulator/worker/worker.go
```diff
@@ -0,0 +1,316 @@
...
+	return ctx.Err()
+}
+
+func (w *worker) waitForBalance(ctx context.Context, stdout bool, minBalance *big.Int) error {
+	for ctx.Err() == nil {
+		if err := w.fetchBalance(ctx); err != nil {
+			return fmt.Errorf("could not get balance: %w", err)
+		}
+
+		if new(big.Int).Sub(w.balance, minBalance).Sign() >= 0 {
```
aaronbuchwald, line 171:
```suggestion
		if w.balance.Cmp(minBalance) >= 0 {
```

cmd/simulator/worker/worker.go
```diff
@@ -0,0 +1,316 @@
...
+		w.balance = new(big.Int).Sub(w.balance, cost)
+		w.balance = new(big.Int).Sub(w.balance, transferAmount)
+		return nil
+	}
+	return ctx.Err()
+}
+
+func (w *worker) work(ctx context.Context, availableWorkers []*worker, fundRequest chan common.Address) error {
+	for ctx.Err() == nil {
+		if new(big.Int).Sub(w.balance, maxTransferCost).Sign() < 0 {
```
aaronbuchwald, line 225:
```suggestion
		if w.balance.Cmp(maxTransferCost) < 0 {
```

cmd/simulator/worker/worker.go
```diff
@@ -0,0 +1,316 @@
...
+		time.Sleep(workDelay)
+	}
+	return ctx.Err()
+}
+
+func (w *worker) fund(ctx context.Context, fundRequest chan common.Address) error {
+	for {
+		select {
+		case recipient := <-fundRequest:
+			if new(big.Int).Sub(w.balance, minFunderBalance).Sign() < 0 {
```
aaronbuchwald, line 248:
```suggestion
			if w.balance.Cmp(minFunderBalance) < 0 {
```

cmd/simulator/key/key.go
```diff
@@ -0,0 +1,85 @@
...
+	"path/filepath"
+
+	"github.com/ethereum/go-ethereum/common"
+	ethcrypto "github.com/ethereum/go-ethereum/crypto"
+)
+
+type Key struct {
+	PrivKey *ecdsa.PrivateKey
+	Address common.Address
+}
```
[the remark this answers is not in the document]
patrick-ogrady: Didn't know this existed lol

cmd/simulator/worker/worker.go
patrick-ogrady: there is a separate loop to wait for funds

cmd/simulator/worker/worker.go
```diff
@@ -0,0 +1,321 @@
...
+		}
+	}
+
+	for len(workers) < desiredWorkers {
+		i := len(workers)
+		worker, err := newWorker(nil, endpoints[i%len(endpoints)])
+		if err != nil {
+			// This should never happen
+			return nil, nil, fmt.Errorf("unable to create worker: %w", err)
+		}
```
aaronbuchwald, line 101: can we remove both `this should never happen` comments. This would happen if it's unable to establish a connection to the endpoint.

aaronbuchwald (approved):
One optional nit, other than that this LGTM

PR #63, opened 2022-03-16
Ensure gas limit in genesis is consistent

patrick-ogrady:
Return an error when the `gasLimit` specified in the genesis header does not match the `gasLimit` specified in the `FeeConfig`.


core/genesis.go
```diff
@@ -310,6 +310,12 @@ func (g *Genesis) ToBlock(db ethdb.Database) *types.Block {
 			head.BaseFee = g.Config.GetFeeConfig().MinBaseFee
 		}
 	}
+	if g.Config.FeeConfig.GasLimit.Uint64() != head.GasLimit {
+		panic(fmt.Sprintf(
+			"gas limit in fee config (%d) does not match gas limit in header (%d)",
+			g.Config.FeeConfig.GasLimit, head.GasLimit,
+		))
+	}
```
aaronbuchwald, line 318: Can we handle this within `SetupGenesisBlock` instead of `ToBlock` so that we can return a fatal error instead of panicking here?

core/genesis.go
```diff
@@ -184,6 +184,11 @@ func SetupGenesisBlock(db ethdb.Database, genesis *Genesis) (*params.ChainConfig
 	if genesis.Config == nil {
 		return nil, errGenesisNoConfig
 	}
+	// Make sure genesis gas limit is consistent
+	gasLimitConfig := genesis.Config.FeeConfig.GasLimit.Uint64()
```
patrick-ogrady, line 188: We could choose to only do this on a go-forward basis. (i.e. if genesis doesn't exist)

patrick-ogrady: This is obviously not a great solution because old validators can continue fine but new validators won't be able to join.

aaronbuchwald (approved):
LGTM

PR #66, opened 2022-03-30
Improve stop flow to reduce frequency of unclean shutdowns

aaronbuchwald:
This PR moves the shutdown manager stop above the other functionality when stopping the blockchain to ensure that the last trie is written as early as possible and reduce the likelihood that state will need to be regenerated on a restart.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -525,7 +525,13 @@ func (bc *BlockChain) Stop() {
 		return
 	}
 
+	log.Info("Shutting down state manager")
+	if err := bc.stateManager.Shutdown(); err != nil {
+		log.Error("Failed to Shutdown state manager", "err", err)
+	}
+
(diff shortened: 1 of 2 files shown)


core/blockchain.go
```diff
@@ -525,7 +525,13 @@ func (bc *BlockChain) Stop() {
 		return
 	}
 
+	log.Info("Shutting down state manager")
+	if err := bc.stateManager.Shutdown(); err != nil {
+		log.Error("Failed to Shutdown state manager", "err", err)
+	}
+
 	// Unsubscribe all subscriptions registered from blockchain.
+	log.Info("Clsoing subscriptions scope")
```
ceyonur, line 534:
```suggestion
	log.Info("Closing subscriptions scope")
```

ceyonur: on the second thought if we don't need this can we remove it so it can be consistent with the coreth?

PR #68, opened 2022-03-30
Add tx allow list

aaronbuchwald:
This PR adds a tx allow list stateful precompile.

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,6 +286,97 @@ If you call `readAllowList(addr)` then you can read the current role of `addr`, 
 
 WARNING: if you remove all of the admins from the allow list, it will no longer be possible to update the allow list without modifying the subnet-evm to schedule a network upgrade.
 
+### Restricting Who Can Submit Transactions
+Similar to restricting contract deployers, this precompile restricts which addresses may submit transactions on chain. Like the previous section, you can activate the precompile by including an `AllowList` configuration in your genesis file:
+```json
+{
+  "config": {
+    "chainId": 99999,
+    "homesteadBlock": 0,
+    "eip150Block": 0,
+    "eip150Hash": "0x2086799aeebeae135c246c65021c82b4e15a2c451340993aacfd2751886514f0",
+    "eip155Block": 0,
+    "eip158Block": 0,
+    "byzantiumBlock": 0,
+    "constantinopleBlock": 0,
+    "petersburgBlock": 0,
+    "istanbulBlock": 0,
+    "muirGlacierBlock": 0,
+    "subnetEVMTimestamp": 0,
+    "feeConfig": {
+      "gasLimit": 20000000,
+      "minBaseFee": 1000000000,
+      "targetGas": 100000000,
+      "baseFeeChangeDenominator": 48,
+      "minBlockGasCost": 0,
+      "maxBlockGasCost": 10000000,
+      "targetBlockRate": 2,
+      "blockGasCostStep": 500000
+    },
+    "txAllowListConfig": {
+      "blockTimestamp": 0,
+      "adminAddresses":["0x8db97C7cEcE249c2b98bDC0226Cc4C2A57BF52FC"]
+    }
+  },
+  "alloc": {
+    "8db97C7cEcE249c2b98bDC0226Cc4C2A57BF52FC": {
+      "balance": "0x52B7D2DCC80CD2E4000000"
+    }
+  },
+  "nonce": "0x0",
+  "timestamp": "0x0",
+  "extraData": "0x00",
+  "gasLimit": "0x1312D00",
+  "difficulty": "0x0",
+  "mixHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
+  "coinbase": "0x0000000000000000000000000000000000000000",
+  "number": "0x0",
+  "gasUsed": "0x0",
+  "parentHash": "0x0000000000000000000000000000000000000000000000000000000000000000"
+}
+```
+
+In this example, `0x8db97C7cEcE249c2b98bDC0226Cc4C2A57BF52FC` is named as the
+`Admin` of the `ContractDeployerAllowList`. This enables them to add other `Admins` or to add
+`Allowed`. Both `Admins` and `Allowed` can submit transactions to the chain.
+
+The `Stateful Precompile` powering the `TxAllowList` adheres to the following Solidity interface at `0x0200000000000000000000000000000000000002` (you can load this interface and interact directly in Remix):
+```solidity
+// (c) 2022-2023, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+// SPDX-License-Identifier: MIT
+
+pragma solidity >=0.8.0;
+
+interface AllowListInterface {
+    // Set [addr] to have the admin role over the allow list
+    function setAdmin(address addr) external;
+
+    // Set [addr] to be enabled on the allow list
+    function setEnabled(address addr) external;
+
+    // Set [addr] to have no role over the allow list
+    function setNone(address addr) external;
+
+    // Read the status of [addr]
+    function readAllowList(address addr) external view returns (uint256);
+}
+```
+
+If you attempt to add an `Allowed` and you are not an `Admin`, you will see
+something like:
+![admin fail](./imgs/admin_fail.png)
+
+If you attempt to submit a transaction but you are not an `Admin` or not
+`Allowed`, you will see something like: `cannot issue transaction from non-allow listed address`
+
+The allow list has three roles: `None`, `Allowed`, and `Admin`.
+
+If you call `readAllowList(addr)` then you can read the current role of `addr`, which will return a `uint256` with a value of 0, 1, or 2, corresponding to the roles `None`, `Allowed`, and `Admin` respectively.
+
+WARNING: if you remove all of the admins from the allow list, it will no longer be possible to update the allow list without modifying the subnet-evm to schedule a network upgrade.
+
 ### Minting Native Coins
 
 You can mint native(gas) coins with a precompiled contract. In order to activate this feature, you can provide `nativeMinterConfig` in genesis:
--- a/contract-examples/hardhat.config.ts
+++ b/contract-examples/hardhat.config.ts
@@ -3,20 +3,22 @@ import { SignerWithAddress } from "@nomiclabs/hardhat-ethers/signers"
 import { BigNumber } from "ethers"
 import "@nomiclabs/hardhat-waffle"
 
-const MINT_ADDRESS = "0x0200000000000000000000000000000000000001";
-const ALLOWLIST_ADDRESS = "0x0200000000000000000000000000000000000000";
-const BLACKHOLE_ADDRESS = "0x0100000000000000000000000000000000000000";
+const BLACKHOLE_ADDRESS = "0x0100000000000000000000000000000000000000"
+const CONTRACT_ALLOW_LIST_ADDRESS = "0x0200000000000000000000000000000000000000"
+const MINT_ADDRESS = "0x0200000000000000000000000000000000000001"
+const TX_ALLOW_LIST_ADDRESS = "0x0200000000000000000000000000000000000002"
+
 
 const ROLES = {
   0: "None",
   1: "Enabled",
   2: "Admin",
-};
+}
 
 const getRole = async (allowList, address) => {
-  const role = await allowList.readAllowList(address);
-  console.log(`${address} has role: ${ROLES[role.toNumber()]}`);
-};
+  const role = await allowList.readAllowList(address)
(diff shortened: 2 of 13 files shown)


precompile/tx_allow_list.go
```diff
@@ -0,0 +1,52 @@
...
+// GetTxAllowListStatus returns the role of [address] for the contract deployer
+// allow list.
+func GetTxAllowListStatus(stateDB StateDB, address common.Address) AllowListRole {
+	return getAllowListStatus(stateDB, TxAllowListAddress, address)
+}
+
+// SetTxAllowListStatus sets the permissions of [address] to [role] for the
+// tx allow list.
+// assumes [role] has already been verified as valid.
+func SetAllowListStatus(stateDB StateDB, address common.Address, role AllowListRole) {
```
aaronbuchwald, line 50: `SetTxAllowListStatus`

contract-examples/hardhat.config.ts
```diff
@@ -3,20 +3,21 @@ import { SignerWithAddress } from "@nomiclabs/hardhat-ethers/signers"
 import { BigNumber } from "ethers"
 import "@nomiclabs/hardhat-waffle"
 
-const MINT_ADDRESS = "0x0200000000000000000000000000000000000001";
-const ALLOWLIST_ADDRESS = "0x0200000000000000000000000000000000000000";
-const BLACKHOLE_ADDRESS = "0x0100000000000000000000000000000000000000";
+const MINT_ADDRESS = "0x0200000000000000000000000000000000000001"
+const CONTRACT_ALLOW_LIST_ADDRESS = "0x0200000000000000000000000000000000000000"
+const TX_ALLOW_LIST_ADDRESS = "0x0200000000000000000000000000000000000002"
+const BLACKHOLE_ADDRESS = "0x0100000000000000000000000000000000000000"
```
aaronbuchwald, line 9: Can we order these by the address?

core/state_processor_test.go
```diff
@@ -300,6 +301,71 @@ func TestStateProcessorErrors(t *testing.T) {
 			}
 		}
 	}
+
+	{
```
aaronbuchwald, line 305: Can we move this to a separate test function?

core/stateful_precompile_test.go
```diff
@@ -258,6 +258,239 @@ func TestContractDeployerAllowListRun(t *testing.T) {
 	}
 }
 
+func TestTransactionAllowListRun(t *testing.T) {
```
aaronbuchwald, line 261:
```suggestion
func TestTxAllowListRun(t *testing.T) {
```

plugin/evm/vm_test.go
```diff
@@ -2288,3 +2288,104 @@ func TestBuildAllowListActivationBlock(t *testing.T) {
...
+
+	// Check that address 0 is whitelisted and address 1 is not
+	role := precompile.GetTxAllowListStatus(genesisState, testEthAddrs[1])
+	if role != precompile.AllowListNoRole {
+		t.Fatalf("Expected allow list status to be set to no role: %s, but found: %s", precompile.AllowListNoRole, role)
+	}
+	role = precompile.GetTxAllowListStatus(genesisState, testEthAddrs[0])
+	if role != precompile.AllowListAdmin {
+		t.Fatalf("Expected allow list status to be set to no role: %s, but found: %s", precompile.AllowListAdmin, role)
+	}
```
aaronbuchwald, line 2333: nit: can we swap the order of these so that it checks address 0 and then address 1?

plugin/evm/vm_test.go
```diff
@@ -2288,3 +2288,104 @@ func TestBuildAllowListActivationBlock(t *testing.T) {
...
+
+	// Submit a successful transaction
+	tx := types.NewTransaction(uint64(0), testEthAddrs[0], new(big.Int).Mul(firstTxAmount, big.NewInt(3)), 21000, big.NewInt(testMinGasPrice*3), nil)
```
aaronbuchwald, line 2336: nit: can we rename these to tx0 and tx1 so that it corresponds to the address numbers?

aaronbuchwald, line 2336: Can we replace the amount argument in these txs with `big.NewInt(1)` instead of the more complicated multiplication? I think this is borrowing logic that as applicable from another test.

We should be able to do the same thing and replace `big.NewInt(testMinGasPrice*3)` with `big.NewInt(testMinGasPrice)`.

aaronbuchwald (commented):
Few small nits

contract-examples/hardhat.config.ts
dasconnor: done

core/state_processor_test.go
dasconnor: done

plugin/evm/vm_test.go
dasconnor: done

dasconnor: done

core/state_processor_test.go
```diff
@@ -302,6 +303,90 @@ func TestStateProcessorErrors(t *testing.T) {
 	}
 }
 
+// TestStateProcessorErrors tests the output from the 'core' errors
+// as defined in core/error.go. This error is generated when the
+// blockchain imports a bad block with a transaction from a
+// non-whitelisted address.
+func TestBadWhitelistedBlock(t *testing.T) {
```
aaronbuchwald, line 310: Nit: can we update the test name to: TestBadTxAllowListBlock and update the comment as well?

dasconnor: fixed

core/state_processor_test.go
```diff
@@ -302,6 +303,90 @@ func TestStateProcessorErrors(t *testing.T) {
 	}
 }
 
+// TestStateProcessorErrors tests the output from the 'core' errors
```
aaronbuchwald, line 306: nit: TestStateProcessorErrors -> TestBadTxAllowListBlock

The comment also mentions the errors in `core/errors.go` which aren't used here, so we should update that part of the comment too.

core/state_transition.go
```diff
@@ -307,9 +308,17 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
 	homestead := st.evm.ChainConfig().IsHomestead(st.evm.Context.BlockNumber)
 	istanbul := st.evm.ChainConfig().IsIstanbul(st.evm.Context.BlockNumber)
 	subnetEVM := st.evm.ChainConfig().IsSubnetEVM(st.evm.Context.Time)
+	isTxAllowList := st.evm.ChainConfig().IsTxAllowList(st.evm.Context.Time)
 
 	contractCreation := msg.To() == nil
 
+	if isTxAllowList {
```
ceyonur, line 315: hm should we do this check before buying gas (i.e in `preCheck`)?

core/state_transition.go
```diff
@@ -307,9 +308,17 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
...
+		txAllowListRole := precompile.GetTxAllowListStatus(st.state, sender.Address())
```
ceyonur, line 316: In the `contractDeployerAllowList` we checked tx.Origin. Should we check it here as well instead of sender?

core/state_transition.go
dasconnor: Thoughts @aaronbuchwald?

core/state_transition.go
dasconnor: is `sender.Address()` equivalent to `msg.sender`? If I understand correctly, tx.origin will always return the EOA that originated the TX and sender.Address will change to a contract if there is a nested call. We probably want to only whitelist EOAs, right? Don't want to need to make people also verify contracts? I think there a couple tests that might need modification if we switch, but we should be clear about intended behavior.

core/state_transition.go
aaronbuchwald: Ya let's move this into `preCheck` that's a much cleaner place to add this check. We should add to the comment above where `preCheck` is called within `TransitionDb()` as well.

To explain a bit:

We're still performing this check before we do any EVM execution takes place, so this doesn't present any DoS vulnerabilities where we have to throw something away after already doing work on it.

As for doing this before buying gas, I think the concern there would be that we return this error before modifying the staetdb in any way? For that, the error returned here means the entire transaction is invalid, not a VM related error, so the statedb won't be re-used with an incorrect modification made either way.

core/state_transition.go
aaronbuchwald: tx.Origin is the name of the variable on the `txContext` within the EVM, so for the contract deployer list, we check that since it is what is available. The `sender.Address()` is the same address as Connor said: the EOA address that originates the transaction.

The txContext is available here as well, but it seems more direct to access it in this case through `sender.Address()`, so I think we can leave this as is.

README.md
```diff
@@ -286,6 +286,97 @@ If you call `readAllowList(addr)` then you can read the current role of `addr`,
...
+If you attempt to add an `Allowed` and you are not an `Admin`, you will see
+something like:
+![admin fail](./imgs/admin_fail.png)
+
+If you attempt to submit a transaction but you are not an `Admin` or not
+`Allowed`, you will see something like: `cannot issue transaction from non-allow listed address`
+
+The allow list has three roles: `None`, `Allowed`, and `Admin`.
+
+If you call `readAllowList(addr)` then you can read the current role of `addr`, which will return a uint256 with a value of 0, 1, or 2, corresponding to the roles `None`, `Allowed`, and `Admin` respectively.
```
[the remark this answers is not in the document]
dasconnor: change made

contract-examples/hardhat.config.ts
```diff
@@ -175,5 +213,22 @@ export default {
         "0x750839e9dbbd2a0910efe40f50b2f3b2f2f59f5580bb4b83bd8c1201cf9a010a"
       ]
     },
+    subnet: {
```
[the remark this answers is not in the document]
dasconnor: In a future commit, I'm using the subnet network to drive tests. I actually need to modify it dynamically, so I'd like to keep it separate. I added a comment to address this.

contract-examples/test/ExampleTxAllowList.ts
```diff
@@ -0,0 +1,189 @@
...
+  let noRole: SignerWithAddress
+  before(async function () {
+    admin = await ethers.getSigner(adminAddress)
+    const contractF: ContractFactory = await ethers.getContractFactory("ExampleTxAllowList", { signer: admin })
+    contract = await contractF.deploy()
+    await contract.deployed()
+    const contractAddress: string = contract.address
+    console.log(`Contract deployed to: ${contractAddress}`)
+
+      ;[, allowed, noRole] = await ethers.getSigners()
```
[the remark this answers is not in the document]
dasconnor: yeah, `getSigners()` returns an array, this just extracts the second and third entry in that array

core/state_transition.go
dasconnor: I moved this, let me know if I did it correctly

core/state_transition.go
dasconnor: When I moved to precheck, I switched to st.msg.From() since that's what the rest of the function used. I think that's acceptable?

core/state_transition.go
ceyonur: it seems good to me

core/state_transition.go
ceyonur: I was just under impression `sender.Address()` was equal to `msg.sender` but seems it was an EOA. it actually verifies that under: 
https://github.com/ava-labs/subnet-evm/blob/f98dcad25f9f2ca08f92e5b50e11a1e19b4f0392/core/state_transition.go#L241-L245
Thanks for clarification!

contract-examples/test/ExampleTxAllowList.ts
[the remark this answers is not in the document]
ceyonur: looks a little bit weird but it might be just my symmetry OCD haha

core/state_transition.go
```diff
@@ -246,6 +247,15 @@ func (st *StateTransition) preCheck() error {
 		if vm.IsProhibited(st.msg.From()) {
 			return fmt.Errorf("%w: address %v", vmerrs.ErrAddrProhibited, st.msg.From())
 		}
+
+		// Check that the sender is on the tx allow list if enabled
+		isTxAllowList := st.evm.ChainConfig().IsTxAllowList(st.evm.Context.Time)
```
aaronbuchwald, line 252: Perfect 🙌

contract-examples/test/ExampleTxAllowList.ts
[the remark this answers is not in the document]
dasconnor: It shouldn't have the indentation, but my editor wouldn't let me remove it lol

ceyonur (approved):
👍

dasconnor (approved):

PR #81, opened 2022-04-18
[AV-1855] Simplify build script to place binary in avalanchego plugin directory

aaronbuchwald:
This PR simplifies the build script to take in either zero or one parameters, where one argument specifies the location to build the plugin binary and zero arguments will attempt to place the plugin binary in the default location within the default build directory of AvalancheGo in the gopath.


scripts/build.sh
```diff
@@ -38,8 +38,10 @@ source "$SUBNET_EVM_PATH"/scripts/constants.sh
 
 if [[ $# -eq 1 ]]; then
     binary_path=$1
+elif [[ $# -eq 0 ]]; then
+    binary_path="$GOPATH/src/github.com/ava-labs/avalanchego/build/plugins/srEXiWaHuhNyGwPUi444Tu47ZEDwxTWrbQiuD7FmgSAQ6X7Dy"
```
ceyonur, line 42: We should print this binary path when built with default paths. If someone does not use this path for their avalanchego, then this may be lost. Personally I think the simpler way is to put that binary under subnet-evm folder i.e `./build/srEXiWaHuhNyGwPUi444Tu47ZEDwxTWrbQiuD7FmgSAQ6X7Dy`. Plus this is probably safer than assuming that avalanchego path exists.

aaronbuchwald: Will add a printout of the path. People can still pass in a binary path if they want, but I'd much prefer to have this functionality to make it easier to use with the expected setup.

darioush: Should we try to infer the avalanchego path as the parent directory of the current repo instead ie, `../avalanchego/`?

aaronbuchwald: I'd prefer to only take advantage of this happy path if avalanchego is in the expected location. If the expected directory is not there, then we would want to error either way and it currently already provides a readable error message in this case, so I don't think this is necessary. If it's not there, users will need to specify the binary path.

PR #82, opened 2022-04-19
*: add e2e tests, replace "runner" with network-runner

gyuho:
[asciicast](https://asciinema.org/a/488625)

Screen Shot 2022-04-19 at 12 18 27 PM

Screen Shot 2022-04-19 at 12 17 58 PM


dasconnor (commented):
looks good, just one small thing

vinay2sonu:
I am not a coder but what if it looks like something techy, good luck with your hard work.

tests/e2e/e2e_test.go
```diff
@@ -0,0 +1,277 @@
...
+// See the file LICENSE for licensing terms.
+
+// e2e implements the e2e tests.
+package e2e_test
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"io/ioutil"
```
danlaine, line 11: From https://pkg.go.dev/io/ioutil:
`As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code. See the specific function documentation for details.`

tests/e2e/e2e_test.go
```diff
@@ -0,0 +1,276 @@
...
+	ginkgo.It("can TODO", func() {
+		if mode != "test" {
+			ginkgo.Skip("mode is not 'test'; skipping...")
+		}
+
+		// TODO: e2e tests specific for subnet-evm
+	})
+})
+
+func createDefaultCtx() (context.Context, context.CancelFunc) {
```
aaronbuchwald, line 233: I think adding a function like this obscures what is actually happening. Since this is a 1 liner anyways, can we just remove this in favor of creating a context with a 2 minute timeout wherever this is called to make it more clear what the default context is?

tests/e2e/e2e_test.go
```diff
@@ -0,0 +1,276 @@
...
+
+func (ci clusterInfo) Save(p string) error {
+	ob, err := yaml.Marshal(ci)
+	if err != nil {
+		return err
+	}
+	return os.WriteFile(p, ob, fsModeWrite)
+}
+
+func getVMID(vmName string) (ids.ID, error) {
```
aaronbuchwald, line 269: Can we add this as a utility within AvalancheGo? We should probably use this as a standard practice to create our VMIDs instead of having a private helper function within this package.

tests/e2e/e2e_test.go
```diff
@@ -0,0 +1,276 @@
...
+	gomega.Expect(ci.Save(outputPath)).Should(gomega.BeNil())
+
+	b, err := os.ReadFile(outputPath)
+	gomega.Expect(err).Should(gomega.BeNil())
+	outf("\n{{blue}}$ cat %s:{{/}}\n%s\n", outputPath, string(b))
+})
+
+var _ = ginkgo.AfterSuite(func() {
+	switch mode {
+	case "test":
```
aaronbuchwald, line 208: can we use constants for `test` / `run` here?

Also, I don't think we should be relying on the test suite as the way to start a local network without shutting it down by passing in a `mode` parameter in this way.

gyuho: Yeah we can remove later when we converge into a single CLI tool. This is to keep the existing feature in runner script :)

tests/e2e/e2e_test.go
gyuho: Left as TODO, will add a util to avalanchego and import here in the following PR.

ceyonur (approved):

PR #121, opened 2022-06-01
init fee config precompile

ceyonur:
adds fee config manager precompile with 2 main functions:
`setFeeConfig(uint256,uint256,uint256,uint256,uint256,uint256,uint256,uint256)` and 
`getFeeConfig returns (uint256,uint256,uint256,uint256,uint256,uint256,uint256,uint256)`
these inputs/output fields represents the fee config struct in order, as:
```
commontype.FeeConfig{
		GasLimit,              
		TargetBlockRate,         
		MinBaseFee,            
		TargetGas,           
		BaseFeeChangeDenominator,
		MinBlockGasCost,      
		MaxBlockGasCost,      
		BlockGasCostStep,         
	}
```
when a fee config is set, it determines next block's fee config. Fee configs are activated in bulk, it means that it's not possible to change only one field. 

This PR also adds a Fee Config Provider. If the precompile is not activated OR if the precompile contract storage is empty OR if all fields are given 0s, then config provider returns the fee config given in genesis (as a part of chain config). Otherwise it reads contract state and returns fee config in the precompile contract.


precompile/fee_config_manager.sol
```diff
@@ -0,0 +1,29 @@
...
+
+interface FeeConfigManagerInterface {
+    // Set [addr] to have the admin role over the minter list
+    function setAdmin(address addr) external;
+
+    // Set [addr] to be enabled on the minter list
+    function setEnabled(address addr) external;
+
+    // Set [addr] to have no role over the minter list
+    function setNone(address addr) external;
```
aaronbuchwald, line 16: can we update these comments to reflect that these roles are for the fee config manager?

precompile/params.go
```diff
@@ -9,8 +9,10 @@ import "github.com/ethereum/go-ethereum/common"
 const (
 	ModifyAllowListGasCost = 20_000
 	ReadAllowListGasCost   = 5_000
-
-	MintGasCost = 30_000
+	MintGasCost            = 30_000
+	// TODO: decide on gas costs
+	SetFeeConfigGasCost = 30_000
```
aaronbuchwald, line 14: Can we increase the gas cost to match 20k per state slot that we need to write to perform this function?

aaronbuchwald: Same comment for read to base this off of the number of state slots that we need to read to fulfill the request

precompile/fee_config_manager.go
```diff
@@ -0,0 +1,230 @@
...
+			return fmt.Errorf("unknown field key %d", i)
+		}
+		stateDB.SetState(FeeConfigManagerAddress, common.Hash{byte(i)}, hashInput)
+	}
+	return nil
+}
+
+// createMintNativeCoin checks if the caller is permissioned for minting operation.
+// The execution function parses the [input] into native coin amount and receiver address.
+func createSetFeeConfig(accessibleState PrecompileAccessibleState, caller common.Address, addr common.Address, input []byte, suppliedGas uint64, readOnly bool) (ret []byte, remainingGas uint64, err error) {
```
aaronbuchwald, line 191: Can we change this function name to setFeeConfig? The reason it is create for the allow list functionality is that we are actually creating the function that will be used, but this function is just used directly.

precompile/fee_config_manager.go
```diff
@@ -0,0 +1,268 @@
...
+	targetGasKey
+	baseFeeChangeDenominatorKey
+	minBlockGasCostKey
+	maxBlockGasCostKey
+	blockGasCostStepKey
+
+	minFeeConfigFieldKey = gasLimitKey
+	maxFeeConfigFieldKey = blockGasCostStepKey
+
+	numFeeConfigField = (maxFeeConfigFieldKey - minFeeConfigFieldKey + 1)
```
patrick-ogrady, line 30: nit: why not just add `_numFeeConfigField` as the last item in the `iota`?

precompile/fee_config_manager.go
```diff
@@ -0,0 +1,268 @@
...
+func GetStoredFeeConfig(stateDB StateDB) (commontype.FeeConfig, bool) {
+	if !stateDB.Exist(FeeConfigManagerAddress) {
+		return commontype.FeeConfig{}, false
+	}
+	feeConfig := commontype.FeeConfig{}
+	isAllZeros := true
+	for i := minFeeConfigFieldKey; i <= maxFeeConfigFieldKey; i++ {
+		val := stateDB.GetState(FeeConfigManagerAddress, common.Hash{byte(i)})
+		// even if one of these vals are non-zero means that this state exist
+		if isAllZeros && val != (common.Hash{0}) {
```
patrick-ogrady, line 145: nit: isn't `common.Hash{}` already all zeros?

precompile/fee_config_manager.go
```diff
@@ -0,0 +1,268 @@
...
+			panic("unknown key")
+		}
+	}
+	if isAllZeros {
+		return commontype.EmptyFeeConfig, false
+	}
+	return feeConfig, true
+}
+
+func StoreFeeConfig(stateDB StateDB, feeConfig commontype.FeeConfig) {
```
patrick-ogrady, line 175: nit: should we verify the byte size of each val in feeConifg is less than 32 bytes? (otherwise common.BigToHash crops left)

precompile/fee_config_manager.go
```diff
@@ -0,0 +1,268 @@
...
+// PackSetFeeConfig packs [feeConfig] with the selector into the appropriate arguments for setting fee config operations.
+func PackSetFeeConfig(feeConfig commontype.FeeConfig) ([]byte, error) {
+	// function selector (4 bytes) + input(feeConfig)
+	return packHelper(feeConfig, true)
+}
+
+func packHelper(feeConfig commontype.FeeConfig, useSelector bool) ([]byte, error) {
+	fullLen := feeConfigInputLen
+	packed := [][]byte{
+		feeConfig.GasLimit.FillBytes(make([]byte, 32)),
```
patrick-ogrady, line 100: nit: use `common.HashLength`

precompile/fee_config_manager.go
```diff
@@ -0,0 +1,268 @@
...
+
+// PackSetFeeConfig packs [feeConfig] with the selector into the appropriate arguments for setting fee config operations.
+func PackSetFeeConfig(feeConfig commontype.FeeConfig) ([]byte, error) {
+	// function selector (4 bytes) + input(feeConfig)
+	return packHelper(feeConfig, true)
+}
+
+func packHelper(feeConfig commontype.FeeConfig, useSelector bool) ([]byte, error) {
+	fullLen := feeConfigInputLen
+	packed := [][]byte{
```
patrick-ogrady, line 99: nit: may want to create utility to take an array of big.Int and perform this transformation.

precompile/fee_config_manager.go
```diff
@@ -0,0 +1,268 @@
...
+}
+
+// UnpackFeeConfigInput attempts to unpack [input] into the arguments to the fee config precompile
+// assumes that [input] does not include selector (omits first 4 bytes in PackSetFeeConfigInput)
+func UnpackFeeConfigInput(input []byte) (commontype.FeeConfig, error) {
+	if len(input) != feeConfigInputLen {
+		return commontype.FeeConfig{}, fmt.Errorf("invalid input length for fee config input: %d", len(input))
+	}
+	return commontype.FeeConfig{
+		GasLimit:                 new(big.Int).SetBytes(returnPackedElement(input, 0)),
```
patrick-ogrady, line 124: nit: can't you use consts at top of file for this?

precompile/fee_config_manager.go
```diff
@@ -0,0 +1,268 @@
...
+		}
+	}
+	if isAllZeros {
+		return commontype.EmptyFeeConfig, false
+	}
+	return feeConfig, true
+}
+
+func StoreFeeConfig(stateDB StateDB, feeConfig commontype.FeeConfig) {
+	for i := minFeeConfigFieldKey; i <= maxFeeConfigFieldKey; i++ {
```
patrick-ogrady, line 176: nit: use 0 here for i and then use `_max` trick from above

precompile/fee_config_manager.go
```diff
@@ -0,0 +1,268 @@
...
+	if err != nil {
+		return nil, remainingGas, err
+	}
+
+	stateDB := accessibleState.GetStateDB()
+	// Verify that the caller is in the allow list and therefore has the right to modify it
+	callerStatus := getAllowListStatus(stateDB, FeeConfigManagerAddress, caller)
+	if !callerStatus.IsEnabled() {
+		return nil, remainingGas, fmt.Errorf("%w: %s", ErrCannotChangeFee, caller)
+	}
```
patrick-ogrady, line 223: nit: we should create a standard func that performs 205-223

precompile/fee_config_manager.go
```diff
@@ -0,0 +1,268 @@
...
+	output, err := PackFeeConfig(feeConfig)
+
+	// Return an empty output and the remaining gas
+	return output, remainingGas, err
+}
+
+// createFeeConfigManagerPrecompile returns a StatefulPrecompiledContract with R/W control of an allow list at [precompileAddr],
+// and a storage setter for fee configs.
+func createFeeConfigManagerPrecompile(precompileAddr common.Address) StatefulPrecompiledContract {
+	setAdmin := newStatefulPrecompileFunction(setAdminSignature, createAllowListRoleSetter(precompileAddr, AllowListAdmin))
```
patrick-ogrady, line 257: nit: make this a utility function

precompile/fee_config_manager.sol
```diff
@@ -0,0 +1,44 @@
+// (c) 2022-2023, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+// SPDX-License-Identifier: MIT
+
+pragma solidity >=0.8.0;
+
+interface FeeConfigManagerInterface {
```
patrick-ogrady, line 8: nit: is there some abstraction in solidity that allows us to inherit the `Admin` interface? like `is AllowList`?

precompile/contract_native_minter.go
```diff
@@ -73,14 +74,14 @@ func UnpackMintInput(input []byte) (common.Address, *big.Int, error) {
 	if len(input) != mintInputLen {
 		return common.Address{}, nil, fmt.Errorf("invalid input length for minting: %d", len(input))
 	}
-	to := common.BytesToAddress(input[:common.HashLength])
```
patrick-ogrady, line 76: ❤️

params/config.go
```diff
@@ -113,54 +114,13 @@ type ChainConfig struct {
...
-	MaxBlockGasCost *big.Int `json:"maxBlockGasCost,omitempty"`
-	// BlockGasCostStep determines how much to increase/decrease the block gas cost depending on the amount of time elapsed since the previous block.
-	// If the block is produced at the target rate, the block gas cost will stay the same as the block gas cost for the parent block.
-	// If it is produced faster/slower, the block gas cost will be increased/decreased by the step value for each second faster/slower than the target
-	// block rate accordingly.
-	// Note: if the BlockGasCostStep is set to a very large number, it effectively requires block production to go no faster than the TargetBlockRate.
-	//
-	// Ex: if a block is produced two seconds faster than the target block rate, the block gas cost will increase by 2 * BlockGasCostStep.
-	BlockGasCostStep *big.Int `json:"blockGasCostStep,omitempty"`
+	FeeManagerConfig                precompile.FeeConfigManagerConfig          `json:"feeManagerConfig,omitempty"`
```
patrick-ogrady, line 123: nit: add inline comment

eth/api_backend.go
```diff
@@ -431,3 +436,7 @@ func (b *EthAPIBackend) StateAtTransaction(ctx context.Context, block *types.Blo
 func (b *EthAPIBackend) MinRequiredTip(ctx context.Context, header *types.Header) (*big.Int, error) {
 	return dummy.MinRequiredTip(b.ChainConfig(), header)
 }
+
+func (b *EthAPIBackend) GetFeeConfig(parent *types.Header) (commontype.FeeConfig, error) {
```
patrick-ogrady, line 440: nit: why have this and `GetFeeConfigAt`?

eth/gasprice/gasprice.go
```diff
@@ -215,6 +240,10 @@ func (oracle *Oracle) EstimateBaseFee(ctx context.Context) (*big.Int, error) {
 	}
 
 	baseFee = math.BigMin(baseFee, nextBaseFee)
+
+	if oracle.backend.ChainConfig().IsFeeConfigManager(new(big.Int).SetUint64(block.Header().Time)) {
```
patrick-ogrady, line 244: We already do this within `estimateNextBaseFee`: https://github.com/ava-labs/subnet-evm/blob/0d8ef8a7ed250a83ed34a3dc9baec165656cb4aa/consensus/dummy/dynamic_fees.go#L107

eth/gasprice/gasprice.go
```diff
@@ -262,6 +297,10 @@ func (oracle *Oracle) SuggestPrice(ctx context.Context) (*big.Int, error) {
 		baseFee = math.BigMin(baseFee, nextBaseFee)
 	}
 
+	if oracle.backend.ChainConfig().IsFeeConfigManager(new(big.Int).SetUint64(block.Header().Time)) {
```
patrick-ogrady, line 300: nit: see above

plugin/evm/vm.go
```diff
@@ -273,7 +274,7 @@ func (vm *VM) Initialize(
 		g.Config = params.SubnetEVMDefaultChainConfig
 	}
 
-	if g.Config.FeeConfig == (params.FeeConfig{}) {
+	if g.Config.FeeConfig == commontype.EmptyFeeConfig {
```
patrick-ogrady, line 277: Does this work with `big.Int` == 0 values? I would think they would need to be nil.

precompile/fee_config_manager.sol
patrick-ogrady, line 8: Why don't we match the interface with what is in contract-examples?

core/tx_pool.go
```diff
@@ -1295,10 +1297,7 @@ func (pool *TxPool) runReorg(done chan struct{}, reset *txpoolResetRequest, dirt
 	if reset != nil {
 		pool.demoteUnexecutables()
 		if reset.newHead != nil && pool.chainconfig.IsSubnetEVM(new(big.Int).SetUint64(reset.newHead.Time)) {
-			_, baseFeeEstimate, err := dummy.EstimateNextBaseFee(pool.chainconfig, reset.newHead, uint64(time.Now().Unix()))
-			if err == nil {
-				pool.priced.SetBaseFee(baseFeeEstimate)
-			}
+			pool.updateBaseFeeAt(reset.newHead)
```
darioush, line 1300: should we check `err` or explicitly ignore it?

core/tx_pool.go
```diff
@@ -1420,6 +1419,15 @@ func (pool *TxPool) reset(oldHead, newHead *types.Header) {
 	pool.pendingNonces = newTxNoncer(statedb)
 	pool.currentMaxGas = newHead.GasLimit
 
+	if pool.chainconfig.IsFeeConfigManager(new(big.Int).SetUint64(newHead.Time)) {
+		feeConfig, err := pool.chain.GetFeeConfigAt(newHead)
+		if err != nil {
+			log.Error("Failed to get fee config state", "err", err, "root", newHead.Root)
+			return
+		}
+		pool.minimumFee = feeConfig.MinBaseFee
```
darioush, line 1428: can we add a comment why we update this?

plugin/evm/block_verification.go
```diff
@@ -142,13 +143,18 @@ func (blockValidatorSubnetEVM) SyntacticVerify(b *Block) error {
...
 		)
 	}
-	expectedGas := b.vm.chainConfig.GetFeeConfig().GasLimit.Uint64()
-	if ethHeader.GasLimit != expectedGas {
-		return fmt.Errorf(
-			"expected gas limit to be %d in subnetEVM but got %d",
-			expectedGas, ethHeader.GasLimit,
-		)
+
+	// if fee config manager is, we cannot check state at this point; so skip checking gas limit.
```
darioush, line 147:
```suggestion
	// if fee config manager is enabled, FeeConfig depends on state. State is not available here, so skip checking the gas limit.
```

commontype/fee_config.go
```diff
@@ -0,0 +1,61 @@
+package commontype
```
aaronbuchwald, line 1: nit: add license

commontype/fee_config.go
```diff
@@ -0,0 +1,61 @@
...
+
+// FeeConfig specifies the parameters for the dynamic fee algorithm, which determines the gas limit, base fee, and block gas cost of blocks
+// on the network.
+//
+// The dynamic fee algorithm simply increases fees when the network is operating at a utilization level above the target and decreases fees
+// when the network is operating at a utilization level below the target.
+// This struct is used by params.Config and precompile.FeeConfigManager
+// any modification of this struct has direct affect on the precompiled contract
+// and changes should be carefully handled in the precompiled contract code.
+type FeeConfig struct {
```
aaronbuchwald, line 13: did we need to move this to a separate package to avoid a circular dependency? I'm not a big fan of having a commonytype package with just this in it.

commontype/fee_config.go
```diff
@@ -0,0 +1,61 @@
...
+	// block rate accordingly.
+	// Note: if the BlockGasCostStep is set to a very large number, it effectively requires block production to go no faster than the TargetBlockRate.
+	//
+	// Ex: if a block is produced two seconds faster than the target block rate, the block gas cost will increase by 2 * BlockGasCostStep.
+	BlockGasCostStep *big.Int `json:"blockGasCostStep,omitempty"`
+}
+
+var EmptyFeeConfig = FeeConfig{}
+
+var ZeroFeeConfig = FeeConfig{
```
aaronbuchwald, line 52: can we add comments to both of these and put them in a single var block?

contract-examples/contracts/ExampleFeeManager.sol
```diff
@@ -0,0 +1,55 @@
...
+
+  function addAllowed(address addr) public onlyOwner {
+    feeManager.setEnabled(addr);
+  }
+
+  function revoke(address addr) public onlyOwner {
+    require(_msgSender() != addr, "cannot revoke own role");
+    feeManager.setNone(addr);
+  }
+}
```
aaronbuchwald, line 55: Should there be a function to set and retrieve the fee config within this example contract?

contract-examples/contracts/IFeeManager.sol
```diff
@@ -0,0 +1,32 @@
+//SPDX-License-Identifier: MIT
+pragma solidity >=0.6.2;
```
aaronbuchwald, line 2: Can we update this and everything else to min version of v0.8.15?

aaronbuchwald: Alternatively, let's move this to at least v0.8.0 and have a separate PR to bump everything to v0.8.15.

core/blockchain_reader.go
```diff
@@ -343,3 +348,31 @@ func (bc *BlockChain) SubscribeAcceptedLogsEvent(ch chan<- []*types.Log) event.S
 func (bc *BlockChain) SubscribeAcceptedTransactionEvent(ch chan<- NewTxsEvent) event.Subscription {
 	return bc.scope.Track(bc.txAcceptedFeed.Subscribe(ch))
 }
+
+func (bc *BlockChain) GetFeeConfigAt(parent *types.Header) (commontype.FeeConfig, error) {
```
aaronbuchwald, line 352: nit: `GetFeeConfigAt` -> `GetFeeConfig`

eth/gasprice/gasprice_test.go
```diff
@@ -90,6 +91,10 @@ func (b *testBackend) SubscribeChainHeadEvent(ch chan<- core.ChainHeadEvent) eve
 	return nil
 }
 
+func (b *testBackend) GetFeeConfigAt(parent *types.Header) (commontype.FeeConfig, error) {
```
aaronbuchwald, line 94: can we remove this in favor of `GetFeeConfig` only

internal/ethapi/api.go
```diff
@@ -608,6 +609,22 @@ func (s *PublicBlockChainAPI) ChainId() (*hexutil.Big, error) {
 	return nil, fmt.Errorf("chain not synced beyond EIP-155 replay-protection fork block")
 }
 
+func (s *PublicBlockChainAPI) FeeConfig(ctx context.Context, blockNrOrHash *rpc.BlockNumberOrHash) (*commontype.FeeConfig, error) {
```
aaronbuchwald, line 612: Nice!

params/config.go
```diff
@@ -411,7 +368,12 @@ func (c *ChainConfig) checkCompatible(newcfg *ChainConfig, headHeight *big.Int,
 
 	// Check that the configuration of the optional stateful precompiles is compatible.
 	if isForkIncompatible(c.TxAllowListConfig.Timestamp(), newcfg.TxAllowListConfig.Timestamp(), headTimestamp) {
-		return newCompatError("AllowList fork block timestamp", c.TxAllowListConfig.Timestamp(), newcfg.TxAllowListConfig.Timestamp())
+		return newCompatError("TxAllowList fork block timestamp", c.TxAllowListConfig.Timestamp(), newcfg.TxAllowListConfig.Timestamp())
+	}
+
+	// Check that the configuration of the optional stateful precompiles is compatible.
```
aaronbuchwald, line 374: Looks like we use the exact same comment for each of the stateful precompiles, can we make each of these comments specific to the relevant precompile?

plugin/evm/gasprice_update.go
```diff
@@ -42,7 +42,7 @@ func (gpu *gasPriceUpdater) start() {
 	if disabled := gpu.handleUpdate(gpu.setter.SetGasPrice, gpu.chainConfig.SubnetEVMTimestamp, big.NewInt(0)); disabled {
 		return
 	}
-	minBaseFee := gpu.chainConfig.GetFeeConfig().MinBaseFee
+	minBaseFee := gpu.chainConfig.FeeConfig.MinBaseFee
```
aaronbuchwald, line 45: should we get the latest min base fee here instead?

precompile/allow_list.go
```diff
@@ -157,13 +157,14 @@ func createAllowListRoleSetter(precompileAddr common.Address, role AllowListRole
 			return nil, remainingGas, vmerrs.ErrWriteProtection
 		}
 
+		stateDB := evm.GetStateDB()
```
aaronbuchwald, line 160: good call!

precompile/fee_config_manager.go
```diff
@@ -0,0 +1,268 @@
...
+	"math/big"
+
+	"github.com/ava-labs/subnet-evm/commontype"
+	"github.com/ava-labs/subnet-evm/vmerrs"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+const (
+	// must preserve order of these fields
+	gasLimitKey = iota
```
aaronbuchwald, line 18: can we start with `iota + 1` ?

precompile/utils.go
```diff
@@ -31,3 +32,52 @@ func deductGas(suppliedGas uint64, requiredGas uint64) (uint64, error) {
...
+	bytesHashes, err := packOrderedHashes(input[1:], hashLen)
+	if err != nil {
+		return nil, err
+	}
+	copy(buf[selectorLen:], bytesHashes)
+
+	return buf, nil
+}
+
+func packOrderedHashes(input [][]byte, fullLength int) ([]byte, error) {
```
aaronbuchwald, line 64: can we add a comment here and to the function `returnPackedElement` ?

precompile/fee_config_manager.go
ceyonur: it's specifically checking all fields to ensure that all slots contain zeros. This can be used as a `reset`, i.e `setFeeConfig(0,0,0,0,0,0,0,0) == reset()` which reverts back to the genesis config.

precompile/fee_config_manager.go
ceyonur: it's actually safe since this is being used after we unpack it with HashLength from inputs:
https://github.com/ava-labs/subnet-evm/blob/fee-config-precompile/precompile/fee_config_manager.go#L213

but wouldn't hurt adding another check

precompile/fee_config_manager.sol
ceyonur: yes we use it in. https://github.com/ava-labs/subnet-evm/blob/fee-config-precompile/contract-examples/contracts/IFeeManager.sol#L5

I think in these files we wanted to explicitly show which functions are available for precompiled contracts.

eth/gasprice/gasprice.go
ceyonur: yes but if we increase `minBaseFee` and `oracle.suggestDynamicFees(ctx)` reports a lower fee (since previous blocks has lower base fees) it will resolved to `nextBaseFee` in the 
https://github.com/ava-labs/subnet-evm/blob/0d8ef8a7ed250a83ed34a3dc9baec165656cb4aa/eth/gasprice/gasprice.go#L297
 which is `nextBaseFee` > minBaseFee.

This is not probably the ideal approach though. We should be changing `oracle.estimateNextBaseFee` to adjust this fee config changes.

plugin/evm/vm.go
ceyonur: no, but when genesis JSON is unmarshalled without the `FeeConfig`, it returns EmptyFeeConfig.

core/tx_pool.go
ceyonur: added a Error log

commontype/fee_config.go
ceyonur: yes :/. I've tried other packages but no luck :/

contract-examples/contracts/ExampleFeeManager.sol
ceyonur: yes this was under WIP, adding them now

contract-examples/hardhat.config.ts
```diff
@@ -33,7 +32,7 @@ export default {
   networks: {
     local: {
       //"http://{ip}:{port}/ext/bc/{chainID}/rpc
-      url: "http://127.0.0.1:9650/ext/bc/dRTfPJh4jEaRZoGkPc7xreeYbDGBrGWRV48WAYVyUgApsmzGo/rpc",
+      url: "http://127.0.0.1:9650/ext/bc/bWeTNmehHHLWZH2jMsCCnqdpMfAXWgb7XEkWcVjLd6eFaCM4Y/rpc",
```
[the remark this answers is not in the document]
ceyonur: local testing. blockchain ID changes at each different run. reverting it now

eth/gasprice/gasprice_test.go
ceyonur: there is also a `getFeeConfig` in the precompile, so thought it would be cleaner with `GetFeeConfigAt` (which also indicates it changes with different headers)

core/blockchain_reader.go
aaronbuchwald, line 352: can we add a comment and make sure the desired behavior is well defined in the comment? Ie. do we assume that the fee config must be defined in the state if it's been enabled or do we fall back to the original config if it's not present in the state?

precompile/fee_config_manager.go
```diff
@@ -0,0 +1,268 @@
...
+
+	// 8 fields in FeeConfig struct
+	feeConfigInputLen = common.HashLength * numFeeConfigField
+)
+
+// FeeConfigManagerConfig wraps [AllowListConfig] and uses it to implement the StatefulPrecompileConfig
+// interface while adding in the contract deployer specific precompile address.
+type FeeConfigManagerConfig struct {
+	AllowListConfig
+}
```
aaronbuchwald, line 52: Hmm I was thinking that we would include an updated fee config here, but this seems cleaner to just have the admin user send a transaction to update it.

precompile/fee_config_manager.go
```diff
@@ -0,0 +1,268 @@
...
+// fee config manager list. assumes [role] has already been verified as valid.
+func SetFeeConfigManagerStatus(stateDB StateDB, address common.Address, role AllowListRole) {
+	setAllowListRole(stateDB, FeeConfigManagerAddress, address, role)
+}
+
+// PackGetFeeConfigInput packs the getFeeConfig signature
+func PackGetFeeConfigInput() []byte {
+	return getFeeConfigSignature
+}
+
```
aaronbuchwald, line 84: Do we need a function for this?

precompile/fee_config_manager.go
aaronbuchwald: I think Patrick's original comment meant to say that we can use `common.Hash{}` instead of `common.Hash{0}` here

eth/gasprice/gasprice.go
```diff
@@ -249,10 +270,24 @@ func (oracle *Oracle) SuggestPrice(ctx context.Context) (*big.Int, error) {
...
+	feeConfig, err := oracle.backend.GetFeeConfigAt(block.Header())
+	if err != nil {
+		return nil, err
+	}
+
 	// We calculate the [nextBaseFee] if a block were to be produced immediately.
 	// If [nextBaseFee] is lower than the estimate from sampling, then we return it
 	// to prevent returning an incorrectly high fee when the network is quiescent.
-	nextBaseFee, err := oracle.estimateNextBaseFee(ctx)
+	nextBaseFee, err := oracle.estimateNextBaseFee(block, feeConfig)
```
aaronbuchwald, line 290: In both places we call `estimateNextBaseFee` we fetch the latest block and the fee config at that block, so we should move both of these calls into `estimateNextBaseFee` and change from passing in these two params to passing in the `ctx`.

plugin/evm/gasprice_update.go
aaronbuchwald: never mind, this is correct

core/blockchain_reader.go
```diff
@@ -343,3 +348,31 @@ func (bc *BlockChain) SubscribeAcceptedLogsEvent(ch chan<- []*types.Log) event.S
...
+		return config.FeeConfig, nil
+	}
+
+	if feeConfig, ok := bc.feeConfigCache.Get(parent.Root); ok {
+		return feeConfig.(commontype.FeeConfig), nil
+	}
+
+	stateDB, err := bc.StateAt(parent.Root)
+	if err != nil {
+		log.Debug("feeConfigManager is activated, but no state found; returning err", "err", err)
```
aaronbuchwald, line 366: Unless we require that the fee config is set when the precompile activates, this is probably a little excessive as well.

aaronbuchwald: What do you think of passing in the state instead of the parent header and then removing the error from the return value? I think this would make it easier to see exactly what is considered an error (retrieving the state) as opposed to retrieving the fee config from the state.

precompile/fee_config_manager.go
```diff
@@ -0,0 +1,268 @@
...
+		MinBaseFee:               new(big.Int).SetBytes(returnPackedElement(input, 2)),
+		TargetGas:                new(big.Int).SetBytes(returnPackedElement(input, 3)),
+		BaseFeeChangeDenominator: new(big.Int).SetBytes(returnPackedElement(input, 4)),
+		MinBlockGasCost:          new(big.Int).SetBytes(returnPackedElement(input, 5)),
+		MaxBlockGasCost:          new(big.Int).SetBytes(returnPackedElement(input, 6)),
+		BlockGasCostStep:         new(big.Int).SetBytes(returnPackedElement(input, 7)),
+	}, nil
+}
+
+// GetStoredFeeConfig returns fee config from contract storage in given state
```
aaronbuchwald, line 135: Can we add a description of the returned boolean to this comment?

precompile/fee_config_manager.go
aaronbuchwald, line 175: Can we add a comment here?

precompile/fee_config_manager.go
```diff
@@ -0,0 +1,268 @@
...
+		return nil, remainingGas, fmt.Errorf("%w: %s", ErrCannotChangeFee, caller)
+	}
+
+	StoreFeeConfig(stateDB, feeConfig)
+
+	// Return an empty output and the remaining gas
+	return []byte{}, remainingGas, nil
+}
+
+// getFeeConfig returns the stored fee config as an output
```
aaronbuchwald, line 231: nit: can we update the comment here to say that this is the execution function for the getFeeConfig call instead of just saying it returns an output

precompile/fee_config_manager.go
```diff
@@ -0,0 +1,268 @@
...
+		// we cannot pack empty fee config, instead use zero fee config
+		output, err := PackFeeConfig(commontype.ZeroFeeConfig)
+		// should never fail to packing an empty fee config
+		if err != nil {
+			panic(err)
+		}
+		return output, remainingGas, nil
+	}
+
+	output, err := PackFeeConfig(feeConfig)
```
aaronbuchwald, line 248: nit: can we add an if condition here to return `nil, remainingGas, err` if the error is non-nil instead of returning the output regardless below?

precompile/fee_config_manager.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+
+	stateDB := accessibleState.GetStateDB()
+	// Verify that the caller is in the allow list and therefore has the right to modify it
+	callerStatus := getAllowListStatus(stateDB, FeeConfigManagerAddress, caller)
+	if !callerStatus.IsEnabled() {
+		return nil, remainingGas, fmt.Errorf("%w: %s", ErrCannotChangeFee, caller)
+	}
+
+	StoreFeeConfig(stateDB, feeConfig)
```
aaronbuchwald, line 225: I think we should verify that the requested fee config is actually valid here ie min is less than max etc.

consensus/dummy/dynamic_fees_test.go
```diff
@@ -265,54 +265,6 @@ func TestLongWindowOverflow(t *testing.T) {
 	}
 }
 
-func TestSelectBigWithinBounds(t *testing.T) {
```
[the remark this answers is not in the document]
ceyonur: it was moved to https://github.com/ava-labs/subnet-evm/pull/121/files#diff-584859deaab0437d9cc94db396497f011a5d61fbb97134787eb2866e7a5f4c00

precompile/fee_config_manager.go
ceyonur: I was thinking that activation/actual fee config setting will be different actions. but with defaulting to chain configs and using a single-source of truth, actually we can add that here too. However I'm afraid that would introduce extra confusion since there will be 2 different fee configs in the genesis.

precompile/fee_config_manager.go
ceyonur: not actually, but prefer to follow our convention here :D

eth/gasprice/gasprice.go
```diff
@@ -176,11 +179,19 @@ func NewOracle(backend OracleBackend, config Config) *Oracle {
 			lastHead = ev.Block.Hash()
 		}
 	}()
-	minBaseFee := backend.ChainConfig().GetFeeConfig().MinBaseFee
+	feeConfig, err := backend.GetFeeConfigAt(backend.LastAcceptedBlock().Header())
+	var minBaseFee *big.Int
+	if err != nil {
+		// resort back to chain config
+		log.Warn("Got error while getting fee config, defaulting to genesis fee config")
+		minBaseFee = backend.ChainConfig().FeeConfig.MinBaseFee
```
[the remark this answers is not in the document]
ceyonur: I'm afraid this one does not return any error, should we change the function return type?

precompile/fee_config_manager.go
ceyonur: you mean for the FeeConfig type?

aaronbuchwald: No I mean right now we are passing in `*big.Int` and sometimes `common.Hash` etc, and pack them individually, so if we had a helper that handled the following interface:

```
type Bytable interface {
     Bytes() []byte
}
```

then we could pass either in.

contract-examples/hardhat.config.ts
```diff
@@ -33,7 +32,7 @@ export default {
   networks: {
     local: {
       //"http://{ip}:{port}/ext/bc/{chainID}/rpc
-      url: "http://127.0.0.1:9650/ext/bc/dRTfPJh4jEaRZoGkPc7xreeYbDGBrGWRV48WAYVyUgApsmzGo/rpc",
+      url: "http://127.0.0.1:9658/ext/bc/vadzRvoCaBbxuab6aq7mAhXC38TZUgGYMXjqmuaqHpdeuwDyD/rpc",
```
aaronbuchwald, line 35: can we revert to port 9650?

core/blockchain.go
```diff
@@ -101,6 +103,11 @@ const (
 	statsReportLimit = 8 * time.Second
 )
 
+type cachableFeeConfig struct {
```
aaronbuchwald, line 106: nit: add comment

precompile/utils.go
```diff
@@ -31,3 +33,63 @@ func deductGas(suppliedGas uint64, requiredGas uint64) (uint64, error) {
...
+}
+
+func bigToHashSafe(b *big.Int) (common.Hash, error) {
+	bytes := b.Bytes()
+	len := len(bytes)
+	if len > common.HashLength {
+		return common.Hash{}, fmt.Errorf("expected %d, got %d length", common.HashLength, len)
+	}
+	return common.BigToHash(b), nil
+}
```
aaronbuchwald, line 95: We should replace this with using `common.BigToHash`

params/config.go
```diff
@@ -548,14 +527,26 @@ func (c *ChainConfig) enabledStatefulPrecompiles() []precompile.StatefulPrecompi
 		statefulPrecompileConfigs = append(statefulPrecompileConfigs, &c.TxAllowListConfig)
 	}
 
+	if c.FeeManagerConfig.Timestamp() != nil {
```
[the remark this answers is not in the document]
ceyonur: I though we wanted to default to genesis config when no config is provided (but precompile is still activated).

tests/e2e/solidity/suites.go
```diff
@@ -46,11 +46,35 @@ var _ = utils.DescribePrecompile("[TX Allow List]", func() {
...
+	// 	runHardhatTests("./test/ExampleDeployerList.ts")
+	// })
+
+	// ginkgo.It("contract native minter", func() {
+	// 	err := startSubnet("./tests/e2e/genesis/contract_native_minter.json")
+	// 	gomega.Expect(err).Should(gomega.BeNil())
+	// 	running := runner.IsRunnerUp()
+	// 	gomega.Expect(running).Should(gomega.BeTrue())
+	// 	runHardhatTests("./test/ERC20NativeMinter.ts")
+	// })
```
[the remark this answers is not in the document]
ceyonur: I initially tried to run them one-by-one but did not succeed. they run fine without others, but seems it keeps state after each spec. I also tried with shutting-down clusters + running them again line in https://github.com/ava-labs/subnet-evm/blob/fee-config-precompile/tests/e2e/e2e_test.go,  but still did not help. Keeping these along with genesises like TODOs. @gyuho or @dasconnor might help with this one.

core/blockchain_reader.go
aaronbuchwald: Ya, this is better as is.

consensus/dummy/dynamic_fees_test.go
```diff
@@ -200,7 +200,7 @@ func testDynamicFeesStaysWithinRange(t *testing.T, test test) {
 	for index, block := range blocks[1:] {
 		tc := params.TestChainConfig
 		tc.FeeConfig.MinBaseFee = params.TestMinBaseFee
-		nextExtraData, nextBaseFee, err := CalcBaseFee(params.TestChainConfig, header, block.timestamp)
+		nextExtraData, nextBaseFee, err := CalcBaseFee(params.TestChainConfig, params.TestChainConfig.FeeConfig, header, block.timestamp)
```
aaronbuchwald, line 203: Can we add a test case where the base fee is both below and above the allowed range after the update?

precompile/fee_config_manager.go
```diff
@@ -0,0 +1,320 @@
...
+
+	ErrCannotChangeFee = errors.New("non-enabled cannot change fee config")
+)
+
+// FeeConfigManagerConfig wraps [AllowListConfig] and uses it to implement the StatefulPrecompileConfig
+// interface while adding in the contract deployer specific precompile address.
+type FeeConfigManagerConfig struct {
+	AllowListConfig
+	commontype.FeeConfig
+}
```
aaronbuchwald, line 55: For the other precompiles we simply directly embed the allow list config since it's the only value present, but here it looks like we're embedding JSON fields from both the allow list and fee config together, which is a little bit weird.

Can we add JSON tags to at least the `FeeConfig` field here? I think it could go either way on if the `AllowListConfig` here should have its own JSON tag as well.

core/types/block.go
```diff
@@ -273,6 +273,7 @@ func (b *Block) GasLimit() uint64     { return b.header.GasLimit }
 func (b *Block) GasUsed() uint64      { return b.header.GasUsed }
 func (b *Block) Difficulty() *big.Int { return new(big.Int).Set(b.header.Difficulty) }
 func (b *Block) Time() uint64         { return b.header.Time }
+func (b *Block) Timestamp() *big.Int  { return new(big.Int).SetUint64(b.header.Time) }
```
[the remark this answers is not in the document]
aaronbuchwald: We should also switch to passing in the header to `CheckConfigurePrecompiles` so that we don't need to use `types.NewBlockWithHeader(head)` in `miner/worker.go` and it contains all of the information needed for this function.

contract-examples/test/ExampleFeeManager.ts
```diff
@@ -0,0 +1,229 @@
...
+    contract = await FeeManager.deploy()
+    await contract.deployed()
+    const contractAddress: string = contract.address
+    console.log(`Contract deployed to: ${contractAddress}`)
+
+    const signers: SignerWithAddress[] = await ethers.getSigners()
+    manager = signers.slice(-1)[0]
+    nonEnabled = signers.slice(-2)[0]
+
+    const feeManager = await ethers.getContractAt("IFeeManager", FEE_MANAGER, owner);
```
dasconnor, line 62: this can also be typed

contract-examples/test/ExampleFeeManager.ts
```diff
@@ -0,0 +1,229 @@
...
+    })
+    await tx.wait()
+  });
+
+  it("should add contract deployer as owner", async function () {
+    const contractOwnerAddr: string = await contract.owner()
+    expect(owner.address).to.equal(contractOwnerAddr)
+  });
+
+  // this contract is not given minter permission yet, so should not mintdraw
```
dasconnor, line 84: bad comment

contract-examples/test/ExampleFeeManager.ts
```diff
@@ -0,0 +1,229 @@
...
+  it("contract should not be able to change fee without enabled", async function () {
+    const managerPrecompile = await ethers.getContractAt("IFeeManager", FEE_MANAGER, owner);
```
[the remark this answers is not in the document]
dasconnor: And maybe `contractOwner`, `contractManager`, and `contractNonEnabled` as well

contract-examples/test/ExampleFeeManager.ts
```diff
@@ -0,0 +1,229 @@
...
+    var res = await contract.getFeeConfigLastChangedAt()
+
+    expect(res).to.equal(txRes.blockNumber)
+  })
+
+  it("should let low fee tx to be in mempool", async function () {
+    var res = await contract.getCurrentFeeConfig()
+    expect(res.minBaseFee).to.be.equal(LOW_FEES.minBaseFee)
+
+    var testMaxFeePerGas = HIGH_FEES.minBaseFee - 10000
```
dasconnor, line 129: It's not clear how this fee relates to the set fee because you grab constants from both HIGH and LOW fees. I'm not sure which is greater. It's hard to judge what this test is doing.

dasconnor (commented):
Nice work, I do have a few suggestions about how you can improve the solidity tests. Not the biggest deal in the world, but I think they'd help with the code's long-term maintainability.

core/types/block.go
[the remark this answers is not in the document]
ceyonur: we cannot use header because it causes cyclic import.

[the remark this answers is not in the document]
ceyonur: I think it's also cleaner to use a function returning bigInt for the `IsForkTransition`. In order to implement the `BlockContext`  with `Time()` we need to implement it in the https://github.com/ava-labs/subnet-evm/blob/fee-config-precompile/core/vm/evm.go#L119 which also conflicts with the field name `Time`. It's like an rabbit hole there :D

params/config.go
```diff
@@ -548,14 +522,28 @@ func (c *ChainConfig) enabledStatefulPrecompiles() []precompile.StatefulPrecompi
...
 // CheckConfigurePrecompiles iterates over any stateful precompile configs that go into effect at some point and configures them
 // if they are activated between [parentTimestamp] and [currentTimestamp].
-func (c *ChainConfig) CheckConfigurePrecompiles(parentTimestamp *big.Int, currentTimestamp *big.Int, statedb precompile.StateDB) {
+func (c *ChainConfig) CheckConfigurePrecompiles(parentTimestamp *big.Int, blockContext precompile.BlockContext, statedb precompile.StateDB) {
 	// Iterate the enabled stateful precompiles and configure them if needed
 	for _, config := range c.enabledStatefulPrecompiles() {
-		precompile.CheckConfigure(parentTimestamp, currentTimestamp, config, statedb)
+		if err := config.Validate(); err != nil {
+			panic(err)
+		}
```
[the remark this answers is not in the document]
ceyonur: we're going to remove this with #139 right?

precompile/allow_list.go
```diff
@@ -147,23 +146,25 @@ func createAllowListRoleSetter(precompileAddr common.Address, role AllowListRole
 			return nil, 0, err
 		}
 
+		if readOnly {
+			return nil, remainingGas, vmerrs.ErrWriteProtection
+		}
+
```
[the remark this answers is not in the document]
ceyonur: hmm yea it's kind-of risky but I see that other precompiles (native minter) use that order. but probably not a wise move to reorder these for the existing ones.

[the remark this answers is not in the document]
ceyonur: I just tried to shortcut this one before converting input

precompile/contract_deployer_allow_list.go
```diff
@@ -32,6 +32,11 @@ func (c *ContractDeployerAllowListConfig) Contract() StatefulPrecompiledContract
 	return ContractDeployerAllowListPrecompile
 }
 
+// Validate validates given config and returns error.
+func (c *ContractDeployerAllowListConfig) Validate() error {
+	return nil // no special verification
```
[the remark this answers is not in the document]
ceyonur: I think this is going to be removed after #139, right?

precompile/contract_native_minter.go
```diff
@@ -45,6 +44,11 @@ func (c *ContractNativeMinterConfig) Contract() StatefulPrecompiledContract {
 	return ContractNativeMinterPrecompile
 }
 
+// Validate validates given config and returns error.
+func (c *ContractNativeMinterConfig) Validate() error {
+	return nil // no special verification
```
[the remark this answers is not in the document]
ceyonur: same answer :D
> I think this is going to be removed after https://github.com/ava-labs/subnet-evm/pull/139, right?

precompile/tx_allow_list.go
```diff
@@ -38,6 +38,11 @@ func (c *TxAllowListConfig) Contract() StatefulPrecompiledContract {
 	return TxAllowListPrecompile
 }
 
+// Validate validates given config and returns error.
+func (c *TxAllowListConfig) Validate() error {
+	return nil // no special verification
```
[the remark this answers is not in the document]
ceyonur: going to be removed after #139

contract-examples/test/ExampleFeeManager.ts
```diff
@@ -0,0 +1,229 @@
...
+  minBlockGasCost: 0, // minBlockGasCost
+  maxBlockGasCost: 10_000_000, // maxBlockGasCost
+  blockGasCostStep: 100_000 // blockGasCostStep
+}
+
+
+describe("ExampleFeeManager", function () {
+  this.timeout("30s")
+
+  let owner: SignerWithAddress
```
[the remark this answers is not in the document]
ceyonur: deployer might be confused with deployer allow list

contract-examples/test/ExampleFeeManager.ts
ceyonur: right!

contract-examples/test/ExampleFeeManager.ts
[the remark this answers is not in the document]
ceyonur: good call!

contract-examples/test/ExampleFeeManager.ts
```diff
@@ -0,0 +1,229 @@
...
+  let contract: Contract
```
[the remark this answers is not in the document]
ceyonur: I tried to add typechain to the repo but failed miserably...

core/blockchain_reader.go
```diff
@@ -343,3 +348,36 @@ func (bc *BlockChain) SubscribeAcceptedLogsEvent(ch chan<- []*types.Log) event.S
...
+func (bc *BlockChain) GetFeeConfigAt(parent *types.Header) (commontype.FeeConfig, *big.Int, error) {
+	config := bc.Config()
+	bigTime := new(big.Int).SetUint64(parent.Time)
+	if !config.IsFeeConfigManager(bigTime) {
+		return config.FeeConfig, common.Big0, nil
+	}
+
+	// try to return it from the cache
+	if cached, hit := bc.feeConfigCache.Get(parent.Root); hit {
+		cachedFeeConfig, ok := cached.(*cachableFeeConfig)
```
aaronbuchwald, line 365: nit: `cacheable`

precompile/fee_config_manager.go
```diff
@@ -0,0 +1,315 @@
...
+	// 8 fields in FeeConfig struct
+	feeConfigInputLen = common.HashLength * numFeeConfigField
+
+	feeConfigLastChangedAtKey = common.Hash{'l', 'c', 'a'}
+
+	ErrCannotChangeFee = errors.New("non-enabled cannot change fee config")
+)
+
+// FeeConfigManagerConfig wraps [AllowListConfig] and uses it to implement the StatefulPrecompileConfig
+// interface while adding in the contract deployer specific precompile address.
```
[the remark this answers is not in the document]
aaronbuchwald: Good catch, updating in a nits PR

eth/gasprice/gasprice.go
```diff
@@ -229,6 +239,14 @@ func (oracle *Oracle) estimateNextBaseFee(ctx context.Context) (*big.Int, error)
 	if err != nil {
 		return nil, err
 	}
+
+	// If the block does have a baseFee, calculate the next base fee
+	// based on the current time and add it to the tip to estimate the
+	// total gas price estimate.
+	feeConfig, _, err := oracle.backend.GetFeeConfigAt(block.Header())
```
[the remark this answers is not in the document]
aaronbuchwald: Good catch

aaronbuchwald (approved):
LGTM

PR #130, opened 2022-06-09
[AV-1764] E2E Test Framework

dasconnor:
Sets up a framework for running end-to-end hardhat tests against subnet-evm.

To add new tests, add them in `tests/e2e/solidity/suites.go`. Right now, I'm running the test suite for the TX Allow List precompile

Limitations: currently we can only deploy one subnet in a test run. The next network runner release will introduce a fix that allows us to deploy multiple subnets. I will add that functionality in a follow-up PR.


tests/e2e/solidity/suites.go
```diff
@@ -0,0 +1,54 @@
...
+	fmt.Println(string(out))
+	gomega.Expect(err).Should(gomega.BeNil())
+}
+
+func startSubnet(genesisPath string) {
+	runner.StartNetwork(vmId, vmName, genesisPath, utils.GetPluginDir())
+	utils.UpdateHardhatConfig()
+}
+
+var _ = utils.DescribePrecompile("[TX Allow List]", func() {
```
ceyonur, line 37: should we add deployer allow list and native minter precompile tests as well? 

Actually I was thinking to spin-up ANR only one time and enable all possible precompiles in the same genesis and run all hardhat tests at the same time. Wouldn't that actually much faster as we wouldn't need to spin up a network for each precompile test? Plus we would catch any edge-cases where all of them are activated at the same time etc.

tests/e2e/utils/parser.go
```diff
@@ -0,0 +1,79 @@
...
+	PID      int      `yaml:"pid"`
+	URIs     []string `yaml:"uris"`
+}
+
+const (
+	HARDHAT_CONFIG = "contract-examples/hardhat.config.ts"
+)
+
+func updateHardhat(rpc string) {
+	input, err := ioutil.ReadFile(HARDHAT_CONFIG)
```
ceyonur, line 41: I think we can create a separate hardhat config instead of updating existing one. hardhat can take `--config` as a specifiable option.

contract-examples/scripts/testAllowList.ts
```diff
@@ -0,0 +1,27 @@
...
+} from "ethers"
+import { ethers } from "hardhat"
+import { abi } from "../artifacts/contracts/IAllowList.sol/IAllowList.json"
+
+const adminAddress: string = "0x8db97C7cEcE249c2b98bDC0226Cc4C2A57BF52FC"
+const TX_ALLOW_LIST_ADDRESS = "0x0200000000000000000000000000000000000002"
+
+
+
+const main = async (): Promise<any> => {
```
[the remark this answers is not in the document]
dasconnor: removed

contract-examples/hardhat.config.ts
```diff
@@ -217,7 +217,7 @@ export default {
     // Data here may be overwritten without warning.
     subnet: {
```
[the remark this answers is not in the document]
dasconnor: done

scripts/run.e2e.sh
```diff
@@ -0,0 +1,132 @@
+#!/usr/bin/env bash
+set -e
```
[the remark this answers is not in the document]
dasconnor: I'm actually not really a fan of `run.sh` as written. Hopefully we can deprecate it in a few weeks and replace with avalanche-cli. I realized I actually broke normal run functionality, so I rewrote it a decent bit. I removed `run.e2e.sh`

tests/e2e/solidity/suites.go
dasconnor: I wanted to focus this PR on setting up the framework more than writing actual tests. I think those should be a follow-up effort.

tests/e2e/utils/parser.go
dasconnor: I rewrote this. Instead of overriding the hardhat config, I write the rpc url to a git ignored file and read it from that file if it exists

ceyonur (approved):
great work!

scripts/run.sh
```diff
@@ -225,31 +219,37 @@ server \
...
+  --avalanchego-log-level=${AVALANCHE_LOG_LEVEL} \
+  --output-path=/tmp/avalanchego-v${VERSION}/output.yaml \
+  --mode=${MODE}
+
+  EXIT_CODE=$?
 else
-  echo "cluster is not ready in time... terminating ${PID}"
-  kill ${PID}
-  exit 255
+  go run scripts/parser/main.go \
```
patrick-ogrady, line 246: With the changes to `main.go`, won't this result in starting the runner twice?

patrick-ogrady (approved):

PR #136, opened 2022-06-23
Add version to code

ceyonur:
adds version to code

### DIFF
--- a/plugin/evm/version.go
+++ b/plugin/evm/version.go
@@ -11,7 +11,7 @@ var (
 	// GitCommit is set by the build script
 	GitCommit string
 	// Version is the version of Subnet EVM
-	Version string
+	Version string = "v0.2.3"
 )
 
 func init() {


plugin/evm/version.go
```diff
@@ -11,7 +11,7 @@ var (
 	// GitCommit is set by the build script
 	GitCommit string
 	// Version is the version of Subnet EVM
-	Version string
+	Version string = "v0.2.3"
```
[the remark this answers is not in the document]
ceyonur: done

plugin/evm/version.go
```diff
@@ -11,7 +11,7 @@ var (
 	// GitCommit is set by the build script
 	GitCommit string
 	// Version is the version of Subnet EVM
-	Version string
```
patrick-ogrady, line 14: I thought we typically provided this as a build flag? Why set it explicitly here?

ceyonur: releaser does not set this correctly. we either had to tweak releaser, or set this explicitly. Coreth also sets this explicitly.

patrick-ogrady: Can't you just put it here? https://github.com/ava-labs/subnet-evm/blob/0abd6b6e55ba4eb0ef94cfa475b064eff64d6594/.goreleaser.yml#L6-L8

aaronbuchwald: Added

.goreleaser.yml
```diff
@@ -5,6 +5,7 @@ builds:
     binary: subnet-evm
     flags:
       - -v
+      - -ldflags "-X github.com/ava-labs/subnet-evm/plugin/evm.Version=v0.2.5"
```
aaronbuchwald, line 8: Note: a further improvement would be to set this by passing this in as an environment variable to the GH Action.

ceyonur: done

aaronbuchwald (approved):
LGTM

aaronbuchwald:
Nice find to use `{{.Version}}` @ceyonur !

darioush (approved):

PR #137, opened 2022-06-28
Nits

aaronbuchwald:
This PR adds a few nit changes primarily to the fee config helper functions.

### DIFF
--- a/precompile/fee_config_manager.go
+++ b/precompile/fee_config_manager.go
@@ -33,14 +33,14 @@ var (
 	// Singleton StatefulPrecompiledContract for setting fee configs by permissioned callers.
 	FeeConfigManagerPrecompile StatefulPrecompiledContract = createFeeConfigManagerPrecompile(FeeConfigManagerAddress)
 
-	setFeeConfigSignature     = CalculateFunctionSelector("setFeeConfig(uint256,uint256,uint256,uint256,uint256,uint256,uint256,uint256)")
-	getFeeConfigSignature     = CalculateFunctionSelector("getFeeConfig()")
-	getLastChangedAtSignature = CalculateFunctionSelector("getLastChangedAt()")
+	setFeeConfigSignature              = CalculateFunctionSelector("setFeeConfig(uint256,uint256,uint256,uint256,uint256,uint256,uint256,uint256)")
+	getFeeConfigSignature              = CalculateFunctionSelector("getFeeConfig()")
+	getFeeConfigLastChangedAtSignature = CalculateFunctionSelector("getFeeConfigLastChangedAt()")
 
 	// 8 fields in FeeConfig struct
 	feeConfigInputLen = common.HashLength * numFeeConfigField
 
-	lastChangedAtKey = common.Hash{'l', 'c', 'a'}
+	feeConfigLastChangedAtKey = common.Hash{'l', 'c', 'a'}
 )
 
 // FeeConfigManagerConfig wraps [AllowListConfig] and uses it to implement the StatefulPrecompileConfig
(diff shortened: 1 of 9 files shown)


precompile/fee_config_manager.go
```diff
@@ -84,41 +84,44 @@ func PackGetFeeConfigInput() []byte {
...
+		common.BigToHash(feeConfig.BaseFeeChangeDenominator),
+		common.BigToHash(feeConfig.MinBlockGasCost),
+		common.BigToHash(feeConfig.MaxBlockGasCost),
+		common.BigToHash(feeConfig.BlockGasCostStep),
 	}
+
 	if useSelector {
-		packed = append([][]byte{setFeeConfigSignature}, packed...)
-		return packOrderedHashesWithSelector(packed, fullLen+selectorLen)
+		res := make([]byte, len(setFeeConfigSignature)+len(hashes)*common.HashLength)
```
ceyonur, line 117: isn't  `len(hashes)*common.HashLength)` equivalent to `feeConfigInputLen`?

precompile/fee_config_manager.go
```diff
@@ -200,52 +203,34 @@ func StoreFeeConfig(stateDB StateDB, feeConfig commontype.FeeConfig, blockContex
...
-	return nil
-}
-
-// getFeeConfigHashes takes [feeConfig] and converts them to an array of hashes, with ordered with key indexes.
-func getFeeConfigHashes(feeConfig commontype.FeeConfig) ([]common.Hash, error) {
-	res := make([]common.Hash, minFeeConfigFieldKey+numFeeConfigField)
-	for i := minFeeConfigFieldKey; i <= numFeeConfigField; i++ {
-		var hashInput common.Hash
-		var err error
+		var input *big.Int
```
ceyonur, line 209: nit: I prefer this to be a hash so that we can try to convert to hash under each key case below. Currently all these fields can be converted a big number but still I feel like it's more natural to do a specific hash conversion for each of these fields.

precompile/fee_config_manager.go
```diff
@@ -178,14 +181,14 @@ func GetStoredFeeConfig(stateDB StateDB) commontype.FeeConfig {
...
-			panic("unknown key")
+			panic(fmt.Sprintf("unknown fee config key: %d", i))
 		}
 	}
 	return feeConfig
 }
 
-func GetStoredLastChangedAt(stateDB StateDB) *big.Int {
-	val := stateDB.GetState(FeeConfigManagerAddress, lastChangedAtKey)
+func GetFeeConfigLastUpdatedAt(stateDB StateDB) *big.Int {
```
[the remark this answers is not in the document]
aaronbuchwald: Good call

precompile/fee_config_manager.go
aaronbuchwald: Yes, updated.

precompile/fee_config_manager.go
```diff
@@ -200,52 +203,34 @@ func StoreFeeConfig(stateDB StateDB, feeConfig commontype.FeeConfig, blockContex
 	if blockNumber == nil {
 		return fmt.Errorf("blockNumber cannot be nil")
 	}
+	stateDB.SetState(FeeConfigManagerAddress, feeConfigLastChangedAtKey, common.BigToHash(blockNumber))
```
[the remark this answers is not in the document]
aaronbuchwald: Ok, going to move the entire block number part below the for loop so that we fetch the block number as close as possible to where it's used.

precompile/fee_config_manager.go
aaronbuchwald: Good point, more future proof. Updated.

precompile/utils.go
```diff
@@ -34,62 +33,37 @@ func deductGas(suppliedGas uint64, requiredGas uint64) (uint64, error) {
...
-	if err != nil {
-		return nil, err
-	}
-	copy(buf[selectorLen:], bytesHashes)
-
-	return buf, nil
+// packOrderedHashesWithSelector packs the function selector and ordered list of hashes into [dst]
+// byte slice.
+// assumes that [dst] has sufficient room for [functionSelector] and [hashes].
+func packOrderedHashesWithSelector(dst []byte, functionSelector []byte, hashes []common.Hash) {
```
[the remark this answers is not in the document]
ceyonur: I was just thinking it's an extension to `packOrderedHashes(dst []byte, hashes []common.Hash)` , so it would take functionSelector as the extra param. but that's really a nit

PR #140, opened 2022-07-06
wip upgrade bytes

ceyonur:
adds upgrade byte handling. It bases on the following template:

```
[
  {
    "blockTimestamp": 123123,
    "enable": {
      "txAllowListConfig": {
        "adminAddresses": [
          "0xdDD7Ec19790C8B1ca44dD78a46a2d4EAbc95023b"
        ]
      },
      "feeManagerConfig": {
        "adminAddresses": [
          "0x8db97C7cEcE249c2b98bDC0226Cc4C2A57BF52FC"
        ]
      }
    }
  }
]
```

where each version is a object in the array.

### DIFF
--- a/plugin/evm/vm.go
+++ b/plugin/evm/vm.go
@@ -28,6 +28,7 @@ import (
 	"github.com/ava-labs/subnet-evm/params"
 	"github.com/ava-labs/subnet-evm/peer"
 	"github.com/ava-labs/subnet-evm/plugin/evm/message"
+	"github.com/ava-labs/subnet-evm/precompile"
 
 	"github.com/prometheus/client_golang/prometheus"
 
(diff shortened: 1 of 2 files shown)


plugin/evm/vm.go
```diff
@@ -279,6 +280,12 @@ func (vm *VM) Initialize(
 		g.Config.FeeConfig = params.DefaultFeeConfig
 	}
 
+	if len(upgradeBytes) != 0 {
+		if err := vm.handleUpgradeBytes(upgradeBytes); err != nil {
```
patrick-ogrady, line 284: We should persist the parsed upgrade for each timestamp and error if it changes. This alone will save us MUCH pain.

ceyonur:
Closing in favor of #148

PR #143, opened 2022-07-07
AV-1878 Multiple tests in e2e suite

felipemadero:


go.mod
```diff
@@ -4,7 +4,7 @@ go 1.17
 
 require (
 	github.com/VictoriaMetrics/fastcache v1.10.0
-	github.com/ava-labs/avalanche-network-runner-sdk v0.0.0
+	github.com/ava-labs/avalanche-network-runner v1.1.3
```
aaronbuchwald, line 7: ❤️

tests/e2e/runner/runner.go
```diff
@@ -97,20 +88,34 @@ done:
...
+		resp, err := cli.Health(cctx)
 		ccancel()
 		if err != nil {
 			cancel()
 			return "", "", err
 		}
 
+		if resp.ClusterInfo == nil {
+			continue
+		}
```
aaronbuchwald, line 100: Is there a case where `Health(ctx)` returns a nil error and `ClusterInfo` is also nil?

tests/e2e/runner/runner.go
```diff
@@ -97,20 +88,34 @@ done:
 
 		outf("{{magenta}}checking custom VM status{{/}}\n")
 		cctx, ccancel := context.WithTimeout(context.Background(), 2*time.Minute)
```
aaronbuchwald, line 90: Is there any reason that we create a separate context here for this request rather than the first one created in `WaitForCustomVm(...)` ?

dasconnor: I think it's because the first timer has a significantly longer timeout than the healthcheck

dasconnor (approved):
LGTM

aaronbuchwald:
Going to merge this in with the fee config precompile PR and use it to separate out the precompile tests.

aaronbuchwald:
Closing since these changes were included in #145

tests/e2e/runner/runner.go
felipemadero: currently not. can fix that

PR #146, opened 2022-07-12
[AV-1895] remove useless check for nil clusterInfo

felipemadero:
Part of E2E code that calls ANR Health() function. If the error is nil, then the field ClusterInfo of the return value, will not be nil.


aaronbuchwald (approved):
LGTM - thanks for following up on this

PR #148, opened 2022-07-12
enable/disable precompiles through upgradeBytes

darioush:
Issue: https://github.com/ava-labs/subnet-evm/issues/154

Adds the ability to specify network upgrade timestamps (eg, SubnetEVMTimestamp) and enabling stateful precompiles in upgrade bytes.
Upgrades in upgrades bytes will override the chain config and previously enabled upgrades as long as those upgrades have not been enabled (checked by block timestamp).
Passing `"Disable": true` to the precompile config will destroy the address space of the precompile and it will no longer be available or in effect.


precompile/upgrades_config.go
```diff
@@ -0,0 +1,270 @@
...
+	"math/big"
+	"reflect"
+
+	"github.com/ava-labs/subnet-evm/utils"
+)
+
+// UpgradesConfig includes a list of network upgrades.
+// Upgrades must be sorted in increasing order of BlockTimestamp.
+type UpgradesConfig struct {
+	Upgrades []Upgrade `json:"upgrades,omitempty"`
```
darioush, line 17: todo: see if we can get rid of additional nesting into `upgrades` here.

aaronbuchwald:
Since we need to maintain backwards compatibility, I don't think that we can make this change as is to the `ChainConfig` serialization.

plugin/evm/vm_test.go
```diff
@@ -2344,21 +2333,159 @@ func TestTxAllowListSuccessfulTx(t *testing.T) {
...
+		t.Fatalf("Expected number of txs to be %d, but found %d", 1, txs.Len())
+	}
+	assert.Equal(t, signedTx0.Hash(), txs[0].Hash())
+
+	// verify the issued block is after the network upgrade
+	assert.True(t, block.Timestamp().Cmp(disableAllowListTimestamp) >= 0)
+
+	// wait between blocks for gas price and txPool to process chain event
+	// TODO: is it possible to fix this?
+	time.Sleep(2 * time.Second)
```
aaronbuchwald, line 2439: If this is just waiting for the tx pool to update to the new head, we can just wait for a new event from `newTxPoolHeadChan`

core/state_processor_test.go
```diff
@@ -350,6 +348,7 @@ func TestBadTxAllowListBlock(t *testing.T) {
 		genesis       = gspec.MustCommit(db)
 		blockchain, _ = NewBlockChain(db, DefaultCacheConfig, gspec.Config, dummy.NewFaker(), vm.Config{}, common.Hash{})
 	)
+	config.UpgradesConfig.AddTxAllowListUpgrade(big.NewInt(0), []common.Address{})
```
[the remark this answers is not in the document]
darioush: we can do that but I think it reads more concisely like this.

darioush:
somewhat unsure if [7e69f17](https://github.com/ava-labs/subnet-evm/pull/148/commits/7e69f17f4d10dd187c86e8709aaa62da744d93be) made this better

params/upgrades_config.go
```diff
@@ -208,8 +197,8 @@ func (c *UpgradesConfig) CheckCompatible(newcfg *UpgradesConfig, headTimestamp *
 			// TODO: verify config?
```
darioush, line 197: should we add a `.Equals()` method to `precompile.StatefulPrecompileConfig`?

params/upgrades_config.go
```diff
@@ -0,0 +1,252 @@
...
+// types that can be retrieved from an upgrade.
+var getters = []getterFn{
+	func(u *Upgrade) precompile.StatefulPrecompileConfig { return u.ContractDeployerAllowListConfig },
+	func(u *Upgrade) precompile.StatefulPrecompileConfig { return u.ContractNativeMinterConfig },
+	func(u *Upgrade) precompile.StatefulPrecompileConfig { return u.TxAllowListConfig },
+	func(u *Upgrade) precompile.StatefulPrecompileConfig { return u.FeeManagerConfig },
+}
+
+// TODO: can we do better?
+func isNil(s precompile.StatefulPrecompileConfig) bool {
```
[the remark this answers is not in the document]
darioush: unfortunately when a nil eg `*TxAllowListConfig` or `* FeeManagerConfig` is returned as a `precompile.StatefulPrecompileConfig` it won't be ` == nil`.

core/tx_pool.go
```diff
@@ -984,7 +984,7 @@ func (pool *TxPool) AddRemotesSync(txs []*types.Transaction) []error {
 }
 
 // This is like AddRemotes with a single transaction, but waits for pool reorganization. Tests use this method.
-func (pool *TxPool) addRemoteSync(tx *types.Transaction) error {
+func (pool *TxPool) AddRemoteSync(tx *types.Transaction) error {
```
aaronbuchwald, line 987: Is there a reason this change is made as part of this PR?

params/test_upgrades_config.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// This file contains methods used to add network upgrades that enable and disable
+// stateful precompiles. These methods should only be called from tests.
+// Non-test code should configure the Updates struct by JSON using the chain config
+// or upgrade bytes instead.
+
+// AddContractDeployerAllowListUpgrade adds a ContractDeployerAllowList upgrade at [blockTimestamp].
+func (c *UpgradesConfig) AddContractDeployerAllowListUpgrade(blockTimestamp *big.Int, admins []common.Address) {
```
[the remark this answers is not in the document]
aaronbuchwald: Can we get rid of these test functions and just set the `PrecompileUpgrades` field where they are being used?

params/upgrade_bytes_config.go
```diff
@@ -0,0 +1,70 @@
...
+package params
+
+import (
+	"encoding/json"
+	"math/big"
+)
+
+// UpgradeBytesConfig represents part of the ChainConfig
+// that can be upgraded via upgradeBytes
+type UpgradeBytesConfig struct {
```
aaronbuchwald, line 13: Can we make this `UpgradeConfig`?

params/upgrades_config.go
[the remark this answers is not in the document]
aaronbuchwald: Can we just have the getter function return a boolean to indicate if it is nil as well? Inside of the getter function the type is known, so the nil check should work correctly and be able to return a boolean that we can check.

params/upgrades_config.go
```diff
@@ -0,0 +1,252 @@
...
+type getterFn func(*Upgrade) precompile.StatefulPrecompileConfig
+
+// getters contains a list of functions that return each of the stateful precompile
+// types that can be retrieved from an upgrade.
+var getters = []getterFn{
+	func(u *Upgrade) precompile.StatefulPrecompileConfig { return u.ContractDeployerAllowListConfig },
+	func(u *Upgrade) precompile.StatefulPrecompileConfig { return u.ContractNativeMinterConfig },
+	func(u *Upgrade) precompile.StatefulPrecompileConfig { return u.TxAllowListConfig },
+	func(u *Upgrade) precompile.StatefulPrecompileConfig { return u.FeeManagerConfig },
+}
```
aaronbuchwald, line 33: Is the ordering of `getters` relevant?

params/upgrades_config.go
```diff
@@ -0,0 +1,252 @@
...
+			}
+		}
+	}
+	return configs
+}
+
+// GetContractDeployerAllowListConfig returns the latest forked ContractDeployerAllowListConfig
+// specified by [c] or nil if it was never enabled.
+func (c *UpgradesConfig) GetContractDeployerAllowListConfig(blockTimestamp *big.Int) *precompile.ContractDeployerAllowListConfig {
+	getter := func(u *Upgrade) precompile.StatefulPrecompileConfig { return u.ContractDeployerAllowListConfig }
```
aaronbuchwald, line 140: The same function is defined above. Can we make these variables above and re-use them in these getter functions?

plugin/evm/vm.go
```diff
@@ -742,3 +751,49 @@ func (vm *VM) readLastAccepted() (common.Hash, error) {
 		return lastAcceptedHash, nil
 	}
 }
+
+// handleUpgradeBytes applies the configuration of upgradeBytes
+// to the chain config. If specified upgradeBytes are not compatible
+// with previously used upgradeBytes, an error is returned.
+// If they are compatible, they are persisted and should be specified
+// again or replaced with a new, compatible configuration.
+func (vm *VM) handleUpgradeBytes(upgradeBytes []byte) error {
```
aaronbuchwald, line 760: Can we instead pass this into `core.BlockChain` to be handled in the constructor in the same place that we check if the chain config is valid?

I think it makes sense to. keep these two checks in the same place.

aaronbuchwald: This would mean we can add the `upgradeBytesKey` to the DB schema and we can simply unmarshal the upgrade bytes into the `UpgradeConfig` and pass the whole `ChainConfig` into the `*core.BlockChain` object when it gets constructed the same way we do now.

params/upgrades_config.go
```diff
@@ -0,0 +1,252 @@
...
+
+// ValidatePrecompileUpgrades checks the PrecompileUpgrades is well formed:
+// - PrecompileUpgrades must specify only one key per Upgrade
+// - the specified blockTimestamps must monotonically increase
+// - the specified blockTimestamps must be compatible with those
+//   specified in the chainConfig by genesis.
+// - check a precompile is disabled before it is re-enabled
+func (c *UpgradesConfig) ValidatePrecompileUpgrades() error {
+	for i, upgrade := range c.PrecompileUpgrades {
+		hasKey := false // used to verify if there is only one key per Upgrade
```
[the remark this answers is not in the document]
darioush: Yes, the main motivation to restrict to 1 key is to keep the increasing timestamp requirement more explicit and simpler to verify

core/genesis.go
```diff
@@ -349,6 +356,7 @@ func (g *Genesis) Commit(db ethdb.Database) (*types.Block, error) {
 	rawdb.WriteHeadBlockHash(db, block.Hash())
 	rawdb.WriteHeadHeaderHash(db, block.Hash())
 	rawdb.WriteChainConfig(db, block.Hash(), config)
+	rawdb.WriteUpgradeConfig(db, block.Hash(), &config.UpgradeConfig)
```
darioush, line 359: probably can merge `WriteUpgradeConfig` and `WriteChainConfig`

params/upgrades_config.go
darioush: not really, switched to an enum / key based approach to avoid repeating the functions.

params/precompile_config.go
```diff
@@ -0,0 +1,273 @@
...
+				return fmt.Errorf("PrecompileUpgrades[%d] cannot have a nil timestamp", i)
+			}
+			hasKey = true
+		}
+		if !hasKey {
+			return fmt.Errorf("empty precompile upgrade at index %d", i)
+		}
+	}
+
+	for _, key := range precompileKeys {
```
aaronbuchwald, line 83: Currently we only assert configs are monotonically increasing for a single precompile key, but we should enforce that they are monotonically increasing for any precompile key as well.

darioush: https://github.com/ava-labs/subnet-evm/pull/168 adds this verification

aaronbuchwald: Merged, thanks for putting this up!

plugin/evm/vm_upgrade_bytes_test.go
```diff
@@ -0,0 +1,264 @@
...
+	// initialize the VM with these upgrade bytes
+	issuer, vm, dbManager, appSender := GenesisVM(t, true, genesisJSONPreSubnetEVM, "", string(upgradeBytesJSON))
+	vm.clock.Set(subnetEVMTimestamp)
+
+	// verify upgrade is applied
+	if !vm.chainConfig.IsSubnetEVM(big.NewInt(subnetEVMTimestamp.Unix())) {
+		t.Fatal("expected subnet-evm network upgrade to have been enabled")
+	}
+
+	// Submit a successful transaction
```
aaronbuchwald, line 171:
```suggestion
	// Submit a successful transaction and build a block to move the chain head past the SubnetEVMTimestamp network upgrade
```

plugin/evm/vm_upgrade_bytes_test.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+
+	issueAndAccept(t, issuer, vm) // make a block
+
+	if err := vm.Shutdown(); err != nil {
+		t.Fatal(err)
+	}
+
+	// VM should not start again without proper upgrade bytes.
+	ctx := NewContext()
```
aaronbuchwald, line 187: Why are we creating a new context instead of re-using the previous one here?

params/upgrade_config_test.go
```diff
@@ -0,0 +1,272 @@
...
+			if tt.expectedErrorString != "" {
+				assert.ErrorContains(t, err, tt.expectedErrorString)
+			} else {
+				assert.NoError(t, err)
+			}
+		})
+	}
+}
+
+func TestCheckCompatibleUpgradeConfigs(t *testing.T) {
```
aaronbuchwald, line 62: ❤️

params/upgrade_config_test.go
```diff
@@ -0,0 +1,272 @@
...
+	}
+
+	tests := map[string]test{
+		"upgrade bytes conflicts with genesis (re-enable without disable)": {
+			expectedErrorString: "disable should be [true]",
+			upgrades: []PrecompileUpgrade{
+				{
+					TxAllowListConfig: precompile.NewTxAllowListConfig(big.NewInt(1), admins),
+				},
+			},
```
aaronbuchwald, line 32: can we add a case where disable is true, but the timestamps are the same as well?

core/genesis.go
darioush: I think the reason to store UpgradeBytes is to ensure it is consistently provided (eg, so you can be sure of what upgrades you are running by examining `config.json` and `update.json`).

I think storing them under the same key complicates the serialization/deserialization.

darioush:
Sketched the idea of merging Read/Write Upgrade with Chain in rawdb: https://github.com/ava-labs/subnet-evm/pull/170/files 

Seems this leaves less room for potential mistakes in the future though I also don't mind the code as is with functions in rawdb having less responsibilities

core/rawdb/database.go
```diff
@@ -173,6 +173,8 @@ func InspectDatabase(db ethdb.Database, keyPrefix, keyStart []byte) error {
 			preimages.Add(size)
 		case bytes.HasPrefix(key, configPrefix) && len(key) == (len(configPrefix)+common.HashLength):
 			metadata.Add(size)
+		case bytes.HasPrefix(key, upgradeConfigPrefix) && len(key) == (len(configPrefix)+common.HashLength):
```
patrick-ogrady, line 176: should be `len(upgradeConfigPrefix)`

core/rawdb/database.go
```diff
@@ -173,6 +173,8 @@ func InspectDatabase(db ethdb.Database, keyPrefix, keyStart []byte) error {
...
+			metadata.Add(size)
```
patrick-ogrady, line 177: Commenting that I agree we probably call this `metadata`.

params/config.go
```diff
@@ -111,15 +117,22 @@ type ChainConfig struct {
 	IstanbulBlock       *big.Int `json:"istanbulBlock,omitempty"`       // Istanbul switch block (nil = no fork, 0 = already on istanbul)
 	MuirGlacierBlock    *big.Int `json:"muirGlacierBlock,omitempty"`    // Eip-2384 (bomb delay) switch block (nil = no fork, 0 = already activated)
 
-	SubnetEVMTimestamp *big.Int `json:"subnetEVMTimestamp,omitempty"` // A placeholder for the latest avalanche forks (nil = no fork, 0 = already activated)
-
-	FeeConfig          commontype.FeeConfig `json:"feeConfig"`                    // Set the configuration for the dynamic fee algorithm
-	AllowFeeRecipients bool                 `json:"allowFeeRecipients,omitempty"` // Allows fees to be collected by block builders.
+	NetworkUpgrades              // Config for timestamps that enable avalanche network upgrades
+	PrecompileUpgrade            // Config for enabling precompiles from genesis
```
patrick-ogrady, line 121: On the fence of whether we should call this `PrecompileActivations`.

Nonetheless, I would make this plural to match `NetworkUpgrades`.

plugin/evm/vm.go
```diff
@@ -314,6 +314,23 @@ func (vm *VM) Initialize(
...
+		var upgradeConfig params.UpgradeConfig
+		if err := json.Unmarshal(upgradeBytes, &upgradeConfig); err != nil {
+			return err
+		}
+		vm.chainConfig.UpgradeConfig = upgradeConfig
+	}
+
+	// create genesisHash after applying upgradeBytes in case
+	// upgradeBytes modifies genesis.
+	vm.genesisHash = ethConfig.Genesis.ToBlock(nil).Hash()
```
patrick-ogrady, line 332: Why move this above Etherbase verifcation?

plugin/evm/vm_test.go
```diff
@@ -329,6 +331,34 @@ func TestVMUpgrades(t *testing.T) {
 	}
 }
 
+func issueAndAccept(t *testing.T, issuer <-chan engCommon.Message, vm *VM) snowman.Block {
```
patrick-ogrady, line 334: ❤️ ❤️

precompile/allow_list.go
```diff
@@ -31,17 +31,11 @@ var (
 	allowListInputLen = common.HashLength
 )
 
-// AllowListConfig specifies the configuration of the allow list.
-// Specifies the block timestamp at which it goes into effect as well as the initial set of allow list admins.
+// AllowListConfig specifies the initial set of allow list admins.
 type AllowListConfig struct {
-	BlockTimestamp *big.Int `json:"blockTimestamp"`
```
patrick-ogrady, line 37: Could anything weird happen by removing this?

utils/fork.go
```diff
@@ -22,3 +22,15 @@ func IsForkTransition(fork *big.Int, parent *big.Int, current *big.Int) bool {
 	currentForked := IsForked(fork, current)
 	return !parentForked && currentForked
 }
+
+// BigNumEqual returns true if x and y are equivalent ie. both nil or both
+// contain the same value.
+func BigNumEqual(x, y *big.Int) bool {
+	if x == nil {
```
patrick-ogrady, line 29: Does this work?
```
if x == nil || y == nil {
  return x == y
}
```

params/precompile_config.go
```diff
@@ -0,0 +1,281 @@
...
+// CheckConfigurePrecompiles checks if any of the precompiles specified by the chain config are enabled or disabled by the block
+// transition from [parentTimestamp] to the timestamp set in [blockContext]. If this is the case, it calls [Configure]
+// or [Deconfigure] to apply the necessary state transitions for the upgrade.
+// This function is called:
+// - within genesis setup to configure the starting state for precompiles enabled at genesis,
+// - during block processing to update the state before processing the given block.
+func (c *ChainConfig) CheckConfigurePrecompiles(parentTimestamp *big.Int, blockContext precompile.BlockContext, statedb precompile.StateDB) {
+	blockTimestamp := blockContext.Timestamp()
+	for _, key := range precompileKeys { // Note: configure precompiles in a deterministic order.
+		for _, config := range c.getActivatingPrecompileConfigs(parentTimestamp, blockTimestamp, key, c.PrecompileUpgrades) {
```
patrick-ogrady, line 271: Will this suicide multiple times if multiple blocks are produced at the same timestamp? Or do we enforce a precompile upgrade is only processed once at a timestamp (maybe only the boundary of less than time to >= time)?

patrick-ogrady: AFAICT this logic enforces we only run once on the boundary and we should be good.

plugin/evm/vm.go
patrick-ogrady: I'm not sure if there are any side effects but feels weird to process a block before completing validation (occurs in `Genesis.ToBlock`).

core/rawdb/database.go
darioush: nice catch. 👍

core/rawdb/database.go
darioush: This just follows the existing pattern.

plugin/evm/vm.go
darioush: Yeah the only requirement is to include the chain config (including the upgrade bytes) in genesis.
Going to rearrange this to keep the verification. Very good idea.

precompile/allow_list.go
darioush: I don't think so. The replacement `UpgradeableConfig` is embedded in the precompile configs similar to how `AllowListConfig` is embedded, so this change only moves it.

params/config.go
darioush: This is a little nuanced since in the (genesis) chain config, multiple upgrades can be specified (one of each kind of precompile).
In `UpgradeConfig`, each `PrecompileUpgrade` is only allowed to specify one key, then we have an array of them which makes `PrecompileUpgrades`.

Definitely happy to rename whichever way is clearer.

ceyonur (approved):
IT'S UPGRADIN' TIME!

plugin/evm/vm.go
```diff
@@ -329,11 +329,23 @@ func (vm *VM) Initialize(
...
-
 	vm.chainConfig = g.Config
 	vm.networkID = ethConfig.NetworkId
 
+	// Apply upgradeBytes (if any) by unmarshalling them into [chainConfig.UpgradeConfig].
+	// Initializing the chain will verify upgradeBytes are compatible with existing values.
+	if len(upgradeBytes) > 0 {
+		var upgradeConfig params.UpgradeConfig
+		if err := json.Unmarshal(upgradeBytes, &upgradeConfig); err != nil {
+			return err
```
aaronbuchwald, line 340:
```suggestion
			return fmt.Errorf("failed to parse upgrade bytes: %w", err)
```

PR #99, opened 2022-05-12
improve run script

ceyonur:
we can omit downloading & extracting steps if we already did them. also waits until sigterm received in run mode so we can shutdown background processes.

PR #155, opened 2022-07-19
Cleaned up ANR install process

### LINKED ISSUE #152: Convert subnet-evm E2E tests to use go install
Subnet-evm's e2e tests currently download the network runner multiple times. Let's migrate to using go install so that we don't need to repeat the install process multiple times.

dasconnor:
Use `go install` instead of unpacking a binary.

Closes #152


dasconnor:
This is blocked on the latest ANR release.

ceyonur:
Does this also close #99 ?

dasconnor:
I don't think this will close #99 because this only addresses the network runner, not avalanchego. One question for you @ceyonur, do you have any idea why the pipeline is failing? Looks like the fee config test. Not sure how/why that could be affected.

dasconnor:
ok, finally passed

aaronbuchwald (approved):
LGTM

darioush (approved):

PR #165, opened 2022-07-25
nit: improve some precompile comments, type checks

darioush:
minor improvements for comments and type checks

### DIFF
--- a/params/config.go
+++ b/params/config.go
@@ -528,8 +528,12 @@ func (c *ChainConfig) enabledStatefulPrecompiles() []precompile.StatefulPrecompi
 	return statefulPrecompileConfigs
 }
 
-// CheckConfigurePrecompiles iterates over any stateful precompile configs that go into effect at some point and configures them
-// if they are activated between [parentTimestamp] and [currentTimestamp].
+// CheckConfigurePrecompiles checks if any of the precompiles specified by [c] is enabled by the block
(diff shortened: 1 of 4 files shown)


params/config.go
```diff
@@ -528,8 +528,12 @@ func (c *ChainConfig) enabledStatefulPrecompiles() []precompile.StatefulPrecompi
 	return statefulPrecompileConfigs
 }
 
-// CheckConfigurePrecompiles iterates over any stateful precompile configs that go into effect at some point and configures them
-// if they are activated between [parentTimestamp] and [currentTimestamp].
+// CheckConfigurePrecompiles checks if any of the precompiles specified by [c] is enabled by the block
+// transition from [parentTimestamp] to the timestamp set in [blockContext]. If this is the case, it calls [Configure]
+// to apply the necessary state transitions for the upgrade.
+// This function is called:
```
aaronbuchwald, line 534:
```suggestion
// CheckConfigurePrecompiles checks if any of the precompiles specified by are enabled by the block
// transition from [parentTimestamp] to the timestamp set in [blockContext]. If this is the case, it calls [Configure]
// to apply the necessary state transitions for the upgrade.
// This function is called:
```

I think it's better not to reference c here since it won't be clear when the comment pops up in another package.

aaronbuchwald (approved):
LGTM - one optional nit

ceyonur (approved):

PR #170, opened 2022-07-27
merge (Read,Write) UpgradeConfig  w/ ChainConfig

darioush:
Sketched the idea of merging Read/Write Upgrade with Chain in rawdb.

Seems this leaves less room for potential mistakes in the future though I also don't mind the code as is with functions in rawdb having less responsibilities


aaronbuchwald (approved):
LGTM

ceyonur (approved):

PR #176, opened 2022-07-29
Reserve address spaces for custom precompiles

### LINKED ISSUE #164: Reserve Addresses for Precompiles
Reserve the following addresses for future precompiles:

- Addresses in the range 0x0100000000000000000000000000000000000000-0x01000000000000000000000000000000000000ff for precompiles from coreth
- Addresses in the range 0x0200000000000000000000000000000000000000-0x02000000000000000000000000000000000000ff for optional subnet-evm precompiles
- Addresses in the range 0x0300000000000000000000000000000000000000-0x03000000000000000000000000000000000000ff for subnet-evm forks

aaronbuchwald:
This PR resolves #164

### DIFF
--- a/precompile/params.go
+++ b/precompile/params.go
@@ -3,7 +3,11 @@
 
 package precompile
 
-import "github.com/ethereum/go-ethereum/common"
+import (
+	"fmt"
+
+	"github.com/ethereum/go-ethereum/common"
+)
 
 // Gas costs for stateful precompiles
 const (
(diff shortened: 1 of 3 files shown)


precompile/params.go
```diff
@@ -40,4 +44,38 @@ var (
 		TxAllowListAddress,
 		FeeConfigManagerAddress,
 	}
+	reservedRanges = []AddressRange{
+		{
+			common.HexToAddress("0x0100000000000000000000000000000000000000"),
```
darioush, line 49: can we also add an explanation for the `0x01` range to the comment above?

darioush (approved):

ceyonur (approved):

precompile/params.go
aaronbuchwald: Good call, just added a comment to explain they come from coreth.

PR #180, opened 2022-08-01
add precompile gen cmd & template

### LINKED ISSUE #159: Add ABI Encoding/Decoding for Stateful Precompile Devs
Stateful precompiles currently require that the developer manually encode/decode arguments as per the ABI encoding schema: https://docs.soliditylang.org/en/latest/abi-spec.html. We should switch to using ABI arguments for encoding/decoding function arguments and return types: https://github.com/ava-labs/subnet-evm/blob/master/accounts/abi/argument.go.

The abigen command can be used to auto-generate golang binding for interacting with a smart contract, but we don't even need to go that far. We can do the following with this existing tool in order to create automatic generation of precompile configs and contracts available to users and make our own precompile development workflow significantly easier.

1. Parse a Solidity interface into the corresponding ABI
2. Generate a golang file containing the ABI as a variable
3. Generate the precompile config with a single block timestamp as the only type in the struct (named based off of Solidity contract name)
4. Iterate over the functions available in the contract ABI to create golang function stubs for each one, which will include the first line to use ABI unpack to unpack the input arguments into the expected inputs (add a line to pack the function return type as well at the end with nil values passed in to start)
5. Create a function stub for the fallback function if specified
6. Create a function to construct the actual stateful precompile at the bottom of the file

This should allow us to support the auto-generation of a fully stubbed out stateful precompile and take all of the hassle and difficulty out of packing and unpacking arguments and return types from the precompile developer.

ceyonur:
Adds precompile gen cmd tool which can take a json ABI file and generates a go-lang precompiled contract. It also can directly detect if the ABI uses `AllowList` precompile and then configures the generated precompile accordingly.

The PR also contains a guideline and highlights for developers to guide them for a complete precompile integration.

closes #159


accounts/abi/bind/precompiletemplate.go
```diff
@@ -0,0 +1,312 @@
...
+	{{- if not .Original.IsConstant}}
+
+	ErrCannot{{.Normalized.Name}} = errors.New("non-enabled cannot {{.Original.Name}}")
+	{{- end}}
+	{{- end}}
+
+	{{.Contract.Type}}ABI abi.ABI // will be filled by init func
+)
+
+// {{.Contract.Type}}Config {{if .Contract.AllowList}}wraps [AllowListConfig] and uses it to implement {{else}}implements{{end}} the StatefulPrecompileConfig
```
darioush, line 83: ok with me if we wanted to leave out the `if` part of this comment

accounts/abi/bind/precompiletemplate.go
```diff
@@ -0,0 +1,312 @@
...
+}
+
+// tmplContract contains the data needed to generate an individual contract binding.
+type tmplPrecompileContract struct {
+	*tmplContract
+	AllowList bool                   // Indicator whether the contract uses AllowList precompile
+	Funcs     map[string]*tmplMethod // Contract functions that include both Calls + Transacts in tmplContract
+}
+
+// tmplSourceGo is the Go precompiled source template.
```
darioush, line 18:
```suggestion
// tmplSourcePrecompileGo is the Go precompiled source template.
```

accounts/abi/bind/precompiletemplate.go
```diff
@@ -0,0 +1,312 @@
...
+}
+
+// Address returns the address of the {{.Contract.Type}}. Addresses reside under the precompile/params.go
+// Select a non-conflicting address and set it in the params.go.
+func (c *{{.Contract.Type}}Config) Address() common.Address {
+	return {{.Contract.Type}}Address
+}
+
+// Configure configures [state] with the initial configuration.
+func (c *{{.Contract.Type}}Config) Configure(_ ChainConfig, state StateDB, _ BlockContext) {
```
darioush, line 159: should we allow this function to return `error`?

accounts/abi/bind/bind.go
```diff
@@ -219,18 +228,66 @@ func Bind(types []string, abis []string, bytecodes []string, fsigs []map[string]
...
+		precompileContract := &tmplPrecompileContract{
+			tmplContract: firstContract,
+			AllowList:    isAllowList,
+			Funcs:        funcs,
+		}
+
+		data = &tmplPrecompileData{
+			Contract: precompileContract,
+			Structs:  structs,
+		}
```
darioush, line 280: nit: may be easier to follow if we made this part into a function and got `data` as the return value (or `data` and `templateSource`)

accounts/abi/bind/precompiletemplate.go
```diff
@@ -0,0 +1,312 @@
...
+	{{- end}}
+
+	// Return an empty output and the remaining gas
+	return packedOutput, remainingGas, nil
+}
+{{end}}
+
+// create{{.Contract.Type}}Precompile returns a StatefulPrecompiledContract
+// with getters and setters for the precompile.
+{{if .Contract.AllowList}} //Access to the getters/setters is controlled by an allow list for [precompileAddr].{{end}}
```
darioush, line 297:
```suggestion
{{if .Contract.AllowList -}} // Access to the getters/setters is controlled by an allow list for [precompileAddr].{{end}}
```

accounts/abi/bind/precompiletemplate.go
```diff
@@ -0,0 +1,312 @@
...
+func create{{.Contract.Type}}Precompile(precompileAddr common.Address) StatefulPrecompiledContract {
+	var functions []*statefulPrecompileFunction
+	{{- if .Contract.AllowList}}
+	functions = append(functions, createAllowListFunctions(precompileAddr)...)
+  {{- end}}
```
darioush, line 302: seems there is tab vs. space here (or otherwise wspace is not aligned)?
also in some other places in this file.

accounts/abi/bind/precompiletemplate.go
```diff
@@ -0,0 +1,312 @@
...
+	type {{.Name}} struct {
+	{{range $field := .Fields}}
+	{{$field.Name}} {{$field.Type}}{{end}}
+	}
+{{- end}}
+
+{{- range .Contract.Funcs}}
+{{ if len .Normalized.Inputs | lt 1}}
+type {{capitalise .Normalized.Name}}Input struct{
+{{range .Normalized.Inputs}} {{capitalise .Name}} {{bindtype .Type $structs}}; {{end}}
```
darioush, line 104: nit: seems `;` is not necessary?

accounts/abi/bind/precompiletemplate.go
```diff
@@ -0,0 +1,312 @@
...
+	// this function does not return an output, leave this one as is
+	packedOutput := []byte{}
+	{{- else}}
+	packedOutput, err := Pack{{.Normalized.Name}}Output(nil)
+	if err != nil {
+		return nil, remainingGas, err
+	}
+	{{- end}}
+
+	// Return an empty output and the remaining gas
```
darioush, line 290:
```suggestion
	// Return the packed output and the remaining gas
```

accounts/abi/abi.go
```diff
@@ -91,6 +91,37 @@ func (abi ABI) Pack(name string, args ...interface{}) ([]byte, error) {
...
+func (abi ABI) getInputs(name string, data []byte) (Arguments, error) {
+	// since there can't be naming collisions with contracts and events,
+	// we need to decide whether we're calling a method or an event
+	var args Arguments
+	if method, ok := abi.Methods[name]; ok {
+		if len(data)%32 != 0 {
+			return nil, fmt.Errorf("abi: improperly formatted input: %s - Bytes: [%+v]", string(data), data)
+		}
+		args = method.Inputs
+	}
```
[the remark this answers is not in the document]
ceyonur: we don't deal with events in precompiles but you're totally right 👍

accounts/abi/bind/precompiletemplate.go
ceyonur: there is a looong chain of methods that should be returning an error:
`StatefulPrecompileConfig.Configure > precompile.Config > CheckConfigurePrecompiles` but still I agree that we should enable errors here (especially for validation of configuration) @aaronbuchwald

accounts/abi/bind/precompiletemplate.go
ceyonur: `-` is not necessarry here, I added the whitespace.

accounts/abi/bind/precompiletemplate.go
ceyonur: hm seems my text-editor got confused.

accounts/abi/bind/precompiletemplate.go
ceyonur: I think this is necessary for the golang, otherwise it will keep adding fields to each other in the same line. like:
` Addr common.Address  Amount *big.Int ` but `;` can separate those and gofmt can just create a new line by deleting `;`

darioush:
LGTM. Would like to propagate the error from configure (can go in as separate PR before this or with this PR)

accounts/abi/bind/bind.go
```diff
@@ -601,3 +629,36 @@ func hasStruct(t abi.Type) bool {
...
+
+	for k, v := range contract.Transacts {
+		funcs[k] = v
+	}
+
+	for k, v := range contract.Calls {
+		funcs[k] = v
+	}
+
+	_, isAllowList := funcs[readAllowListFuncKey]
```
aaronbuchwald, line 644: I wonder if this may unintentionally interfere with anyone else's contracts and if there's another way to do this. Is it possible to parse the contract to see if it is something like:

```
interface NameOfContract is IAllowList
```

instead of checking for a specific function name?

accounts/abi/bind/bind.go
```diff
@@ -224,12 +232,30 @@ func Bind(types []string, abis []string, bytecodes []string, fsigs []map[string]
...
-		Structs:   structs,
+
+	var data interface{}
+	var templateSource string
+
+	// Generate the contract template data according to contract type (precompile/non)
+	if isPrecompile {
+		if lang == LangJava {
+			return "", errors.New("java binding for precompiled contracts is not supported yet")
+		}
```
aaronbuchwald, line 243: Should we change this to check to ensure that it is always golang? There's another option for ObjC that would not be caught here and we only support golang anyways.

accounts/abi/bind/bind.go
```diff
@@ -224,12 +232,30 @@ func Bind(types []string, abis []string, bytecodes []string, fsigs []map[string]
...
-		Package:   pkg,
-		Contracts: contracts,
-		Libraries: libs,
-		Structs:   structs,
+
+	var data interface{}
+	var templateSource string
+
+	// Generate the contract template data according to contract type (precompile/non)
+	if isPrecompile {
```
aaronbuchwald, line 240: Would it be possible to create a separate function for binding precompiles and create helper functions to share between them rather than add an argument to the existing `Bind` function? Changing the signature of `Bind` may result in any code that relies on the golang ABI running into issues.

accounts/abi/bind/precompiletemplate.go
```diff
@@ -0,0 +1,311 @@
...
+}
+
+// Equal returns true if [s] is a [*{{.Contract.Type}}Config] and it has been configured identical to [c].
+func (c *{{.Contract.Type}}Config) Equal(s StatefulPrecompileConfig) bool {
+	// typecast before comparison
+	other, ok := (s).(*{{.Contract.Type}}Config)
+	if !ok {
+		return false
+	}
+	return c.UpgradeableConfig.Equal(&other.UpgradeableConfig) {{if .Contract.AllowList}} && c.AllowListConfig.Equal(&other.AllowListConfig) {{end}}
```
aaronbuchwald, line 149: Should we add a TODO here to add whatever additional checks are necessary to ensure that the configs are actually equal?

And clarify that if the timestamp / allow list is the only thing in the config then no additional changes are necessary.

aaronbuchwald (changes requested):

params/config.go
```diff
@@ -520,6 +532,8 @@ func (c *ChainConfig) AvalancheRules(blockNum, blockTimestamp *big.Int) Rules {
 	rules.IsContractNativeMinterEnabled = c.IsContractNativeMinter(blockTimestamp)
 	rules.IsTxAllowListEnabled = c.IsTxAllowList(blockTimestamp)
 	rules.IsFeeConfigManagerEnabled = c.IsFeeConfigManager(blockTimestamp)
+	// ADD YOUR PRECOMPILE HERE
+	// rules.IsYourPrecompileEnabled = c.IsYourPrecompile(blockTimestamp)
```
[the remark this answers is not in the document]
ceyonur: yep changed all of them

accounts/abi/abi.go
```diff
@@ -91,6 +91,40 @@ func (abi ABI) Pack(name string, args ...interface{}) ([]byte, error) {
...
+		return nil, fmt.Errorf("method '%s' not found", name)
+	}
+	arguments, err := method.Outputs.Pack(intf...)
+	if err != nil {
+		return nil, err
+	}
+	return arguments, nil
+}
+
+func (abi ABI) getInputs(name string, data []byte) (Arguments, error) {
```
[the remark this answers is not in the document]
ceyonur: I think it's confusing because `getArguments` is not actually getting arguments (which typically referred to inputs) but getting outputs. weird.

accounts/abi/abi.go
```diff
@@ -105,11 +139,20 @@ func (abi ABI) getArguments(name string, data []byte) (Arguments, error) {
 		args = event.Inputs
 	}
 	if args == nil {
-		return nil, errors.New("abi: could not locate named method or event")
+		return nil, fmt.Errorf("abi: could not locate named method or event: %s", name)
```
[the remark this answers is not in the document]
ceyonur: @darioush would you like the honors :D ?

accounts/abi/abi.go
```diff
@@ -91,6 +91,40 @@ func (abi ABI) Pack(name string, args ...interface{}) ([]byte, error) {
 	return append(method.ID, arguments...), nil
 }
 
+// PackOutput packs the given intf as the output of [method] to conform the ABI.
+// This does not include method ids.
```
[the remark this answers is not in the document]
ceyonur: I agree `method` is much more clear than the `name` but let's keep it `name` since it's being used in other functions. I updated the comment accordingly.

cmd/precompilegen/main.go
```diff
@@ -0,0 +1,153 @@
...
+import (
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/ava-labs/subnet-evm/accounts/abi/bind"
+	"github.com/ava-labs/subnet-evm/internal/flags"
+	"github.com/ethereum/go-ethereum/cmd/utils"
```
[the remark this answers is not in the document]
ceyonur: hm i wouldn't prefer it unless we modify that code.

accounts/abi/bind/bind.go
ceyonur: Generator parses ABI not the contract. it's kind-of a janky way to fetch this from ABI but I don't think ABI includes inheritance info. Parsing the contract would be a very different thing than the current approach. We can make it stricter with checking all of AllowList funcs (`readAllowList` + `setAdmin` + setEnabled` + `setNone`).

accounts/abi/bind/bind.go
ceyonur: I think that check is a no-op since we statically set it in the precompilegen. but right better to add a not golang check instead java check
https://github.com/ava-labs/subnet-evm/blob/precompile-gen/cmd/precompilegen/main.go#L91

accounts/abi/bind/bind.go
ceyonur: my first attempt to use it as an separate function but then I was under impression that it would make it much harder for code migration.

accounts/abi/bind/bind.go
aaronbuchwald: I'd be in favor of making it stricter.

Either way, we should also document that overlapping with these function names will result in unintended behavior.

accounts/abi/bind/precompiletemplate.go
```diff
@@ -0,0 +1,315 @@
...
+	{{.Contract.Type}}RawABI = "{{.Contract.InputABI}}"
+)
+
+// Singleton StatefulPrecompiledContract and signatures.
+var (
+	_ StatefulPrecompileConfig = &{{.Contract.Type}}Config{}
+
+	{{.Contract.Type}}Precompile StatefulPrecompiledContract = create{{.Contract.Type}}Precompile({{.Contract.Type}}Address)
+	{{- range .Contract.Funcs}}
+	{{decapitalise .Normalized.Name}}Signature = CalculateFunctionSelector("{{.Original.Sig}}")
```
aaronbuchwald, line 73: Can we switch from using the helper function `CalculateFunctionSelector` to using the `Method.ID` from the ABI ?

accounts/abi/bind/precompiletemplate.go
```diff
@@ -0,0 +1,315 @@
...
+
+	"github.com/ava-labs/subnet-evm/accounts/abi"
+	"github.com/ava-labs/subnet-evm/vmerrs"
+
+	"github.com/ethereum/go-ethereum/common"
+)
+
+const (
+	{{- range .Contract.Funcs}}
+	{{.Normalized.Name}}GasCost uint64 = -1 // SET A GAS COST HERE
```
aaronbuchwald, line 60: Can we set this to 0 instead of -1 so the compiler doesn't show this as an error when the file is generated?

Not sure if this was intentional to prevent users from setting it incorrectly, but I think it would be better if the default compiled even if it isn't recommended to run with gas costs set to 0.

accounts/abi/bind/precompiletemplate.go
```diff
@@ -0,0 +1,315 @@
...
+// Singleton StatefulPrecompiledContract and signatures.
+var (
+	_ StatefulPrecompileConfig = &{{.Contract.Type}}Config{}
+
+	{{.Contract.Type}}Precompile StatefulPrecompiledContract = create{{.Contract.Type}}Precompile({{.Contract.Type}}Address)
+	{{- range .Contract.Funcs}}
+	{{decapitalise .Normalized.Name}}Signature = CalculateFunctionSelector("{{.Original.Sig}}")
+	{{- if not .Original.IsConstant}}
+
+	ErrCannot{{.Normalized.Name}} = errors.New("non-enabled cannot {{.Original.Name}}")
```
aaronbuchwald, line 76: Where are these errors meant to be used?

precompile/contract.go
```diff
@@ -86,18 +86,13 @@ func newStatefulPrecompileFunction(selector []byte, execute RunStatefulPrecompil
 // off responsibilities to internal execution functions.
 // Note: because we only ever read from [functions] there no lock is required to make it thread-safe.
 type statefulPrecompileWithFunctionSelectors struct {
-	fallback  *statefulPrecompileFunction
+	fallback  RunStatefulPrecompileFunc
```
aaronbuchwald, line 89: Good call

accounts/abi/bind/precompiletemplate.go
```diff
@@ -0,0 +1,366 @@
...
+// There are some must-be-done changes waiting in the file. These are highlighted by CAPITAL WORDS.
+// Additionally there are other files you need to edit to activate your precompile.
+// These areas are highlighted with comments "ADD YOUR PRECOMPILE HERE".
+// For testing take a look at other precompile tests in core/stateful_precompile_test.go
+
+/* General guidelines for precompile development:
+1- Read the comment and set a suitable contract address in precompile/params.go. E.g:
+	{{.Contract.Type}}Address = common.HexToAddress("ASUITABLEHEXADDRESS")
+2- Set gas costs here
+3- Unless you know what you're doing write your custom code only after highlighted areas with "CUSTOM CODE STARTS HERE".
```
aaronbuchwald, line 33:
```suggestion
3- It is recommended to only modify code in the highlighted areas marked with "CUSTOM CODE STARTS HERE". Modifying code outside of these areas should be done with caution and with a deep understanding of how these changes may impact the EVM.
```

accounts/abi/bind/precompiletemplate.go
```diff
@@ -0,0 +1,366 @@
...
+	callerStatus := getAllowListStatus(stateDB, {{$contract.Type}}Address, caller)
+	if !callerStatus.IsEnabled() {
+		return nil, remainingGas, fmt.Errorf("%w: %s", Err{{$contract.Type}}CannotFallback, caller)
+	}
+	// allow list code ends here.
+	{{- end}}
+
+	// CUSTOM CODE STARTS HERE
+
+	// Fallback function cannot return an output, return remainingGas only.
```
aaronbuchwald, line 339: A fallback function can return an output: https://docs.soliditylang.org/en/latest/contracts.html?highlight=fallback%20function. It is just the receive function which is not allowed to return an output.

accounts/abi/bind/precompiletemplate.go
```diff
@@ -0,0 +1,366 @@
...
+10- Run your e2e precompile Solidity tests with 'E2E=true ./scripts/run.sh'
+
+*/
+
+package precompile
+
+import (
+	"errors"
+	"fmt"
+	"strings"
```
aaronbuchwald, line 50: Missing import for `*big.Int`, should we do the same thing that's done in the binding template to import a few more things that might get used and include a var block below to ensure that they don't trigger unused imports?

accounts/abi/bind/precompiletemplate.go
```diff
@@ -0,0 +1,366 @@
...
+	}
+	// allow list code ends here.
+  {{end}}
+	// CUSTOM CODE STARTS HERE
+	// use inputStruct ...
+	{{- if len .Normalized.Outputs | eq 0}}
+	// this function does not return an output, leave this one as is
+	packedOutput := []byte{}
+	{{- else}}
+	packedOutput, err := Pack{{.Normalized.Name}}Output(nil)
```
aaronbuchwald, line 300: Can we include the empty struct in question here instead of nil, so that the code compiles once it's generated and returns an empty result instead of not compiling?

accounts/abi/bind/precompiletemplate.go
```diff
@@ -0,0 +1,366 @@
...
+	{{- if len .Normalized.Inputs | eq 0}}
+	// no input provided for this function
+	{{else}}
+	// attempts to unpack [input] into the arguments to the {{.Normalized.Name}}Input.
+	// Assumes that [input] does not include selector
+	// You can use unpacked [inputStruct] variable in your code
+	inputStruct, err := Unpack{{capitalise .Normalized.Name}}Input(input)
+	if err != nil{
+		return nil, remainingGas, err
+	}
```
aaronbuchwald, line 279: Can we include something below here so that there's no complaint about `inputStruct` being unused?

Perhaps:

```
_ = inputStruct // CUSTOM CODE OPERATES ON INPUT
```

accounts/abi/bind/precompiletemplate.go
```diff
@@ -0,0 +1,366 @@
...
+	Funcs     map[string]*tmplMethod // Contract functions that include both Calls + Transacts in tmplContract
+}
+
+// tmplSourcePrecompileGo is the Go precompiled source template.
+const tmplSourcePrecompileGo = `
+// Code generated
+// This file is a generated precompile contract with stubbed abstract functions.
+// The file is generated by a template. Please inspect every code and comment in this file before use.
+
+// There are some must-be-done changes waiting in the file. These are highlighted by CAPITAL WORDS.
```
aaronbuchwald, line 24:
```suggestion
// There are some must-be-done changes waiting in the file. Each area requiring you to add your code is marked with CUSTOM CODE to make them easy to find and modify.
```

aaronbuchwald: Want to make it easy for users to search for them instead of needing to scan the file looking for capital words

accounts/abi/bind/precompiletemplate.go
ceyonur: ooh nice

accounts/abi/bind/precompiletemplate.go
ceyonur: for allowlist check to return an error

accounts/abi/bind/precompiletemplate.go
```diff
@@ -0,0 +1,366 @@
...
+}
+
+// Contract returns the singleton stateful precompiled contract to be used for {{.Contract.Type}}.
+func (c *{{.Contract.Type}}Config) Contract() StatefulPrecompiledContract {
+	return {{.Contract.Type}}Precompile
+}
+
+{{if .Contract.AllowList}}
+// Get{{.Contract.Type}}Status returns the role of [address] for the {{.Contract.Type}} list.
+func Get{{.Contract.Type}}Status(stateDB StateDB, address common.Address) AllowListRole {
```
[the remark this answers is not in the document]
ceyonur: good call!

aaronbuchwald (approved):
LGTM - great work!

accounts/abi/bind/precompiletemplate.go
```diff
@@ -0,0 +1,390 @@
+// (c) 2019-2022, Ava Labs, Inc. All rights reserved.
```
aaronbuchwald, line 1: nit: can we rename this file to `precompile_template.go` ?

accounts/abi/bind/bind.go
```diff
@@ -224,12 +232,30 @@ func Bind(types []string, abis []string, bytecodes []string, fsigs []map[string]
...
 	}
-	// Generate the contract template data content and render it
-	data := &tmplData{
-		Package:   pkg,
-		Contracts: contracts,
-		Libraries: libs,
-		Structs:   structs,
+
+	var data interface{}
+	var templateSource string
```
aaronbuchwald, line 237: Can we combine this into one var block?

aaronbuchwald:
Would it be possible to add a unit test for the precompile generation?

accounts/abi/bind/precompiletemplate.go
```diff
@@ -0,0 +1,390 @@
...
+	_ = big.NewInt
+	_ = strings.NewReader
+)
+
+{{$contract := .Contract}}
+// Singleton StatefulPrecompiledContract and signatures.
+var (
+	_ StatefulPrecompileConfig = &{{.Contract.Type}}Config{}
+
+	{{.Contract.Type}}Precompile StatefulPrecompiledContract = create{{.Contract.Type}}Precompile({{.Contract.Type}}Address)
```
aaronbuchwald, line 83: This needs to be moved into the `init` function since calling this here will take place before `init()` gets invoked.

This results in the ABI not being populated when the create function is called, such that each referenced `method.ID` is empty.

aaronbuchwald: To catch this in the future, we should also check that the method names actually exist within the `create` function so that if the requested function does not exist, it will detect it there immediately.

ceyonur: fixed & added some checks

accounts/abi/bind/bind.go
ceyonur: done

accounts/abi/bind/precompiletemplate.go
ceyonur: done

accounts/abi/bind/precompiletemplate.go
aaronbuchwald: Great, would you mind making a PR to geth as well with the moved error checks in the ABI package to fail earlier?

aaronbuchwald (approved):

accounts/abi/argument.go
```diff
@@ -196,6 +196,9 @@ func (arguments Arguments) UnpackValues(data []byte) ([]interface{}, error) {
 	virtualArgs := 0
 	for index, arg := range nonIndexedArgs {
 		marshalledValue, err := toGoType((index+virtualArgs)*32, arg.Type, data)
+		if err != nil {
+			return nil, err
+		}
```
darioush, line 201: should this be made as a PR to upstream?

darioush (approved):
lgtm, very exciting

accounts/abi/bind/precompiletemplate.go
ceyonur: yep i will also create one for coreth, or are we going to include this in precompile gen PR?

accounts/abi/argument.go
ceyonur: yep both to geth and coreth

PR #172, opened 2022-07-28
*: add simulator e2e tests

gyuho:
```
Load simulator for subnet-evm + C-chain

Usage:
  simulator [flags]

Flags:
  -f, --base-fee uint                  Base fee (default 25)
  -c, --concurrency int                Concurrency (default 10)
  -e, --endpoints strings              If non-empty, it loads the endpoints from the YAML and overwrites --config
  -h, --help                           help for simulator
  -k, --keys string                    Directory for key files (default ".simulator/keys")
  -o, --network-runner-output string   If non-empty, it loads the endpoints from the YAML and overwrites --config
  -p, --priority-fee uint              Base fee (default 1)
  -t, --timeout duration               Duration to run simulator (default 1m0s)
```

Adding my old e2e tests to spin up a cluster https://github.com/ava-labs/subnet-evm/blob/fed7e3c2649f328f775821046a1d7693c78aa768/tests/e2e/e2e_test.go, and integrate with simulator.

PR #182, opened 2022-08-03
worker: support dynamic-fee-tx-data-size configurable

gyuho:

c.f., https://github.com/ava-labs/subnet-evm/pull/172

### DIFF
--- a/cmd/simulator/worker/worker.go
+++ b/cmd/simulator/worker/worker.go
@@ -184,7 +184,7 @@ func (w *worker) waitForBalance(ctx context.Context, stdout bool, minBalance *bi
 	return ctx.Err()
 }
 
-func (w *worker) sendTx(ctx context.Context, recipient common.Address, value *big.Int) error {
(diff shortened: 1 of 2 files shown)


cmd/simulator/worker/worker.go
```diff
@@ -194,7 +194,7 @@ func (w *worker) sendTx(ctx context.Context, recipient common.Address, value *bi
 			GasFeeCap: feeCap,
 			GasTipCap: priorityFee,
 			Value:     value,
-			Data:      []byte{},
+			Data:      make([]byte, dynamicFeeTxDataN),
```
ceyonur, line 197: just speculating: should we set random bytes here?

gyuho: Per @patrick-ogrady and for the sake of testing, zero bytes should be enough (testing the gas limit with simulated loads).

gyuho: Unless we do some processing with `Data` field in subnet-evm?

PR #184, opened 2022-08-05
Bump go version to 1.18

aaronbuchwald:
This PR bumps the minimum golang version for building subnet-evm to v1.18.1.

### DIFF
--- a/.github/workflows/lint-tests-release.yml
+++ b/.github/workflows/lint-tests-release.yml
@@ -26,7 +26,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        go: ["1.17"]
+        go: ["1.18"]
         os: [ubuntu-20.04]
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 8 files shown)


darioush (approved):

ceyonur (approved):

hexfusion:
Do we need to bump this to 1.18? 

https://github.com/ava-labs/subnet-evm/blob/master/.github/workflows/lint-tests-release.yml#L29

aaronbuchwald:
> ~Do we need to bump this to 1.18?~
> 
> https://github.com/ava-labs/subnet-evm/blob/master/.github/workflows/lint-tests-release.yml#L29
> 
> disregard I just noticed the test below shows 1.17
> 
> image

Good catch, this was actually the cause of the issue. The unit test with the name Golang Unit Tests v1.17 was marked as a required status check to merge to master. Since it was using the matrix option to set the name of the test, when the version changed to v1.18 it didn't recognize the new golang version unit test as the expected and required github action to pass CI. I updated the required status check, so that it is now marked as ready to merge. Nice catch!

Note: this will mean that other PRs will have the same blocked status until they rebase off of master.

PR #186, opened 2022-08-08
fix nil assert state functions panic in test

ceyonur:
`assertState` can throw panic if no such function is defined in tests. This PR fixes this behavior by adding a nil check


aaronbuchwald (approved):
LGTM - good idea

darioush (approved):
LGTM. we could also be more explicit with a boolean like `skipAssertState` to avoid a passing test if we forget to pass `assertState`, but that can be overkill too.

PR #187, opened 2022-08-08
generate FeeConfigManagerPC and add tests

ceyonur:
This PR is test-only and should not be merged. The PR includes a generated precompile `fee_config_managerpc` which has same behavior as `fee_config_manager` precompile. The PR also contains an unit test to check if the generated precompile works as expected.

You can re-create the precompile with 
```
go run ./cmd/precompilegen/main.go --out ./precompile/fee_config_managerpc.go --abi ./IFeeManagerPC.abi --type FeeConfigManagerPC
```


contract-examples/contracts/IFeeManagerPC.sol
```diff
@@ -0,0 +1,43 @@
...
+            uint256 blockGasCostStep
+        );
+
+    // Get the last block number changed the fee config from the contract storage
+    function getFeeConfigPCLastChangedAt()
+        external
+        view
+        returns (uint256 blockNumber);
+
+    function getTestPC(uint256 testt)
```
darioush, line 39: is this intentionally called `testt`?

ceyonur: nope but it really doesn't matter since that contract is just for the test. (or should i say `testt`)

aaronbuchwald:
Closing since this was opened as an example

PR #198, opened 2022-08-15
refactor logger & add json log format

### LINKED ISSUE #196: add json log format
We have added json log format to coreth. We can port those changes to Subnet-EVM ad well.

ceyonur:
Ports coreth log init and adds json log format. This one also removes MultiHandler with `vm.ctx.Log` since we don't support `io.Writer` in plugin context loggers. 

closes #196


plugin/evm/config.go
```diff
@@ -35,6 +35,7 @@ const (
 	defaultPriorityRegossipTxsPerAddress          = 16
 	defaultOfflinePruningBloomFilterSize   uint64 = 512 // Default size (MB) for the offline pruner to use
 	defaultLogLevel                               = "info"
+	defaultLogJSONFormat                          = false
```
aaronbuchwald, line 38:
```suggestion
	defaultLogJSONFormat                          = true
```

aaronbuchwald (approved):
LGTM - we should update to enable JSON logging by default

plugin/evm/config.go
ceyonur: coreth/avalanchego logs are not json by default. making this json by default would seem a little weird among others.

plugin/evm/vm.go
```diff
@@ -243,14 +209,20 @@ func (vm *VM) Initialize(
...
 	}
 
-	// Set log level
-	logLevel, err := log.LvlFromString(vm.config.LogLevel)
+	vm.ctx = ctx
+
+	// Create logger
+	alias, err := vm.ctx.BCLookup.PrimaryAlias(vm.ctx.ChainID)
+	if err != nil {
+		alias = vm.ctx.ChainID.String()
```
darioush, line 217: can we log an error here (or is this expected to occur as a fallback and in that case can we clarify with a comment)?

darioush (approved):

plugin/evm/config.go
aaronbuchwald: Ah ok, I misunderstood something @StephenButtolph mentioned and thought there was a request to switch everything to use JSON logs by default.

plugin/evm/vm.go
ceyonur: done

PR #214, opened 2022-08-19
Avalanchego bump anr

darioush:


tests/e2e/runner/runner.go
```diff
@@ -46,12 +47,28 @@ func GetClient() client.Client {
...
+	// Create the logger
+	logLevel, err := logging.ToLevel(networkRunnerLogLevel)
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+
+	logFactory := logging.NewFactory(logging.Config{
+		DisplayLevel: logLevel,
+		LogLevel:     logLevel,
```
felipemadero, line 60: probably don't want to create a log file. in this case set this to logging.Off

tests/e2e/runner/runner.go
```diff
@@ -46,12 +47,28 @@ func GetClient() client.Client {
 func InitializeRunner(execPath_ string, grpcEp string, networkRunnerLogLevel string) error {
 	execPath = execPath_
 
-	var err error
+	// TODO: is this logic to create the logger correct?
```
felipemadero, line 50: seems ok IMHO

felipemadero (commented):
thanks for taking care of this update

tests/e2e/runner/runner.go
```diff
@@ -46,12 +47,25 @@ func GetClient() client.Client {
...
-	var err error
+	// Create the logger
+	logLevel, err := logging.ToLevel(networkRunnerLogLevel)
+	if err != nil {
+		return err
+	}
+
+	logFactory := logging.NewFactory(logging.Config{
+		DisplayLevel: logLevel,
+		LogLevel:     logging.Off,
```
darioush, line 58: can we add a comment why should this be `logging.Off`? does not seem intuitive

aaronbuchwald: Added

felipemadero: just want the logs on stdout, and want to prevent the client logger to create a log file, which is not needed in CI, and also usually not needed on local executions

felipemadero (approved):
LGTM

aaronbuchwald (approved):

PR #221, opened 2022-08-20
Add comment to clarify log line won't index out of bounds

aaronbuchwald:
This PR adds a small comment to state that a suspicious line will not cause an index out of bounds because the arguments will be normalized in advance.

### DIFF
--- a/plugin/evm/log.go
+++ b/plugin/evm/log.go
@@ -75,7 +75,7 @@ func SubnetEVMJSONFormat(alias string) log.Format {
 			if !ok {
 				props[errorKey] = fmt.Sprintf("%+v is not a string key", r.Ctx[i])
 			} else {
-				props[k] = formatJSONValue(r.Ctx[i+1]) // The number of arguments is normalized prior to this to prevent index out of bounds
+				props[k] = formatJSONValue(r.Ctx[i+1]) // The number of arguments is normalized from the geth logger to prevent an index out of bounds here
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


plugin/evm/log.go
```diff
@@ -75,7 +75,7 @@ func SubnetEVMJSONFormat(alias string) log.Format {
 			if !ok {
 				props[errorKey] = fmt.Sprintf("%+v is not a string key", r.Ctx[i])
 			} else {
-				props[k] = formatJSONValue(r.Ctx[i+1])
+				props[k] = formatJSONValue(r.Ctx[i+1]) // The number of arguments is normalized prior to this to prevent index out of bounds
```
ceyonur, line 78: should we add a line says that's being done by geth logger?

aaronbuchwald, line 78:
```suggestion
				props[k] = formatJSONValue(r.Ctx[i+1]) // The number of arguments is normalized from the geth logger to prevent an index out of bounds here
```

plugin/evm/log.go
```diff
@@ -75,7 +75,7 @@ func SubnetEVMJSONFormat(alias string) log.Format {
 			if !ok {
 				props[errorKey] = fmt.Sprintf("%+v is not a string key", r.Ctx[i])
 			} else {
-				props[k] = formatJSONValue(r.Ctx[i+1])
+				props[k] = formatJSONValue(r.Ctx[i+1]) // The number of arguments is normalized from the geth logger to prevent an index out of bounds here
```
ceyonur, line 78:
```suggestion
                                // The number of arguments is normalized from the geth logger
                                // to prevent an index out of bounds here
				props[k] = formatJSONValue(r.Ctx[i+1])
```

ceyonur: uh hope this looks good

ceyonur: copied same comment from coreth. hope this looks good.

PR #226, opened 2022-08-24
add initial precompile configs and handle them

ceyonur:
Adds initial precomple configs to existing precompile so they can be activated and perform their operations through configs, without requiring an interaction with actual contracts.


precompile/allow_list.go
```diff
@@ -43,12 +43,17 @@ var (
...
 type AllowListConfig struct {
-	AllowListAdmins []common.Address `json:"adminAddresses"`
+	AllowListAdmins  []common.Address `json:"adminAddresses"`
+	EnabledAddresses []common.Address `json:"enabledAddresses,omitempty"` // initial enabled addresses
 }
 
 // Configure initializes the address space of [precompileAddr] by initializing the role of each of
 // the addresses in [AllowListAdmins].
 func (c *AllowListConfig) Configure(state StateDB, precompileAddr common.Address) {
+	// First set enabled roles so these can be upgraded to admin addresses below.
```
aaronbuchwald, line 53: Hmm good call, should we be more strict here and require that there are no overlaps between the two lists?

precompile/fee_config_manager.go
```diff
@@ -56,6 +56,7 @@ var (
 type FeeConfigManagerConfig struct {
 	AllowListConfig // Config for the fee config manager allow list
 	UpgradeableConfig
+	InitialFeeConfig *commontype.FeeConfig `json:"initialFeeConfig,omitempty"` // initial fee config to be immediately activated
```
aaronbuchwald, line 59: Should we verify that this is valid on startup, so that if there's an invalid fee config then it will cause the node to refuse to start when it's first passed in as opposed to causing a panic when the upgrade activates?

ceyonur: yeah good point

precompile/allow_list.go
ceyonur: added verify to allowlist

precompile/fee_config_manager.go
ceyonur: done with Verify()

params/config.go
```diff
@@ -287,7 +287,7 @@ func (c *ChainConfig) Verify() error {
 	}
 
 	// Verify the precompile upgrades are internally consistent given the existing chainConfig.
-	if err := c.VerifyPrecompileUpgrades(); err != nil {
+	if err := c.verifyPrecompileUpgrades(); err != nil {
```
aaronbuchwald, line 290: good call

params/precompile_config_test.go
```diff
@@ -57,30 +59,132 @@ func TestValidateWithChainConfig(t *testing.T) {
...
+			upgrade: PrecompileUpgrade{
+				FeeManagerConfig: &precompile.FeeConfigManagerConfig{
+					AllowListConfig:   precompile.AllowListConfig{AllowListAdmins: admins},
+					UpgradeableConfig: precompile.UpgradeableConfig{BlockTimestamp: big.NewInt(3)},
+					InitialFeeConfig: &commontype.FeeConfig{
+						GasLimit: big.NewInt(-1),
+					},
+				},
+			},
+			expectedError: "gasLimit = -1 cannot be less than or equal to 0",
```
aaronbuchwald, line 164: Can we add a test case for 0 as well?

precompile/allow_list.go
```diff
@@ -70,6 +72,28 @@ func (c *AllowListConfig) Equal(other *AllowListConfig) bool {
...
+				return fmt.Errorf("cannot set address %s as both admin and enabled", adminAddr)
+			}
+		}
+	}
+	return nil
+}
+
+// Enum constants for valid AllowListRole
+type AllowListRole common.Hash
+
```
aaronbuchwald, line 96: Can we move all of the AllowList type, functions and the declarations of the valid ones into one file?

I preferred the type declaration right above the declaration of the enum values, but I think it's even cleaner if we move it to a separate file

precompile/allow_list.go
aaronbuchwald: No that we added this within `Verify()` can we remove this comment?

precompile/allow_list.go
```diff
@@ -125,6 +149,10 @@ func setAllowListRole(stateDB StateDB, precompileAddr, address common.Address, r
 	// Generate the state key for [address]
 	addressKey := address.Hash()
 	// Assign [role] to the address
+	// This stores the [role] in the contract storage with address [precompileAddr]
+	// and [addressKey] hash. It means that any reusage of the [addressKey] for different value
+	// conflicts with the same slot [role] is stored.
+	// Precompile implementations must use a different key than [addressKey]
```
aaronbuchwald, line 155: Good point, since this is something that a user of the `AllowList` needs to be aware of, can we add a comment to the AllowList type above?

In retrospect, I guess we should have used an allow list prefix, but if other implementations take that approach then it won't matter.

precompile/contract_native_minter.go
```diff
@@ -35,6 +36,7 @@ var (
 type ContractNativeMinterConfig struct {
 	AllowListConfig
 	UpgradeableConfig
+	InitialMint map[common.Address]*math.HexOrDecimal256 `json:"initialMint,omitempty"` // initial mint config to be immediately minted
```
aaronbuchwald, line 39: why are we using `*math.HexOrDecimal256` instead of `*big.Int` here?

precompile/stateful_precompile_config.go
```diff
@@ -35,6 +35,9 @@ type StatefulPrecompileConfig interface {
 	// Contract returns a thread-safe singleton that can be used as the StatefulPrecompiledContract when
 	// this config is enabled.
 	Contract() StatefulPrecompiledContract
+	// Verify is called before configuring stateful precompile config. It is important to capture any invalid configuration
+	// before activation.
```
aaronbuchwald, line 39:
```suggestion
	// Verify is called before configuring stateful precompile config. Configure can assume the Config has passed verification.
```

precompile/allow_list.go
```diff
@@ -70,6 +72,28 @@ func (c *AllowListConfig) Equal(other *AllowListConfig) bool {
 	return true
 }
 
+// Verify returns an error if there is an overlapping address between admins and enableds
+func (c *AllowListConfig) Verify() error {
+	// check if both lists are empty
```
aaronbuchwald, line 77: I think we can improve this by either:

1. Switch to returning early if one of the lists is empty since there can be no overlaps and move the more complex functionality out of the body of the conditional
2. This is cheap and only called once on startup, so we could also remove the check entirely

precompile/allow_list.go
aaronbuchwald: Can we also add unit tests for verify within the `precompile` package?

precompile/allow_list.go
```diff
@@ -43,12 +40,17 @@ var (
 
 // AllowListConfig specifies the initial set of allow list admins.
 type AllowListConfig struct {
-	AllowListAdmins []common.Address `json:"adminAddresses"`
+	AllowListAdmins  []common.Address `json:"adminAddresses"`
+	EnabledAddresses []common.Address `json:"enabledAddresses,omitempty"` // initial enabled addresses
```
aaronbuchwald, line 44:
```suggestion
	EnabledAddresses []common.Address `json:"enabledAddresses"` // initial enabled addresses
```

Adding `omitempty` here will only result in setting `EnabledAddresses` to `nil` instead of an empty list. Since we don't differentiate between these two within the allow list, we should keep it consistent with `AllowListAdmins`.

params/precompile_config.go
```diff
@@ -66,7 +66,7 @@ func (p *PrecompileUpgrade) getByKey(key precompileKey) (precompile.StatefulPrec
 // - the specified blockTimestamps must be compatible with those
 //   specified in the chainConfig by genesis.
 // - check a precompile is disabled before it is re-enabled
-func (c *ChainConfig) VerifyPrecompileUpgrades() error {
+func (c *ChainConfig) verifyPrecompileUpgrades() error {
```
darioush, line 69: nit: should update comment to `verifyPrecompileUpgrades`

params/precompile_config_test.go
```diff
@@ -7,19 +7,21 @@ import (
...
 	"testing"
 
+	"github.com/ava-labs/subnet-evm/commontype"
 	"github.com/ava-labs/subnet-evm/precompile"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestValidateWithChainConfig(t *testing.T) {
```
darioush, line 17: should we rename to `TestVerify` or `TestVerifyChainConfig`?

params/precompile_config_test.go
```diff
@@ -7,19 +7,21 @@ import (
...
-	config := &ChainConfig{
-		PrecompileUpgrade: PrecompileUpgrade{
-			TxAllowListConfig: &precompile.TxAllowListConfig{
-				UpgradeableConfig: precompile.UpgradeableConfig{
-					BlockTimestamp: big.NewInt(2),
-				},
+	baseConfig := *SubnetEVMDefaultChainConfig
+	config := &baseConfig
+	config.PrecompileUpgrade = PrecompileUpgrade{
+		TxAllowListConfig: &precompile.TxAllowListConfig{
```
darioush, line 22: nit: can we use `precompile.NewTxAllowListConfig` here for brevity?

params/precompile_config_test.go
```diff
@@ -57,30 +59,132 @@ func TestValidateWithChainConfig(t *testing.T) {
...
+			if tt.expectedError == "" {
+				require.NoError(err)
+			} else {
+				require.ErrorContains(err, tt.expectedError)
+			}
+		})
+	}
 }
 
 func TestValidateRequiresSortedTimestamps(t *testing.T) {
```
darioush, line 184: same idea here re. renaming `Validate` to `Verify`

precompile/allow_list.go
```diff
@@ -70,6 +72,28 @@ func (c *AllowListConfig) Equal(other *AllowListConfig) bool {
 	return true
 }
 
+// Verify returns an error if there is an overlapping address between admins and enableds
```
darioush, line 75:
```suggestion
// Verify returns an error if there is an overlapping address between admin and enabled roles
```

precompile/contract_native_minter.go
ceyonur: To make it similar with existing alloc mapping in genesis.

precompile/allow_list.go
ceyonur: done

params/precompile_config_test.go
ceyonur: done

precompile/allow_list.go
ceyonur: You mean moving `AllowListRole` declaration and funcs into a new file? Or you mean moving all of AllowList types (like ContractDeployer/TxAllowList) into one file?

precompile/allow_list.go
ceyonur: added to precompile template

precompile/allow_list.go
aaronbuchwald: Yup I meant moving `AllowListRole` declaration and functions into a new file

core/stateful_precompile_test.go
```diff
@@ -581,6 +583,27 @@ func TestContractNativeMinterRun(t *testing.T) {
...
+				return precompile.PackReadAllowList(noRoleAddr)
+			},
+			suppliedGas: precompile.ReadAllowListGasCost,
+			readOnly:    false,
+			expectedRes: common.Hash(precompile.AllowListNoRole).Bytes(),
+			assertState: func(t *testing.T, state *state.StateDB) {
+				res := precompile.GetContractNativeMinterStatus(state, allowAddr)
+				assert.Equal(t, precompile.AllowListEnabled, res)
+
+				assert.Equal(t, common.Big2, state.GetBalance(allowAddr), "expected minted funds")
```
aaronbuchwald, line 604: I was confused by this address being required to be enabled here since it's not included in the config. Should we switch from performing some setup at the start of `t.Run` to always including the config in each of these tests so that it can configure the state as necessary?

precompile/allow_list.go
ceyonur: I believe we do first one by checking `if len(c.EnabledAddresses) != 0 && len(c.AllowListAdmins) != 0 ` 
but yeah let's return early for readability

params/precompile_config_test.go
ceyonur: done

params/precompile_config_test.go
ceyonur: done

params/precompile_config_test.go
ceyonur: done

precompile/allow_list.go
```diff
@@ -70,44 +72,25 @@ func (c *AllowListConfig) Equal(other *AllowListConfig) bool {
...
-func (s AllowListRole) IsNoRole() bool {
-	switch s {
-	case AllowListNoRole:
-		return true
-	default:
-		return false
+	enabledMap := make(map[common.Address]struct{})
+	for _, enabledAddr := range c.EnabledAddresses {
+		if _, ok := enabledMap[enabledAddr]; !ok {
+			enabledMap[enabledAddr] = struct{}{}
```
aaronbuchwald, line 84: Can we ensure that there are no repeats within either list as well?

params/precompile_config_test.go
```diff
@@ -57,30 +55,147 @@ func TestValidateWithChainConfig(t *testing.T) {
...
+				},
+				{
+					TxAllowListConfig: precompile.NewDisableTxAllowListConfig(big.NewInt(2)),
+				},
+				{
+					TxAllowListConfig: &precompile.TxAllowListConfig{
+						AllowListConfig:   precompile.AllowListConfig{AllowListAdmins: admins, EnabledAddresses: admins},
+						UpgradeableConfig: precompile.UpgradeableConfig{BlockTimestamp: big.NewInt(3)},
+					},
+				},
```
aaronbuchwald, line 95: can we use constructors to set up all of the precompile configs within the precompile config validation tests?

precompile/config_test.go
```diff
@@ -0,0 +1,78 @@
...
+func TestVerifyPrecompileUpgrades(t *testing.T) {
+	admins := []common.Address{{1}}
+	tests := []struct {
+		name          string
+		config        StatefulPrecompileConfig
+		expectedError string
+	}{
+		{
+			name: "invalid allow list config in tx allowlist",
+			config: &TxAllowListConfig{
```
aaronbuchwald, line 24: Can we use the precompile config constructors in this file as well?

core/stateful_precompile_test.go
ceyonur: yup removed unneccessary ones

precompile/allow_list.go
ceyonur: hm that wouldn't be back-compatible if there are any chains activated a precompile with duplicate admin addresses.

accounts/abi/bind/precompile_template.go
```diff
@@ -192,6 +192,20 @@ func (c *{{.Contract.Type}}Config) Contract() StatefulPrecompiledContract {
 	return {{.Contract.Type}}Precompile
 }
 
+// Verify tries to verify {{.Contract.Type}}Config and returns an error accordingly.
+func (c *{{.Contract.Type}}Config) Verify() error {
+	{{if .Contract.AllowList}}
+	// Verify AllowList first
+	if err := c.AllowListConfig.Verify(); err != nil{
```
darioush, line 199:
```suggestion
	if err := c.AllowListConfig.Verify(); err != nil {
```

precompile/tx_allow_list.go
```diff
@@ -26,10 +26,13 @@ type TxAllowListConfig struct {
 }
 
 // NewTxAllowListConfig returns a config for a network upgrade at [blockTimestamp] that enables
-// TxAllowList with the given [admins] as members of the allowlist.
-func NewTxAllowListConfig(blockTimestamp *big.Int, admins []common.Address) *TxAllowListConfig {
+// TxAllowList with the given [admins] and [enableds] as members of the allowlist.
+func NewTxAllowListConfig(blockTimestamp *big.Int, admins []common.Address, enableds []common.Address) *TxAllowListConfig {
```
darioush, line 30: nice

precompile/fee_config_manager.go
```diff
@@ -96,8 +101,14 @@ func (c *FeeConfigManagerConfig) Equal(s StatefulPrecompileConfig) bool {
 // Configure configures [state] with the desired admins based on [c].
 func (c *FeeConfigManagerConfig) Configure(chainConfig ChainConfig, state StateDB, blockContext BlockContext) {
 	// Store the initial fee config into the state when the fee config manager activates.
-	if err := StoreFeeConfig(state, chainConfig.GetFeeConfig(), blockContext); err != nil {
-		panic(fmt.Sprintf("fee config should have been verified in genesis: %s", err))
+	if c.InitialFeeConfig != nil {
+		if err := StoreFeeConfig(state, *c.InitialFeeConfig, blockContext); err != nil {
+			panic(fmt.Sprintf("invalid feeConfig provided: %s", err))
```
darioush, line 106: can we add a comment that explains this panic will not occur since verify is passed?

precompile/contract_native_minter.go
```diff
@@ -35,14 +36,19 @@ var (
...
 }
 
 // NewContractNativeMinterConfig returns a config for a network upgrade at [blockTimestamp] that enables
-// ContractNativeMinter with the given [admins] as members of the allowlist.
-func NewContractNativeMinterConfig(blockTimestamp *big.Int, admins []common.Address) *ContractNativeMinterConfig {
+// ContractNativeMinter with the given [admins] and [enableds] as members of the allowlist with [initialMint] as initial mint operation.
```
darioush, line 43:
```suggestion
// ContractNativeMinter with the given [admins] and [enableds] as members of the allowlist. Also mints balances according to [initialMint] when the upgrade is activated.
```

precompile/contract_native_minter.go
```diff
@@ -64,6 +70,13 @@ func (c *ContractNativeMinterConfig) Address() common.Address {
 
 // Configure configures [state] with the desired admins based on [c].
 func (c *ContractNativeMinterConfig) Configure(_ ChainConfig, state StateDB, _ BlockContext) {
+	for to, amount := range c.InitialMint {
+		if amount != nil {
+			bigIntAmount := (*big.Int)(amount)
+			state.AddBalance(to, bigIntAmount)
+		}
+	}
+
```
darioush, line 79: should check `InitialMint` in Equals as well

precompile/fee_config_manager.go
```diff
@@ -96,8 +101,14 @@ func (c *FeeConfigManagerConfig) Equal(s StatefulPrecompileConfig) bool {
 // Configure configures [state] with the desired admins based on [c].
 func (c *FeeConfigManagerConfig) Configure(chainConfig ChainConfig, state StateDB, blockContext BlockContext) {
 	// Store the initial fee config into the state when the fee config manager activates.
-	if err := StoreFeeConfig(state, chainConfig.GetFeeConfig(), blockContext); err != nil {
-		panic(fmt.Sprintf("fee config should have been verified in genesis: %s", err))
+	if c.InitialFeeConfig != nil {
```
darioush, line 104: we should check `InitialFeeConfig` in Equals as well

precompile/contract_deployer_allow_list.go
```diff
@@ -23,10 +23,13 @@ type ContractDeployerAllowListConfig struct {
 }
 
 // NewContractDeployerAllowListConfig returns a config for a network upgrade at [blockTimestamp] that enables
-// ContractDeployerAllowList with the given [admins] as members of the allowlist.
-func NewContractDeployerAllowListConfig(blockTimestamp *big.Int, admins []common.Address) *ContractDeployerAllowListConfig {
+// ContractDeployerAllowList with the given [admins] and [enableds] as members of the allowlist.
```
darioush, line 26:
```suggestion
// ContractDeployerAllowList with [admins] and [enableds] as members of the allowlist.
```

precompile/allow_list.go
```diff
@@ -70,44 +72,25 @@ func (c *AllowListConfig) Equal(other *AllowListConfig) bool {
...
-// Valid returns true iff [s] represents a valid role.
-func (s AllowListRole) Valid() bool {
-	switch s {
-	case AllowListNoRole, AllowListEnabled, AllowListAdmin:
-		return true
-	default:
-		return false
+// Verify returns an error if there is an overlapping address between admin and enabled roles
+func (c *AllowListConfig) Verify() error {
+	// check if both lists are empty
```
darioush, line 77:
```suggestion
	// return early if either list is empty
```

precompile/contract_native_minter.go
```diff
@@ -64,6 +70,13 @@ func (c *ContractNativeMinterConfig) Address() common.Address {
 
 // Configure configures [state] with the desired admins based on [c].
 func (c *ContractNativeMinterConfig) Configure(_ ChainConfig, state StateDB, _ BlockContext) {
+	for to, amount := range c.InitialMint {
+		if amount != nil {
+			bigIntAmount := (*big.Int)(amount)
+			state.AddBalance(to, bigIntAmount)
```
darioush, line 76: should we check for overflow?

core/stateful_precompile_test.go
```diff
@@ -755,7 +755,14 @@ func TestContractNativeMinterRun(t *testing.T) {
 			precompile.SetContractNativeMinterStatus(state, adminAddr, precompile.AllowListAdmin)
 			precompile.SetContractNativeMinterStatus(state, allowAddr, precompile.AllowListEnabled)
 			precompile.SetContractNativeMinterStatus(state, noRoleAddr, precompile.AllowListNoRole)
+			assert.Equal(t, precompile.AllowListAdmin, precompile.GetContractNativeMinterStatus(state, adminAddr))
+			assert.Equal(t, precompile.AllowListEnabled, precompile.GetContractNativeMinterStatus(state, allowAddr))
```
darioush, line 759: nit: should we rename to `enabledAddr` for consistency?

darioush:
todo: documentation PR for initial config options

params/precompile_config.go
ceyonur: done

precompile/fee_config_manager.go
ceyonur: done

precompile/contract_native_minter.go
ceyonur: yeah nice catch 👍

precompile/contract_native_minter.go
ceyonur: hm nice question, I don't normally see any overflow guard for big ints. what do you think? @aaronbuchwald

aaronbuchwald: We should add overflow guards to protect against this since this is up to the user now.

Every place where we do this, it's up to a trusted contract that can mint funds, but we should add overflow checks within the native minter to make sure that nobody minting very aggressively runs into overflow issues.

aaronbuchwald: Created an issue here to track: https://github.com/ava-labs/subnet-evm/issues/248. We have the option to either add overflow guards or make sure that it's clearly documented (and that it's a reasonable expectation) to say that the user of the native minter precompile will ensure there is never on the order of enough funds to overflow a uint256.

I think it should actually be reasonable to require that of the user of the native minter precompile.

precompile/allow_list.go
```diff
@@ -59,55 +61,44 @@ func (c *AllowListConfig) Equal(other *AllowListConfig) bool {
...
 	}
-	if len(c.AllowListAdmins) != len(other.AllowListAdmins) {
+	if !checkList(c.AllowListAdmins, other.AllowListAdmins) {
+		return false
+	}
+
+	return checkList(c.EnabledAddresses, other.EnabledAddresses)
+}
+
+func checkList(current []common.Address, other []common.Address) bool {
```
aaronbuchwald, line 71: Can we add a comment and update the name to `isEqualAddressLists` ? Check leaves it a bit vague as to what is being checked.

precompile/allow_list.go
```diff
@@ -59,55 +61,44 @@ func (c *AllowListConfig) Equal(other *AllowListConfig) bool {
 	if other == nil {
 		return false
 	}
-	if len(c.AllowListAdmins) != len(other.AllowListAdmins) {
+	if !checkList(c.AllowListAdmins, other.AllowListAdmins) {
```
aaronbuchwald, line 64: can we update this to use the syntax `if condition := checkList(...,...); !condition {`

precompile/fee_config_manager.go
```diff
@@ -90,14 +95,31 @@ func (c *FeeConfigManagerConfig) Equal(s StatefulPrecompileConfig) bool {
...
 	}
-	return c.UpgradeableConfig.Equal(&other.UpgradeableConfig) && c.AllowListConfig.Equal(&other.AllowListConfig)
+	eq := c.UpgradeableConfig.Equal(&other.UpgradeableConfig) && c.AllowListConfig.Equal(&other.AllowListConfig)
+	if !eq {
+		return false
+	}
+
+	if c.InitialFeeConfig == nil {
+		return other.InitialFeeConfig == nil
+	}
```
aaronbuchwald, line 105: Can we switch this to use a length check? Otherwise, I believe that this will differentiate between an empty map and a nil map.

Can we add a test to confirm the behavior for that case?

core/stateful_precompile_test.go
```diff
@@ -543,8 +520,9 @@ func TestContractNativeMinterRun(t *testing.T) {
 	}
 
 	adminAddr := common.HexToAddress("0x8db97C7cEcE249c2b98bDC0226Cc4C2A57BF52FC")
-	allowAddr := common.HexToAddress("0xAb5801a7D398351b8bE11C439e05C5B3259aeC9B")
+	enabledAddr := common.HexToAddress("0xAb5801a7D398351b8bE11C439e05C5B3259aeC9B")
 	noRoleAddr := common.HexToAddress("0xF60C45c607D0f41687c94C314d300f483661E13a")
+	testAddr := common.BigToAddress(common.Big2)
```
aaronbuchwald, line 525: Can we switch to using the same format of `common.HexToAddress(...)` to create each of the addresses?

commontype/fee_config.go
```diff
@@ -81,6 +82,22 @@ func (f *FeeConfig) Verify() error {
 	return f.checkByteLens()
 }
 
+// Equal checks if given [other] is same with this FeeConfig.
+func (f *FeeConfig) Equal(other *FeeConfig) bool {
```
aaronbuchwald, line 86: Can we add unit tests for each of these cases as well?

precompile/config_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+func TestVerifyPrecompileUpgrades(t *testing.T) {
+	admins := []common.Address{{1}}
+	tests := []struct {
+		name          string
+		config        StatefulPrecompileConfig
+		expectedError string
+	}{
+		{
+			name:          "invalid allow list config in tx allowlist",
```
aaronbuchwald, line 23: These all use `NewTxAllowListConfig`. Can we update this to use each allow list based precompile as well as for the allow list config itself?

We can change the type for `config` above to just `Verify() error`

aaronbuchwald: Can we also add test cases just for the allow list config for:

- both are empty and it passes
- both are non-empty with no overlaps and they pass

precompile/config_test.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	"math/big"
+	"testing"
+
+	"github.com/ava-labs/subnet-evm/commontype"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/stretchr/testify/require"
+)
+
+func TestVerifyPrecompileUpgrades(t *testing.T) {
```
aaronbuchwald, line 15: Can we add a similar test suite to verify that the `Equal` function works as expected for each of the different config types as well?

precompile/allow_list.go
ceyonur: done

precompile/fee_config_manager.go
ceyonur: this is not a map, it's a struct. were you looking at this one?
https://github.com/ava-labs/subnet-evm/blob/2d3a51702ba1bfe1c27dcb58d9d634e9c4dc9a4f/precompile/contract_native_minter.go#L101-L103

core/stateful_precompile_test.go
ceyonur: done

precompile/config_test.go
ceyonur: done

precompile/contract_native_minter.go
```diff
@@ -79,7 +93,28 @@ func (c *ContractNativeMinterConfig) Equal(s StatefulPrecompileConfig) bool {
...
+			return false
+		}
+		bigIntAmount := (*big.Int)(amount)
+		bigIntVal := (*big.Int)(val)
+		if !utils.BigNumEqual(bigIntAmount, bigIntVal) {
+			return false
+		}
+	}
+
+	return true
```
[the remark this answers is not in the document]
ceyonur: done

commontype/fee_config.go
ceyonur: added

precompile/config_test.go
ceyonur: done

precompile/fee_config_manager.go
aaronbuchwald: Yes I was, my bad.

commontype/fee_config_test.go
```diff
@@ -0,0 +1,140 @@
...
+			config:        func() *FeeConfig { c := validFeeConfig; c.BaseFeeChangeDenominator = big.NewInt(0); return &c }(),
+			expectedError: "baseFeeChangeDenominator = 0 cannot be less than or equal to 0",
+		},
+		{
+			name:          "invalid MinBlockGasCost in FeeConfig",
+			config:        func() *FeeConfig { c := validFeeConfig; c.MinBlockGasCost = big.NewInt(-1); return &c }(),
+			expectedError: "minBlockGasCost = -1 cannot be less than 0",
+		},
+		{
+			name:          "valid  FeeConfig",
```
darioush, line 63:
```suggestion
			name:          "valid FeeConfig",
```

precompile/allow_list.go
```diff
@@ -59,55 +61,45 @@ func (c *AllowListConfig) Equal(other *AllowListConfig) bool {
 	if other == nil {
 		return false
 	}
-	if len(c.AllowListAdmins) != len(other.AllowListAdmins) {
+	if !areEqualAddressLists(c.AllowListAdmins, other.AllowListAdmins) {
+		return false
+	}
+
+	return areEqualAddressLists(c.EnabledAddresses, other.EnabledAddresses)
```
darioush, line 68: doesn't this make it impossible to change the order of addresses?

precompile/allow_list.go
```diff
@@ -43,12 +40,17 @@ var (
...
 type AllowListConfig struct {
-	AllowListAdmins []common.Address `json:"adminAddresses"`
+	AllowListAdmins  []common.Address `json:"adminAddresses"`
+	EnabledAddresses []common.Address `json:"enabledAddresses"` // initial enabled addresses
 }
 
 // Configure initializes the address space of [precompileAddr] by initializing the role of each of
 // the addresses in [AllowListAdmins].
 func (c *AllowListConfig) Configure(state StateDB, precompileAddr common.Address) {
+	// First set enabled roles so these can be upgraded to admin addresses below.
```
darioush, line 50: should we remove this comment since overlap is no longer allowed?

precompile/contract_native_minter.go
darioush: can we add a comment about this decision here?

precompile/allow_list.go
ceyonur: yes. do you think we should flex this?

darioush: We could also add a function to the interface that returns a canonical representation that can be used in hashing.

ceyonur: I don't think we need to sort these lists. I think it's ok to expect the same order every time since `hash(list[b,c,a]) != hash(list[a,b,c])`. This is because the byte representation of those lists are not equal. JSON arrays are also ordered so users should intuitively know that providing any different list can cause issues.

aaronbuchwald: Since `Equal` will be used by the node locally to determine the validity of an upgrade config, if we change this in the future to be "looser" or to sort these lists, then the upgrade process should still be straight forward ie.

we push an upgrade that changes the `Equal` check to to define equality to ensure that the sorted list is equal and an upgrade that would have previously been invalid (ie a list in a different order, but containing the same set of addresses) will become valid, which is fine since it's just for the node's upgrade config.

We should also make sure that we do not allow duplicates in the list. Created a ticket for this here, since I think this PR is ready to merge: https://github.com/ava-labs/subnet-evm/issues/269.

Issue #223, opened 2022-08-24
Switch from using HardHat to Foundry


aaronbuchwald:
~~This ticket is to switch from using HardHat to the Foundry tool with the goal of switching from JS to Rust dependency for security reasons.~~

Use `Foundry` cli tooling with [DS-test](https://github.com/dapphub/ds-test) instead of HardHat. Benefits include (but are not limited to):
 - removes dependency on npm/node-modules as a potential attack vector on CI
 
 Implementation plan
 - [ ] Make sure DS-test works with a local subnet
 - [ ] add docs for installing foundry locally 
 - [ ] add script(s) for running local tests locally
 - [ ] add github action (CI) for foundry tests (https://github.com/foundry-rs/foundry-toolchain)
 - [ ] Migrate precompile tests to solidity-based tests (DS-test) using Foundry/Forge for deployment 
 - [ ] deprecate HardHat tests in CI

gyuho:
https://github.com/ava-labs/subnet-evm/pull/238 is the first attempt to integrate with foundry (awesome tool btw).

And for testing precompiles, rather than hardcode individual genesis files for each case (how it's done now in the `tests/e2e/solidity`), I suggest we automate the upgrade config test using chain config (e.g., write fee config manager file and restart the process and ensures that it's upgraded), to simulate the real world operations.

aaronbuchwald:
As a first pass, we should migrate from starting an entire network  to the approach outlined in this ticket to make running e2e tests of the VM behavior (where consensus/networking does not need to be tested) https://github.com/ava-labs/subnet-evm/issues/358.

This should reduce the complexity of orchestrating the e2e tests significantly.

aaronbuchwald:
We have migrated all of the e2e precompile tests to run on a single node network as in #358 and simplified the code to construct a network from a given genesis and then execute the HardHat tests on that subnet (ref: https://github.com/ava-labs/subnet-evm/blob/master/tests/precompile/solidity/suites.go#L29-L34).

We should be able to use the code from @gyuho 's previous pass at integrating Foundry here: https://github.com/ava-labs/subnet-evm/blob/c4d0e2d6200477a3cdf5d2ca047271656a74519c/tests/e2e/precompile-upgrade/suites.go#L212

PR #238, opened 2022-08-29
*: basic solidity, precompile upgrades e2e tests with foundry

gyuho:
ref. https://github.com/ava-labs/subnet-evm/issues/223

Adding a simple counter test with foundry


gyuho:
Also removing some existing npx tests from `tests/e2e` as it's been flaky and not running in CI at the moment.

gyuho:
In the following PR, we should remove `parser` (unnecessary network-runner dependency), all genesis based upgrade tests (now done via upgrade e2e tests, no need to set up different clusters for each precompile config when we already e2e tests upgrades), `npx` dependency.

scripts/run.sh
```diff
@@ -264,21 +277,10 @@ run_simulator() {
...
 }
 
-if [[ ${SKIP_NETWORK_RUNNER_START} == false ]]; then
-  echo "running ginkgo"
-  run_ginkgo
-  # to fail the script if ginkgo failed
-  EXIT_CODE=$?
-else
-  echo "running scripts/parser/main.go"
-  go run scripts/parser/main.go \
```
gyuho, line 274: This is really hacky... we should deprecate this and just unify to `tests/e2e/e2e_test.go`...

tests/e2e/precompile-upgrade/suites.go
```diff
@@ -0,0 +1,330 @@
...
+				utils.Outf("{{magenta}}writing chain config for %q{{/}}: %q\n", name, configPath)
+				gomega.Expect(os.WriteFile(configPath, defaultChainCfgBytes, 0777)).Should(gomega.BeNil())
+
+				utils.Outf("{{magenta}}restarting the node %q{{/}}\n", name)
+				ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
+				resp, err := runnerCli.RestartNode(ctx, name)
+				cancel()
+				gomega.Expect(err).Should(gomega.BeNil())
+
+				time.Sleep(20 * time.Second)
```
aaronbuchwald, line 86: can we switch from a 20 second sleep to increasing the timeout on the health check?

tests/e2e/precompile-upgrade/suites.go
```diff
@@ -0,0 +1,330 @@
...
+				utils.Outf("{{magenta}}writing chain config for %q{{/}}: %q\n", name, upgradeJSONPath)
+				gomega.Expect(os.WriteFile(upgradeJSONPath, upgradeCfgAllowsEwoqBytes, 0777)).Should(gomega.BeNil())
+
+				utils.Outf("{{magenta}}restarting the node %q{{/}}\n", name)
+				ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
+				resp, err := runnerCli.RestartNode(ctx, name)
+				cancel()
+				gomega.Expect(err).Should(gomega.BeNil())
+
+				time.Sleep(20 * time.Second)
```
aaronbuchwald, line 169: same comment here

tests/e2e/solidity/counter/suites.go
```diff
@@ -0,0 +1,86 @@
...
+				contractAddr = strings.Replace(line, "Deployed to: ", "", 1)
+				break
+			}
+		}
+		gomega.Expect(contractAddr).ShouldNot(gomega.BeEmpty())
+		utils.Outf("{{green}}counter contract address:{{/}} %q\n", contractAddr)
+
+		utils.Outf("{{green}}set the current counter number{{/}}\n")
+		s, err = utils.RunCommand(
+			2*time.Minute,
```
aaronbuchwald, line 58: can we change from using a deadline here to passing in a context with a deadline?

tests/e2e/utils/command.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/go-cmd/cmd"
+)
+
+func RunCommand(timeout time.Duration, bin string, args ...string) (cmd.Status, error) {
```
aaronbuchwald, line 15: can we switch from taking in a timeout to taking in a ctx?

tests/e2e/precompile-upgrade/suites.go
```diff
@@ -0,0 +1,330 @@
...
+
+	"github.com/ethereum/go-ethereum/common"
+	ethcrypto "github.com/ethereum/go-ethereum/crypto"
+	ginkgo "github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
+)
+
+type chainConfig struct {
+	evm.Config
+}
```
aaronbuchwald, line 30: what's the purpose of wrapping the `evm.Config` like this? This is a bit confusing because `chainConfig` is also the term used to `*params.ChainConfig`

tests/e2e/precompile-upgrade/suites.go
```diff
@@ -0,0 +1,330 @@
...
+		gomega.Expect(runnerCli).ShouldNot(gomega.BeNil())
+
+		defaultChainCfg := chainConfig{}
+		defaultChainCfg.Config.SetDefaults()
+		defaultChainCfg.Config.LogJSONFormat = true
+		defaultChainCfgBytes, err := json.Marshal(defaultChainCfg)
+		gomega.Expect(err).Should(gomega.BeNil())
+
+		// "ewoq" key with "0x8db97C7cEcE249c2b98bDC0226Cc4C2A57BF52FC"
+		ewoqPrivKey := "56289e99c94b6912bfc12adc093c9b51124f0dc54ac7a766b2bc5ccf558d8027"
```
aaronbuchwald, line 49: nit: make this a constant

aaronbuchwald: Could also use the constant from AvalancheGo genesis directory and convert to the appropriate type here

tests/e2e/precompile-upgrade/suites.go
```diff
@@ -0,0 +1,330 @@
...
+			}
+		})
+
+		evmCli := new(utils.EvmClient)
+		ginkgo.By("makes txs after upgrade", func() {
+			ci := utils.GetClusterInfo()
+			gomega.Expect(len(ci.URIs) > 0).Should(gomega.BeTrue())
+			utils.Outf("{{magenta}}sending txs to subnet-evm endpoints:{{/}} %q\n", ci.SubnetEVMRPCEndpoints)
+
+			ep := ci.SubnetEVMRPCEndpoints[rand.Intn(100)%len(ci.SubnetEVMRPCEndpoints)]
```
aaronbuchwald, line 102: why don't we use `rand.Intn(len(ci.SubnetEVMRPCEndpoints)` ?

tests/e2e/precompile-upgrade/suites.go
```diff
@@ -0,0 +1,330 @@
...
+			gomega.Expect(err).Should(gomega.BeNil())
+
+			ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
+			prevBal, err := evmCli.FetchBalance(ctx, ewoqAddr)
+			cancel()
+			gomega.Expect(err).Should(gomega.BeNil())
+
+			if prevBal.Cmp(new(big.Int).SetInt64(0)) == 0 {
+				ginkgo.Skip("no balance... skipping tests...")
+			}
```
aaronbuchwald, line 115: why do we skip the test if the balance is zero? This seems like it could hide an unexpected case

tests/e2e/precompile-upgrade/suites.go
```diff
@@ -0,0 +1,330 @@
...
+			gomega.Expect(err).Should(gomega.BeNil())
+
+			utils.Outf("{{magenta}}successfully sent:{{/}} after balance %v (old balance %v)\n", newBal, prevBal)
+		})
+
+		// only allow txs from "ewoq"
+		upgradeCfgAllowsEwoq := chainUpgradeConfig{
+			PrecompileUpgrades: []*params.PrecompileUpgrade{
+				{
+					TxAllowListConfig: &precompile.TxAllowListConfig{
```
aaronbuchwald, line 135: nit: use `precompile.NewTxAllowListConfig` to construct the config here

tests/e2e/precompile-upgrade/suites.go
```diff
@@ -0,0 +1,330 @@
...
+				time.Sleep(5 * time.Second)
+
+				ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
+				prevBal, err := evmCli.FetchBalance(ctx, ewoqAddr)
+				cancel()
+				gomega.Expect(err).Should(gomega.BeNil())
+
+				if prevBal.Cmp(new(big.Int).SetInt64(0)) == 0 {
+					ginkgo.Skip("no balance... skipping tests...")
+				}
```
aaronbuchwald, line 189: This could hide an unexpected failure case, what's the rationale for including this?

aaronbuchwald (changes requested):

tests/e2e/precompile-upgrade/suites.go
gyuho: Changing to `evm.Config`

tests/e2e/precompile-upgrade/suites.go
gyuho: Was trying to be generic so it can be run against any cluster. On second thought, will just document instead assuming it will run against local cluster with ewoq wallet having enough balance.

tests/e2e/precompile-upgrade/suites.go
gyuho: Removed

tests/e2e/precompile-upgrade/suites.go
gyuho: Now assume ewoq already has balance and funds addr2

gyuho:
> Run foundry-rs/foundry-toolchain@v1
  with:
    version: 1.0.0
Error: Error: Unexpected HTTP response: [4](https://github.com/ava-labs/subnet-evm/actions/runs/3153016130/jobs/5128997940#step:3:5)04

> Version to install, e.g. nightly or 1.0.0. Note: Foundry only has nightly builds for the time being.


foundry 1.0.0 doesn't seem to work... reverting back to nightly

gyuho:
Fix https://github.com/ava-labs/subnet-evm/issues/251

scripts/run.sh
```diff
@@ -270,21 +281,10 @@ run_simulator() {
...
-    $CHAIN_ID $GENESIS_ADDRESS \
-    $BASEDIR/avalanchego-${VERSION}/avalanchego \
-    ${AVALANCHEGO_PLUGIN_DIR} \
-    "0.0.0.0:12342" \
-    "$BASEDIR/genesis.json"
-fi
+echo "running ginkgo"
+run_ginkgo
+# to fail the script if ginkgo failed
+EXIT_CODE=$?
```
aaronbuchwald, line 287: Is this intended to run the ginkgo tests and then the simulator as well?

gyuho: This exit code will only be overwritten if the simulator tests are enabled. Otherwise, only used for failing the run.sh script if ginkgo fails (in this case by default, ginkgo fails to launch subnet-evm in the local network).

So, only runs simulator when RUN_SIMULATOR=true

Issue #228, opened 2022-08-24
Gas Fee Estimation problem after setFeeConfig is called


ilkerulutas:
**Describe the bug**

After the Fee Config Manager's setFeeConfig function is called and the fee config values are updated the gas fee estimation does not reflect latest changes.

In our case the initial "minBaseFee" was 1,000,000,000, after increasing it to 50,000,000,000 we were not able to send any transactions with the default fee estimations using metamask or core (Transactions were being rejected because of low fee). As a workaround we had to override and set "Max Priority Fee" and "Max Fee" to higher numbers in order to push one transaction. Once a new block was accepted by the network the estimations were fixed, Metamask and Core started suggesting better values.

**To Reproduce**

1. Initial feeConfig:

``` json
"feeConfig": {
        "gasLimit": 20000000,
        "targetBlockRate": 2,
        "minBaseFee": 1000000000,
        "targetGas": 100000000,
        "baseFeeChangeDenominator": 48,
        "minBlockGasCost": 0,
        "maxBlockGasCost": 10000000,
        "blockGasCostStep": 500000
}
```

2. Observe fee estimations work using Metamask/Core

3. Update feeConfig:

``` json
"feeConfig": {
        "gasLimit": 35000000,
	"targetBlockRate": 2,
        "minBaseFee": 50000000000,
        "targetGas": 100000000,
        "baseFeeChangeDenominator": 48,
        "minBlockGasCost": 0,
        "maxBlockGasCost": 10000000,
        "blockGasCostStep": 1000000
}
```

4. Observe fee estimations of Metamask/Core are still the same for same type of operation

**Expected behavior**

Fee estimations should be updated accordingly

PR #241, opened 2022-08-30
reset lastBaseFee if fee config changes

darioush:
The computation in suggestDynamicFees holds on to lastBaseFee even if the fee was modified in the most recent block. This causes an issue for estimating the fee of the block immediately following a block with a change to the min base fee.

closes: https://github.com/ava-labs/subnet-evm/issues/228


eth/gasprice/gasprice_test.go
```diff
@@ -467,3 +472,62 @@ func TestSuggestGasPricePreAP3(t *testing.T) {
...
+	require.NoError(t, err)
+
+	// before issuing the block changing the fee into the chain, the fee estimation should
+	// follow the fee config in genesis.
+	backend := newTestBackend(t, &chainConfig, 0, func(i int, b *core.BlockGen) {})
+	oracle, err := NewOracle(backend, config)
+	require.NoError(t, err)
+	got, err := oracle.SuggestPrice(context.Background())
+	require.NoError(t, err)
+	assert.Equal(t, chainConfig.FeeConfig.MinBaseFee, got)
```
aaronbuchwald, line 502: Can we use require throughout the test and do `require.New(t)` at the top so that we don't include `t` as a parameter?

eth/gasprice/gasprice.go
```diff
@@ -333,7 +336,14 @@ func (oracle *Oracle) suggestDynamicFees(ctx context.Context) (*big.Int, *big.In
 		tipResults     []*big.Int
 		baseFeeResults []*big.Int
 	)
-
+	// if the fee config has changed in this block, use the new base fee
+	// if it is higher than this block's base fee.
+	if feeLastChangedAt != nil && feeLastChangedAt.Uint64() == number {
+		if lastBaseFee.Cmp(feeConfig.MinBaseFee) < 0 {
+			lastBaseFee = feeConfig.MinBaseFee
+			// TODO: should we also reset lastPrice?
```
aaronbuchwald, line 344: Since `lastPrice` is derived from the tip or amount above base fee people are willing to pay, I think we could go either way on this depending on how we expect people's willingness to pay above the base fee to be impacted by the fee config change.

So I don't think we necessarily need to reset it here.

aaronbuchwald: I believe we need to handle excluding base fees that are from a different "fee config era" completely rather than only setting the `lastBaseFee` if there's an exact match.

darioush: that is already handled in line #348

aaronbuchwald (approved):
LGTM

eth/gasprice/gasprice.go
aaronbuchwald: Ah I see

anusha-ctrl (approved):

PR #243, opened 2022-08-30
geth import allowlist

darioush:
Prevents accidental imports from geth or coreth when we are porting upstream changes.


scripts/geth-allowed-packages.txt
```diff
@@ -0,0 +1,17 @@
+"github.com/ethereum/go-ethereum/accounts/abi"
```
aaronbuchwald, line 1: can we remove `abi` and `params` ?

aaronbuchwald (approved):
LGTM

PR #247, opened 2022-09-02
tests: implement tx allow list e2e tests

gyuho:


gyuho:
Moving tests to https://github.com/ava-labs/subnet-evm/pull/238.

PR #252, opened 2022-09-05
Revert PR 249

holisticode:


felipemadero (approved):
LGTM

ramilexe:
Can we get it merged asap?

ceyonur (approved):

darioush (approved):

aaronbuchwald (approved):

hexfusion:
Just FYI we could consider nixing the shared-timer[1] logic which is now back as part of avalanchego. We should come up  with a general plan for replacement as the it is defined as deprecated but just a thought before this merges.

[1] https://github.com/ava-labs/subnet-evm/pull/249/files#diff-08c0341b9a66a2439b8a1ae1403c49af7621c34d89d3e0b0023c11077d0463ed

aaronbuchwald:
> Just FYI we could consider nixing the shared-timer[1] logic which is now back as part of avalanchego. We should come up with a general plan for replacement as the it is defined as deprecated but just a thought before this merges.
> 
> [1] https://github.com/ava-labs/subnet-evm/pull/249/files#diff-08c0341b9a66a2439b8a1ae1403c49af7621c34d89d3e0b0023c11077d0463ed

Good callout, adding an issue to track this here. We should be able to remove all timeouts from the block building logic. We have already gotten rid of the two staged timer in coreth, which will be migrated to subnet-evm soon.

PR #253, opened 2022-09-05
GetBlockIDAtHeight: return database.ErrNotFound

darioush:
This PR updates GetBlockIDAtHeight to return the correct error expected by the engine's interface.
This was causing an issue where nodes that have state synced would not be able to parse some ancestors during bootstrapping.

### DIFF
--- a/plugin/evm/vm.go
+++ b/plugin/evm/vm.go
@@ -674,10 +674,12 @@ func (vm *VM) VerifyHeightIndex() error {
 // GetBlockIDAtHeight retrieves the blkID of the canonical block at [blkHeight]
 // if [blkHeight] is less than the height of the last accepted block, this will return
 // a canonical block. Otherwise, it may return a blkID that has not yet been accepted.
+// Note: Engine's interface requires this function returns database.ErrNotFound
+// if a block is not found.
 func (vm *VM) GetBlockIDAtHeight(blkHeight uint64) (ids.ID, error) {
 	ethBlock := vm.blockChain.GetBlockByNumber(blkHeight)
 	if ethBlock == nil {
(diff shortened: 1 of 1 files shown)


ceyonur (approved):

abi87 (changes requested):
@darioush let me see if I got the bug right:
1. `GetAncestors` call may be served by a `GetBlock` call which uses `database.ErrNotFound` as a special flag to signal ancestors are missing, a typical situation with state sync. Any other error would cause the `GetAncestor` message to be dropped.
2. When it comes to C-chain, calls are intercepted by `chain.State` object, specifically its `getBlock func(ids.ID) (snowman.Block, error)` attribute. `State.getBlock` retrieves its block via `evm.GetBlock` method then caches it via `State.addBlockOutsideConsensus` method **and returns the error from the latter**.
3. `State.addBlockOutsideConsensus` queries  evm again to retrieve block status and place it in the right cache. This is done via the attribute `State.getStatus` which is initialized via `produceGetStatus` call and uses `GetBlockIDAtHeight` when available. This is available in evm.

So really we need `return ids.ID{}, database.ErrNotFound` because of the way chain.State does caching.
I wonder if a better solution would be to expand `State.getBlock` to return block state as well. We'd get rid of the need to be specific with the error and avoid polling twice for the same block (by ID the first time, by height the second).
Longer term, I also wonder if we should `errors.Is()` instead of exact error comparison to handle flags. Iirc @aaronbuchwald had a PR made with this

abi87 (approved):
Moved `chain.State` concerns to https://github.com/ava-labs/avalanchego-internal/pull/1949.

PR #258, opened 2022-09-12
use last accepted block for check compatibility in genesis

ceyonur:
changes header block with last accepted block in Genesis `CheckCompatible`. Header block prevents one to enable a precompile retroactively with the header timestamp. However the node can still activate it gracefully since lastBlock.timestamp < headerBlock.timestmap. This enables one to activate a precompile with the context of last block timestamp.


core/genesis.go
```diff
@@ -240,12 +240,20 @@ func SetupGenesisBlock(db ethdb.Database, genesis *Genesis) (*params.ChainConfig
...
 	// Check config compatibility and write the config. Compatibility errors
 	// are returned to the caller unless we're already at block zero.
-	headBlock := rawdb.ReadHeadBlock(db)
-	if headBlock == nil {
-		return newcfg, fmt.Errorf("missing head block")
+	// headBlock := rawdb.ReadHeadBlock(db)
+	// if headBlock == nil {
+	// 	return newcfg, fmt.Errorf("missing head block")
+	// }
+
```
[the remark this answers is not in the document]
ceyonur: left this one intentionally, so we can see the difference between geth and this one.

core/genesis.go
```diff
@@ -240,12 +240,20 @@ func SetupGenesisBlock(db ethdb.Database, genesis *Genesis) (*params.ChainConfig
...
+	// we use last accepted block for cfg compatibility check to handle
+	// retroactive upgrades gracefully.
```
darioush, line 249:
```suggestion
	// we use last accepted block for cfg compatibility check. Note this allows
	// the node to continue making progress if it had halted previously due to
	// a missing upgrade configuration.
```
or something like that?

core/genesis.go
```diff
@@ -240,12 +240,21 @@ func SetupGenesisBlock(db ethdb.Database, genesis *Genesis) (*params.ChainConfig
...
 	// Check config compatibility and write the config. Compatibility errors
 	// are returned to the caller unless we're already at block zero.
-	headBlock := rawdb.ReadHeadBlock(db)
-	if headBlock == nil {
-		return newcfg, fmt.Errorf("missing head block")
+	// headBlock := rawdb.ReadHeadBlock(db)
+	// if headBlock == nil {
+	// 	return newcfg, fmt.Errorf("missing head block")
+	// }
+
```
aaronbuchwald, line 247: Can we remove this commented out code?

core/genesis.go
```diff
@@ -240,12 +240,21 @@ func SetupGenesisBlock(db ethdb.Database, genesis *Genesis) (*params.ChainConfig
...
+	// we use last accepted block for cfg compatibility check. Note this allows
+	// the node to continue making progress if it had halted previously due to
+	// a missing upgrade configuration.
+	lastBlock := ReadBlockByHash(db, lastAcceptedHash)
+	if lastBlock == nil {
+		return newcfg, fmt.Errorf("missing last accepted block")
 	}
```
aaronbuchwald, line 254: What's returned here if we are starting up for the first time?

core/genesis.go
ceyonur: again I intentionally left it as comment out to show difference between actual behaviour vs modified one. but since this was asked twice, I guess this looks ugly :D deleting it.

core/genesis.go
ceyonur: genesis hash https://github.com/ava-labs/subnet-evm/blob/master/plugin/evm/vm.go#L843-L845

params/precompile_config.go
```diff
@@ -235,9 +235,12 @@ func (c *ChainConfig) GetActivePrecompiles(blockTimestamp *big.Int) PrecompileUp
 // Returns a ConfigCompatError if upgrades already forked at [headTimestamp] are missing from
 // [precompileUpgrades]. Upgrades not already forked may be modified or absent from [precompileUpgrades].
 // Returns nil if [precompileUpgrades] is compatible with [c].
-func (c *ChainConfig) CheckPrecompilesCompatible(precompileUpgrades []PrecompileUpgrade, headTimestamp *big.Int) *ConfigCompatError {
+// Assumes given timestamp is the last accepted block timestamp.
+// Otherwise it might prevent an retroactive upgrade to be activated,
+// even though it can be actually activated at that time.
```
aaronbuchwald, line 240:
```suggestion
// This ensures that as long as the node has not accepted a block with a different rule set it will allow a new upgrade to be applied as long as it activates after the last accepted block.
```

core/genesis.go
```diff
@@ -381,3 +386,11 @@ func GenesisBlockForTesting(db ethdb.Database, addr common.Address, balance *big
...
 	return g.MustCommit(db)
 }
+
+func ReadBlockByHash(db ethdb.Reader, hash common.Hash) *types.Block {
```
aaronbuchwald, line 390: Can we add a comment here?

core/genesis.go
aaronbuchwald: Thanks for removing

core/genesis_test.go
```diff
@@ -204,7 +209,7 @@ func TestStatefulPrecompilesConfigure(t *testing.T) {
 			}
 
 			db := rawdb.NewMemoryDatabase()
-			_, err := SetupGenesisBlock(db, genesis)
+			_, err := SetupGenesisBlock(db, genesis, common.Hash{})
```
aaronbuchwald, line 212: Should we use the genesis hash here? It looks like this still passes because it's setting up the database for the first time, so it never hits the point where it checks the config against the last accepted hash.

core/genesis.go
aaronbuchwald: I see, forgot about this. Can we add a comment that mentions this requirement here?

ceyonur: done

core/genesis.go
ceyonur: done

core/genesis_test.go
ceyonur: done

aaronbuchwald (approved):
LGTM - could we add a regression test as well to ensure that when there is a head block past the last accepted block, we are still able to correctly apply an upgrade?

core/genesis_test.go
```diff
@@ -223,3 +230,61 @@ func TestStatefulPrecompilesConfigure(t *testing.T) {
...
+	// Advance header to block #4, past the ContractDeployerAllowListConfig.
+	blocks, _, _ := GenerateChain(customg.Config, genesis, dummy.NewFullFaker(), db, 4, 25, nil)
+
+	require := require.New(t)
+	_, err := bc.InsertChain(blocks)
+	require.NoError(err)
+
+	// accept up to block #2
+	for _, block := range blocks[:2] {
+		if err := bc.Accept(block); err != nil {
```
darioush, line 258: nit: can we switch to using require?

core/genesis_test.go
```diff
@@ -223,3 +230,61 @@ func TestStatefulPrecompilesConfigure(t *testing.T) {
...
+
+	// assert block is after the activation block
+	require.Greater(block.Time(), contractDeployerConfig.Timestamp().Uint64())
+	// assert last accepted block is before the activation block
+	require.Less(bc.lastAccepted.Time(), contractDeployerConfig.Timestamp().Uint64())
+
+	// This should not return any error since the last accepted block is before the activation block.
+	config, _, err := setupGenesisBlock(db, &activatedGenesis, bc.lastAccepted.Hash())
+	if err != nil {
+		t.Fatal(err)
```
darioush, line 285: same comment about require

darioush (approved):

params/config.go
```diff
@@ -264,18 +264,18 @@ func (c *ChainConfig) Is{YourPrecompile}(blockTimestamp *big.Int) bool {
 // CheckCompatible checks whether scheduled fork transitions have been imported
 // with a mismatching chain configuration.
 func (c *ChainConfig) CheckCompatible(newcfg *ChainConfig, height uint64, timestamp uint64) *ConfigCompatError {
-	bhead := new(big.Int).SetUint64(height)
-	bheadTimestamp := new(big.Int).SetUint64(timestamp)
+	bNumber := new(big.Int).SetUint64(height)
+	bTimestamp := new(big.Int).SetUint64(timestamp)
```
aaronbuchwald, line 268: Looks like this rename was unnecessary

PR #260, opened 2022-09-16
Reward Manager precompile

ceyonur:
Adds fee recipients precompile to determine how rewards should be handled

1- Directly to a single address 
2- Allowed Fee Recipients so minters can claim rewards for themselves
3- Disabled and fees directed to blackhole

- [x] Added precompile template and contract
- [x] Verify blocks according to reward handler in state 
- [x] Build blocks according to reward handler in state (set coinbase correctly)
- [ ] ~~Add contract call mechanism~~
- [x] Unit Tests 
- [x] Example contract with solidity tests


precompile/reward_manager.go
```diff
@@ -0,0 +1,473 @@
...
+	_ StatefulPrecompileConfig = &RewardManagerConfig{}
+
+	ErrCannotAllowFeeRecipients      = errors.New("non-enabled cannot allowFeeRecipients")
+	ErrCannotAreFeeRecipientsAllowed = errors.New("non-enabled cannot areFeeRecipientsAllowed")
+	ErrCannotCurrentRewardAddress    = errors.New("non-enabled cannot currentRewardAddress")
+	ErrCannotDisableRewards          = errors.New("non-enabled cannot disableRewards")
+	ErrCannotSetRewardAddress        = errors.New("non-enabled cannot setRewardAddress")
+
+	ErrCannotEnableBothRewards = errors.New("cannot enable both fee recipients and reward address at the same time")
+	ErrCannotEmptyRewardAddr   = errors.New("cannot set empty reward address")
```
darioush, line 51: seems this is not used

precompile/reward_manager.go
```diff
@@ -0,0 +1,473 @@
...
+func (c *InitialRewardConfig) Equal(other *InitialRewardConfig) bool {
+	if other == nil {
+		return false
+	}
+
+	return c.AllowFeeRecipients == other.AllowFeeRecipients && c.RewardAddress == other.RewardAddress
+}
+
+// RewardManagerConfig implements the StatefulPrecompileConfig
+// interface while adding in the RewardManager specific precompile address.
```
darioush, line 85:
```suggestion
// interface while adding in the RewardManager specific precompile config.
```

precompile/reward_manager.go
```diff
@@ -0,0 +1,473 @@
...
+// Equal returns true if [s] is a [*RewardManagerConfig] and it has been configured identical to [c].
+func (c *RewardManagerConfig) Equal(s StatefulPrecompileConfig) bool {
+	// typecast before comparison
+	other, ok := (s).(*RewardManagerConfig)
+	if !ok {
+		return false
+	}
+	// CUSTOM CODE STARTS HERE
+	// modify this boolean accordingly with your custom RewardManagerConfig, to check if [other] and the current [c] are equal
+	// if RewardManagerConfig contains only UpgradeableConfig  and AllowListConfig  you can skip modifying it.
```
darioush, line 134:
```suggestion
	// if RewardManagerConfig contains only UpgradeableConfig and AllowListConfig you can skip modifying it.
```

precompile/reward_manager.go
```diff
@@ -0,0 +1,473 @@
...
+	return getAllowListStatus(stateDB, RewardManagerAddress, address)
+}
+
+// SetRewardManagerAllowListStatus sets the permissions of [address] to [role] for the
+// RewardManager list. Assumes [role] has already been verified as valid.
+func SetRewardManagerAllowListStatus(stateDB StateDB, address common.Address, role AllowListRole) {
+	setAllowListRole(stateDB, RewardManagerAddress, address, role)
+}
+
+// PackAllowFeeRecipients packs the include selector (first 4 func signature bytes).
```
darioush, line 211:
```suggestion
// PackAllowFeeRecipients packs the function selector (first 4 func signature bytes).
```

core/stateful_precompile_test.go
```diff
@@ -254,16 +241,16 @@ func TestContractDeployerAllowListRun(t *testing.T) {
...
 			blockContext := &mockBlockContext{blockNumber: common.Big0}
-			ret, remainingGas, err := precompile.ContractDeployerAllowListPrecompile.Run(&mockAccessibleState{state: state, blockContext: blockContext}, test.caller, test.precompileAddr, test.input(), test.suppliedGas, test.readOnly)
+			ret, remainingGas, err := precompile.ContractDeployerAllowListPrecompile.Run(&mockAccessibleState{state: state, blockContext: blockContext}, test.caller, precompile.ContractDeployerAllowListAddress, test.input(), test.suppliedGas, test.readOnly)
 			if len(test.expectedErr) != 0 {
 				if err == nil {
-					assert.Failf(t, "run expectedly passed without error", "expected error %q", test.expectedErr)
+					require.Failf(t, "run expectedly passed without error", "expected error %q", test.expectedErr)
 				} else {
-					assert.True(t, strings.Contains(err.Error(), test.expectedErr), "expected error (%s) to contain substring (%s)", err, test.expectedErr)
+					require.True(t, strings.Contains(err.Error(), test.expectedErr), "expected error (%s) to contain substring (%s)", err, test.expectedErr)
```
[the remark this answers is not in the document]
ceyonur: done

core/stateful_precompile_test.go
```diff
@@ -755,20 +715,20 @@ func TestContractNativeMinterRun(t *testing.T) {
...
 			}
-			ret, remainingGas, err := precompile.ContractNativeMinterPrecompile.Run(&mockAccessibleState{state: state, blockContext: blockContext}, test.caller, test.precompileAddr, test.input(), test.suppliedGas, test.readOnly)
+			ret, remainingGas, err := precompile.ContractNativeMinterPrecompile.Run(&mockAccessibleState{state: state, blockContext: blockContext}, test.caller, precompile.ContractNativeMinterAddress, test.input(), test.suppliedGas, test.readOnly)
 			if len(test.expectedErr) != 0 {
 				if err == nil {
-					assert.Failf(t, "run expectedly passed without error", "expected error %q", test.expectedErr)
+					require.Failf(t, "run expectedly passed without error", "expected error %q", test.expectedErr)
 				} else {
-					assert.True(t, strings.Contains(err.Error(), test.expectedErr), "expected error (%s) to contain substring (%s)", err, test.expectedErr)
+					require.True(t, strings.Contains(err.Error(), test.expectedErr), "expected error (%s) to contain substring (%s)", err, test.expectedErr)
```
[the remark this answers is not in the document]
ceyonur: done

core/stateful_precompile_test.go
```diff
@@ -477,15 +453,15 @@ func TestTxAllowListRun(t *testing.T) {
...
 			blockContext := &mockBlockContext{blockNumber: common.Big0}
-			ret, remainingGas, err := precompile.TxAllowListPrecompile.Run(&mockAccessibleState{state: state, blockContext: blockContext}, test.caller, test.precompileAddr, test.input(), test.suppliedGas, test.readOnly)
+			ret, remainingGas, err := precompile.TxAllowListPrecompile.Run(&mockAccessibleState{state: state, blockContext: blockContext}, test.caller, precompile.TxAllowListAddress, test.input(), test.suppliedGas, test.readOnly)
 			if len(test.expectedErr) != 0 {
 				if err == nil {
-					assert.Failf(t, "run expectedly passed without error", "expected error %q", test.expectedErr)
+					require.Failf(t, "run expectedly passed without error", "expected error %q", test.expectedErr)
 				} else {
-					assert.True(t, strings.Contains(err.Error(), test.expectedErr), "expected error (%s) to contain substring (%s)", err, test.expectedErr)
+					require.True(t, strings.Contains(err.Error(), test.expectedErr), "expected error (%s) to contain substring (%s)", err, test.expectedErr)
```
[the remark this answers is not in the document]
ceyonur: done

core/stateful_precompile_test.go
```diff
@@ -1060,12 +1006,341 @@ func TestFeeConfigManagerRun(t *testing.T) {
 			if test.config != nil {
 				test.config.Configure(params.TestChainConfig, state, blockContext)
 			}
-			ret, remainingGas, err := precompile.FeeConfigManagerPrecompile.Run(&mockAccessibleState{state: state, blockContext: blockContext}, test.caller, test.precompileAddr, test.input(), test.suppliedGas, test.readOnly)
+			ret, remainingGas, err := precompile.FeeConfigManagerPrecompile.Run(&mockAccessibleState{state: state, blockContext: blockContext}, test.caller, precompile.FeeConfigManagerAddress, test.input(), test.suppliedGas, test.readOnly)
+			if len(test.expectedErr) != 0 {
+				if err == nil {
+					require.Failf(t, "run expectedly passed without error", "expected error %q", test.expectedErr)
+				} else {
+					require.True(t, strings.Contains(err.Error(), test.expectedErr), "expected error (%s) to contain substring (%s)", err, test.expectedErr)
```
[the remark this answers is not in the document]
ceyonur: done

core/stateful_precompile_test.go
```diff
@@ -1060,12 +1006,341 @@ func TestFeeConfigManagerRun(t *testing.T) {
...
+		},
+		"get current reward address from no role succeeds": {
+			caller: noRoleAddr,
+			preCondition: func(t *testing.T, state *state.StateDB) {
+				precompile.StoreRewardAddress(state, testAddr)
+			},
+			input: func() []byte {
+				input, err := precompile.PackCurrentRewardAddress()
+				if err != nil {
+					panic(err)
```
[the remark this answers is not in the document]
ceyonur: yep, good catch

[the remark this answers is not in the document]
ceyonur: done

core/stateful_precompile_test.go
```diff
@@ -1060,12 +1006,341 @@ func TestFeeConfigManagerRun(t *testing.T) {
...
+			suppliedGas: precompile.ModifyAllowListGasCost,
+			readOnly:    false,
+			expectedErr: precompile.ErrCannotModifyAllowList.Error(),
+		},
+	} {
+		t.Run(name, func(t *testing.T) {
+			db := rawdb.NewMemoryDatabase()
+			state, err := state.New(common.Hash{}, state.NewDatabase(db), nil)
+			if err != nil {
+				t.Fatal(err)
```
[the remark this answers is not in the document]
ceyonur: done

core/stateful_precompile_test.go
```diff
@@ -1060,12 +1006,341 @@ func TestFeeConfigManagerRun(t *testing.T) {
...
+				test.config.Configure(params.TestChainConfig, state, blockContext)
+			}
+			ret, remainingGas, err := precompile.RewardManagerPrecompile.Run(&mockAccessibleState{state: state, blockContext: blockContext}, test.caller, precompile.RewardManagerAddress, test.input(), test.suppliedGas, test.readOnly)
 			if len(test.expectedErr) != 0 {
 				if err == nil {
-					assert.Failf(t, "run expectedly passed without error", "expected error %q", test.expectedErr)
+					require.Failf(t, "run expectedly passed without error", "expected error %q", test.expectedErr)
 				} else {
-					assert.True(t, strings.Contains(err.Error(), test.expectedErr), "expected error (%s) to contain substring (%s)", err, test.expectedErr)
+					require.True(t, strings.Contains(err.Error(), test.expectedErr), "expected error (%s) to contain substring (%s)", err, test.expectedErr)
```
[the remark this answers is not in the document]
ceyonur: done

plugin/evm/vm.go
```diff
@@ -325,17 +325,13 @@ func (vm *VM) Initialize(
...
-		} else {
-			return errors.New("cannot specify a custom fee recipient on this blockchain")
-		}
-	} else if g.Config.AllowFeeRecipients {
-		log.Warn("Chain enabled `AllowFeeRecipients`, but chain config has not specified any coinbase address. Defaulting to the blackhole address.")
+		address := common.HexToAddress(vm.config.FeeRecipient)
+		log.Info("Setting fee recipient", "address", address)
+		vm.ethConfig.Miner.Etherbase = address
+	} else {
+		log.Warn("Config has not specified any coinbase address. Defaulting to the blackhole address.")
```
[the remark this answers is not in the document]
ceyonur: It will still be handled in miner/worker while setting header coinbase. Before, it's guaranteed to know if this was activated in genesis, so we were able to show a warning message or set etherbase to blackhole address accordingly. i.e if allow fee, show warning if no address is specified; otherwise set etherbase it to blackhole address. However with that precompile a network can always switch to allow fee recipients after the genesis. In that case we cannot forcefully set the etherbase address to blackhole address otherwise they would always send fees to blackhole address if allow fee is enabled in the future. so we have to keep etherbase as it is. instead of setting the etherbase here we set the coinbase in the miner/worker just before committing the header.

precompile/reward_manager.go
ceyonur: yup correct

plugin/evm/vm_test.go
```diff
@@ -2576,3 +2577,142 @@ func TestAllowFeeRecipientEnabled(t *testing.T) {
...
+
+func TestRewardManagerPrecompileSetRewardAddress(t *testing.T) {
+	genesis := &core.Genesis{}
+	if err := genesis.UnmarshalJSON([]byte(genesisJSONSubnetEVM)); err != nil {
+		t.Fatal(err)
+	}
+	genesis.Config.RewardManagerConfig = precompile.NewRewardManagerConfig(common.Big0, testEthAddrs[0:1], nil, nil)
+	genesis.Config.AllowFeeRecipients = true // enable this in genesis to test if this is recognized by the reward manager
+	genesisJSON, err := genesis.MarshalJSON()
+	if err != nil {
```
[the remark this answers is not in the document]
ceyonur: done

plugin/evm/vm_test.go
```diff
@@ -2576,3 +2577,142 @@ func TestAllowFeeRecipientEnabled(t *testing.T) {
...
+func TestRewardManagerPrecompileAllowFeeRecipieints(t *testing.T) {
+	genesis := &core.Genesis{}
+	if err := genesis.UnmarshalJSON([]byte(genesisJSONSubnetEVM)); err != nil {
+		t.Fatal(err)
+	}
+	genesis.Config.RewardManagerConfig = precompile.NewRewardManagerConfig(common.Big0, testEthAddrs[0:1], nil, nil)
+	genesis.Config.AllowFeeRecipients = false // disable this in genesis
+	genesisJSON, err := genesis.MarshalJSON()
+	if err != nil {
+		t.Fatal(err)
```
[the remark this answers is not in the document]
ceyonur: done

plugin/evm/vm_test.go
```diff
@@ -2576,3 +2577,142 @@ func TestAllowFeeRecipientEnabled(t *testing.T) {
...
+	require.NoError(t, err)
+
+	balance := blkState.GetBalance(testAddr)
+	assert.Equal(t, 1, balance.Cmp(common.Big0))
+}
+
+func TestRewardManagerPrecompileAllowFeeRecipieints(t *testing.T) {
+	genesis := &core.Genesis{}
+	if err := genesis.UnmarshalJSON([]byte(genesisJSONSubnetEVM)); err != nil {
+		t.Fatal(err)
```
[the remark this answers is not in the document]
ceyonur: done

miner/worker.go
```diff
@@ -188,11 +203,11 @@ func (w *worker) commitNewWork() (*types.Block, error) {
 	}
 	if len(localTxs) > 0 {
 		txs := types.NewTransactionsByPriceAndNonce(env.signer, localTxs, header.BaseFee)
-		w.commitTransactions(env, txs, w.coinbase)
+		w.commitTransactions(env, txs, header.Coinbase)
 	}
 	if len(remoteTxs) > 0 {
 		txs := types.NewTransactionsByPriceAndNonce(env.signer, remoteTxs, header.BaseFee)
-		w.commitTransactions(env, txs, w.coinbase)
+		w.commitTransactions(env, txs, header.Coinbase)
```
aaronbuchwald, line 210: Shouldn't this be the same as `w.coinbase` ? Can we avoid changing this line?

ceyonur: we don't actuall change `w.coinbase` according to `GetCoinbaseAt`. That's because if node operator defines a fee recipient address we should keep it without modifying. `w.coinbase` is set once in `Initialize`. So if we modify it in this function and then fee recipients are enabled then the node operator's initial custom fee recipients adress won't be available. So coinbase in headers can be changed according to the rules in the conract but the coinbase in the miner should not change so we can re-use a custom fee address if possible.

consensus/dummy/consensus.go
```diff
@@ -69,6 +70,26 @@ func NewFullFaker() *DummyEngine {
 	}
 }
 
+// verifyCoinbase checks that the coinbase is valid for the given [header] and [parent].
+func (self *DummyEngine) verifyCoinbase(config *params.ChainConfig, header *types.Header, parent *types.Header, chain consensus.ChainHeaderReader) error {
+	// get the coinbase configured at parent
+	configuredAddressAtParent, isAllowFeeRecipients, err := chain.GetCoinbaseAt(parent)
```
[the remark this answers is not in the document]
ceyonur: yes

contract-examples/test/ExampleRewardDistributor.ts
```diff
@@ -0,0 +1,203 @@
...
+
+    // Send a transaction to mine a new block
+    const tx = await owner.sendTransaction({
+      to: signer1.address,
+      value: ethers.utils.parseEther("0.0001")
+    })
+    await tx.wait()
+
+    balance = await ethers.provider.getBalance(contract.address)
+    expect(balance).to.be.equal(0)
```
aaronbuchwald, line 72: Can we verify the updated balance of the blackhole address here as well?

plugin/evm/vm_test.go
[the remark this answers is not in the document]
ceyonur: pushed

contract-examples/tasks.ts
```diff
@@ -215,11 +216,55 @@ task("feeManager:get", "gets fee config")
...
+// npx hardhat feeManager:readRole --network local --address [address]
 task("feeManager:readRole", "a task to get the network deployer minter list")
   .addParam("address", "the address you want to know the minter role for")
   .setAction(async (args, hre) => {
     const allowList = await hre.ethers.getContractAt("IFeeManager", FEE_MANAGER_ADDRESS)
     await getRole(allowList, args.address)
   })
 
+
+// npx hardhat rewardManager:readRole --network local --address [address]
```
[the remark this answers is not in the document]
ceyonur: yep, fixed it thanks.

contract-examples/tasks.ts
```diff
@@ -215,11 +216,55 @@ task("feeManager:get", "gets fee config")
...
+    const rewardManager = await hre.ethers.getContractAt("IRewardManager", REWARD_MANAGER_ADDDRESS)
+    const result = await rewardManager.currentRewardAddress()
+    if (result == "0x0000000000000000000000000000000000000000") {
+      console.log("Custom Fee Recipients are allowed.")
+    } else {
+      console.log(`Current reward address is ${result}`)
+    }
+  })
+
+// npx hardhat rewardManager:areFeeRecipientsAllowed --network local --address [address]
```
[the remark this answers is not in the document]
ceyonur: done

contract-examples/tasks.ts
```diff
@@ -215,11 +216,55 @@ task("feeManager:get", "gets fee config")
...
+
+// npx hardhat rewardManager:allowFeeRecipients --network local
+task("rewardManager:allowFeeRecipients", "allows custom fee recipients to receive rewards")
+  .setAction(async (_, hre) => {
+    const rewardManager = await hre.ethers.getContractAt("IRewardManager", REWARD_MANAGER_ADDDRESS)
+    const result = await rewardManager.allowFeeRecipients()
+    console.log(result)
+  })
+
+task("rewardManager:disableRewards", "disables all rewards, and starts burning fees.")
```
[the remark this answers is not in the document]
ceyonur: done

contract-examples/test/ExampleRewardDistributor.ts
```diff
@@ -0,0 +1,203 @@
...
+
+  it("should be able to revoke a reward address", async function () {
+    const rewardAddress = signer1
+    let tx = await contract.revoke(rewardAddress.address);
+    await tx.wait()
+    const isRewardAddress = await contract.isRewardAddress(rewardAddress.address)
+    expect(isRewardAddress).to.be.equal(false)
+  })
+
+  it("should be able to receive rewards after revoke", async function () {
```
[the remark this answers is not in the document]
ceyonur: actually the test itself should be different. completed it.

contract-examples/contracts/ExampleRewardDistributor.sol
```diff
@@ -0,0 +1,72 @@
...
+    require(rewardAddresses[addr] == 0, "Already a reward address");
+
+    rewardAddresses[addr] = block.number;
+  }
+
+  function revoke(address addr) public onlyOwner _isRewardAddress(addr) {
+    uint256 reward = _estimateReward(addr);
+    delete rewardAddresses[addr];
+
+    payable(addr).transfer(reward);
```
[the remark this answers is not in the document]
ceyonur: yep, nice catch!

precompile/config_test.go
```diff
@@ -375,3 +388,75 @@ func TestEqualFeeConfigManagerConfig(t *testing.T) {
...
+			expected: false,
+		},
+		{
+			name:     "different type",
+			config:   NewRewardManagerConfig(big.NewInt(3), admins, enableds, nil),
+			other:    NewTxAllowListConfig(big.NewInt(3), []common.Address{{1}}, []common.Address{{2}}),
+			expected: false,
+		},
+		{
+			name:     "different version",
```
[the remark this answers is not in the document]
ceyonur: fixed them

precompile/reward_manager.go
```diff
@@ -0,0 +1,472 @@
...
+)
+
+// Reference imports to suppress errors if they are not otherwise used.
+var (
+	_ = errors.New
+	_ = big.NewInt
+	_ = strings.NewReader
+)
```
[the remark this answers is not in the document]
ceyonur: removed

consensus/consensus.go
```diff
@@ -57,6 +57,9 @@ type ChainHeaderReader interface {
 
 	// GetFeeConfigAt retrieves the fee config and last changed block number at block header.
 	GetFeeConfigAt(parent *types.Header) (commontype.FeeConfig, *big.Int, error)
+
+	// GetCoinbaseAt retrieves the configured coinbase address at [header]. If fee recipients are allowed, returns true in the second return value.
+	GetCoinbaseAt(header *types.Header) (common.Address, bool, error)
```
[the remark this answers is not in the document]
ceyonur: makes sense, changed it

contract-examples/test/ExampleRewardDistributor.ts
```diff
@@ -0,0 +1,203 @@
...
+    const Contract: ContractFactory = await ethers.getContractFactory("ExampleRewardDistributor", { signer: owner })
+    contract = await Contract.deploy(rewardRate)
+    await contract.deployed()
+    const contractAddress: string = contract.address
+    console.log(`Contract deployed to: ${contractAddress}`)
+
+    // Send a transaction to mine a new block
+    const tx = await owner.sendTransaction({
+      to: signer1.address,
+      value: ethers.utils.parseEther("10")
```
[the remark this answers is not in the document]
ceyonur: it was just to make sure that `signer1` won't out-of funds

contract-examples/contracts/ExampleRewardDistributor.sol
```diff
@@ -0,0 +1,72 @@
+//SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "./IAllowList.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
+
+contract ExampleRewardDistributor is Ownable {
```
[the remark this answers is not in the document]
ceyonur: yup added

consensus/dummy/consensus.go
```diff
@@ -175,6 +196,12 @@ func (self *DummyEngine) verifyHeader(chain consensus.ChainHeaderReader, header
 	if err := self.verifyHeaderGasFields(config, header, parent, chain); err != nil {
 		return err
 	}
+	// Ensure that coinbase is valid if reward manager is enabled
+	if config.IsRewardManager(timestamp) {
+		if err := self.verifyCoinbase(config, header, parent, chain); err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
ceyonur: just wanted to short-circuit it since we already check that in syntactic verify. but sure we can remove this condition.

contract-examples/contracts/IRewardManager.sol
```diff
@@ -0,0 +1,15 @@
...
+interface IRewardManager is IAllowList {
+  function setRewardAddress(address addr) external;
+
+  function allowFeeRecipients() external;
+
+  function disableRewards() external;
+
+  function currentRewardAddress() external view returns (address rewardAddress);
+
+  function areFeeRecipientsAllowed() external view returns (bool isAllowed);
```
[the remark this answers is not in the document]
ceyonur: I was thinking it would be a little bit complicated with that case:

```
setRewards(address addr, bool allowFeeRecipieints)
```

```
getRewards() (address rewardAddress, bool isAllowed)
```

also disabling it would be harder since it should be called specifically like that:
```
 setRewards(<blackhole addr>, false)
```

Could you provide what's in your mind?

plugin/evm/block_verification.go
```diff
@@ -95,9 +96,12 @@ func (v blockValidator) SyntacticVerify(b *Block, rules params.Rules) error {
...
 	}
-	// Coinbase must be zero, if AllowFeeRecipients is not enabled and not SubnetEVM
-	if !(rules.IsSubnetEVM && b.vm.chainConfig.AllowFeeRecipients) && b.ethBlock.Coinbase() != constants.BlackholeAddr {
-		return fmt.Errorf("invalid coinbase %v does not match required blackhole address %v", ethHeader.Coinbase, constants.BlackholeAddr)
+	// if reward manager is enabled, Coinbase depends on state. State is not available here, so skip checking the coinbase here.
+	if !rules.IsRewardManagerEnabled {
+		// If Subnet EVM or AllowFeeRecipients is not enabled, Coinbase must be BlackholeAddr.
+		if (!rules.IsSubnetEVM || !b.vm.chainConfig.AllowFeeRecipients) && b.ethBlock.Coinbase() != constants.BlackholeAddr {
+			return fmt.Errorf("%w: %v does not match required blackhole address %v", vmerrs.ErrInvalidCoinbase, ethHeader.Coinbase, constants.BlackholeAddr)
+		}
```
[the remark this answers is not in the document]
ceyonur: makes sense. removing it.

plugin/evm/vm_test.go
```diff
@@ -2506,15 +2507,11 @@ func TestAllowFeeRecipientDisabled(t *testing.T) {
 
 	<-issuer
 
-	_, err = vm.BuildBlock()
-	assert.ErrorContains(t, err, "block failed verification")
-
-	vm.miner.SetEtherbase(constants.BlackholeAddr) // set blackhole address
+	blk, err := vm.BuildBlock()
+	require.NoError(t, err) // this won't return an error since miner will set the etherbase to blackhole address
```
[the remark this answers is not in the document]
ceyonur: good idea! added.

plugin/evm/vm_test.go
```diff
@@ -2576,3 +2573,137 @@ func TestAllowFeeRecipientEnabled(t *testing.T) {
...
+	genesis.Config.AllowFeeRecipients = false // disable this in genesis
+	genesisJSON, err := genesis.MarshalJSON()
+	require.NoError(t, err)
+	etherBase := common.HexToAddress("0x0123456789") // give custom ether base
+	c := Config{}
+	c.SetDefaults()
+	c.FeeRecipient = etherBase.String()
+	configJSON, err := json.Marshal(c)
+	require.NoError(t, err)
+	issuer, vm, _, _ := GenesisVM(t, true, string(genesisJSON), string(configJSON), "")
```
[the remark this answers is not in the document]
ceyonur: added

precompile/allow_list.go
```diff
@@ -47,6 +47,7 @@ type AllowListConfig struct {
 // Configure initializes the address space of [precompileAddr] by initializing the role of each of
 // the addresses in [AllowListAdmins].
 func (c *AllowListConfig) Configure(state StateDB, precompileAddr common.Address) {
+	// First set enabled roles so these can be upgraded to admin addresses below.
```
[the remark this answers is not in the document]
ceyonur: yup removed it

contract-examples/contracts/ExampleRewardDistributor.sol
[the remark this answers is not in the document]
ceyonur: hmm I think that would actually equivalent to calling precompile contracts directly. I specifically put this one to show how fees can be distributed to multiple addresses. I can remove this example contract and modify tests to call precompile directly?

contract-examples/test/ExampleRewardDistributor.ts
ceyonur: added

consensus/dummy/consensus.go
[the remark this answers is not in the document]
ceyonur: hmm apperantly we cannot always verify it here since it breaks unit tests :/. they don't expect any coinbase verification or set the coinbase to blackhole, so eventually they fail.
https://github.com/ava-labs/subnet-evm/actions/runs/3427844912/jobs/5711340511

precompile/allow_list.go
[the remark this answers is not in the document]
ceyonur: restored the verification since it breaks too many unit tests :/

plugin/evm/block_verification.go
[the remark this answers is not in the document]
ceyonur: restored the verification since it breaks too many unit tests :/

contract-examples/contracts/ExampleRewardDistributor.sol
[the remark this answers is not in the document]
aaronbuchwald: It would still be different than calling the precompile directly since it would be calling the precompile from a smart contract rather than an EOA.

My fear is only that people will mis-interpret this as an example of how to handle fee rewards, and I'm not sure if we would want that here.

plugin/evm/vm_test.go
```diff
@@ -2506,15 +2508,27 @@ func TestAllowFeeRecipientDisabled(t *testing.T) {
...
+		modifiedHeader,
+		internalBlk.ethBlock.Transactions(),
+		nil,
+		nil,
+		new(trie.Trie),
+	)
+
+	modifiedBlk := vm.newBlock(modifiedBlock)
+
+	require.ErrorIs(t, modifiedBlk.Verify(), vmerrs.ErrInvalidCoinbase)
```
aaronbuchwald, line 2531: Nice! 🔥

contract-examples/contracts/IRewardManager.sol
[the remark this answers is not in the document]
aaronbuchwald: I think you're right here actually that it will be better and simpler to implement it this way as discussed offline, marking this as resolved.

aaronbuchwald, line 14: Can we add comments to these functions?

plugin/evm/vm_test.go
```diff
@@ -2566,13 +2580,145 @@ func TestAllowFeeRecipientEnabled(t *testing.T) {
...
+		require.NoError(t, vm.Shutdown())
+	}()
+
+	newTxPoolHeadChan := make(chan core.NewTxPoolReorgEvent, 1)
+	vm.txPool.SubscribeNewReorgEvent(newTxPoolHeadChan)
+
+	data, err := precompile.PackAllowFeeRecipients()
+	require.NoError(t, err)
+
+	tx := types.NewTransaction(uint64(0), precompile.RewardManagerAddress, big.NewInt(1), 21240+precompile.SetRewardAddressGasCost, big.NewInt(testMinGasPrice), data)
```
aaronbuchwald, line 2688: What's the reason for 21240?

plugin/evm/vm_test.go
```diff
@@ -2566,13 +2580,145 @@ func TestAllowFeeRecipientEnabled(t *testing.T) {
...
+	require.Equal(t, ethBlock.Coinbase(), testAddr) // reward address was activated at previous block
+	// Verify that etherBase has received fees
+	blkState, err := vm.blockChain.StateAt(ethBlock.Root())
+	require.NoError(t, err)
+
+	balance := blkState.GetBalance(testAddr)
+	require.Equal(t, 1, balance.Cmp(common.Big0))
+}
+
+func TestRewardManagerPrecompileAllowFeeRecipieints(t *testing.T) {
```
aaronbuchwald, line 2662: Can we add the following test cases as well:

- reward manager gets disabled and the rewards go back to the blackhole address
- reward manager gets disable and the rewards go back to genesis configuration of allow fee recipients and miner uses its original specified coinbase address

precompile/reward_manager.go
```diff
@@ -0,0 +1,472 @@
...
+
+// Singleton StatefulPrecompiledContract and signatures.
+var (
+	_ StatefulPrecompileConfig = &RewardManagerConfig{}
+
+	ErrCannotAllowFeeRecipients      = errors.New("non-enabled cannot allowFeeRecipients")
+	ErrCannotAreFeeRecipientsAllowed = errors.New("non-enabled cannot areFeeRecipientsAllowed")
+	ErrCannotCurrentRewardAddress    = errors.New("non-enabled cannot currentRewardAddress")
+	ErrCannotDisableRewards          = errors.New("non-enabled cannot disableRewards")
+	ErrCannotSetRewardAddress        = errors.New("non-enabled cannot setRewardAddress")
```
aaronbuchwald, line 42: should these be `cannot call ...`

precompile/reward_manager.go
```diff
@@ -0,0 +1,472 @@
...
+// Address returns the address of the RewardManager. Addresses reside under the precompile/params.go
+// Select a non-conflicting address and set it in the params.go.
+func (c *RewardManagerConfig) Address() common.Address {
+	return RewardManagerAddress
+}
+
+// Configure configures [state] with the initial configuration.
+func (c *RewardManagerConfig) Configure(chainConfig ChainConfig, state StateDB, _ BlockContext) {
+	c.AllowListConfig.Configure(state, RewardManagerAddress)
+	// CUSTOM CODE STARTS HERE
```
aaronbuchwald, line 149: Can we remove all `CUSTOM CODE` labels from this? I think we should generally remove these from production code.

precompile/reward_manager.go
```diff
@@ -0,0 +1,472 @@
...
+	ErrCannotSetRewardAddress        = errors.New("non-enabled cannot setRewardAddress")
+
+	ErrCannotEnableBothRewards = errors.New("cannot enable both fee recipients and reward address at the same time")
+	ErrEmptyRewardAddress      = errors.New("reward address cannot be empty")
+
+	RewardManagerABI        abi.ABI                     // will be initialized by init function
+	RewardManagerPrecompile StatefulPrecompiledContract // will be initialized by init function
+
+	rewardAddressStorageKey        = common.Hash{'r', 'a', 's', 'k'}
+	allowFeeRecipientsAddressValue = common.Address{'a', 'f', 'r', 'a', 'v'}
```
aaronbuchwald, line 51: Since this is an address it may collide with an address value supplied by a user accidentally, which seems like an unintended behavior.

Could we fix this by using a hash value here that populates the left most 12 bytes? Since an address converted to a hash takes up only the right most 20 bytes, this will guarantee that a user supplied address will not result in the same value.

core/stateful_precompile_test.go
```diff
@@ -1060,22 +941,321 @@ func TestFeeConfigManagerRun(t *testing.T) {
...
+				input, err := precompile.PackSetRewardAddress(testAddr)
+				require.NoError(t, err)
+
+				return input
+			},
+			suppliedGas: precompile.SetRewardAddressGasCost,
+			readOnly:    false,
+			expectedRes: []byte{},
+		},
+		"disable rewards from enabled succeeds": {
```
aaronbuchwald, line 1042: Should we add an assertion on the state to this test?

core/stateful_precompile_test.go
```diff
@@ -1060,22 +941,321 @@ func TestFeeConfigManagerRun(t *testing.T) {
...
+				input, err := precompile.PackAllowFeeRecipients()
+				require.NoError(t, err)
+
+				return input
+			},
+			suppliedGas: precompile.AllowFeeRecipientsGasCost,
+			readOnly:    false,
+			expectedRes: []byte{},
+		},
+		"set reward address from enabled succeeds": {
```
aaronbuchwald, line 1030: Should we add an assertion on the state to the end of this test?

core/stateful_precompile_test.go
```diff
@@ -1060,22 +941,321 @@ func TestFeeConfigManagerRun(t *testing.T) {
...
+				input, err := precompile.PackDisableRewards()
+				require.NoError(t, err)
+
+				return input
+			},
+			suppliedGas: precompile.DisableRewardsGasCost,
+			readOnly:    false,
+			expectedErr: precompile.ErrCannotDisableRewards.Error(),
+		},
+		"set allow fee recipients from enabled succeeds": {
```
aaronbuchwald, line 1018: Should we add an assertion on the state to the end of this test?

consensus/consensus.go
```diff
@@ -57,6 +57,9 @@ type ChainHeaderReader interface {
 
 	// GetFeeConfigAt retrieves the fee config and last changed block number at block header.
 	GetFeeConfigAt(parent *types.Header) (commontype.FeeConfig, *big.Int, error)
+
+	// GetCoinbaseAt retrieves the configured coinbase address at [parent]. If fee recipients are allowed, returns true in the second return value.
+	GetCoinbaseAt(parent *types.Header) (common.Address, bool, error)
```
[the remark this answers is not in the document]
ceyonur: changed returning empty address. we now return whatever address stored in the contract (a predefined address).

contract-examples/contracts/IRewardManager.sol
ceyonur: yep added

plugin/evm/vm_test.go
ceyonur: Do you mean tests should cover reward manager precompile itself is disabled or rewards are disabled through the precompile? In the second one I don’t think we should resort back to genesis configuration since it is the intended way to disable whole rewards and send them to blackhole.

precompile/reward_manager.go
ceyonur: yep added to both template and here

precompile/reward_manager.go
ceyonur: yep, makes utter sense. changed it to a reserved address.

core/stateful_precompile_test.go
ceyonur: added

core/stateful_precompile_test.go
ceyonur: added

precompile/reward_manager.go
aaronbuchwald: Can we switch to using a `common.Hash` value instead of a `common.Address` value instead? Since we left-pad the address values with zero bytes to create a 32 byte value, if we use a hash value that has a non-zero value in the left-most 12 bytes, then we will guarantee not to have any collisions without needing to reserve an additional address.

precompile/reward_manager.go
```diff
@@ -0,0 +1,455 @@
...
+		DisableFeeRewards(state)
+	} else { // configure the RewardManager according to chainConfig
+		if chainConfig.AllowedFeeRecipients() {
+			EnableAllowFeeRecipients(state)
+			return
+		}
+		// chainConfig does not have any reward address
+		// if it does not enable fee recipients, disable rewards
+		DisableFeeRewards(state)
+	}
```
aaronbuchwald, line 173: can we clean the conditions in this block of code up?

tests/e2e/solidity/suites.go
```diff
@@ -92,14 +92,25 @@ var _ = utils.DescribePrecompile(func() {
...
+	})
+
 	// ADD YOUR PRECOMPILE HERE
 	/*
 			ginkgo.It("your precompile", ginkgo.Label("solidity-with-npx"), func() {
 			err := startSubnet("./tests/e2e/genesis/{your_precompile}.json")
 			gomega.Expect(err).Should(gomega.BeNil())
 			running := runner.IsRunnerUp(grpcEp)
 			gomega.Expect(running).Should(gomega.BeTrue())
-			runHardhatTests("./test/Example{YourPrecompile}Test.ts")
```
aaronbuchwald, line 102: Should this be reverted back since we still use `Example{...}` in most of our file names?

precompile/reward_manager.go
```diff
@@ -0,0 +1,457 @@
...
+type InitialRewardConfig struct {
+	AllowFeeRecipients bool           `json:"allowFeeRecipients"`
+	RewardAddress      common.Address `json:"rewardAddress,omitempty"`
+}
+
+func (i *InitialRewardConfig) Verify() error {
+	switch {
+	case i.AllowFeeRecipients && i.RewardAddress != (common.Address{}):
+		return ErrCannotEnableBothRewards
+		// shall we also check blackhole address here?
```
aaronbuchwald, line 63: I think we can just check this condition and treat the blackhole address the same as any other address. Can we remove the question here?

precompile/reward_manager.go
```diff
@@ -0,0 +1,457 @@
...
+}
+
+func (i *InitialRewardConfig) Configure(state StateDB) {
+	// enable allow fee recipients
+	if i.AllowFeeRecipients {
+		EnableAllowFeeRecipients(state)
+	} else if i.RewardAddress == (common.Address{}) {
+		// if reward address is empty and allow fee recipients is false
+		// then disable rewards
+		DisableFeeRewards(state)
```
aaronbuchwald, line 84: Since `DisableFeeRewards` is identical to setting the reward address to the burn address, would it make sense to get rid of `DisableFeeRewards` helper and just store the specified reward address with no further check? I don't think it should be necessary to re-route the zero address to the burn address that we have used in the past.

We could also get rid of the `disableRewards` function since it does the same thing as `setRewardAddress` with a special case as well.

contract-examples/tasks.ts
```diff
@@ -215,11 +216,57 @@ task("feeManager:get", "gets fee config")
...
 task("feeManager:readRole", "a task to get the network deployer minter list")
   .addParam("address", "the address you want to know the minter role for")
   .setAction(async (args, hre) => {
     const allowList = await hre.ethers.getContractAt("IFeeManager", FEE_MANAGER_ADDRESS)
     await getRole(allowList, args.address)
   })
 
+
+// npx hardhat rewardManager:currentRewardAddress --network local
+task("rewardManager:currentRewardAddress", "a task to get the current configured rewarding address")
```
darioush, line 229:
```suggestion
task("rewardManager:currentRewardAddress", "a task to get the current configured reward address")
```

contract-examples/tasks.ts
```diff
@@ -215,11 +216,57 @@ task("feeManager:get", "gets fee config")
...
+    const result = await rewardManager.currentRewardAddress()
+    if (areFeeRecipientsAllowed) {
+      console.log("Custom Fee Recipients are allowed. (%s)", result)
+    } else {
+      console.log(`Current reward address is ${result}`)
+    }
+  })
+
+// npx hardhat rewardManager:areFeeRecipientsAllowed --network local
+task("rewardManager:areFeeRecipientsAllowed", "a task to get whether the fee recipients are allowed")
```
darioush, line 242:
```suggestion
task("rewardManager:areFeeRecipientsAllowed", "a task to get whether custom fee recipients are allowed to receive rewards")
```

darioush: should we drop the "a task to..." from these since the other descriptions don't have it?

contract-examples/test/ExampleRewardManager.ts
```diff
@@ -0,0 +1,154 @@
...
+import { expect } from "chai";
+import {
+  BigNumber,
+  Contract,
+  ContractFactory,
+} from "ethers"
+import { ethers } from "hardhat"
+import ts = require("typescript");
+
+// make sure this is always an admin for minter precompile
```
darioush, line 14: should this be reward manager precompile?

params/config.go
```diff
@@ -547,7 +555,14 @@ func (c *ChainConfig) AvalancheRules(blockNum, blockTimestamp *big.Int) Rules {
 	return rules
 }
 
-// GetFeeConfig returns the FeeConfig
+// GetFeeConfig returns the original FeeConfig contained in the ChainConfig (will not be modified by FeeConfigManager.
```
darioush, line 558: nit: missing `)`

params/config.go
```diff
@@ -547,7 +555,14 @@ func (c *ChainConfig) AvalancheRules(blockNum, blockTimestamp *big.Int) Rules {
...
+// Implements precompile.ChainConfig interface.
 func (c *ChainConfig) GetFeeConfig() commontype.FeeConfig {
 	return c.FeeConfig
 }
+
+// AllowedFeeRecipients returns whether the original AllowedFeeRecipients parameter (will not be modified by the RewardManager.
```
darioush, line 564: same as above

precompile/reward_manager.go
```diff
@@ -0,0 +1,457 @@
...
...ddress\",\"name\":\"rewardAddress\",\"type\":\"address\"}],\"stateMutability\":\"view\",\"type\":\"function\"},{\"inputs\":[],\"name\":\"disableRewards\",\"outputs\":[],\"stateMutability\":\"nonpayable\",\"type\":\"function\"},{\"inputs\":[{\"internalType\":\"address\",\"name\":\"addr\",\"type\":\"address\"}],\"name\":\"readAllowList\",\"outputs\":[{\"internalType\":\"uint256\",\"name\":\"role\",\"type\":\"uint256\"}],\"stateMutability\":\"view\",\"type\":\"function\"},{\"inputs\":[{\"internalType\":\"address\",\"name\":\"addr\",\"type\":\"address\"}],\"name\":\"setAdmin\",\"outputs\":[],\"stateMutability\":\"nonpayable\",\"type\":\"function\"},{\"inputs\":[{\"internalType\":\"address\",\"name\":\"addr\",\"type\":\"address\"}],\"name\":\"setEnabled\",\"outputs\":[],\"stateMutability\":\"nonpayable\",\"type\":\"function\"},{\"inputs\":[{\"internalType\":\"address\",\"name\":\"addr\",\"type\":\"address\"}],\"name\":\"setNone\",\"outputs\":[],\"stateMutability\":\"nonpayable\",\"type\":\"function\"},{\"inputs\":[{\"internalType\":\"address\",\"name\":\"addr\",\"type\":\"address\"}],\"name\":\"setRewardAddress\",\"outputs\":[],\"stateMutability\":\"nonpayable\",\"type\":\"function\"}]"
```
darioush, line 31: wonder if we should switch to embedding these?
separate work from this of course, @aaronbuchwald @ceyonur what do you think?

darioush (commented):
Just some nits

tests/e2e/solidity/suites.go
ceyonur: This is not a hard requirement. There is also `ERC20NativeMinter`, so I did not want to imply it should start with `Example`.

precompile/reward_manager.go
ceyonur: removed question

precompile/reward_manager.go
ceyonur: In the code-wise you're absolutely correct, just wanted to improve UX by adding this function. I don't think it's very practical to call the function with blackhole address each time, especially from external contracts. We don't provide burn address in EVM to contracts, so I think it's very likely to have a mistake in those contracts. Like having a wrong blackhole address etc.

precompile/reward_manager.go
ceyonur: ABI is required in the code to encode/decode inputs/outputs. alternatively we can put them in a separate file and read here.

params/config.go
ceyonur: reworded those

contract-examples/test/ExampleRewardManager.ts
ceyonur: yup changed it

params/config.go
ceyonur: reworded

precompile/reward_manager.go
aaronbuchwald: I agree it would be nice to move this to a separate file and use go embedding and update the precompilegen to do that for you.

ceyonur: I will create a separate PR for that

contract-examples/test/ExampleRewardManager.ts
```diff
@@ -0,0 +1,154 @@
...
+
+    // Send a transaction to mine a new block
+    const tx = await owner.sendTransaction({
+      to: signer1.address,
+      value: ethers.utils.parseEther("10")
+    })
+    await tx.wait()
+  });
+
+  it("should add contract deployer as owner", async function () {
```
aaronbuchwald, line 53: Can we add one final hardhat test for `disableRewards()`? Seems this is not tested in the HardHat tests atm.

ceyonur: added

aaronbuchwald (approved):
LGTM

PR #265, opened 2022-09-20
Nits

aaronbuchwald:
This PR addresses a few nits on the coreth migration.

### DIFF
--- a/plugin/evm/block_verification.go
+++ b/plugin/evm/block_verification.go
@@ -54,18 +54,6 @@ func (v blockValidator) SyntacticVerify(b *Block, rules params.Rules) error {
 		)
 	}
 
-	// if this is SubnetEVM and FeeConfigManager is enabled we cannot check this.
-	// FeeConfig depends on state. State is not available here, so skip checking the gas limit.
-	if !(rules.IsSubnetEVM && rules.IsFeeConfigManagerEnabled) {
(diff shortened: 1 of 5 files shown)


plugin/evm/block_verification.go
```diff
@@ -54,18 +54,6 @@ func (v blockValidator) SyntacticVerify(b *Block, rules params.Rules) error {
...
-	if !(rules.IsSubnetEVM && rules.IsFeeConfigManagerEnabled) {
-		expectedGas := b.vm.chainConfig.FeeConfig.GasLimit.Uint64()
-		if ethHeader.GasLimit != expectedGas {
-			return fmt.Errorf(
-				"expected gas limit to be %d in subnetEVM but got %d",
-				expectedGas, ethHeader.GasLimit,
-			)
-		}
-	}
-
```
ceyonur, line 68: Yup that was totally wrong, nice catch. but I don't think we need to remove this altogether. Wouldn't changing condition to 
`rules.IsSubnetEVM && !rules.IsFeeConfigManagerEnabled` work?

aaronbuchwald: Ya this would work to, but we already enforce this correctly in the consensus code where we verify the gas limit, so I thought it would be simpler to remove from here instead.

PR #275, opened 2022-09-22
Last accepted compatibility reg fail

ceyonur:
Demonstrates failing regression test for #258
This PR uses header instead of last block in genesis, to intentionally fail regression test.


aaronbuchwald:
Thanks for putting this up to verify.

PR #276, opened 2022-09-22
add duplicate check in allow list

### LINKED ISSUE #269: Require Allow List Config does not have duplicates
We should add a requirement to the `Verify()` function for the allow list config that there are no duplicated addresses.

ceyonur:
Adds duplicate check to allow list

closes #269

### DIFF
--- a/precompile/allow_list.go
+++ b/precompile/allow_list.go
@@ -87,26 +87,25 @@ func (c *AllowListConfig) Verify() error {
 		return nil
 	}
 
-	enabledMap := make(map[common.Address]struct{})
+	addressMap := make(map[common.Address]bool)
 	for _, enabledAddr := range c.EnabledAddresses {
 		// check for duplicates
-		if _, ok := enabledMap[enabledAddr]; ok {
+		if _, ok := addressMap[enabledAddr]; ok {
 			return fmt.Errorf("duplicate address %s in enabled list", enabledAddr)
 		}
-		enabledMap[enabledAddr] = struct{}{}
+		addressMap[enabledAddr] = false
 	}
 
-	adminMap := make(map[common.Address]struct{})
 	for _, adminAddr := range c.AllowListAdmins {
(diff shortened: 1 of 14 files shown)


precompile/config_test.go
```diff
@@ -64,6 +64,16 @@ func TestVerifyPrecompileUpgrades(t *testing.T) {
 			config:        NewContractNativeMinterConfig(big.NewInt(3), admins, admins, nil),
 			expectedError: "cannot set address",
 		},
+		{
+			name:          "duplicate admins in config in native minter allowlist",
+			config:        NewContractNativeMinterConfig(big.NewInt(3), append(admins, admins[0]), enableds, nil),
+			expectedError: "duplicate address",
+		},
+		{
+			name:          "duplicate admins in config in native minter allowlist",
```
darioush, line 73:
```suggestion
			name:          "duplicate enableds in config in native minter allowlist",
```

darioush (approved):

precompile/allow_list.go
```diff
@@ -86,16 +86,27 @@ func (c *AllowListConfig) Verify() error {
...
+
+	adminMap := make(map[common.Address]struct{})
 	for _, adminAddr := range c.AllowListAdmins {
+		// check for overlap between enabled and admin lists
 		if _, ok := enabledMap[adminAddr]; ok {
 			return fmt.Errorf("cannot set address %s as both admin and enabled", adminAddr)
 		}
+		// check for duplicates
+		if _, ok := adminMap[adminAddr]; ok {
+			return fmt.Errorf("duplicate address %s in enabled list", adminMap)
```
aaronbuchwald, line 107: Since we do not allow the same address to be included in both admins and enableds, can we switch to using a single map to ensure that we do not have any repeats across both?

ceyonur: done

aaronbuchwald (approved):

PR #279, opened 2022-09-26
bump to avago 1.8.6

felipemadero:


.goreleaser.yml
```diff
@@ -6,16 +6,29 @@ builds:
...
-        goarch: 386
+    goos:
+      - linux
+      - darwin
+    goarch:
+      - amd64
+      - arm64
+    env:
+      - CGO_ENABLED=1
+      - CGO_FLAGS="-O -D__BLST_PORTABLE__"
```
darioush, line 17: should we add this flag to any instructions we have to build from source @yulin-dong?

felipemadero: this seems to be a good workaround for the pretty strange blst-related runtime error `Caught SIGILL in blst_cgo_init, consult <blst>/bindings/go/README.md.`  eg https://github.com/ava-labs/subnet-evm/actions/runs/3130327361/jobs/5080451590

scripts/build.sh
```diff
@@ -50,4 +50,5 @@ fi
 
 # Build Subnet EVM, which is run as a subprocess
 echo "Building Subnet EVM Version: $subnet_evm_version at $binary_path"
+export CGO_CFLAGS="-O -D__BLST_PORTABLE__"
```
patrick-ogrady, line 53: Could you copy the comment for this from AvalancheGo?

ceyonur (commented):
we should also update avago version in `scripts/versions.sh`

.goreleaser.yml
yulin-dong: Currently our developer doc are more focused on avalanche-cli instead of subnet-evm directly. So I think we are fine not updating doc.

scripts/run.sh
```diff
@@ -55,6 +55,9 @@ echo ENABLE_SOLIDITY_TESTS: ${ENABLE_SOLIDITY_TESTS}
 echo GINKGO_SKIP_FLAGS: ${GINKGO_SKIP_FLAGS}
 echo AVALANCHE_LOG_LEVEL: ${AVALANCHE_LOG_LEVEL}
 
+# avoid blst related invalid operation errors
+export CGO_CFLAGS="-O -D__BLST_PORTABLE__"
```
ceyonur: could we move these flags into `constants.sh` and instead import it in `run.sh` and `build_test.sh`?

felipemadero: already did. maybe you are looking at a previous version of the code?

ceyonur: yup my bad, thanks!

.github/workflows/lint-tests-release.yml
```diff
@@ -95,6 +95,19 @@ jobs:
...
+        run: sudo apt-get -y install gcc-aarch64-linux-gnu
+      - name: Checkout osxcross
+        uses: actions/checkout@v2
+        with:
+          repository: tpoechtrager/osxcross
+          path: osxcross
+      - name: Build osxcross
+        run: |
+          cd osxcross
+          wget https://github.com/phracker/MacOSX-SDKs/releases/download/11.3/MacOSX11.3.sdk.tar.xz -O tarballs/MacOSX11.3.sdk.tar.xz
```
darioush, line 108: should we verify a checksum on this file?

darioush (approved):

.github/workflows/lint-tests-release.yml
felipemadero: not a bad idea. adding

PR #283, opened 2022-09-30
*: track accepted transaction counts

gyuho:
Screen Shot 2022-09-29 at 5 48 52 PM


aaronbuchwald:
I'm not a big fan of the pattern that is used here, but if we are going to add this we should make two changes:

1. Make an equivalent PR to coreth and go-ethereum
2. Follow the same style that is used for existing metrics that is present here: https://github.com/ava-labs/coreth/blob/master/core/tx_pool.go#L111

PR #292, opened 2022-10-05
use anr pid in e2e test yaml

felipemadero:
Currently e2e test binary is writing its own pid in the yaml it generates with ANR info. This PR changes that so that the yaml contains ANR server pid. 
Also changes simulator flag name so the yaml is not interpreted as ANR-generated.


darioush (approved):

darioush:
can you add a description to the pr?

tests/e2e/runner/runner.go
```diff
@@ -136,20 +140,23 @@ done:
...
+	}
+	if pid == 0 {
+		return "", "", pid, errors.New("pid not found")
 	}
-	return blockchainID, logsDir, nil
+	return blockchainID, logsDir, pid, nil
 }
 
-func GetClusterInfo(blockchainId string, logsDir string) (clusterInfo, error) {
+func GetClusterInfo(blockchainId string, logsDir string, pid int) (clusterInfo, error) {
```
aaronbuchwald, line 159: We should not need to pass a process ID into a getter function to return in the output.

tests/e2e/runner/runner.go
```diff
@@ -92,8 +92,9 @@ func startRunner(vmName string, genesisPath string, pluginDir string) error {
 	return nil
 }
 
-func WaitForCustomVm(vmId ids.ID) (string, string, error) {
+func WaitForCustomVm(vmId ids.ID) (string, string, int, error) {
```
aaronbuchwald, line 95: Can we move this logic inside of the avalanche-network-runner? I would think that we can move the for loop here into the avalanche-network-runner and simplify this code which will likely need to be duplicated in any project doing this similar and most likely very common task.

aaronbuchwald (approved):

tests/e2e/runner/runner.go
felipemadero: both logsDir and pid are passed to this probably ill-named function that also ask server for uris and save cluster info data (including given logsDir+pid) to disk before returning it. Renaming to SaveClusterInfo

tests/e2e/runner/runner.go
felipemadero: This is already included for second ANR refactoring, but will anyway make a explicit issue now and work first on it. https://github.com/ava-labs/avalanche-network-runner/issues/302
After doing that, will add issues to modify CLI/subnet-evm

PR #298, opened 2022-10-06
fix goreleaser problem from v0.4.0 

felipemadero:
Goreleaser failing on real releases. there is a check for the git repositoty to not being dirty, and  that seems to only be checked when doing the release, not when doing snapshots. That's why it was not seen before.
The repo was dirty because osxcross was inside it. this PR sets osxcross and subnet-evm side by side.


aaronbuchwald (approved):
LGTM

PR #306, opened 2022-10-14
Review nits for porting corethv0.10.1

darioush:


internal/ethapi/api.go
```diff
@@ -138,18 +138,9 @@ func (s *EthereumAPI) Syncing() (interface{}, error) {
...
-
-func (s *BlockChainAPI) GetChainConfig(ctx context.Context) GetChainConfigResponse {
-	config := s.b.ChainConfig()
-	resp := GetChainConfigResponse{
-		ChainConfig:   config,
-		UpgradeConfig: config.UpgradeConfig,
-	}
-	return resp
+// GetChainConfig returns the chain config.
+func (s *EthereumAPI) GetChainConfig(ctx context.Context) *params.ChainConfig {
```
ceyonur, line 142: duplicated API?

darioush: I am not sure how this is supposed to work? should this be defined on EthereumAPI or BlockchainAPI?

ceyonur: ChainID is under the BlockchainAPI so I thought it makes sense to put this under BlockchainAPI.

aaronbuchwald: I think this makes more sense under `BlockchainAPI` and we should make sure that it still includes the `UpgradeConfig` change so that the upgrade bytes are included in the JSON response correctly

darioush: Should we make the same change in coreth?

aaronbuchwald (approved):
LGTM

internal/ethapi/api.go
aaronbuchwald: Ah I see, yes we should, we actually documented it this way as well: https://docs.avax.network/apis/avalanchego/apis/subnet-evm#eth_getchainconfig

ceyonur (approved):

PR #311, opened 2022-10-20
Save cache periodically

aaronbuchwald:
This PR adds back the original functionality from geth to save the trie clean cache to disk and load it on startup to reduce lower performance on startup due to warming up the cache.

This includes a change in behavior from the original geth code to save the cache to disk from `SaveCachePeriodically` when a stop signal is sent, so that it does not require an explicit call to `SaveCache` from `core/blockchain.go`.


trie/database.go
```diff
@@ -893,3 +899,50 @@ func (db *Database) Size() (common.StorageSize, common.StorageSize) {
...
+		log.Error("Failed to persist clean trie cache", "error", err)
+		return err
+	}
+	log.Info("Persisted the clean trie cache", "path", dir, "elapsed", common.PrettyDuration(time.Since(start)))
+	return nil
+}
+
+// SaveCache atomically saves fast cache data to the given dir using all
+// available CPU cores.
+func (db *Database) SaveCache(dir string) error {
```
darioush, line 923: since this behavior requires a comment (explaining the use of all available CPUs)
should we just inline this method and its comment in SaveCachePeriodically and not have an exported method here?

darioush (commented):
should we add a UT?

trie/database.go
```diff
@@ -893,3 +899,50 @@ func (db *Database) Size() (common.StorageSize, common.StorageSize) {
...
+	return db.saveCache(dir, runtime.GOMAXPROCS(0))
```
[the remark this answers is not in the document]
aaronbuchwald: Why specifically one less? Definitely open to not using every core.

trie/database.go
aaronbuchwald: Since this is just copied from geth, I'd prefer to leave as is.

aaronbuchwald:
> should we add a UT?

I don't think it's 100% necessary, but would be good to add. There's not currently one in geth and it may require a dirty hack since `core/blockchain.go` will currently sanitize a frequency of less than a minute.

trie/database.go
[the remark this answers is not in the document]
darioush: Leaving at least one for everything else :) Also open to other ideas.

trie/database.go
```diff
@@ -893,3 +899,49 @@ func (db *Database) Size() (common.StorageSize, common.StorageSize) {
...
+		case <-ticker.C:
+			db.saveCache(dir, 1)
+		case <-stopCh:
+			// Write the latest contents of the cache to disk after receiving a stop request.
+			// Note: this is different than geth, which requires an explicit
+			// call to save the cache on shutdown.
+			err := db.SaveCache(dir)
+			if err != nil {
+				log.Warn("Failed to save cache after stop requested", "err", err)
+			}
```
darioush, line 944:
```suggestion
			}
			return
```

trie/database.go
```diff
@@ -298,7 +300,11 @@ func NewDatabase(diskdb ethdb.KeyValueStore) *Database {
 func NewDatabaseWithConfig(diskdb ethdb.KeyValueStore, config *Config) *Database {
 	var cleans *fastcache.Cache
 	if config != nil && config.Cache > 0 {
-		cleans = fastcache.New(config.Cache * 1024 * 1024)
+		if config.Journal == "" {
+			cleans = fastcache.New(config.Cache * 1024 * 1024)
+		} else {
+			cleans = fastcache.LoadFromFileOrNew(config.Journal, config.Cache*1024*1024)
```
ceyonur, line 306: `config.Cache` * 1024 * 1024 => `config.Cache`  * `units.MiB?` ignore if we want to minimize the diff.

trie/database.go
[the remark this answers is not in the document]
ceyonur: maybe half?

ceyonur (commented):
some nits but ignore if we want to minimize the diff.

core/blockchain.go
```diff
@@ -326,6 +339,20 @@ func NewBlockChain(
 	// Start processing accepted blocks effects in the background
 	go bc.startAcceptor()
 
+	// If periodic cache journal is required, spin it up.
+	if bc.cacheConfig.TrieCleanRejournal > 0 && len(bc.cacheConfig.TrieCleanJournal) > 0 {
+		if bc.cacheConfig.TrieCleanRejournal < time.Minute {
```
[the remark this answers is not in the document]
aaronbuchwald: Yup, good call

plugin/evm/config.go
```diff
@@ -90,10 +90,12 @@ type Config struct {
 	RPCTxFeeCap float64 `json:"rpc-tx-fee-cap"`
 
 	// Cache settings
-	TrieCleanCache        int `json:"trie-clean-cache"`         // Size of the trie clean cache (MB)
-	TrieDirtyCache        int `json:"trie-dirty-cache"`         // Size of the trie dirty cache (MB)
-	TrieDirtyCommitTarget int `json:"trie-dirty-commit-target"` // Memory limit to target in the dirty cache before performing a commit (MB)
-	SnapshotCache         int `json:"snapshot-cache"`           // Size of the snapshot disk layer clean cache (MB)
+	TrieCleanCache        int      `json:"trie-clean-cache"`         // Size of the trie clean cache (MB)
+	TrieCleanJournal      string   `json:"trie-clean-journal"`       // Directory to use to save the trie clean cache (must be populated to enable journaling the trie clean cache)
+	TrieCleanRejournal    Duration `json:"trie-clean-rejournal"`     // Frequency to re-journal the trie clean cache to disk (minimum 1 minute, must be populated to enable journaling the trie clean cache)
```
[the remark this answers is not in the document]
aaronbuchwald: I'd prefer to minimize the diff and add clear comments

trie/database.go
aaronbuchwald: I'd prefer to minimize the diff rather than import an avalanchego constant here.

trie/database.go
```diff
@@ -893,3 +899,54 @@ func (db *Database) Size() (common.StorageSize, common.StorageSize) {
...
+	err := db.cleans.SaveToFileConcurrent(dir, threads)
+	if err != nil {
+		log.Error("Failed to persist clean trie cache", "error", err)
+		return err
+	}
+	log.Info("Persisted the clean trie cache", "path", dir, "elapsed", common.PrettyDuration(time.Since(start)))
+	return nil
+}
+
+// SaveCache atomically saves fast cache data to the given dir using all
```
darioush, line 921:
```suggestion
// SaveCache atomically saves fast cache data to the given dir using half
```

darioush (approved):

PR #313, opened 2022-10-21
expose stats from fastcache

darioush:


utils/metered_cache.go
```diff
@@ -0,0 +1,90 @@
...
+	return mc
+}
+
+// updateStats updates metrics from fastcache
+func (mc *MeteredCache) updateStats() {
+	s := fastcache.Stats{}
+	mc.UpdateStats(&s)
+	for statName, stat := range mc.stats {
+		stat.Update(int64(statsMap[statName](&s)))
+	}
```
aaronbuchwald, line 84: This seems a bit overly complex to use in one place. Could we simplify this and just call update with the correct values from stats rather than create a map of functions to return the uint64 values above?

trie/database.go
```diff
@@ -281,8 +282,9 @@ func expandNode(hash hashNode, n node) node {
 
 // Config defines all necessary options for database.
 type Config struct {
-	Cache     int  // Memory allowance (MB) to use for caching trie nodes in memory
-	Preimages bool // Flag whether the preimage of trie key is recorded
+	Cache          int    // Memory allowance (MB) to use for caching trie nodes in memory
+	Preimages      bool   // Flag whether the preimage of trie key is recorded
+	StatsNamespace string // Namespace to use for stats
```
aaronbuchwald, line 287: Can we add a comment to explain why we use a namespace in the config here whereas in most other places we rely on a global namespace?

utils/metered_cache.go
```diff
@@ -0,0 +1,114 @@
...
+	if mc.namespace == "" {
+		return
+	}
+	mc.ops++
+	if mc.ops%mc.updateFrequency != 0 {
+		return
+	}
+
+	s := fastcache.Stats{}
+	mc.UpdateStats(&s)
```
patrick-ogrady, line 72: Do we want to store a metric for how long this takes to run?

darioush: made a change. or did you mean a cumulative metric to measure the total time spent updating stats?

patrick-ogrady: I meant the latter (want to make sure this doesn't take too long)

utils/metered_cache.go
```diff
@@ -54,6 +57,7 @@ func NewMeteredCache(size int, journal string, namespace string, updateFrequency
 		mc.gets = metrics.GetOrRegisterGauge(fmt.Sprintf("%s/gets", namespace), nil)
 		mc.sets = metrics.GetOrRegisterGauge(fmt.Sprintf("%s/sets", namespace), nil)
 		mc.misses = metrics.GetOrRegisterGauge(fmt.Sprintf("%s/misses", namespace), nil)
```
patrick-ogrady, line 59: aren't some of these counts?

utils/metered_cache.go
```diff
@@ -54,6 +57,7 @@ func NewMeteredCache(size int, journal string, namespace string, updateFrequency
...
+		mc.statsTime = metrics.GetOrRegisterGauge(fmt.Sprintf("%s/statsTime", namespace), nil)
```
patrick-ogrady, line 60: I think this is should be `RegisteredResettingTimer`?

utils/metered_cache.go
darioush: gets, sets, misses and collisions are monotonically increasing unless `Reset` is called (which is not used in the codebase).
should we switch them to meters?

patrick-ogrady: I think that makes a lot more sense IMO. We won't call reset?

You may want to add a comment about that.

patrick-ogrady: nvm

utils/metered_cache.go
darioush: Reverted that change since using a timer would be more suited for individual measurement. We can use gauge or meter for the cumulative time

patrick-ogrady (approved):

utils/metered_cache.go
```diff
@@ -0,0 +1,120 @@
...
+	mc.bytesSize.Update(int64(s.BytesSize))
+	mc.collisions.Update(int64(s.Collisions))
+	mc.gets.Update(int64(s.GetCalls))
+	mc.sets.Update(int64(s.SetCalls))
+	mc.misses.Update(int64(s.Misses))
+	mc.statsTime.Inc(int64(time.Since(start))) // cumulative metric
+}
+
+func (mc *MeteredCache) Del(k []byte) {
+	defer mc.updateStatsIfNeeded()
```
aaronbuchwald, line 88: Can we remove all of these defer statements? I'm not sure there's any reason to perform this at the end of the function call instead of the beginning.

aaronbuchwald (approved):
LGTM

PR #314, opened 2022-10-21
adds a UT for journal clean cache on shutdown

darioush:


aaronbuchwald (approved):
LGTM - great catch and unit test

Issue #245, opened 2022-08-31
flaky test TestArchiveBlockChain/TestStatefulPrecompiles


darioush:
seen this test failure a couple times now. eg, https://github.com/ava-labs/subnet-evm/runs/8120806684?check_suite_focus=true
```
    --- FAIL: TestArchiveBlockChain/TestStatefulPrecompiles (0.06s)
        test_blockchain.go:166: head state missing 1:0x9bd1529846788581c59ad30279e44797e094f6f00d391d4a48cc4324a91dad01
```

PR #316, opened 2022-10-24
fixes flakes in TestArchiveBlockChainSnapsDisabled

darioush:
tested by `go test -count=100 -run=TestArchiveBlockChainSnapsDisabled ./core`

closes: https://github.com/ava-labs/subnet-evm/issues/245

If we don't synchronize on accept (leave it async) we can hit this line:
https://github.com/ava-labs/subnet-evm/blob/992d683beea10ad52c030f462fd172091668ab3d/core/test_blockchain.go#L112

where LastAcceptedBlock differs from LastConsensusAcceptedBlock and the state for the lastAccepted would not be available (since Accept could be processing concurrently writes may not have made it to the DB).


ceyonur (approved):

aaronbuchwald:
Could you add to the description how this fixes the issue?

We should be able to call `DrainAcceptorQueue` after calling `Accept` on a sequence of blocks to ensure that they have been fully processed.

aaronbuchwald (approved):
LGTM

PR #319, opened 2022-10-26
Save cache periodically test

aaronbuchwald:
- Remove the minimum enforced trie clean rejournal interval of 1 minute
- Add unit test to ensure that trie clean rejournaling generates files correctly

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -347,10 +347,8 @@ func NewBlockChain(
 
 	// If periodic cache journal is required, spin it up.
 	if bc.cacheConfig.TrieCleanRejournal > 0 && len(bc.cacheConfig.TrieCleanJournal) > 0 {
-		if bc.cacheConfig.TrieCleanRejournal < time.Minute {
-			log.Warn("Sanitizing invalid trie cache journal time", "provided", bc.cacheConfig.TrieCleanRejournal, "updated", time.Minute)
-			bc.cacheConfig.TrieCleanRejournal = time.Minute
-		}
+		log.Info("Starting to save trie clean cache periodically", "journalDir", bc.cacheConfig.TrieCleanJournal, "freq", bc.cacheConfig.TrieCleanRejournal)
+
 		triedb := bc.stateCache.TrieDB()
 		bc.rejournalWg.Add(1)
 		go func() {
--- a/core/blockchain_test.go
+++ b/core/blockchain_test.go
@@ -21,6 +21,7 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/fsnotify/fsnotify"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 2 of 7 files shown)


core/blockchain_test.go
```diff
@@ -70,6 +72,77 @@ func TestArchiveBlockChain(t *testing.T) {
...
+
+	time.Sleep(2 * time.Second) // Sleep for 2 seconds to ensure that there is time for a clean trie rejournal
+
+	dirEntries, err := os.ReadDir(trieCleanJournal)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if len(dirEntries) == 0 {
+		t.Fatalf("Expected trie clean journal to generate non-zero number of entries in the trie clean journal directory")
```
darioush, line 142: nit: can we switch to using `require` instead of `t.Fatal`

core/blockchain_test.go
```diff
@@ -70,6 +72,77 @@ func TestArchiveBlockChain(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	for _, block := range chain {
+		if err := blockchain.Accept(block); err != nil {
+			t.Fatal(err)
+		}
+	}
+	blockchain.DrainAcceptorQueue()
+
+	time.Sleep(2 * time.Second) // Sleep for 2 seconds to ensure that there is time for a clean trie rejournal
```
darioush, line 134: should we have a channel to listen on instead? seems this could be flaky or block the tests for 2 seconds.

core/blockchain.go
```diff
@@ -347,10 +347,8 @@ func NewBlockChain(
 
 	// If periodic cache journal is required, spin it up.
 	if bc.cacheConfig.TrieCleanRejournal > 0 && len(bc.cacheConfig.TrieCleanJournal) > 0 {
-		if bc.cacheConfig.TrieCleanRejournal < time.Minute {
```
darioush, line 350: could we keep this as a var and modify it in tests?

aaronbuchwald: I think it's cleaner to avoid modifying it specifically for tests and don't think that this condition is really necessary. Generally I'm not a big fan of sanitizing configs and replacing the values rather than returning an error for an invalid config.

core/blockchain_test.go
aaronbuchwald: Good idea, looking into this now. Seems very useful for tests.

aaronbuchwald: Put up a change, this library seems a little bit flaky, but will see if this pattern works. Please take a look and see if you have any cleaner ideas of how to use it correctly.

darioush (approved):
If the test becomes flaky we can either switch to waiting for 1 event and shutting down the chain or explicitly adding a wait callback / channel.

aaronbuchwald:
> If the test becomes flaky we can either switch to waiting for 1 event and shutting down the chain or explicitly adding a wait callback / channel.

When waiting for only one event, this seems to fail. Unclear why, the fsnotify library gives me some strange results, but happy to merge this suggested fix if this test ends up being flaky at some point: https://github.com/ava-labs/subnet-evm/pull/322.

PR #334, opened 2022-11-02
Added compatibility tracking

dasconnor:
Added a `compatibility.json` similar to the one being added in avalanchego. Will allow external apps to determine which version of avalanchego to use without needing to parse the readme. Also added a test to make sure the json parses.

This file will need to be updated with each new subnet-evm release.


utils/compatibility_test.go
```diff
@@ -0,0 +1,26 @@
...
+	expectedRPCVersion := 17
+
+	compat, err := os.ReadFile("../compatibility.json")
+	assert.NoError(t, err)
+
+	var parsedCompat rpcChainCompatibility
+	err = json.Unmarshal(compat, &parsedCompat)
+	assert.NoError(t, err)
+	assert.Equal(t, expectedRPCVersion, parsedCompat.RPCChainVMProtocolVersion[subevmVersion])
+}
```
aaronbuchwald, line 26: can we move this test to `plugin/evm/version_test.go` and use the actual version and actual imported `rpcchainvm.ProtocolVersion` from AvalancheGo, so that if we mess this up for the latest version this. test will fail?

utils/compatibility_test.go
```diff
@@ -0,0 +1,26 @@
...
+
+type rpcChainCompatibility struct {
+	RPCChainVMProtocolVersion map[string]int `json:"rpcChainVMProtocolVersion"`
+}
+
+func TestCompatibility(t *testing.T) {
+	subevmVersion := "v0.4.0"
+	expectedRPCVersion := 17
+
+	compat, err := os.ReadFile("../compatibility.json")
```
aaronbuchwald, line 19: Can we use JSON embed here? For example: https://github.com/ava-labs/coreth/blob/master/plugin/evm/ext_data_hashes.go#L11-L13

utils/compatibility_test.go
dasconnor: Added a test, for it to work tho you will need to remember to update the version in `plugins/evm/version.go`.

utils/compatibility_test.go
dasconnor: `go:embed` only works with files in a subdirectory of the test. I don't think we want to bury this json file in subdirectories, so I'm not going to make this change.

plugin/evm/version_test.go
```diff
@@ -0,0 +1,40 @@
...
+func TestCompatibilityCurrentVersion(t *testing.T) {
+	compat, err := os.ReadFile(compatibilityFile)
+	assert.NoError(t, err)
+
+	var parsedCompat rpcChainCompatibility
+	err = json.Unmarshal(compat, &parsedCompat)
+	assert.NoError(t, err)
+
+	_, valueInJSON := parsedCompat.RPCChainVMProtocolVersion[Version]
+	assert.True(t, valueInJSON)
```
aaronbuchwald, line 39: We should assert that the actual value matches the rpcchainvm protocol version that we have imported from avalanchego as well.

aaronbuchwald (approved):
LGTM - thanks for putting this up 🙌

utils/compatibility_test.go
aaronbuchwald: That's the goal 😉

cmd/simulator/go.mod
```diff
@@ -15,7 +15,7 @@ replace github.com/ava-labs/subnet-evm => ../..
 
 require (
 	github.com/VictoriaMetrics/fastcache v1.10.0 // indirect
-	github.com/ava-labs/avalanchego v1.9.1 // indirect
+	github.com/ava-labs/avalanchego v1.9.2-rc.7 // indirect
```
patrick-ogrady, line 18: remove rc

patrick-ogrady (approved):

Issue #160, opened 2022-07-22
Remove WAGMI Airdop JSON


aaronbuchwald:
Remove the JSON file that was used as the airdrop for the WAGMI genesis and switch to allowing users to specify a path to an airdrop genesis that WAGMI can use to maintain backwards comaptibility.

PR #338, opened 2022-11-02
Make airdrop file configurable

darioush:
closes: https://github.com/ava-labs/subnet-evm/issues/160


aaronbuchwald (approved):

core/genesis.go
```diff
@@ -80,6 +73,7 @@ type Genesis struct {
 	Alloc         GenesisAlloc        `json:"alloc"      gencodec:"required"`
 	AirdropHash   common.Hash         `json:"airdropHash"`
 	AirdropAmount *big.Int            `json:"airdropAmount"`
+	AirdropData   []byte              `json:"-"` // provided in a separate file, not serialized in this struct.
```
ceyonur, line 76: does that mean we need to restart our WAGMI nodes with the new configuration? Can we also remove the json file in this PR?

ceyonur: I think we need to provide a configuration for wagmi nodes in any way (even if we keep it in the subnet-evm repo). So it's probably best to move them to a correct place (wagmi folder under `public-chain-assets`).

aaronbuchwald (approved):

PR #349, opened 2022-11-07
Start v0.4.4 release cycle

aaronbuchwald:
This PR starts the v0.4.4 release cycle.

### DIFF
--- a/compatibility.json
+++ b/compatibility.json
@@ -1,5 +1,6 @@
 {
   "rpcChainVMProtocolVersion": {
+    "v0.4.4": 19,
     "v0.4.3": 19,
     "v0.4.2": 18,
     "v0.4.1": 18,
--- a/plugin/evm/version.go
+++ b/plugin/evm/version.go
@@ -11,7 +11,7 @@ var (
 	// GitCommit is set by the build script
 	GitCommit string
 	// Version is the version of Subnet EVM
-	Version string = "v0.4.3"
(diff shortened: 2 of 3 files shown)


compatibility.json
```diff
@@ -1,5 +1,6 @@
 {
   "rpcChainVMProtocolVersion": {
+    "v0.4.4": 19,
```
ceyonur, line 3: should we add this to README as well? if we feel like this is a duplicate with README I think we can move readme's compatibility chart here and add avago versions here

darioush (approved):

compatibility.json
aaronbuchwald: Hmm I'm not sure we should add this to the README until it's released since we may change it before actually doing the release. Maybe we shouldn't have it here either until the release, but that would change our ability to update `plugin/evm/version.go` at the start of the release cycle as opposed to the end. I thought about removing it from the README entirely, but think that it is more user friendly to keep the documentation there in the normal format rather than refer people to a JSON file.

cc @dasconnor does including an unreleased version in this JSON file break anything for the CLI? The intention would be to merge this to the master branch at the start of the release cycle and if a PR needs to update the avalanchego dependency and the RPCChainVMProtocol verison that same PR would need to update this JSON file, but it may have unreleased and unfinalized information for the "latest" until then.

ceyonur (approved):

Issue #199, opened 2022-08-16
Add log print-out for network upgrade


yulin-dong:
During the WAGMI network upgrade, it was noticed that there was no log print-out on the timestamp of network upgrade. It would be helpful to print a message right after the timestamp to indicate an upgrade just happened and other related info.

aaronbuchwald:
We should emit this log during Accept by using the changes from this PR: https://github.com/ava-labs/coreth-internal/pull/993 so that we can check what precompiles are activated when the block is accepted and emit a log if any new ones are present.

PR #352, opened 2022-11-07
add prints to activating precompiles

ceyonur:
adds logs to activating/deactivating upgrades

### DIFF
--- a/params/precompile_config.go
+++ b/params/precompile_config.go
@@ -4,7 +4,6 @@
 package params
 
 import (
-	"encoding/json"
 	"fmt"
 	"math/big"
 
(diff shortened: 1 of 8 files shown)


params/precompile_config.go
```diff
@@ -320,13 +342,18 @@ func (c *ChainConfig) CheckConfigurePrecompiles(parentTimestamp *big.Int, blockC
...
 				// Calling Finalise here effectively commits Suicide call and wipes the contract state.
 				// This enables re-configuration of the same contract state in the same block.
 				// Without an immediate Finalise call after the Suicide, a reconfigured precompiled state can be wiped out
 				// since Suicide will be committed after the reconfiguration.
 				statedb.Finalise(true)
 			} else {
+				// TODO: very janky solution, instead implement stringer in the interface
+				// currently skipping this so we don't break current precompiles
+				configJSON, _ := json.Marshal(config)
+				log.Info("Activating new precompile", "name", key, "config", string(configJSON))
```
aaronbuchwald, line 356: I think it would be cleaner to simply add a log within the `Configure` function of each of the precompiles.

ceyonur: I'd like to keep them in a more common place, also we cannot log from `Configure` when they're disabled. Instead we should implement `fmt.Stringer`.

ceyonur: I implemented the stringer interface and simply returned json representation of configs

darioush (approved):

params/precompile_config.go
aaronbuchwald: Ah good point, you're 100% right

aaronbuchwald (approved):

aaronbuchwald:
Resolves https://github.com/ava-labs/subnet-evm/issues/199

PR #353, opened 2022-11-07
Fix e2e tests for github workflow

anusha-ctrl:
# Fix E2E tests for GitHub workflow
## Explanation of Bugs
Only one major bug was found after running through some crazy rabbit holes. The fix was more straightforward than expected. 

### What Was Happening
The run script would start up the Avalanche-Network-Runner server and then start up a subnet using the genesis file from the run script. The genesis file from the run script had none of the precompiles enabled. This caused immediate failures on the e2e tests. Enabling those precompiles caused tests to pass but incorrect behavior that needed extra unnecessary fixes (those are removed now). 

Starting up the network in the run script would override the individual subnets that each of the `suites.go` precompile tests had. This led to those individual subnets never actually starting up because the the network was already bootstrapped.

### The Fix
We should not start up the network in the run script. 

1. This means we should not run `go run scripts/parser/main.go` as it starts up the network. Deleted the file as well, as it is NOT necessary for solidity tests or simulator tests.  Instead, both simulator and solidity tests use `e2e_test.go`  and `e2e_test.go` now starts up a network (or not) depending on the `SKIP_NETWORK_RUNNER_START` flag. See fix number 2. 

2.  Another important fix is that if we have the `SKIP_NETWORK_RUNNER_START` set to true, we should truly avoid starting up the network AND then checking up on network health, VM readiness, saving cluster info etc. This is the line that fixes the `BEFORE_SUITE`.

 ``` go
if skipNetworkRunnerStart {
	return
}
```

3. Set Avalanchego path. 

``` go
// Set AVALANCHEGO_PATH for the solidity suite
os.Setenv("AVALANCHEGO_PATH", avalanchegoExecPath)
```

This was never set so the network had issues starting up for the solidity tests. 

``` go 
// startSubnet starts a test network and launches a subnetEVM instance with the genesis file at [genesisPath]
func startSubnet(genesisPath string) error {
	fmt.Println("AVALANCHEGO_PATH:", os.Getenv("AVALANCHEGO_PATH"))
	_, err := runner.StartNetwork("0.0.0.0:12342", os.Getenv("AVALANCHEGO_PATH"), evm.ID, vmName, genesisPath, utils.GetPluginDir())
	return utils.UpdateHardhatConfig()
}


```

When we don't start up the network in the run script, each of the individual tests in `suites.go` start up a unique subnet with their separate genesis files that have the appropriate precompile enabled and conduct the tests


contract-examples/test/ExampleTxAllowList.ts
```diff
@@ -137,13 +138,15 @@ describe("ExampleTxAllowList", function () {
...
-  })
+  // FIX: Delete this test as we assign deploy permissions in ContractDeployAllowList and not TxAllowList
+  // Check line 499-500 in core/vm/evm.go
+  // it("should let allowed address deploy", async function () {    
+  //   const Token: ContractFactory = await ethers.getContractFactory("ERC20NativeMinter", { signer: allowed })
+  //   let token: Contract
+  //   token = await Token.deploy(11111)
+  //   await token.deployed()
+  //   expect(token.address).not.null
+  // })
```
anusha-ctrl, line 149: @reviewers please confirm that I can delete this test

contract-examples/hardhat.config.ts
```diff
@@ -52,7 +52,9 @@ export default {
         "0xcdbfd34f687ced8c6968854f8a99ae47712c4f4183b78dcc4a903d1bfe8cbf60",
         "0x86f78c5416151fe3546dece84fda4b4b1e36089f2dbc48496faf3a950f16157c",
         "0x750839e9dbbd2a0910efe40f50b2f3b2f2f59f5580bb4b83bd8c1201cf9a010a"
-      ]
+      ],
+      gas: "auto", 
+      gas_price: "auto",
```
ceyonur, line 57: I believe defaults are auto. do we need to explicitly provide these?

scripts/run.sh
```diff
@@ -146,7 +146,28 @@ echo "creating genesis"
...
+    },
+    "contractNativeMinterConfig": {
+      "blockTimestamp": 0,
+      "adminAddresses": ["0x8db97C7cEcE249c2b98bDC0226Cc4C2A57BF52FC"]
+    },
+    "feeManagerConfig": {
+      "blockTimestamp": 0,
+      "adminAddresses": [
+        "0x8db97C7cEcE249c2b98bDC0226Cc4C2A57BF52FC"
+      ]
```
ceyonur, line 168: why do we activate these here?

contract-examples/test/ERC20NativeMinter.ts
```diff
@@ -83,6 +84,13 @@ describe("ERC20NativeMinter", function () {
     await mintTx.wait()
     contractRole = await minterList.readAllowList(contract.address);
     expect(contractRole).to.be.equal(ROLES.MINTER)
+
+    // Enable minter address on TxAllowList to allow deposits
+    const allowList = await ethers.getContractAt("IAllowList", TX_ALLOW_LIST_ADDRESS, owner)
+    let mintEnableTx = await allowList.setEnabled(minter.address);
+    await mintEnableTx.wait()
+    contractRole = await allowList.readAllowList(minter.address);
+    expect(contractRole).to.be.equal(ROLES.MINTER)
```
ceyonur, line 93: these tests should not contain any other precompiles. Tests are not in perfect shape currently (i.e they're not stateless) but at least we should keep them separate from other precompiles.

tests/e2e/runner/runner.go
```diff
@@ -57,21 +62,24 @@ func startRunner(grpcEp string, execPath string, vmName string, genesisPath stri
...
-		return "", "", 0, err
-	}
+	// cli, err := client.New(client.Config{
+	// 	LogLevel:    "info",
+	// 	Endpoint:    grpcEp,
+	// 	DialTimeout: 10 * time.Second,
+	// })
+	// if err != nil {
+	// 	return "", "", 0, err
+	// }
```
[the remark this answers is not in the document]
anusha-ctrl: Hey was still making fixes, so I kept that commented in case. Will do a big clean up. Thanks!

scripts/run.sh
anusha-ctrl: Removed. will provide a comprehensive write up of the bugs I discovered. thanks!

tests/e2e/runner/runner.go
```diff
@@ -32,18 +34,21 @@ func (ci clusterInfo) Save(p string) error {
...
-	if err != nil {
-		return err
-	}
+	return err
+}
 
+func startRunner(grpcEp string, execPath string, vmName string, genesisPath string, pluginDir string) error {
 	utils.Outf("{{green}}tests/e2e/runner sending 'start' with binary path:{{/}} %q\n", execPath)
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
+	fmt.Println("DO YOU EVEN REACH HERE 1")
```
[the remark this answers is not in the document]
anusha-ctrl: Was still debugging stuff thanks!

tests/e2e/runner/runner.go
```diff
@@ -57,21 +62,24 @@ func startRunner(grpcEp string, execPath string, vmName string, genesisPath stri
 	)
 	cancel()
 	if err != nil {
+		fmt.Println("DO YOU EVEN REACH HERE 1.5")
+		fmt.Println(err)
```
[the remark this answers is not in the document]
anusha-ctrl: removed thanks :)

tests/e2e/runner/runner.go
```diff
@@ -57,21 +62,24 @@ func startRunner(grpcEp string, execPath string, vmName string, genesisPath stri
...
 		return err
 	}
+	fmt.Println("DO YOU EVEN REACH HERE 2")
```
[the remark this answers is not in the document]
anusha-ctrl: removed thanks :)

tests/e2e/runner/runner.go
```diff
@@ -187,19 +198,19 @@ func StartNetwork(grpcEp string, execPath string, vmId ids.ID, vmName string, ge
...
-	}
+func StopNetwork() error {
+	// cli, err := client.New(client.Config{
+	// 	LogLevel:    "info",
+	// 	Endpoint:    grpcEp,
+	// 	DialTimeout: 10 * time.Second,
+	// })
+	// if err != nil {
+	// 	return err
+	// }
```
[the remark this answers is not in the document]
anusha-ctrl: removed thanks :)

contract-examples/hardhat.config.ts
anusha-ctrl: Ah yes no need to explicitly provide these. Thanks!

tests/e2e/runner/runner.go
```diff
@@ -32,18 +34,21 @@ func (ci clusterInfo) Save(p string) error {
 	return os.WriteFile(p, ob, fsModeWrite)
 }
 
-func startRunner(grpcEp string, execPath string, vmName string, genesisPath string, pluginDir string) error {
-	cli, err := client.New(client.Config{
+func InitializeRunner(grpcEp string) error {
```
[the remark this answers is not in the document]
anusha-ctrl: Reverted to original

scripts/run.sh
```diff
@@ -280,7 +254,7 @@ run_simulator() {
   --priority-fee=1
 }
 
-if [[ ${SKIP_NETWORK_RUNNER_START} == true ]]; then
+if [[ ${ENABLE_SOLIDITY_TESTS} != true ]]; then
```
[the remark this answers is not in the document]
anusha-ctrl: Agree, maybe we can do this in another pr? Just trying to get CI working again :)

scripts/run.sh
```diff
@@ -166,53 +166,6 @@ echo "creating genesis"
 }
 EOF
 
-# If you'd like to try the airdrop feature, use the commented genesis
```
aaronbuchwald, line 169: 🔥

scripts/run.sh
```diff
@@ -56,7 +56,7 @@ if [[ ${ENABLE_SOLIDITY_TESTS} == true ]]; then
 fi
 
 echo "Running with:"
-echo AVALANCE_VERSION: ${VERSION}
+echo AVALANCHE_VERSION: ${VERSION}
```
aaronbuchwald, line 59: nice catch

contract-examples/test/ExampleTxAllowList.ts
```diff
@@ -138,6 +137,14 @@ describe("ExampleTxAllowList", function () {
...
   })
 
+  it("should let allowed address deploy", async function () {
+    const Token: ContractFactory = await ethers.getContractFactory("ERC20NativeMinter", { signer: allowed })
+    let token: Contract
+    token = await Token.deploy(11111)
+    await token.deployed()
+    expect(token.address).not.null
+  })
+
```
ceyonur, line 147: I think this should not test this

tests/e2e/e2e_test.go
```diff
@@ -142,39 +142,40 @@ var _ = ginkgo.BeforeSuite(func() {
...
-							Genesis: utils.GetVmGenesisPath(),
-						},
-					},
-				))
-			cancel()
-			gomega.Expect(err).Should(gomega.BeNil())
-			utils.Outf("{{green}}successfully started:{{/}} %+v\n", resp.ClusterInfo.NodeNames)
-		}
-	})
+	os.Setenv("AVALANCHEGO_PATH", avalanchegoExecPath)
```
aaronbuchwald, line 145: why is this necessary? Could we add a comment?

tests/e2e/e2e_test.go
```diff
@@ -142,39 +142,40 @@ var _ = ginkgo.BeforeSuite(func() {
...
+
+	if skipNetworkRunnerStart {
+		utils.Outf("{{green}}skipped 'start'{{/}}\n")
+		return
+	}
```
aaronbuchwald, line 150: can we separate out the `BeforeSuite` function into smaller functions to make it easier to follow?

Prior to this, it was starting the network runner and then also spinning up a subnet, but since each test should be responsible for launching its own subnet, we should separate these into smaller functions.

Additionally, previously this would check the health of the network runner server itself, so I think if `skipNetworkRunnerStart` is true, we should still check to confirm that the network runner server is actually health and accessible since we assume that if we're not starting it, then it's because it should already be running in the background if I understand the flow correctly.

cc @gyuho to confirm

scripts/run.sh
```diff
@@ -280,7 +233,9 @@ run_simulator() {
   --priority-fee=1
 }
 
-if [[ ${SKIP_NETWORK_RUNNER_START} == true ]]; then
+# We should only run scripts/parser/main.go if SKIP_NETWORK_RUNNER_START is set to 
+# false as this script starts up the network. 
+if [[ ${SKIP_NETWORK_RUNNER_START} != true ]]; then
```
aaronbuchwald, line 238: @gyuho could you confirm this is the right way to handle this? seems like this changes the previous behavior.

tests/e2e/e2e_test.go
gyuho: Yes, we can separate this into smaller ones for better readability. The global variables are set before `BeforeSuite` and those variables can be used as parameters to the smaller functions.

contract-examples/test/ExampleTxAllowList.ts
anusha-ctrl: tests are now untouched thanks

tests/e2e/e2e_test.go
anusha-ctrl: Done. We need to set this for the solidity suites.go, as we grab the Avalanchego path there. This was never set anywhere and causes the network to not start up properly for `startSubnet`.

tests/e2e/e2e_test.go
anusha-ctrl: Sure

scripts/run.sh
anusha-ctrl: Removed `parser/main.go` because the e2e_test.go starts up the network (or doesn't depending on the flag) there is no need for this file anymore -> `scripts/parser/main.go` as all it did was start up the network. Now the `SKIP_NETWORK_RUNNER_START`  flag works as expected.

tests/e2e/e2e_test.go
```diff
@@ -168,11 +175,15 @@ var _ = ginkgo.BeforeSuite(func() {
...
+	if skipNetworkRunnerStart {
+		return
+	}
+
 	// TODO: network runner health should imply custom VM healthiness
 	// or provide a separate API for custom VM healthiness
 	// "start" is async, so wait some time for cluster health
 	utils.Outf("\n{{magenta}}sleeping before checking custom VM status...{{/}}\n")
-	time.Sleep(2 * time.Minute)
+	time.Sleep(10 * time.Second)
```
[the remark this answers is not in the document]
anusha-ctrl: I'll look at improving this in a later PR thanks

scripts/parser/main.go
```diff
@@ -1,117 +0,0 @@
-// (c) 2022 Ava Labs, Inc. All rights reserved.
```
aaronbuchwald, line 1: Should we remove this entire directory if this file is going to be removed?

anusha-ctrl: yes

tests/e2e/e2e_test.go
```diff
@@ -28,6 +28,10 @@ func TestE2E(t *testing.T) {
 	ginkgo.RunSpecs(t, "subnet-evm e2e test suites")
 }
 
+type networkClient struct {
+	client runner_sdk.Client
+}
```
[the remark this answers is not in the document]
anusha-ctrl: I wanted to use wrapper struct for the client and then create the helper functions around that struct.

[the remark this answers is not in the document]
aaronbuchwald: Good idea, much cleaner with this change 👍

gyuho (approved):
Lgtm. Thanks!

PR #360, opened 2022-11-17
Add counter for verified transactions

aaronbuchwald:
## Why this should be merged

This PR adds a metric for the number of verified transactions in the same style that we track processed and accepted gas.

## How this works

This PR adds an additional global metric for processed transactions next to accepted transactions and increments it when we finish inserting a block.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -85,7 +85,8 @@ var (
 	acceptedBlockGasUsedCounter  = metrics.NewRegisteredCounter("chain/block/gas/used/accepted", nil)
 	badBlockCounter              = metrics.NewRegisteredCounter("chain/block/bad/count", nil)
 
-	acceptedTxsCounter = metrics.NewRegisteredCounter("chain/txs/accepted/count", nil)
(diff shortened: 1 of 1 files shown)


core/blockchain.go
```diff
@@ -85,7 +85,8 @@ var (
 	acceptedBlockGasUsedCounter  = metrics.NewRegisteredCounter("chain/block/gas/used/accepted", nil)
 	badBlockCounter              = metrics.NewRegisteredCounter("chain/block/bad/count", nil)
 
-	acceptedTxsCounter = metrics.NewRegisteredCounter("chain/txs/accepted/count", nil)
+	acceptedTxsCounter  = metrics.NewRegisteredCounter("chain/txs/accepted", nil)
+	processedTxsCounter = metrics.NewRegisteredCounter("chain/txs/processed", nil)
```
ceyonur, line 89: nit: we used `inserts` for block should we use `insert` here as well?

aaronbuchwald: Hmm I followed the same way that we did this for the gas consumption here. Could go either way, but I'd prefer to leave this as is unless you feel strongly

darioush (approved):

PR #373, opened 2022-11-28
Pack topics

aaronbuchwald:
## Why this should be merged

This PR adds `PackTopics` to the abi package to support packing events. This is intended to be made available for precompiles to pack events, so that they can create logs.

## How this works

This works by using the existing code for packing topics used in the abi package to support creating filter queries and breaks out the actual creation of a topic hash into a helper function, so that we can re-purpose it to create an array of topics as would be created by an event log rather than the type `[][]common.Hash` that would be used when creating a filter query.

## How this was tested

Adds a unit test in `abi_test.go` to test the encoding of an event without indexed args, with indexed args, and of an anonymous event.

## Reference for Event Encoding

https://docs.soliditylang.org/en/v0.8.17/abi-spec.html#indexed-event-encoding

### DIFF
--- a/accounts/abi/abi.go
+++ b/accounts/abi/abi.go
@@ -96,26 +96,47 @@ func (abi ABI) Pack(name string, args ...interface{}) ([]byte, error) {
 // hashes derived from indexed arguments and the packed data of non-indexed args according to
 // the event ABI specification.
 // https://docs.soliditylang.org/en/v0.8.17/abi-spec.html#indexed-event-encoding.
-func (abi ABI) PackEvent(name string, indexedArgs []interface{}, nonIndexedArgs []interface{}) ([]common.Hash, []byte, error) {
+// Note: PackEvent does not support array(both fixed size and dynamic-size) and struct types.
+func (abi ABI) PackEvent(name string, args []interface{}) ([]common.Hash, []byte, error) {
 	event, exist := abi.Events[name]
 	if !exist {
 		return nil, nil, fmt.Errorf("event '%s' not found", name)
 	}
-	arguments, err := event.Inputs.NonIndexed().Pack(nonIndexedArgs...)
+	if len(args) != len(event.Inputs) {
+		return nil, nil, fmt.Errorf("event '%s' unexpected number of inputs %d", name, len(args))
+	}
+
+	var (
+		nonIndexedInputs = make([]interface{}, 0)
+		indexedInputs    = make([]interface{}, 0)
+		nonIndexedArgs   Arguments
+		indexedArgs      Arguments
+	)
+
+	for i, arg := range event.Inputs {
+		if arg.Indexed {
+			indexedArgs = append(indexedArgs, arg)
+			indexedInputs = append(indexedInputs, args[i])
+		} else {
+			nonIndexedArgs = append(nonIndexedArgs, arg)
+			nonIndexedInputs = append(nonIndexedInputs, args[i])
+		}
+	}
+
+	packedArguments, err := nonIndexedArgs.Pack(nonIndexedInputs...)
 	if err != nil {
 		return nil, nil, err
 	}
-
 	topics := make([]common.Hash, 0, len(indexedArgs)+1)
 	if !event.Anonymous {
 		topics = append(topics, event.ID)
 	}
-	indexedTopics, err := PackTopics(indexedArgs)
+	indexedTopics, err := PackTopics(indexedInputs)
 	if err != nil {
 		return nil, nil, err
 	}
 
-	return append(topics, indexedTopics...), arguments, nil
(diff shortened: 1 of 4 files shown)


accounts/abi/topics.go
```diff
@@ -37,71 +37,99 @@ import (
...
+			return common.Hash{}, fmt.Errorf("unsupported indexed type: %T", rule)
+		}
+	}
+	return topic, nil
+}
+
+// PackTopics attempts to pack the array of filters into an array of corresponding topics
+// according to the Solidity documentation.
+// Note: PackTopics does not support array(both fixed size and dynamic-size) and struct types.
+func PackTopics(filter []interface{}) ([]common.Hash, error) {
```
aaronbuchwald, line 110: We should move this into the ABI package and have a single function to pack both the topics and event input data.

accounts/abi/topics.go
```diff
@@ -37,71 +37,99 @@ import (
...
-				case val.Kind() == reflect.Array && reflect.TypeOf(rule).Elem().Kind() == reflect.Uint8:
-					reflect.Copy(reflect.ValueOf(topic[:val.Len()]), val)
-				default:
-					return nil, fmt.Errorf("unsupported indexed type: %T", rule)
-				}
+			topic, err := packTopic(rule)
+			if err != nil {
+				return nil, err
 			}
+
```
darioush, line 132: nit: could we remove the wspace to minimize the diff?

accounts/abi/topics.go
```diff
@@ -37,71 +37,99 @@ import (
...
+		case val.Kind() == reflect.Array && reflect.TypeOf(rule).Elem().Kind() == reflect.Uint8:
+			reflect.Copy(reflect.ValueOf(topic[:val.Len()]), val)
+		default:
+			return common.Hash{}, fmt.Errorf("unsupported indexed type: %T", rule)
+		}
+	}
+	return topic, nil
+}
+
+// PackTopics attempts to pack the array of filters into an array of corresponding topics
```
darioush, line 107:
```suggestion
// PackTopics packs the array of filters into an array of corresponding topics
```

accounts/abi/topics.go
```diff
@@ -37,71 +37,99 @@ import (
...
+// according to the Solidity documentation.
+// Note: PackTopics does not support array(both fixed size and dynamic-size) and struct types.
```
darioush, line 109:
```suggestion
// Note: PackTopics does not support array (fixed or dynamic-size) or struct types.
```

accounts/abi/topics.go
darioush: not sure I follow this comment.

aaronbuchwald: Updated the PR with what I was thinking.

Logs include both topics `[]common.Hash` and event data, which is more similar to the way function arguments are encoded and `PackTopics` only converts arguments to the corresponding topic hash based off of this.

Now there is `PackEvent` which will take in all of the arguments both indexed and non-indexed and return both the resulting topics and event data.

accounts/abi/abi.go
```diff
@@ -91,6 +91,54 @@ func (abi ABI) Pack(name string, args ...interface{}) ([]byte, error) {
 	return append(method.ID, arguments...), nil
 }
 
+// PackEvent packs the given event name to conform the ABI.
```
darioush, line 94:
```suggestion
// PackEvent packs the given event name and arguments to conform the ABI.
```

accounts/abi/abi_test.go
```diff
@@ -39,6 +39,7 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/math"
 	"github.com/ethereum/go-ethereum/crypto"
+	"gotest.tools/assert"
```
darioush, line 42: Usually we import stretchr

accounts/abi/topics.go
```diff
@@ -37,70 +37,97 @@ import (
 	"github.com/ethereum/go-ethereum/crypto"
 )
 
+// packTopic attempts to pack rule into the corresponding hash value for a log's topic
```
darioush, line 40:
```suggestion
// packTopic packs rule into the corresponding hash value for a log's topic
```

darioush (approved):
LGTM after the test import is fixed, other comments optional

accounts/abi/abi_test.go
aaronbuchwald: Nice catch

aaronbuchwald:
Added one more simplifying change to this PR to have only one argument array passed in to `PackEvent`, so I re-requested your guys' reviews @ceyonur @darioush

darioush (approved):

PR #374, opened 2022-11-28
nit: remove gotest.tools dependency

darioush:
## Why this should be merged
Generally we import the assert functionality of the stretchr package, keeps the code more consistent.

## How this was tested
CI tests


aaronbuchwald (approved):
LGTM - nice catch

PR #375, opened 2022-11-28
Update AddLog function signature to support usage from precompiles

aaronbuchwald:
## Why this should be merged

This PR updates the function signature of statedb `AddLog(...)` function to:
```
```

Previously, a fully formed log was passed directly in with only these params populated. This PR is intended to avoid needing to import `core/types/` package into the `precompile/` package which currently creates a circular import.

## How this was tested

This is intended as a no-op change, so no unit tests were added.

### DIFF
--- a/core/vm/instructions.go
+++ b/core/vm/instructions.go
@@ -29,7 +29,6 @@ package vm
 import (
 	"sync/atomic"
 
-	"github.com/ava-labs/subnet-evm/core/types"
 	"github.com/ava-labs/subnet-evm/params"
 	"github.com/ava-labs/subnet-evm/vmerrs"
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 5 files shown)


core/vm/instructions.go
```diff
@@ -859,14 +858,12 @@ func makeLog(size int) executionFunc {
 		}
 
 		d := scope.Memory.GetCopy(int64(mStart.Uint64()), int64(mSize.Uint64()))
-		interpreter.evm.StateDB.AddLog(&types.Log{
-			Address: scope.Contract.Address(),
-			Topics:  topics,
-			Data:    d,
-			// This is a non-consensus field, but assigned here because
-			// core/state doesn't know the current block number.
```
darioush, line 867: can we preserve this information by adding a comment for `func (s *StateDB) AddLog...`?

darioush (approved):

anusha-ctrl (approved):

core/vm/instructions.go
aaronbuchwald: Yup, good idea

Issue #280, opened 2022-09-29
Legacy pre-eip-155 transactions not supported


Uponn:
**Describe the bug**
Hey, I'm following [this ](https://docs.avax.network/subnets/deploy-a-gnosis-safe-on-your-evm#custom-network-workflow) guide from the site for deploying a Gnosis Safe on my subnet. I'm getting the following error:
```
C:\Users\Steliyan\Desktop\safe-contracts>yarn hardhat deploy --network subnet           
yarn run v1.22.18
$ C:\Users\Steliyan\Desktop\safe-contracts\node_modules\.bin\hardhat deploy --network subnet
Nothing to compile
sending eth to create2 contract deployer address (0x3fab184622dc19b6109349b94811493bf2a45362) (tx: 0x571c252c750e3cf8c86138035b7c8007fd7dde7c928178fb3524740338a7678d)...
deploying create2 deployer contract (at 0x4e59b44847b379578588920ca78fbf26c0b4956c) using deterministic deployment (https://github.com/Arachnid/deterministic-deployment-proxy)An unexpected error occurred:

Error: ERROR processing C:\Users\Steliyan\Desktop\safe-contracts\src\deploy\deploy_accessors.ts:
Error: legacy pre-eip-155 transactions not supported (error={"name":"ProviderError","code":-32000,"_isProviderError":true}, method="sendTransaction", transaction="0xf8a58085174876e800830186a08080b853604580600e600039806000f350fe7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe03601600081602082378035828234f58015156039578182fd5b8082525050506014600cf31ba02222222222222222222222222222222222222222222222222222222222222222a02222222222222222222222222222222222222222222222222222222222222222", code=UNSUPPORTED_OPERATION, version=providers/5.4.5)
    at Logger.makeError (C:\Users\Steliyan\Desktop\safe-contracts\node_modules\hardhat-deploy\node_modules\@ethersproject\logger\src.ts\index.ts:225:28)
    at Logger.throwError (C:\Users\Steliyan\Desktop\safe-contracts\node_modules\hardhat-deploy\node_modules\@ethersproject\logger\src.ts\index.ts:237:20)
    at checkError (C:\Users\Steliyan\Desktop\safe-contracts\node_modules\hardhat-deploy\node_modules\@ethersproject\providers\src.ts\json-rpc-provider.ts:74:16)
[...]

PR #377, opened 2022-11-29
Migrate replay protection bypass for Nick's Method Signature txs

aaronbuchwald:
## Why this should be merged

This PR migrates the change from coreth to allow transactions intended to not have replay protection to be issued. This includes transactions with equivalent R and S values (these should not be equal unless intentionally set to be equal as per Nick's Method for generating fake signatures that do not correspond to an actual private key) and special case transaction hashes from known transactions meant to be deployed on multiple chains ie. EIP-1820 transaction.

## How this was tested

This was tested by adding a unit test with special case transactions signed using Nick's Method.


darioush (approved):

aaronbuchwald:
Resolves https://github.com/ava-labs/subnet-evm/issues/280

PR #379, opened 2022-11-29
Plumbing snow.Context and evm Call to precompile

minghinmatthewlam:
This PR is for plumbing the `snow.Context` and `evm.Call()` to the precompiles through `PrecompileAccessibleState`.


params/config.go
```diff
@@ -88,8 +89,47 @@ var (
...
+		EIP158Block:         big.NewInt(0),
+		ByzantiumBlock:      big.NewInt(0),
+		ConstantinopleBlock: big.NewInt(0),
+		PetersburgBlock:     big.NewInt(0),
+		IstanbulBlock:       big.NewInt(0),
+		MuirGlacierBlock:    big.NewInt(0),
+		NetworkUpgrades:     NetworkUpgrades{big.NewInt(0)},
+		PrecompileUpgrade:   PrecompileUpgrade{},
+		UpgradeConfig:       UpgradeConfig{},
+		SnowCtx:             snow.DefaultContextTest(),
```
aaronbuchwald, line 110: Instead of adding a new type to the chain config here, we should add `SnowCtx` to the AvalancheContext type.

eth/backend.go
```diff
@@ -115,6 +117,7 @@ func roundUpCacheSize(input int, allocSize int) int {
 // New creates a new Ethereum object (including the
 // initialisation of the common Ethereum object)
 func New(
+	ctx *snow.Context,
```
aaronbuchwald, line 120: nit: remove in favor of setting this via the chain config's AvalancheContext

core/stateful_precompile_test.go
```diff
@@ -56,6 +57,12 @@ func (m *mockAccessibleState) GetStateDB() precompile.StateDB { return m.state }
 
 func (m *mockAccessibleState) GetBlockContext() precompile.BlockContext { return m.blockContext }
 
+func (m *mockAccessibleState) GetSnowContext() *snow.Context { return snow.DefaultContextTest() }
+
```
aaronbuchwald, line 61: can we add ctx as a field to `mockAccessibleState` and return that here?

params/config.go
```diff
@@ -122,6 +162,8 @@ type ChainConfig struct {
 	NetworkUpgrades              // Config for timestamps that enable avalanche network upgrades
 	PrecompileUpgrade            // Config for enabling precompiles from genesis
 	UpgradeConfig     `json:"-"` // Config specified in upgradeBytes (avalanche network upgrades or enable/disabling precompiles). Skip encoding/decoding directly into ChainConfig.
+
+	SnowCtx *snow.Context
```
aaronbuchwald, line 166: Can we move the new `SnowCtx` into `AvalancheContext`?

eth/backend.go
aaronbuchwald, line 120: can we remove this parameter in favor of setting the `snow.Context` on the AvalancheContext prior to initializing the ethereum backend?

precompile/contract.go
```diff
@@ -15,12 +17,19 @@ const (
 	selectorLen = 4
 )
 
+// ContractRef is a reference to the contract's backing object
+type ContractRef interface {
+	Address() common.Address
+}
```
aaronbuchwald, line 23: Can we move this back into its original location and add an extra `CallFromPrecompile` function which will take in a normal `common.Address` type instead of `ContractRef` to reduce the amount of EVM code that needs to be changed?

PR #381, opened 2022-11-29
Plumbing snow.Context for precompiles

minghinmatthewlam:
This PR is for plumbing the snow.Context to the precompiles through PrecompileAccessibleState.


eth/backend.go
```diff
@@ -142,6 +142,7 @@ func New(
 	if genesisErr != nil {
 		return nil, genesisErr
 	}
+
```
aaronbuchwald, line 145: Can we remove this to reduce the diff?

params/config.go
```diff
@@ -88,8 +89,45 @@ var (
 		},
 	}
 
-	TestChainConfig        = &ChainConfig{AvalancheContext{common.Hash{1}}, big.NewInt(1), DefaultFeeConfig, false, big.NewInt(0), big.NewInt(0), common.Hash{}, big.NewInt(0), big.NewInt(0), big.NewInt(0), big.NewInt(0), big.NewInt(0), big.NewInt(0), big.NewInt(0), NetworkUpgrades{big.NewInt(0)}, PrecompileUpgrade{}, UpgradeConfig{}}
-	TestPreSubnetEVMConfig = &ChainConfig{AvalancheContext{common.Hash{1}}, big.NewInt(1), DefaultFeeConfig, false, big.NewInt(0), big.NewInt(0), common.Hash{}, big.NewInt(0), big.NewInt(0), big.NewInt(0), big.NewInt(0), big.NewInt(0), big.NewInt(0), big.NewInt(0), NetworkUpgrades{}, PrecompileUpgrade{}, UpgradeConfig{}}
+	TestChainConfig = &ChainConfig{
+		AvalancheContext:    AvalancheContext{common.Hash{1}, snow.DefaultContextTest()},
```
aaronbuchwald, line 93: Can we update the formatting for this so that both values are on two different lines? Also, now that I think about it, the first value in `AvalancheContext` is the `blockchainID` which is included in `snow.Context` so we can actually remove the `BlockchainID` parameter entirely.

params/config.go
```diff
@@ -88,8 +89,45 @@ var (
...
+		PetersburgBlock:     big.NewInt(0),
+		IstanbulBlock:       big.NewInt(0),
+		MuirGlacierBlock:    big.NewInt(0),
+		NetworkUpgrades:     NetworkUpgrades{big.NewInt(0)},
+		PrecompileUpgrade:   PrecompileUpgrade{},
+		UpgradeConfig:       UpgradeConfig{},
+	}
+
+	TestPreSubnetEVMConfig = &ChainConfig{
+		AvalancheContext:    AvalancheContext{common.Hash{1}, snow.DefaultContextTest()},
```
aaronbuchwald, line 113: Same comment on formatting here

aaronbuchwald (commented):
Few comments and then I think this is good to merge, thanks for putting this up 🙌

minghinmatthewlam:
Discussed to move the addition of `PrecompileAccessibleState.Call()` to another PR to deal with cycling imports

params/config.go
minghinmatthewlam: yup nice catch, removed `BlockchainID`

aaronbuchwald (approved):
LGTM

params/config.go
```diff
@@ -139,7 +177,7 @@ type UpgradeConfig struct {
 
 // AvalancheContext provides Avalanche specific context directly into the EVM.
 type AvalancheContext struct {
-	BlockchainID common.Hash
```
michaelkaplan13, line 142: Sorry I'm just seeing this PR now, but should we leave the `BlockchainID` in the AvalancheContext here?

I think we will need it for teleporter because messages are sent to/from specific chains, rather than from subnets as we say commonly.

minghinmatthewlam: we decided to remove `BlockchainID` here since that value is set through the same snow context's `ChainID` value, and since now we save the snow context in AvalancheContext, opted to remove the extra `BlockchainID` field.

michaelkaplan13: Ah gotcha, didn't realize it was available elsewhere already. Sounds good, thanks.

PR #382, opened 2022-11-30
Embed abis

ceyonur:
## Why this should be merged

Moves inline ABI strings in precompiles to embedded separate string files.

## How this works

Changes how we handle `abi.Bind` function and adds a hook to provide a way for expanding the function without expanding the standard inputs. Creates inputted ABI files in the same directory with output precompile file, and embeds those ABI files in precompile.  If the output precompile file is STDOUT then it will inline ABI strings into the precompile (old behaviour).

Changes precompile template to reflect those changes. 

## How this was tested

Created a RewardManager precompile with precompile gen tool and run `stateful_precompile_test.go` tests. Also confirmed the ABI is inlined when output is STDOUT.


ceyonur:
closing in favor of #383

PR #383, opened 2022-11-30
move inline string ABIs to separate files and embed them

ceyonur:
## Why this should be merged

Moves inline ABI strings in precompiles to embedded separate string files.

## How this works

Changes how we handle `abi.Bind` function and adds a hook to provide a way for expanding the function without expanding the standard inputs. Creates inputted ABI files in the same directory with output precompile file, and embeds those ABI files in precompile.  If the output precompile file is STDOUT then it will inline ABI strings into the precompile (old behaviour).

Changes precompile template to reflect those changes. 

## How this was tested

Created a RewardManager precompile with precompile gen tool and run `stateful_precompile_test.go` tests. Also confirmed the ABI is inlined when output is STDOUT.


accounts/abi/bind/bind.go
```diff
@@ -291,19 +292,14 @@ func Bind(types []string, abis []string, bytecodes []string, fsigs []map[string]
...
 
-	// Generate the contract template data according to contract type (precompile/non)
-	if isPrecompile {
-		if lang != LangGo {
-			return "", errors.New("only GoLang binding for precompiled contracts is supported yet")
-		}
-
-		if len(contracts) != 1 {
-			return "", errors.New("cannot generate more than 1 contract")
+	// Generate the contract template data according to  hook
```
darioush, line 295:
```suggestion
	// Generate the contract template data according to hook
```

precompile/reward_manager.go
```diff
@@ -26,9 +28,6 @@ const (
...
...ddress\",\"name\":\"rewardAddress\",\"type\":\"address\"}],\"stateMutability\":\"view\",\"type\":\"function\"},{\"inputs\":[],\"name\":\"disableRewards\",\"outputs\":[],\"stateMutability\":\"nonpayable\",\"type\":\"function\"},{\"inputs\":[{\"internalType\":\"address\",\"name\":\"addr\",\"type\":\"address\"}],\"name\":\"readAllowList\",\"outputs\":[{\"internalType\":\"uint256\",\"name\":\"role\",\"type\":\"uint256\"}],\"stateMutability\":\"view\",\"type\":\"function\"},{\"inputs\":[{\"internalType\":\"address\",\"name\":\"addr\",\"type\":\"address\"}],\"name\":\"setAdmin\",\"outputs\":[],\"stateMutability\":\"nonpayable\",\"type\":\"function\"},{\"inputs\":[{\"internalType\":\"address\",\"name\":\"addr\",\"type\":\"address\"}],\"name\":\"setEnabled\",\"outputs\":[],\"stateMutability\":\"nonpayable\",\"type\":\"function\"},{\"inputs\":[{\"internalType\":\"address\",\"name\":\"addr\",\"type\":\"address\"}],\"name\":\"setNone\",\"outputs\":[],\"stateMutability\":\"nonpayable\",\"type\":\"function\"},{\"inputs\":[{\"internalType\":\"address\",\"name\":\"addr\",\"type\":\"address\"}],\"name\":\"setRewardAddress\",\"outputs\":[],\"stateMutability\":\"nonpayable\",\"type\":\"function\"}]"
```
darioush, line 31: ❤️

accounts/abi/bind/bind.go
```diff
@@ -101,7 +103,11 @@ func isKeyWord(arg string) bool {
 // to be used as is in client code, but rather as an intermediate struct which
 // enforces compile time type safety and naming convention opposed to having to
 // manually maintain hard coded strings that break on runtime.
-func Bind(types []string, abis []string, bytecodes []string, fsigs []map[string]string, pkg string, lang Lang, libs map[string]string, aliases map[string]string, isPrecompile bool) (string, error) {
+func Bind(types []string, abis []string, bytecodes []string, fsigs []map[string]string, pkg string, lang Lang, libs map[string]string, aliases map[string]string) (string, error) {
+	return BindHelper(types, abis, bytecodes, fsigs, pkg, lang, libs, aliases, nil)
+}
+
+func BindHelper(types []string, abis []string, bytecodes []string, fsigs []map[string]string, pkg string, lang Lang, libs map[string]string, aliases map[string]string, bindHook BindHook) (string, error) {
```
darioush, line 110: does `BindHelper` need to be exported?

accounts/abi/bind/bind.go
```diff
@@ -291,19 +292,14 @@ func Bind(types []string, abis []string, bytecodes []string, fsigs []map[string]
...
+		var err error
+		data, templateSource, err = bindHook(lang, types, contracts, structs)
+		if err != nil {
+			return "", err
 		}
-		precompileType := types[0]
-		firstContract := contracts[precompileType]
-		data, templateSource = createPrecompileDataAndTemplate(firstContract, structs)
-	} else {
+	} else { // default to generate contract binding
```
darioush, line 302: an alternative would be to pass this code in as `bindHook` for the default (`nil`) case. however there is the downside of making it more difficult to trace upstream changes. Not suggesting we make a change, just highlighting the tradeoff

ceyonur: yeah, true. the reason I added this hook is to trace upstream changes more easily.

accounts/abi/bind/bind.go
ceyonur: nope, good catch. unexported it.

darioush (approved):

aaronbuchwald (approved):
LGTM - can we start a `precompile` branch to batch our planned precompile improvements into, so we can do one pass of updating the documentation when they are done?

PR #384, opened 2022-11-30
Add CallFromPrecompile to PrecompileAccessibleState

minghinmatthewlam:
This PR adds a method `CallFromPrecompile` to the interface PrecompileAccessibleState.
- CallFromPrecompile creates an `AccountRef`, which implements ContractRef, then invokes `Call` for the contract

### DIFF
--- a/core/vm/evm.go
+++ b/core/vm/evm.go
@@ -307,6 +307,11 @@ func (evm *EVM) Call(caller ContractRef, addr common.Address, input []byte, gas 
 	return ret, gas, err
 }
 
+// CallFromPrecompile invokes Call to execute the contract with the given input parameters.
+func (evm *EVM) CallFromPrecompile(caller common.Address, addr common.Address, input []byte, gas uint64, value *big.Int) (ret []byte, leftOverGas uint64, err error) {
+	return evm.Call(AccountRef(caller), addr, input, gas, value)
+}
+
 // CallCode executes the contract associated with the addr with the given input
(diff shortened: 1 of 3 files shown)


cam-schultz (approved):
LGTM

aaronbuchwald (approved):
LGTM

core/vm/evm.go
```diff
@@ -307,6 +307,11 @@ func (evm *EVM) Call(caller ContractRef, addr common.Address, input []byte, gas
 	return ret, gas, err
 }
 
+// CallFromPrecompile invokes Call to execute the contract with the given input parameters.
+func (evm *EVM) CallFromPrecompile(caller common.Address, addr common.Address, input []byte, gas uint64, value *big.Int) (ret []byte, leftOverGas uint64, err error) {
+	return evm.Call(AccountRef(caller), addr, input, gas, value)
```
michaelkaplan13, line 312: Should we be converting `caller` to a `ContractRef` expected by `Call`, rather than an `AccountRef`? I assume there's an implicit conversion available, but may as well use what's expected I think.

darioush (approved):

core/vm/evm.go
aaronbuchwald: `ContractRef` is an interface implemented by `AccountRef`. Do you mean to use `Contract` which also implements `ContractRef` ?

michaelkaplan13: Doesn't matter either way.

PR #386, opened 2022-12-01
Improved eth_getLogs caching

patrick-ogrady:
## Why this should be merged
This PR separates the `indexed` and `unindexed` block log caches. We do this because we want the storage of indexed logs to evict unindexed logs (which are much more important to keep in memory for low RPC latency).

## How this works
When serving `eth_getLogs` requests, the node may iterate over any number of blocks looking for events. It significantly improves performance to keep these blocks in memory and separated between indexed/unindexed logs (requires iteration of all unindexed logs for each request).

### DIFF
--- a/eth/ethconfig/config.go
+++ b/eth/ethconfig/config.go
@@ -52,18 +52,19 @@ var DefaultConfig = NewDefaultConfig()
 
 func NewDefaultConfig() Config {
 	return Config{
-		NetworkId:             1,
-		TrieCleanCache:        512,
-		TrieDirtyCache:        256,
-		TrieDirtyCommitTarget: 20,
-		SnapshotCache:         256,
-		FilterLogCacheSize:    32,
-		Miner:                 miner.Config{},
-		TxPool:                core.DefaultTxPoolConfig,
-		RPCGasCap:             25000000,
-		RPCEVMTimeout:         5 * time.Second,
-		GPO:                   DefaultFullGPOConfig,
-		RPCTxFeeCap:           1,
+		NetworkId:                   1,
+		TrieCleanCache:              512,
+		TrieDirtyCache:              256,
+		TrieDirtyCommitTarget:       20,
+		SnapshotCache:               256,
+		IndexedFilterLogCacheSize:   32,
+		UnindexedFilterLogCacheSize: 32,
+		Miner:                       miner.Config{},
+		TxPool:                      core.DefaultTxPoolConfig,
+		RPCGasCap:                   25000000,
+		RPCEVMTimeout:               5 * time.Second,
+		GPO:                         DefaultFullGPOConfig,
+		RPCTxFeeCap:                 1,
 	}
 }
 
@@ -101,8 +102,10 @@ type Config struct {
 	SnapshotCache         int
 	Preimages             bool
 
-	// This is the number of blocks for which logs will be cached in the filter system.
-	FilterLogCacheSize int
+	// This is the number of indexed blocks for which logs will be cached in the filter system.
+	IndexedFilterLogCacheSize int
+	// This is the number of unindexed blocks for which logs will be cached in the filter system.
+	UnindexedFilterLogCacheSize int
 
 	// Mining options
 	Miner miner.Config
--- a/eth/filters/filter.go
+++ b/eth/filters/filter.go
@@ -266,21 +266,21 @@ func (f *Filter) unindexedLogs(ctx context.Context, end uint64) ([]*types.Log, e
 func (f *Filter) blockLogs(ctx context.Context, header *types.Header, skipBloom bool, indexed bool) ([]*types.Log, error) {
 	// Fast track: no filtering criteria
 	if len(f.addresses) == 0 && len(f.topics) == 0 {
-		list, err := f.sys.cachedGetLogs(ctx, header.Hash(), header.Number.Uint64(), !skipBloom)
+		list, err := f.sys.cachedGetLogs(ctx, header.Hash(), header.Number.Uint64(), indexed)
 		if err != nil {
 			return nil, err
 		}
 		return flatten(list), nil
 	} else if skipBloom || bloomFilter(header.Bloom, f.addresses, f.topics) {
-		return f.checkMatches(ctx, header)
+		return f.checkMatches(ctx, header, indexed)
 	}
 	return nil, nil
 }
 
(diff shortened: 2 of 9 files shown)


eth/ethconfig/config.go
```diff
@@ -52,18 +52,19 @@ var DefaultConfig = NewDefaultConfig()
...
-		RPCEVMTimeout:         5 * time.Second,
-		GPO:                   DefaultFullGPOConfig,
-		RPCTxFeeCap:           1,
+		NetworkId:                   1,
+		TrieCleanCache:              512,
+		TrieDirtyCache:              256,
+		TrieDirtyCommitTarget:       20,
+		SnapshotCache:               256,
+		IndexedFilterLogCacheSize:   32,
+		UnindexedFilterLogCacheSize: 32,
```
patrick-ogrady, line 61: We don't increase the default here because it could require more memory. This should be tuned by API providers.

eth/filters/filter_system.go
```diff
@@ -50,16 +50,20 @@ import (
...
+	Timeout               time.Duration // how long filters stay active (default: 5min)
 }
 
 func (cfg Config) withDefaults() Config {
 	if cfg.Timeout == 0 {
 		cfg.Timeout = 5 * time.Minute
 	}
-	if cfg.LogCacheSize == 0 {
-		cfg.LogCacheSize = 32
+	if cfg.IndexedLogCacheSize == 0 {
```
patrick-ogrady, line 62: We should never pass 0 here based on the way configs are loaded, but we defend just in case.

eth/filters/filter.go
```diff
@@ -261,10 +263,10 @@ func (f *Filter) unindexedLogs(ctx context.Context, end uint64) ([]*types.Log, e
 }
 
 // blockLogs returns the logs matching the filter criteria within a single block.
-func (f *Filter) blockLogs(ctx context.Context, header *types.Header, skipBloom bool) ([]*types.Log, error) {
+func (f *Filter) blockLogs(ctx context.Context, header *types.Header, skipBloom bool, indexed bool) ([]*types.Log, error) {
 	// Fast track: no filtering criteria
 	if len(f.addresses) == 0 && len(f.topics) == 0 {
-		list, err := f.sys.cachedGetLogs(ctx, header.Hash(), header.Number.Uint64())
+		list, err := f.sys.cachedGetLogs(ctx, header.Hash(), header.Number.Uint64(), !skipBloom)
```
patrick-ogrady, line 269: This is a mistake. Fixing now.

eth/filters/filter.go
```diff
@@ -278,7 +280,7 @@ func (f *Filter) blockLogs(ctx context.Context, header *types.Header, skipBloom
 // checkMatches checks if the receipts belonging to the given header contain any log events that
 // match the filter criteria. This function is called when the bloom filter signals a potential match.
 func (f *Filter) checkMatches(ctx context.Context, header *types.Header) ([]*types.Log, error) {
-	logsList, err := f.sys.cachedGetLogs(ctx, header.Hash(), header.Number.Uint64())
+	logsList, err := f.sys.cachedGetLogs(ctx, header.Hash(), header.Number.Uint64(), true)
```
patrick-ogrady, line 283: mistake. Fixing now.

eth/filters/filter_system.go
```diff
@@ -626,7 +640,7 @@ func (es *EventSystem) lightFilterLogs(header *types.Header, addresses []common.
 		// Get the logs of the block
 		ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
 		defer cancel()
-		logsList, err := es.sys.cachedGetLogs(ctx, header.Hash(), header.Number.Uint64())
+		logsList, err := es.sys.cachedGetLogs(ctx, header.Hash(), header.Number.Uint64(), true)
```
patrick-ogrady, line 643: We should never hit this but we'll make sure it is correct.

eth/filters/filter.go
```diff
@@ -128,7 +128,9 @@ func (f *Filter) Logs(ctx context.Context) ([]*types.Log, error) {
 		if header == nil {
 			return nil, errors.New("unknown block")
 		}
-		return f.blockLogs(ctx, header, false)
+		size, sections := f.sys.backend.BloomStatus()
+		lastIndexed := size * sections
+		return f.blockLogs(ctx, header, false, lastIndexed > header.Number.Uint64())
```
darioush, line 133: unsure that it makes a difference, is `>` or `>=` correct here?

eth/ethconfig/config.go
```diff
@@ -101,8 +102,10 @@ type Config struct {
 	SnapshotCache         int
 	Preimages             bool
 
-	// This is the number of blocks for which logs will be cached in the filter system.
-	FilterLogCacheSize int
```
patrick-ogrady, line 105: Safe to change because this was never actually exposed to the operator before.

darioush (approved):

eth/filters/filter.go
patrick-ogrady: Elsewhere, it is >

patrick-ogrady: image

patrick-ogrady: I thought about creating a `LastIndexed` function but wanted to tread lightly. idk.

darioush (approved):

eth/filters/filter_system.go
```diff
@@ -626,7 +641,9 @@ func (es *EventSystem) lightFilterLogs(header *types.Header, addresses []common.
 		// Get the logs of the block
 		ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
 		defer cancel()
-		logsList, err := es.sys.cachedGetLogs(ctx, header.Hash(), header.Number.Uint64())
+		indexed := es.sys.backend.LastBloomIndex()
+		headerNumber := header.Number.Uint64()
+		logsList, err := es.sys.cachedGetLogs(ctx, header.Hash(), headerNumber, indexed > headerNumber)
```
aaronbuchwald, line 646: Noticing we could actually remove `lightFilterLogs` completely, out of scope for this PR though

aaronbuchwald: Also no real need to

patrick-ogrady: Yeah, I thought the same.

aaronbuchwald (approved):

PR #387, opened 2022-12-02
Cache using accepted depth

patrick-ogrady:
## Why this should be merged
Improvement to #386

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -165,6 +165,7 @@ type CacheConfig struct {
 	SnapshotVerify                  bool          // Verify generated snapshots
 	SkipSnapshotRebuild             bool          // Whether to skip rebuilding the snapshot in favor of returning an error (only set to true for tests)
 	Preimages                       bool          // Whether to store preimage of trie key to the disk
+	AcceptedCacheSize               int           // Number of blocks of accepted headers/logs to cache at the accepted tip (for eth_getLogs)
 }
 
 var DefaultCacheConfig = &CacheConfig{
@@ -175,6 +176,7 @@ var DefaultCacheConfig = &CacheConfig{
 	CommitInterval:        4096,
 	AcceptorQueueLimit:    64, // Provides 2 minutes of buffer (2s block target) for a commit delay
 	SnapshotLimit:         256,
+	AcceptedCacheSize:     32,
 }
 
 // BlockChain represents the canonical chain given a database with a genesis
@@ -309,17 +311,19 @@ func NewBlockChain(
 			Preimages:   cacheConfig.Preimages,
 			StatsPrefix: trieCleanCacheStatsNamespace,
 		}),
-		bodyCache:      bodyCache,
-		receiptsCache:  receiptsCache,
-		blockCache:     blockCache,
-		txLookupCache:  txLookupCache,
-		feeConfigCache: feeConfigCache,
-		engine:         engine,
-		vmConfig:       vmConfig,
-		badBlocks:      badBlocks,
-		senderCacher:   newTxSenderCacher(runtime.NumCPU()),
-		acceptorQueue:  make(chan *types.Block, cacheConfig.AcceptorQueueLimit),
-		quit:           make(chan struct{}),
+		bodyCache:            bodyCache,
+		receiptsCache:        receiptsCache,
+		blockCache:           blockCache,
+		txLookupCache:        txLookupCache,
+		feeConfigCache:       feeConfigCache,
+		engine:               engine,
+		vmConfig:             vmConfig,
+		badBlocks:            badBlocks,
+		senderCacher:         newTxSenderCacher(runtime.NumCPU()),
+		acceptorQueue:        make(chan *types.Block, cacheConfig.AcceptorQueueLimit),
+		quit:                 make(chan struct{}),
+		acceptedHeadersCache: linkedhashmap.New[uint64, *types.Header](),
+		acceptedLogsCache:    linkedhashmap.New[common.Hash, [][]*types.Log](),
 	}
 	bc.validator = NewBlockValidator(chainConfig, bc, engine)
 	bc.prefetcher = newStatePrefetcher(chainConfig, bc, engine)
@@ -450,11 +454,11 @@ func (bc *BlockChain) warmAcceptedCaches() {
 		lastAccepted = bc.LastAcceptedBlock().NumberU64()
 		seen         = 0
 	)
-	for i := lastAccepted; i >= 1 && seen < 100; i-- {
+	for i := lastAccepted; i >= 1 /* don't load genesis */ && seen < bc.cacheConfig.AcceptedCacheSize; i-- {
 		header := bc.GetHeaderByNumber(i)
 		if header == nil {
 			// This could happen if a node state-synced
-			log.Info("Exiting accepted cache warming early", "height", i, "t", time.Since(startTime))
+			log.Info("Exiting accepted cache warming early because header is nil", "height", i, "t", time.Since(startTime))
 			break
 		}
 		bc.acceptedHeadersCache.Put(header.Number.Uint64(), header)
@@ -462,7 +466,7 @@ func (bc *BlockChain) warmAcceptedCaches() {
 		bc.acceptedLogsCache.Put(header.Hash(), logs)
 		seen++
 	}
-	log.Info("Warmed accepted caches", "t", time.Since(startTime))
+	log.Info("Warmed accepted caches", "items", seen, "t", time.Since(startTime))
 }
 
 // startAcceptor starts processing items on the [acceptorQueue]. If a [nil]
@@ -486,20 +490,19 @@ func (bc *BlockChain) startAcceptor() {
 		}
 
 		// Ensure [acceptedHeadersCache] and [acceptedLogsCache] have latest content
+		//
+		// We know we can always remove the oldest because the cache is already
+		// full after startup (we always prewarm it).
 		bc.acceptedHeadersCache.Put(next.NumberU64(), next.Header())
-		if bc.acceptedHeadersCache.Len() > 100 {
-			oldest, _, _ := bc.acceptedHeadersCache.Oldest()
-			bc.acceptedHeadersCache.Delete(oldest)
-		}
+		oldestNumber, _, _ := bc.acceptedHeadersCache.Oldest()
+		bc.acceptedHeadersCache.Delete(oldestNumber)
 		logs := rawdb.ReadLogs(bc.db, next.Hash(), next.NumberU64())
 		bc.acceptedLogsCache.Put(next.Hash(), logs)
-		if bc.acceptedLogsCache.Len() > 100 {
-			oldest, _, _ := bc.acceptedLogsCache.Oldest()
-			bc.acceptedLogsCache.Delete(oldest)
-		}
+		oldestHash, _, _ := bc.acceptedLogsCache.Oldest()
+		bc.acceptedLogsCache.Delete(oldestHash)
 
 		// Update accepted feeds
-		bc.chainAcceptedFeed.Send(ChainEvent{Block: next, Hash: next.Hash(), Logs: flatten(logs)})
+		bc.chainAcceptedFeed.Send(ChainEvent{Block: next, Hash: next.Hash(), Logs: FlattenLogs(logs)})
 		if len(logs) > 0 {
 			bc.logsAcceptedFeed.Send(logs)
 		}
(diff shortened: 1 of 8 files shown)


patrick-ogrady:
TODO:
```
[node5] panic: event: wrong type in Send got [][]*types.Log, want []*types.Log
[node5] 
[node5] goroutine 65 [running]:
[node5] github.com/ethereum/go-ethereum/event.(*Feed).Send(0xc000f662a0, {0xe4e180?, 0xc000818780?})
[node5] 	/home/runner/go/pkg/mod/github.com/ethereum/go-ethereum@v1.10.26/event/feed.go:142 +0xb2e
[node5] github.com/ava-labs/subnet-evm/core.(*BlockChain).startAcceptor(0xc000f66000)
[node5] 	/home/runner/work/subnet-evm/subnet-evm/core/blockchain.go:508 +0x7ac
[node5] created by github.com/ava-labs/subnet-evm/core.NewBlockChain
[node5] 	/home/runner/work/subnet-evm/subnet-evm/core/blockchain.go:390 +0x106c
```

core/blockchain_reader.go
```diff
@@ -76,7 +76,19 @@ func (bc *BlockChain) GetHeaderByHash(hash common.Hash) *types.Header {
...
+//
+// We optimistically attempt to fetch the header from [acceptedHeadersCache] to
+// avoid a number to hash db lookup to access the inner cache. This path is
+// commonly accessed when processing eth_getLogs requests.
+//
+// We explicitly choose to maintain both hashes because the inner cache can be
+// used with the unprocessed tip, whereas eth_getLogs is only used with
+// accepted blocks by default.
 func (bc *BlockChain) GetHeaderByNumber(number uint64) *types.Header {
+	h, ok := bc.acceptedHeadersCache.Get(number)
```
darioush, line 88: can we add a comment that the cache is thread safe?

core/blockchain.go
```diff
@@ -1949,3 +1994,12 @@ func (bc *BlockChain) ResetToStateSyncedBlock(block *types.Block) error {
 	bc.initSnapshot(head)
 	return nil
 }
+
+// FlattenLogs converts a nested array of logs to a single array of logs.
+func FlattenLogs(list [][]*types.Log) []*types.Log {
```
darioush, line 2000: nit: could this be unexported?

darioush (approved):
This switches reading from `rawdb.ReadReceipts` to `rawdb.ReadLogs`. I think they are equivalent.

core/blockchain.go
patrick-ogrady: We use this in `eth/filter`. I moved it to avoid duplicating the code.

core/fifo_cache.go
```diff
@@ -0,0 +1,39 @@
...
+
+func NewFIFOCache[K comparable, V any](limit int) *FIFOCache[K, V] {
+	return &FIFOCache[K, V]{
+		limit: limit,
+		m:     linkedhashmap.New[K, V](),
+	}
+}
+
+func (f *FIFOCache[K, V]) Put(key K, val V) {
+	if f.m.Len() >= f.limit {
```
patrick-ogrady, line 28: This should be better protected before merging. It isn't a race because we only call `Put` in a single thread.

core/fifo_cache.go
```diff
@@ -0,0 +1,39 @@
...
+// TODO: Move this to AvalancheGo
+type FIFOCache[K, V any] struct {
+	limit int
+
+	m linkedhashmap.LinkedHashmap[K, V]
+
+	// [Linkedhashmap] is thread-safe, so no additional locking is required
+}
+
+func NewFIFOCache[K comparable, V any](limit int) *FIFOCache[K, V] {
```
patrick-ogrady, line 20: ❤️ generics

core/blockchain.go
```diff
@@ -174,6 +175,7 @@ var DefaultCacheConfig = &CacheConfig{
 	CommitInterval:        4096,
 	AcceptorQueueLimit:    64, // Provides 2 minutes of buffer (2s block target) for a commit delay
 	SnapshotLimit:         256,
+	AcceptedCacheSize:     32,
```
patrick-ogrady, line 178: We may want to increase this default substantially. It isn't a lot of data AFAICT.

core/bounded_buffer.go
```diff
@@ -42,6 +47,6 @@ func (b *BoundedBuffer) Insert(h common.Hash) {
 
 // Last retrieves the last item added to the buffer.
 // If no items have been added to the buffer, Last returns an empty hash.
-func (b *BoundedBuffer) Last() common.Hash {
+func (b *BoundedBuffer[K]) Last() K {
 	return b.buffer[b.lastPos]
```
patrick-ogrady, line 51: There is a bug here...fixing now (can be -1)

core/bounded_buffer.go
```diff
@@ -3,45 +3,55 @@
...
 // Last retrieves the last item added to the buffer.
-// If no items have been added to the buffer, Last returns an empty hash.
-func (b *BoundedBuffer) Last() common.Hash {
-	return b.buffer[b.lastPos]
+//
+// If no items have been added to the buffer, Last returns the default value of
+// [K] and [false].
+func (b *BoundedBuffer[K]) Last() (K, bool) {
+	if b.lastPos == -1 {
+		return *new(K), false
```
patrick-ogrady, line 54: only weird part about generics

core/bounded_buffer.go
```diff
@@ -3,45 +3,55 @@
...
 	}
 }
 
-// Insert adds a new common.Hash to the buffer. If the buffer is full, the
-// oldest common.Hash will be evicted and [callback] will be invoked.
-//
-// WARNING: BoundedBuffer does not support the insertion of empty common.Hash.
-// Inserting such data will cause unintended behavior.
-func (b *BoundedBuffer) Insert(h common.Hash) {
-	nextPos := (b.lastPos + 1) % b.size // the first item added to the buffer will be at position 1
```
patrick-ogrady, line 35: Removed this weird off-by-1 start and now explicitly return in `Last` if the value was set

core/fifo_cache.go
```diff
@@ -18,10 +20,16 @@ type FIFOCache[K comparable, V any] struct {
 	m      map[K]V
 }
 
+// NewFIFOCache creates a new First-In-First-Out cache of size [limit].
+//
+// If [limit] is less than 16, [limit] will be overwritten with 16.
```
patrick-ogrady, line 25: Alternatively, we could have a no-op cache if 0 is passed in (which occurs in some tests).

core/fifo_cache.go
```diff
@@ -7,52 +7,69 @@ import (
...
+	Get(K) (V, bool)
 }
 
 // NewFIFOCache creates a new First-In-First-Out cache of size [limit].
 //
-// If [limit] is less than 16, [limit] will be overwritten with 16.
-func NewFIFOCache[K comparable, V any](limit int) *FIFOCache[K, V] {
-	if limit < minCacheSize {
-		limit = minCacheSize
+// If a [limit] of 0 is passed as an argument, a no-op cache is returned that
```
patrick-ogrady, line 24: Some tests run with `AcceptedCacheSize=0`. We can either:
* Pipe a `nil` check on this cache through the code (gross)
* Ensure all configs run with a cache > 0
* Enforce a min cache size (to overwrite 0 limit)
* Do this

core/blockchain_reader.go
```diff
@@ -76,7 +76,19 @@ func (bc *BlockChain) GetHeaderByHash(hash common.Hash) *types.Header {
...
+// commonly accessed when processing eth_getLogs requests.
+//
+// We explicitly choose to maintain both hashes because the inner cache can be
+// used with the unprocessed tip, whereas eth_getLogs is only used with
+// accepted blocks by default.
 func (bc *BlockChain) GetHeaderByNumber(number uint64) *types.Header {
+	h, ok := bc.acceptedHeadersCache.Get(number) // this cache is thread-safe
+	if ok {
+		return h
+	}
```
aaronbuchwald, line 91: Can we move this cache into `hc` since there are already header caches there?

At first glance, I thought this may be a duplicate, but see that this maps an accepted number to the header itself without requiring the hash to number lookup, but still think this would be more clear if we moved it into `hc`.

core/fifo_cache.go
darioush: This looks nice.

core/blockchain.go
```diff
@@ -435,6 +447,36 @@ func (bc *BlockChain) flattenSnapshot(postAbortWork func() error, hash common.Ha
 	return bc.snaps.Flatten(hash)
 }
 
+// warmAcceptedCaches fetches previously accepted headers and logs from disk to
+// pre-populate [acceptedHeadersCache] and [acceptedLogsCache].
+func (bc *BlockChain) warmAcceptedCaches() {
```
aaronbuchwald, line 452: Is it necessary to warm this cache? It should do so once it receives requests, so I'm not sure adding this piece of code is worth it.

core/fifo_cache.go
```diff
@@ -0,0 +1,75 @@
...
+func NewFIFOCache[K comparable, V any](limit int) FIFOCache[K, V] {
+	if limit == 0 {
+		return &NoOpFIFOCache[K, V]{}
+	}
+
+	c := &BufferFIFOCache[K, V]{
+		limit: limit,
+		m:     make(map[K]V, limit),
+	}
+	c.buffer = NewBoundedBuffer[K](limit, c.remove)
```
aaronbuchwald, line 35:
```suggestion
	c.buffer = NewBoundedBuffer(limit, c.remove)
```

my linter is telling me this is an unnecessary type argument

eth/filters/filter_system.go
```diff
@@ -641,9 +610,7 @@ func (es *EventSystem) lightFilterLogs(header *types.Header, addresses []common.
 		// Get the logs of the block
 		ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
 		defer cancel()
-		indexed := es.sys.backend.LastBloomIndex()
```
aaronbuchwald, line 644: Is `LastBloomIndex()` still necessary or can we revert this change from the prior PR?

core/blockchain.go
aaronbuchwald: Could we move this to next to `types.Log` ?

eth/filters/filter_system.go
patrick-ogrady: We could revert. I considered doing this as well.

core/blockchain.go
patrick-ogrady: Prewarming 8192 blocks over gRPC takes ~9s (just for context)

patrick-ogrady: We need some notion of this or I need to add items to the cache during request processing to avoid a pretty bad cold start issue on startup IMO.

eth/filters/filter_system.go
patrick-ogrady: Removed

eth/api_backend.go
```diff
@@ -252,7 +252,10 @@ func (b *EthAPIBackend) GetReceipts(ctx context.Context, hash common.Hash) (type
 }
 
 func (b *EthAPIBackend) GetLogs(ctx context.Context, hash common.Hash, number uint64) ([][]*types.Log, error) {
-	return rawdb.ReadLogs(b.eth.chainDb, hash, number), nil
+	if err := ctx.Err(); err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
patrick-ogrady: We have this for almost all other functions. I believe this was missed when we did this pass (or added later).

core/blockchain_reader.go
patrick-ogrady: This annoyed me as well. I moved to keep all header caches together.

core/blockchain.go
```diff
@@ -305,24 +310,25 @@ func NewBlockChain(
...
+		quit:              make(chan struct{}),
+		acceptedLogsCache: NewFIFOCache[common.Hash, [][]*types.Log](cacheConfig.AcceptedCacheSize),
 	}
 	bc.validator = NewBlockValidator(chainConfig, bc, engine)
 	bc.prefetcher = newStatePrefetcher(chainConfig, bc, engine)
 	bc.processor = NewStateProcessor(chainConfig, bc, engine)
 
 	var err error
-	bc.hc, err = NewHeaderChain(db, chainConfig, engine)
+	bc.hc, err = NewHeaderChain(db, chainConfig, cacheConfig, engine)
```
patrick-ogrady, line 331: This will allow us to tune other header caches directly in the future.

darioush (approved):

core/blockchain_reader.go
```diff
@@ -76,6 +76,8 @@ func (bc *BlockChain) GetHeaderByHash(hash common.Hash) *types.Header {
 
 // GetHeaderByNumber retrieves a block header from the database by number,
 // caching it (associated with its hash) if found.
+//
+// This path is commonly accessed when processing eth_getLogs requests.
```
aaronbuchwald, line 80: Can we remove this comment? I don't think this needs to be here and many API calls hit `GetHeaderByNumber`

patrick-ogrady: Fair

aaronbuchwald (approved):
LGTM

PR #391, opened 2022-12-05
use generic set

danlaine:
## Why this should be merged

use new generic set functionality from avalanchego

## How this was tested

CI


aaronbuchwald (approved):
LGTM once CI is fixed with the AvalancheGo release

PR #395, opened 2022-12-06
replace getByKey with getByAddress

ceyonur:
## Why this should be merged

Removes duplicate key/identifier for configs. Reduces // ADD YOUR PRECOMPILE lines.

## How this works

Removes precompile keys from upgrade configs, uses addresses instead of keys to iterate/fetch configs from upgrades. replaces `getByKey` with `getByAddress `.

## How this was tested

Locally tested with a 5 node setup which has a later-activated upgrade.


params/precompile_config.go
```diff
@@ -41,7 +42,7 @@ func (p *PrecompileUpgrade) getByAddress(address common.Address) (precompile.Sta
 	// ADD YOUR PRECOMPILE HERE
 	/*
 		case precompile.{YourPrecompile}Address:
-		return p.{YourPrecompile}Config , p.{YourPrecompile}Config  != nil
+		  return p.{YourPrecompile}Config , p.{YourPrecompile}Config != nil
```
darioush, line 45:
```suggestion
		return p.{YourPrecompile}Config , p.{YourPrecompile}Config != nil
```

darioush (approved):

params/precompile_config.go
aaronbuchwald: Is there an extra space before the comma here?

aaronbuchwald (approved):
LGTM 🔥

PR #396, opened 2022-12-06
generalize precompile specific upgrade checks

ceyonur:
## Why this should be merged

Reduces mandatory manual code addition for new precompiles ("ADD YOUR PRECOMPILE HERE")

## How this works

Removes precompile specific functions from chain config and uses a generalized check function which takes precompile address.

## How this was tested

Locally tested.


darioush (approved):

aaronbuchwald (approved):
LGTM

PR #398, opened 2022-12-07
Add Inner to ethdb.Batch interface

aaronbuchwald:
## Why this should be merged

This PR will allow the `ethdb.Batch` to implement the AvalancheGo `database.Batch` interface so that we can apply an `ethdb.Batch` alongside shared memory operations.

## How this works

This works by implementing the `Inner()` function from the AvalancheGo `database.Batch` interface. This needs to return a `database.Batch` interface from AvalancheGo not the `ethdb.Batch` interface type.

## How this was tested

This is intended as a no-op change, so this was tested via CI.

### DIFF
--- a/core/rawdb/table.go
+++ b/core/rawdb/table.go
@@ -168,6 +168,11 @@ func (b *tableBatch) Reset() {
 	b.batch.Reset()
 }
 
+// Inner returns itself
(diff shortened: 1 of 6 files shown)


aaronbuchwald:
This approach does not currently work because `Inner()` needs to return the avalanchego type `database.Batch` not `ethdb.Batch`.

PR #400, opened 2022-12-07
core/*: fix go staticcheck warnings (lint)

gyuho:

## Why this should be merged

Fixing linter warnings.

## How this works

N/A

## How this was tested

Unit tests

### DIFF
--- a/core/test_blockchain.go
+++ b/core/test_blockchain.go
@@ -216,7 +216,7 @@ func TestInsertChainAcceptSingleBlock(t *testing.T, create func(db ethdb.Databas
 	signer := types.HomesteadSigner{}
 	// Generate chain of blocks using [genDB] instead of [chainDB] to avoid writing
 	// to the BlockChain's database while generating blocks.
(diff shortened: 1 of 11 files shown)


core/test_blockchain.go
```diff
@@ -1004,8 +1004,10 @@ func TestReorgReInsert(t *testing.T, create func(db ethdb.Database, chainConfig
 // Insert two different chains that result in the identical state root.
 // Once we accept one of the chains, we insert and accept A3 on top of the shared
 // state root
-//   G   (genesis)
-//  / \
+//
+//	 G   (genesis)
+//	/ \
+//
```
darioush, line 1010: could we maintain the diagram's ascii art alignment if we make this change?
(or yet better not modify the alignment in these comments)

darioush (approved):

aaronbuchwald:
I'd prefer not to apply linters to geth code that they are not going to merge in since it increases the diff for an unnecessary change.

gyuho:
> I'd prefer not to apply linters to geth code

Ah makes sense. Agreeing to this point :)

PR #401, opened 2022-12-07
*: make tx pool configurable

gyuho:
## Why this should be merged

To avoid making manual patches for load testing: https://github.com/ava-labs/subnet-evm/tree/fix-load-test

## How this works

Can be set via chain config

## How this was tested

Unit tests.


aaronbuchwald (approved):
LGTM

PR #410, opened 2022-12-12
Fix load test

aaronbuchwald:
This PR improves the existing load test


core/tx_pool.go
```diff
@@ -193,10 +193,10 @@ var DefaultTxPoolConfig = TxPoolConfig{
...
 	PriceBump:  10,
 
-	AccountSlots: 16,
-	GlobalSlots:  4096 + 1024, // urgent + floating queue capacity with 4:1 ratio
-	AccountQueue: 64,
-	GlobalQueue:  1024,
+	AccountSlots: 64,
+	GlobalSlots:  100_000, // urgent + floating queue capacity with 4:1 ratio
+	AccountQueue: 128,
+	GlobalQueue:  100_000,
```
aaronbuchwald, line 199: move these settings to the chain config rather than changing the default

tests/e2e/utils/params.go
```diff
@@ -7,7 +7,7 @@ import (
 	"os"
 	"sync"
 
-	runner_sdk "github.com/ava-labs/avalanche-network-runner-sdk"
+	runner_sdk "github.com/ava-labs/avalanche-network-runner/client"
```
aaronbuchwald, line 10: what's the reason for the switch from the sdk to the network runner client?

tests/e2e/runner/runner.go
```diff
@@ -33,11 +34,22 @@ func (ci clusterInfo) Save(p string) error {
...
+		return err
+	}
+	logFactory := logging.NewFactory(logging.Config{
+		DisplayLevel: logLevel,
+		LogLevel:     logLevel,
+	})
+	log, err := logFactory.Make("main")
+	if err != nil {
+		return err
+	}
```
aaronbuchwald, line 48: if we need to keep this, we should add a helper function for this

PR #413, opened 2022-12-15
Safe balance ops

ceyonur:
## Why this should be merged

With the Native Minter Precompile there can be an overflow in balances. This one adds an error to `AddBalance` and checks the overflow in necessary places.

## How this works

Adds an overflow check to `AddBalance` and other necessary places.g

## How this was tested

Added unit test


precompile/contract_native_minter.go
```diff
@@ -75,7 +75,9 @@ func (c *ContractNativeMinterConfig) Configure(_ ChainConfig, state StateDB, _ B
 	for to, amount := range c.InitialMint {
 		if amount != nil {
 			bigIntAmount := (*big.Int)(amount)
-			state.AddBalance(to, bigIntAmount)
+			if err := state.AddBalance(to, bigIntAmount); err != nil {
+				panic(fmt.Errorf("failed to mint initial balance for %s: %w", to.String(), err))
```
darioush, line 79: Should we consider adding error to Configure, or setting the balance to MaxInt here instead?

Maybe this is fine too, but would like to get thoughts on it, @ceyonur , @aaronbuchwald?

darioush (commented):
Looks great, just the one comment on panic vs. error.

core/evm.go
```diff
@@ -121,14 +123,24 @@ func GetHashFn(ref *types.Header, chain ChainContext) func(n uint64) common.Hash
...
+// and if operation ends up with an overflow.
 // This does not take the necessary gas in to account to make the transfer valid.
-func CanTransfer(db vm.StateDB, addr common.Address, amount *big.Int) bool {
-	return db.GetBalance(addr).Cmp(amount) >= 0
+func CanTransfer(db vm.StateDB, from common.Address, to common.Address, amount *big.Int) error {
+	if db.GetBalance(from).Cmp(amount) < 0 {
+		return vmerrs.ErrInsufficientBalance
+	}
+	// check for overflow. if this is a transfer to self, then we don't need to check for overflow
+	if _, ok := utils.SafeSumUint256(db.GetBalance(to), amount); from != to && !ok {
```
aaronbuchwald, line 134: can we separate the `from` `to` check so that it is prior to doing the safe sum calculation?

core/state/statedb_test.go
```diff
@@ -963,3 +966,21 @@ func TestFlushOrderDataLoss(t *testing.T) {
...
+		memdb    = rawdb.NewMemoryDatabase()
+		statedb  = NewDatabase(memdb)
+		state, _ = New(common.Hash{}, statedb, nil)
+	)
+
+	addr := common.BytesToAddress([]byte("test"))
+
+	err := state.AddBalance(addr, ethmath.MaxBig256)
+	require.NoError(t, err)
+	// add it again and expect overflow
```
aaronbuchwald, line 981:
```suggestion
	// add 1 to trigger overflow and assert we get the correct error
```

core/state_transition.go
```diff
@@ -326,7 +326,7 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
 		contractCreation = msg.To() == nil
 	)
 
-	// Check clauses 4-5, subtract intrinsic gas if everything is correct
+	// Check clauses 4-6, subtract intrinsic gas if everything is correct
```
aaronbuchwald, line 329: Should this be 5-6 instead of 4-6 ?

core/state_transition.go
```diff
@@ -356,8 +356,16 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
 		st.state.SetNonce(msg.From(), st.state.GetNonce(sender.Address())+1)
 		ret, st.gas, vmerr = st.evm.Call(sender, st.to(), st.data, st.gas, st.value)
 	}
+
+	// refund gas first in case refunded address is the coinbase.
+	// otherwise refunding after adding balance to coinbase might result into
+	// an overflow
```
aaronbuchwald, line 362: refunding gas could trigger an overflow in the case that a transaction execution results in increasing the balance of the caller, so should this function return an error?

core/state_transition.go
```diff
@@ -378,7 +386,8 @@ func (st *StateTransition) refundGas(subnetEVM bool) {
 	}
 	// Return ETH for remaining gas, exchanged at the original rate.
 	remaining := new(big.Int).Mul(new(big.Int).SetUint64(st.gas), st.gasPrice)
-	st.state.AddBalance(st.msg.From(), remaining)
+	// This is fine against overflow since we already subtracted the fee from the sender
```
aaronbuchwald, line 389: this occurs at the end of the transaction execution, so I don't think this error is necessarily safe to ignore.

core/vm/evm.go
```diff
@@ -425,7 +428,8 @@ func (evm *EVM) StaticCall(caller ContractRef, addr common.Address, input []byte
 	// This doesn't matter on Mainnet, where all empties are gone at the time of Byzantium,
 	// but is the correct thing to do and matters on other networks, in tests, and potential
 	// future scenarios
-	evm.StateDB.AddBalance(addr, big0)
+	// This fine against overflow, since amount is zero
+	_ = evm.StateDB.AddBalance(addr, big0)
```
aaronbuchwald, line 432: `AddBalance` should return a nil error here, would it be better to check the error value and rely on the fact that it returns nil (with the value 0) rather than ignore the error in the caller?

core/vm/evm.go
```diff
@@ -256,6 +258,7 @@ func (evm *EVM) Call(caller ContractRef, addr common.Address, input []byte, gas
 		}
 		evm.StateDB.CreateAccount(addr)
 	}
+	// we have already checked overflow before capturing snapshot, so it's safe to ignore the error here
```
aaronbuchwald, line 261: This comment is a bit confusing since the function `Transfer` does not actually return an error. Could we instead add a comment like:

```suggestion
	// It is safe to call Transfer since we have not mutated balances since calling CanTransfer
```

core/vm/evm.go
```diff
@@ -520,6 +524,8 @@ func (evm *EVM) create(caller ContractRef, codeAndHash *codeAndHash, gas uint64,
 	if evm.chainRules.IsEIP158 {
 		evm.StateDB.SetNonce(address, 1)
 	}
+
+	// We have already checked overflow with transfer so it's safe to ignore the error here
```
aaronbuchwald, line 528: same comment here

precompile/contract_native_minter.go
aaronbuchwald: If someone messes up their EVM with the over-usage of the native minter, then it would be preferable for things to work mostly as expected. I'd be in favor of changing the behavior from returning an invalidating error to simply setting the balance to maxuint256 if increasing an address' balance causes an overflow.

This way if an address hits max uint 256, sending funds to that address will result in burning the overflowed funds, but will not incorrectly set the address' balance.

As for panic/error, we use the invariant that `Verify` is called on any precompile updates prior to including them in the chain config, so we can assume that `Verify()` has passed. However, since we don't do balance checks the initial minter is an exception. I'd still be in favor of changing the behavior of `AddBalance` to set the balance to maxuint256 if it encounters an overflow rather than returning an error.

For errors that should never occur, I'd be in favor of moving them to the caller to decide what to do with them. We can opt for a better behavior to indicate that this is a critical error without panic-ing.

core/state_transition.go
ceyonur: fixed

precompile/contract_native_minter.go
ceyonur: Changed it to setting maxuint256

core/evm.go
ceyonur: we don't need that anymore

core/state_transition.go
ceyonur: we set to maxuint256

core/state_transition.go
ceyonur: this also changed

core/vm/evm.go
ceyonur: changed it

darioush (commented):
Are we okay to change this behavior for the entire VM @aaronbuchwald? my original understanding was we'd only use the set max logic for configure in the precompile activation mint.
I think the logs are fine, but we can be consistent with error / warn

ceyonur:
> Are we okay to change this behavior for the entire VM @aaronbuchwald? my original understanding was we'd only use the set max logic for configure in the precompile activation mint. I think the logs are fine, but we can be consistent with error / warn

IMHO the behavior should be consistent across all components. 
For the logs I think we can set it to lower levels like INFO or even DEBUG (probably not ideal).

core/vm/evm.go
```diff
@@ -425,7 +429,8 @@ func (evm *EVM) StaticCall(caller ContractRef, addr common.Address, input []byte
 	// This doesn't matter on Mainnet, where all empties are gone at the time of Byzantium,
 	// but is the correct thing to do and matters on other networks, in tests, and potential
 	// future scenarios
-	evm.StateDB.AddBalance(addr, big0)
+	// This fine against overflow, since amount is zero
```
aaronbuchwald, line 432:
```suggestion
	// This is fine against overflow, since the amount is zero
```

aaronbuchwald, line 432:
```suggestion
	// This is fine against overflow, since the amount is zero
```

PR #416, opened 2022-12-18
add coinbase config cache

ceyonur:
## Why this should be merged

Adds a cache and reduces fetching data from state.

## How this works

Adds a lru cache with limit 256. Uses it with the parent root in `GetCoinbaseAt` to reduce the number of reads of configured coinbase from statedb.


aaronbuchwald (approved):

anusha-ctrl (approved):
lgtm, should we add/update a test?

aaronbuchwald (approved):

PR #417, opened 2022-12-20
Update reward manager

ceyonur:
## Why this should be merged

Fixes block verification logic in case reward manager is disabled

## How this works

Removes static check and verify coinbase depending on the parent 

## How this was tested

added unit tests to vm_test and locally tested with a real subnet.

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -117,22 +117,23 @@ func (w *worker) commitNewWork() (*types.Block, error) {
 	defer w.mu.RUnlock()
 
 	tstart := w.clock.Time()
-	timestamp := tstart.Unix()
+	timestamp := uint64(tstart.Unix())
 	parent := w.chain.CurrentBlock()
 	// Note: in order to support asynchronous block production, blocks are allowed to have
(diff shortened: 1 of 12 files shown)


miner/worker.go
```diff
@@ -117,22 +117,30 @@ func (w *worker) commitNewWork() (*types.Block, error) {
...
 
+	bigTimestamp := new(big.Int).SetUint64(timestamp)
 	var gasLimit uint64
+	// We need to get the fee config from the parent block because the fee config is
+	// not yet set in the current block.
+	// This is because the fee config might be set with the state in the FeeManager precompile.
+	// Which could change the fee configs in the current block.
+	// But we don't we don't know what the state will be for the current block.
+	// meaning that the tx that changes the precompile state itself could be also invalidated by the
+	// fee config that it sets.
```
aaronbuchwald, line 137:
```suggestion
	// The fee config manager relies on the state of the parent block to set the fee config
	// because the fee config may be changed by the current block.
```

aaronbuchwald (approved):

darioush (approved):

PR #418, opened 2022-12-20
rework on panics in precompiles

ceyonur:
## Why this should be merged

Removes panics in precompile functions and returns errors instead. Also refactors precompile create functions to make it more compact.

## How this works

Changes `Configure` and `CheckConfigurePrecompiles` functions to return an error instead of panicking. 

Also refactors `create{Name}Precompile` functions to be more compact and clean-looking.

## How this was tested

Locally generated reward manager precompile with tool and compared.

### DIFF
--- a/core/state_processor.go
+++ b/core/state_processor.go
@@ -81,8 +81,7 @@ func (p *StateProcessor) Process(block *types.Block, parent *types.Header, state
 	// Configure any stateful precompiles that should go into effect during this block.
 	err := p.config.ConfigurePrecompiles(new(big.Int).SetUint64(parent.Time), block, statedb)
 	if err != nil {
-		werr := fmt.Errorf("could not configure precompiles: %w", err)
-		log.Error("failed to process the state changes", "err", werr)
+		log.Error("failed to process the state changes", "err", err)
 		return nil, nil, 0, err
 	}
 
--- a/core/state_processor.go
+++ b/core/state_processor.go
@@ -37,6 +37,7 @@ import (
 	"github.com/ava-labs/subnet-evm/params"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/log"
 )
 
 // StateProcessor is a basic Processor, which takes care of transitioning
@@ -79,9 +80,10 @@ func (p *StateProcessor) Process(block *types.Block, parent *types.Header, state
 
 	// Configure any stateful precompiles that should go into effect during this block.
 	err := p.config.ConfigurePrecompiles(new(big.Int).SetUint64(parent.Time), block, statedb)
-	// ASK: Should we panic instead?
 	if err != nil {
-		return nil, nil, 0, fmt.Errorf("could not configure precompiles: %w", err)
+		werr := fmt.Errorf("could not configure precompiles: %w", err)
+		log.Error("failed to process the state changes", "err", werr)
+		return nil, nil, 0, err
 	}
 
 	blockContext := NewEVMBlockContext(header, p.bc, nil)
--- a/core/state_processor.go
+++ b/core/state_processor.go
@@ -78,7 +78,11 @@ func (p *StateProcessor) Process(block *types.Block, parent *types.Header, state
 	)
 
 	// Configure any stateful precompiles that should go into effect during this block.
-	p.config.CheckConfigurePrecompiles(new(big.Int).SetUint64(parent.Time), block, statedb)
+	err := p.config.ConfigurePrecompiles(new(big.Int).SetUint64(parent.Time), block, statedb)
+	// ASK: Should we panic instead?
+	if err != nil {
+		return nil, nil, 0, fmt.Errorf("could not configure precompiles: %w", err)
+	}
 
 	blockContext := NewEVMBlockContext(header, p.bc, nil)
 	vmenv := vm.NewEVM(blockContext, vm.TxContext{}, statedb, p.config, cfg)
--- a/core/state_processor.go
+++ b/core/state_processor.go
@@ -81,7 +81,7 @@ func (p *StateProcessor) Process(block *types.Block, parent *types.Header, state
 	// Configure any stateful precompiles that should go into effect during this block.
 	err := p.config.ConfigurePrecompiles(new(big.Int).SetUint64(parent.Time), block, statedb)
 	if err != nil {
-		log.Error("failed to process the state changes", "err", err)
+		log.Error("failed to configure precompiles processing block", "hash", block.Hash(), "number", block.NumberU64(), "timestamp", block.Time(), "err", err)
 		return nil, nil, 0, err
 	}
 
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -186,9 +186,8 @@ func (w *worker) commitNewWork() (*types.Block, error) {
 	// Configure any stateful precompiles that should go into effect during this block.
 	err = w.chainConfig.ConfigurePrecompiles(new(big.Int).SetUint64(parent.Time()), types.NewBlockWithHeader(header), env.state)
 	if err != nil {
-		werr := fmt.Errorf("failed to configure precompiles: %w", err)
-		log.Error("failed to commit new work", "err", werr)
-		return nil, werr
+		log.Error("failed to commit new work", "err", err)
+		return nil, err
 	}
 
 	// Fill the block with all available pending transactions.
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -186,7 +186,7 @@ func (w *worker) commitNewWork() (*types.Block, error) {
 	// Configure any stateful precompiles that should go into effect during this block.
 	err = w.chainConfig.ConfigurePrecompiles(new(big.Int).SetUint64(parent.Time()), types.NewBlockWithHeader(header), env.state)
 	if err != nil {
-		log.Error("failed to commit new work", "err", err)
+		log.Error("failed to configure precompiles mining new block", "parent", parent.Hash(), "number", header.Number, "timestamp", header.Time, "err", err)
 		return nil, err
 	}
 
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -186,7 +186,9 @@ func (w *worker) commitNewWork() (*types.Block, error) {
 	// Configure any stateful precompiles that should go into effect during this block.
 	err = w.chainConfig.ConfigurePrecompiles(new(big.Int).SetUint64(parent.Time()), types.NewBlockWithHeader(header), env.state)
 	if err != nil {
-		return nil, fmt.Errorf("failed to configure precompiles: %w", err)
+		werr := fmt.Errorf("failed to configure precompiles: %w", err)
+		log.Error("failed to commit new work", "err", werr)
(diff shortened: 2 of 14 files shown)


core/state_processor.go
```diff
@@ -78,7 +78,11 @@ func (p *StateProcessor) Process(block *types.Block, parent *types.Header, state
 	)
 
 	// Configure any stateful precompiles that should go into effect during this block.
-	p.config.CheckConfigurePrecompiles(new(big.Int).SetUint64(parent.Time), block, statedb)
+	err := p.config.ConfigurePrecompiles(new(big.Int).SetUint64(parent.Time), block, statedb)
+	// ASK: Should we panic instead?
+	if err != nil {
+		return nil, nil, 0, fmt.Errorf("could not configure precompiles: %w", err)
+	}
```
aaronbuchwald, line 85: We don't NEED to panic, but an error here would indicate that the subnet is most likely borked.

Could we change this to log an error and not panic? If a subnet ends up borked for this reason, it will start to emit error logs, but I think that's better than panic-ing.

miner/worker.go
```diff
@@ -184,7 +184,10 @@ func (w *worker) commitNewWork() (*types.Block, error) {
 		return nil, fmt.Errorf("failed to create new current environment: %w", err)
 	}
 	// Configure any stateful precompiles that should go into effect during this block.
-	w.chainConfig.CheckConfigurePrecompiles(new(big.Int).SetUint64(parent.Time()), types.NewBlockWithHeader(header), env.state)
+	err = w.chainConfig.ConfigurePrecompiles(new(big.Int).SetUint64(parent.Time()), types.NewBlockWithHeader(header), env.state)
+	if err != nil {
+		return nil, fmt.Errorf("failed to configure precompiles: %w", err)
+	}
```
aaronbuchwald, line 190: Same comment on logging an error here since this should never happen

precompile/allow_list.go
```diff
@@ -219,7 +220,12 @@ func createReadAllowList(precompileAddr common.Address) RunStatefulPrecompileFun
 func createAllowListPrecompile(precompileAddr common.Address) StatefulPrecompiledContract {
 	// Construct the contract with no fallback function.
 	allowListFuncs := createAllowListFunctions(precompileAddr)
-	contract := newStatefulPrecompileWithFunctionSelectors(nil, allowListFuncs)
+	contract, err := NewStatefulPrecompileContract(nil, allowListFuncs)
+	// Change this to be returned as an error after refactoring this precompile
+	// to use the new precompile template.
```
aaronbuchwald, line 225:
```suggestion
	// TODO Change this to be returned as an error after refactoring this precompile
	// to use the new precompile template.
```

precompile/contract.go
```diff
@@ -109,12 +109,15 @@ func newStatefulPrecompileWithFunctionSelectors(fallback RunStatefulPrecompileFu
 	for _, function := range functions {
 		_, exists := contract.functions[string(function.selector)]
 		if exists {
-			panic(fmt.Errorf("cannot create stateful precompile with duplicated function selector: %q", function.selector))
+			return nil, fmt.Errorf("cannot create stateful precompile with duplicated function selector: %q", function.selector)
+		}
+		if function == nil {
```
aaronbuchwald, line 114: why do we add the nil check for the function? This should never happen if implemented properly and will panic later in the creation of the contract if an error is returned here, so I don't think this check is necessary.

precompile/fee_config_manager.go
```diff
@@ -231,6 +231,7 @@ func UnpackFeeConfigInput(input []byte) (commontype.FeeConfig, error) {
 		case blockGasCostStepKey:
 			feeConfig.BlockGasCostStep = new(big.Int).SetBytes(packedElement)
 		default:
+			// this should not ever happen. keep this as panic.
```
aaronbuchwald, line 234:
```suggestion
			// This should never encounter an unknown fee config key
```

precompile/fee_config_manager.go
```diff
@@ -260,6 +261,7 @@ func GetStoredFeeConfig(stateDB StateDB) commontype.FeeConfig {
 		case blockGasCostStepKey:
 			feeConfig.BlockGasCostStep = new(big.Int).Set(val.Big())
 		default:
+			// this should not ever happen. keep this as panic.
```
aaronbuchwald, line 264: Can we change all of the comments that say "keep this as panic" ? This sounds more like a PR review comment explaining why it is KEPT this way than as a comment that belongs permanently in the code and explains why the code IS this way.

precompile/fee_config_manager.go
```diff
@@ -298,6 +300,7 @@ func StoreFeeConfig(stateDB StateDB, feeConfig commontype.FeeConfig, blockContex
 		case blockGasCostStepKey:
 			input = common.BigToHash(feeConfig.BlockGasCostStep)
 		default:
+			// this should not ever happen. keep this as panic.
```
aaronbuchwald, line 303: same comment here

precompile/fee_config_manager.go
```diff
@@ -386,6 +389,11 @@ func createFeeConfigManagerPrecompile(precompileAddr common.Address) StatefulPre
 
 	feeConfigManagerFunctions = append(feeConfigManagerFunctions, setFeeConfigFunc, getFeeConfigFunc, getFeeConfigLastChangedAtFunc)
 	// Construct the contract with no fallback function.
-	contract := newStatefulPrecompileWithFunctionSelectors(nil, feeConfigManagerFunctions)
+	contract, err := NewStatefulPrecompileContract(nil, feeConfigManagerFunctions)
+	// Change this to be returned as an error after refactoring this precompile
```
aaronbuchwald, line 393:
```suggestion
	// TODO Change this to be returned as an error after refactoring this precompile
```

core/state_processor.go
ceyonur: added logs

miner/worker.go
ceyonur: done

precompile/contract.go
ceyonur: I thought this could be go well with the existing function check. removed it.

precompile/fee_config_manager.go
ceyonur: changed

precompile/fee_config_manager.go
ceyonur: changed

core/state_processor.go
```diff
@@ -78,7 +79,12 @@ func (p *StateProcessor) Process(block *types.Block, parent *types.Header, state
 	)
 
 	// Configure any stateful precompiles that should go into effect during this block.
-	p.config.CheckConfigurePrecompiles(new(big.Int).SetUint64(parent.Time), block, statedb)
+	err := p.config.ConfigurePrecompiles(new(big.Int).SetUint64(parent.Time), block, statedb)
+	if err != nil {
+		werr := fmt.Errorf("could not configure precompiles: %w", err)
+		log.Error("failed to process the state changes", "err", werr)
+		return nil, nil, 0, err
+	}
```
aaronbuchwald, line 87: Can we remove the wrapping of this error here? I think it would make sense to wrap this error inside of `ConfigurePrecompiles` if anywhere.

miner/worker.go
```diff
@@ -184,7 +184,12 @@ func (w *worker) commitNewWork() (*types.Block, error) {
 		return nil, fmt.Errorf("failed to create new current environment: %w", err)
 	}
 	// Configure any stateful precompiles that should go into effect during this block.
-	w.chainConfig.CheckConfigurePrecompiles(new(big.Int).SetUint64(parent.Time()), types.NewBlockWithHeader(header), env.state)
+	err = w.chainConfig.ConfigurePrecompiles(new(big.Int).SetUint64(parent.Time()), types.NewBlockWithHeader(header), env.state)
+	if err != nil {
+		werr := fmt.Errorf("failed to configure precompiles: %w", err)
```
aaronbuchwald, line 189: same comment here since we're using the same wrapping we can just move this into `ConfigurePrecompiles`

precompile/fee_config_manager.go
```diff
@@ -298,6 +300,7 @@ func StoreFeeConfig(stateDB StateDB, feeConfig commontype.FeeConfig, blockContex
 		case blockGasCostStepKey:
 			input = common.BigToHash(feeConfig.BlockGasCostStep)
 		default:
+			// this should not ever happen.
```
aaronbuchwald, line 303: can we make this comment match the others? fwiw I think we could get rid of these comments here since it should be obvious from the panic that this should not happen

ceyonur: I'd prefer to keep this, as we're panicking in default case which is not ideal for a "default" :D

miner/worker.go
ceyonur: changed

core/state_processor.go
ceyonur: done

precompile/utils.go
```diff
@@ -17,6 +17,7 @@ var functionSignatureRegex = regexp.MustCompile(`[\w]+\(((([\w]+)?)|((([\w]+),)+
 // CalculateFunctionSelector returns the 4 byte function selector that results from [functionSignature]
 // Ex. the function setBalance(addr address, balance uint256) should be passed in as the string:
 // "setBalance(address,uint256)"
+// TODO: remove this after moving to ABI based function selectors.
```
aaronbuchwald, line 20: Can we remove this TODO? I think this will still be useful to have

miner/worker.go
```diff
@@ -184,7 +184,11 @@ func (w *worker) commitNewWork() (*types.Block, error) {
 		return nil, fmt.Errorf("failed to create new current environment: %w", err)
 	}
 	// Configure any stateful precompiles that should go into effect during this block.
-	w.chainConfig.CheckConfigurePrecompiles(new(big.Int).SetUint64(parent.Time()), types.NewBlockWithHeader(header), env.state)
+	err = w.chainConfig.ConfigurePrecompiles(new(big.Int).SetUint64(parent.Time()), types.NewBlockWithHeader(header), env.state)
+	if err != nil {
+		log.Error("failed to commit new work", "err", err)
```
aaronbuchwald, line 189:
```suggestion
		log.Error("failed to configure precompiles mining new block", "hash", block.Hash(), "number", block.NumberU64(), "timestamp", block.Time(), "err", err)
```

core/state_processor.go
```diff
@@ -78,7 +79,11 @@ func (p *StateProcessor) Process(block *types.Block, parent *types.Header, state
 	)
 
 	// Configure any stateful precompiles that should go into effect during this block.
-	p.config.CheckConfigurePrecompiles(new(big.Int).SetUint64(parent.Time), block, statedb)
+	err := p.config.ConfigurePrecompiles(new(big.Int).SetUint64(parent.Time), block, statedb)
+	if err != nil {
+		log.Error("failed to process the state changes", "err", err)
```
aaronbuchwald, line 84:
```suggestion
		log.Error("failed to configure precompiles processing block", "hash", block.Hash(), "number", block.NumberU64(), "timestamp", block.Time(), "err", err)
```

aaronbuchwald (approved):
Few small comments and then this LGTM

PR #419, opened 2022-12-20
Add requirement that Subnet-evm upgrade is enabled at genesis

anusha-ctrl:
## Why this should be merged
- We should add a requirement that the SubnetEVM upgrade is enabled so that we can deprecate support for it being disabled. It should always be enabled from genesis.

## How this works
- We check the blockTimestamp at genesis and see if it is greater than or equal to `SubnetEVMTimestamp`. `SubnetEVMTimestamp` should be set to 0 so the SubnetEVM upgrade is enabled at genesis. 
- This should error if the `genesisJSON` did not include `subnetEVMTimestamp` or  `subnetEVMTimestamp` was set to some timestamp > 0
- We add a config flag so the user can skip this subnetevm upgrade check for the time being.  

Here is an example of the chain config
```
// SubnetEVMDefaultChainConfig is the default configuration
	SubnetEVMDefaultChainConfig = &ChainConfig{
		ChainID:            SubnetEVMChainID,
		FeeConfig:          DefaultFeeConfig,
		AllowFeeRecipients: false,

		HomesteadBlock:      big.NewInt(0),
		EIP150Block:         big.NewInt(0),
		EIP150Hash:          common.HexToHash("0x2086799aeebeae135c246c65021c82b4e15a2c451340993aacfd2751886514f0"),
		EIP155Block:         big.NewInt(0),
		EIP158Block:         big.NewInt(0),
		ByzantiumBlock:      big.NewInt(0),
		ConstantinopleBlock: big.NewInt(0),
		PetersburgBlock:     big.NewInt(0),
		IstanbulBlock:       big.NewInt(0),
		MuirGlacierBlock:    big.NewInt(0),

		NetworkUpgrades: NetworkUpgrades{
			SubnetEVMTimestamp: big.NewInt(0),
		},
	}
```

## How this was tested

- 2 unit tests were added


darioush (approved):

plugin/evm/vm_test.go
```diff
@@ -369,6 +370,44 @@ func issueAndAccept(t *testing.T, issuer <-chan engCommon.Message, vm *VM) snowm
...
+		[]*engCommon.Fx{},
+		nil,
+	)
+
+	// we expect an error when subnet evm upgrade is nil in chain config
+	require.Error(t, err)
+}
+
+func TestSubnetEVMUpgradeNotEnabledAtGenesis(t *testing.T) {
+	ctx, dbManager, genesisBytes, issuer := setupGenesis(t, genesisJSONSubnetEVMLateEnablement)
```
darioush, line 393: alternatively we can parse the json for `genesisJSONSubnetEVM` and modify the value in the test.

plugin/evm/vm.go
```diff
@@ -379,9 +379,15 @@ func (vm *VM) Initialize(
...
-	// create genesisHash after applying upgradeBytes in case
+	// create genesisBlock after applying upgradeBytes in case
 	// upgradeBytes modifies genesis.
-	vm.genesisHash = vm.ethConfig.Genesis.ToBlock(nil).Hash()
+	genesisBlock := vm.ethConfig.Genesis.ToBlock(nil)
+	vm.genesisHash = genesisBlock.Hash()
+
+	// check that subnetEVM upgrade is enabled from genesis
+	if !vm.chainConfig.IsSubnetEVM(genesisBlock.Number()) {
+		return fmt.Errorf("SubnetEVM upgrade is not enabled")
```
darioush, line 389:
```suggestion
		return fmt.Errorf("SubnetEVM upgrade is not enabled in genesis")
```

anusha-ctrl: thanks!

plugin/evm/vm.go
```diff
@@ -379,9 +379,15 @@ func (vm *VM) Initialize(
...
 
-	// create genesisHash after applying upgradeBytes in case
+	// create genesisBlock after applying upgradeBytes in case
 	// upgradeBytes modifies genesis.
-	vm.genesisHash = vm.ethConfig.Genesis.ToBlock(nil).Hash()
+	genesisBlock := vm.ethConfig.Genesis.ToBlock(nil)
+	vm.genesisHash = genesisBlock.Hash()
+
+	// check that subnetEVM upgrade is enabled from genesis
+	if !vm.chainConfig.IsSubnetEVM(genesisBlock.Number()) {
```
darioush, line 388: this should be timestamp or hardcoded to 0

anusha-ctrl: thanks

anusha-ctrl:
Need to fix some existing tests.  With this new guard in place, it should no longer be possible to abort an Subnet-EVM upgrade. It's all or nothing now.

plugin/evm/vm_upgrade_bytes_test.go
```diff
@@ -154,10 +154,10 @@ func TestVMUpgradeBytesNetworkUpgrades(t *testing.T) {
 
 	// Get a json specifying a Network upgrade at genesis
 	// to apply as upgradeBytes.
-	subnetEVMTimestamp := time.Unix(10, 0)
+	subnetEVMTimestamp := big.NewInt(0)
```
anusha-ctrl, line 157: subnetEVMTimestamp must be 0 to be activated at genesis

plugin/evm/vm_upgrade_bytes_test.go
```diff
@@ -240,17 +230,13 @@ func TestVMUpgradeBytesNetworkUpgradesWithGenesis(t *testing.T) {
...
-	assert.True(t, vm.chainConfig.IsSubnetEVM(big.NewInt(subnetEVMTimestamp.Unix())))
+	assert.True(t, vm.chainConfig.IsSubnetEVM(subnetEVMTimestamp))
 
 	if err := vm.Shutdown(context.Background()); err != nil {
 		t.Fatal(err)
 	}
 
-	// abort a fork specified in genesis
-	upgradeConfig.NetworkUpgrades.SubnetEVMTimestamp = nil
+	upgradeConfig.NetworkUpgrades = nil
```
anusha-ctrl, line 239: no longer testing abort, but now testing the fallback genesis bytes

plugin/evm/vm.go
```diff
@@ -379,9 +379,15 @@ func (vm *VM) Initialize(
...
-	// create genesisHash after applying upgradeBytes in case
+	// create genesisBlock after applying upgradeBytes in case
 	// upgradeBytes modifies genesis.
-	vm.genesisHash = vm.ethConfig.Genesis.ToBlock(nil).Hash()
+	genesisBlock := vm.ethConfig.Genesis.ToBlock(nil)
+	vm.genesisHash = genesisBlock.Hash()
+
+	// check that subnetEVM upgrade is enabled from genesis
+	if !vm.chainConfig.IsSubnetEVM(genesisBlock.Timestamp()) {
+		return fmt.Errorf("SubnetEVM upgrade is not enabled in genesis")
```
ceyonur, line 389: no need to `Errorf` as we don't use any directives. we can use `errors.new` in a global variable.

plugin/evm/vm_test.go
```diff
@@ -369,6 +370,44 @@ func issueAndAccept(t *testing.T, issuer <-chan engCommon.Message, vm *VM) snowm
...
+		genesisBytes,
+		[]byte(""),
+		[]byte(""),
+		issuer,
+		[]*engCommon.Fx{},
+		nil,
+	)
+
+	// we expect an error when subnet evm upgrade is not enabled at genesis and at a later block instead
+	require.Error(t, err)
```
ceyonur, line 408: can we test the exact error? (after moving the error to a global)

plugin/evm/vm_upgrade_bytes_test.go
```diff
@@ -227,10 +210,10 @@ func TestVMUpgradeBytesNetworkUpgradesWithGenesis(t *testing.T) {
 
 	// Get a json specifying a Network upgrade at genesis
 	// to apply as upgradeBytes.
-	subnetEVMTimestamp := time.Unix(10, 0)
+	subnetEVMTimestamp := big.NewInt(0)
```
ceyonur, line 213: nit: can we use `common.Big0` for these?

plugin/evm/vm_test.go
```diff
@@ -369,6 +370,44 @@ func issueAndAccept(t *testing.T, issuer <-chan engCommon.Message, vm *VM) snowm
...
+		issuer,
+		[]*engCommon.Fx{},
+		nil,
+	)
+
+	// we expect an error when subnet evm upgrade is nil in chain config
+	require.Error(t, err)
+}
+
+func TestSubnetEVMUpgradeNotEnabledAtGenesis(t *testing.T) {
```
ceyonur, line 392: can we merge these two test and use different test cases like 
```
for _, test := range tests{
t.Run(...
...
```

plugin/evm/vm_upgrade_bytes_test.go
```diff
@@ -189,36 +181,27 @@ func TestVMUpgradeBytesNetworkUpgrades(t *testing.T) {
...
-
-	// VM should not start if fork is moved forward
-	upgradeConfig.NetworkUpgrades.SubnetEVMTimestamp = big.NewInt(30)
-	upgradeBytesJSON, err = json.Marshal(upgradeConfig)
-	if err != nil {
-		t.Fatalf("could not marshal upgradeConfig to json: %s", err)
-	}
-	err = vm.Initialize(context.Background(), vm.ctx, dbManager, []byte(genesisJSONPreSubnetEVM), upgradeBytesJSON, []byte{}, issuer, []*common.Fx{}, appSender)
-	assert.ErrorContains(t, err, "mismatching SubnetEVM fork block timestamp in database")
+	assert.ErrorContains(t, err, "SubnetEVM upgrade is not enabled in genesis")
```
ceyonur, line 195: hmm these actually testing network upgrades, and enforcing subnet-evm to be activated at genesis leaves us no other testable network upgrade. Do you think it would be possible if we add a `testMode` check to vm and if so skip checking the subnet-evm activation at genesis? cc @aaronbuchwald @darioush

aaronbuchwald: I don't think that's worth it at this time. I'd prefer to re-add these tests back when there is a future network upgrade to include.

plugin/evm/vm_upgrade_bytes_test.go
aaronbuchwald: Can we require that this is set directly in the genesis rather than in the upgrade bytes to make this simpler?

plugin/evm/vm.go
aaronbuchwald, line 389: Can we require this is set in the genesisBytes and not allow the option to set it via the upgradeBytes to simplify?

anusha-ctrl: thanks

plugin/evm/vm_test.go
anusha-ctrl: sure

plugin/evm/vm_test.go
anusha-ctrl: thanks

plugin/evm/vm_upgrade_bytes_test.go
anusha-ctrl: sure

anusha-ctrl: thanks

plugin/evm/vm.go
anusha-ctrl: yes

plugin/evm/vm_upgrade_bytes_test.go
anusha-ctrl: I'll actually revert this back to the original and comment them out. We can uncomment these tests when a new NetworkUpgrade is added

ceyonur (approved):
LGTM!

plugin/evm/vm_upgrade_bytes_test.go
```diff
@@ -145,125 +144,128 @@ func TestVMUpgradeBytesPrecompile(t *testing.T) {
...
-
-	if err := vm.Shutdown(context.Background()); err != nil {
-		t.Fatal(err)
-	}
-}
+// Commenting out the NetworkUpgrade Tests until we get a new NetworkUpgrade.
+// The existing and only NetworkUpgrade that exists is SubnetEVM Upgrade which is now required at genesis.
+
+// func TestVMUpgradeBytesNetworkUpgrades(t *testing.T) {
+// 	// Hack: registering metrics uses global variables, so we need to disable metrics here so that we can initialize the VM twice.
```
aaronbuchwald, line 151: Can we switch from commenting these out to using
```
t.Skip("no non-required network upgrades to test network upgrade abort")
```

aaronbuchwald (approved):
One small requested change to skipping the SubnetEVM network upgrade tests and then this LGTM

plugin/evm/vm_upgrade_bytes_test.go
anusha-ctrl: yes

aaronbuchwald (approved):
LGTM

ceyonur (commented):
code generally lgtm. I just wonder if we should add a flag to avago-chain config to skip this check? We can emphasize this is just for testing purposes and should not be used in producion. it can also still be used for chains that are bot upgraded to subnet-evm.

plugin/evm/vm.go
```diff
@@ -369,6 +370,11 @@ func (vm *VM) Initialize(
 	vm.chainConfig = g.Config
 	vm.networkID = vm.ethConfig.NetworkId
 
+	// check that subnetEVM upgrade is enabled from genesis before upgradeBytes
+	if !vm.chainConfig.IsSubnetEVM(common.Big0) {
+		return errSubnetEVMUpgradeNotEnabled
+	}
+
```
ceyonur, line 377: Shall we move this below to` upgradeBytes` parsing, in case this was activated by a upgrade at genesis? Could this be a possibility? @aaronbuchwald @darioush

aaronbuchwald: They could include it in the upgrade bytes to activate at timestamp 0, but it SHOULD always be included in the genesis, so I think it's preferable to error here rather than after applying the `upgradeBytes`. I actually asked @anusha-ctrl to make this change to put it here.

aaronbuchwald:
> code generally lgtm. I just wonder if we should add a flag to avago-chain config to skip this check? We can emphasize this is just for testing purposes and should not be used in producion. it can also still be used for chains that are bot upgraded to subnet-evm.

Good idea, this way we do not need to create a new release if anyone runs into this unexpectedly. I confirmed in https://github.com/ava-labs/public-chain-assets/ that none of the networks here should be impacted.

@anusha-ctrl could you add a config flag that will allow a user to override this check in case anyone runs into it along with a TODO to remove once we confirm that there are no complaints after making this change?

plugin/evm/vm.go
ceyonur: just thinking someone might want to fix their genesis via this method (or by removing the checks via config flag).

ceyonur (approved):
Nice 👍

plugin/evm/vm_upgrade_bytes_test.go
```diff
@@ -146,6 +146,9 @@ func TestVMUpgradeBytesPrecompile(t *testing.T) {
 }
 
 func TestVMUpgradeBytesNetworkUpgrades(t *testing.T) {
+	// Skipping the NetworkUpgrade Tests until we get a new NetworkUpgrade.
+	// The existing and only NetworkUpgrade that exists is SubnetEVM Upgrade which is now required at genesis.
+	t.Skip("no non-required network upgrades to test network upgrade abort")
```
darioush, line 151: could we re-enable these tests using the new config flag override?

anusha-ctrl: done

plugin/evm/vm_test.go
```diff
@@ -376,6 +377,73 @@ func issueAndAccept(t *testing.T, issuer <-chan engCommon.Message, vm *VM) snowm
...
+		require.ErrorIs(t, err, errSubnetEVMUpgradeNotEnabled)
+	}
+}
+
+func TestOverrideSkipSubnetEVMUpgrade(t *testing.T) {
+	genesisTests := []struct {
+		genesisJSON string
+	}{
+		{
+			// we expect an error when subnet evm upgrade is nil in chain config
```
darioush, line 418: can we update these comments or merge the 2 tests by adding `configJSON` and `expectedErr` in the test struct?

anusha-ctrl: yes

darioush (approved):

aaronbuchwald (approved):
LGTM

PR #421, opened 2022-12-30
Refactor limit order

fdgod09:
## Why this should be merged
I have moved out limit order transaction related logic in a separate file. 
I was working on writing tests for limit_order.go. 
For example in testing RunMatchingEngine() when we find matching limit_orders it makes a call to orderBookContract to execute those orders. The tests then need the correct setup for making those transaction including creating those accounts and adding gas.
We want to test only if the matching logic is working as we can test the code for making orderbook contract call separately. Here i can now start using mocks and check if the correct calls are being when matching orders are found and are not being made when no limit orders match.
Similarly for parsing blocks and storing limit_orders we can test each getting blocks and parsing it separately.
This will help me write better unit tests.
If we also want to add integration tests which test whole flow(we have one in OrderBook.ts) we can add those also separately.

## How this was tested
It was tested using integration test in orderbook.ts.


fdgod09:
Sorry my bad. Again raised a PR here instead of our forked repo 😢

aaronbuchwald:
No worries, cool stuff

PR #426, opened 2023-01-05
Generalized upgrades

ceyonur:


precompile/rewardmanager/config.go
```diff
@@ -150,3 +162,20 @@ func (c *RewardManagerConfig) String() string {
...
+}
+
+// UnmarshalJSON implements the json.Unmarshaler interface.
+func (c *RewardManagerConfig) UnmarshalJSON(b []byte) error {
+	type Alias RewardManagerConfig
+	if err := json.Unmarshal(b, (*Alias)(c)); err != nil {
+		return err
+	}
+	return nil
+}
```
darioush, line 181: can we add a comment that explains this pattern?

precompile/stateful_precompile_config.go
```diff
@@ -4,16 +4,28 @@
...
+	// Contract returns a thread-safe singleton that can be used as the StatefulPrecompiledContract when
+	// this config is enabled.
+	Contract() StatefulPrecompiledContract
+	// Key returns the unique key for the stateful precompile.
+	Key() string
+	// New returns a new instance of the stateful precompile config.
+	New() StatefulPrecompileConfig
+}
+
+// StatefulPrecompileConfig defines the interface for a stateful precompile to
```
darioush, line 26: could we complete the sentence? seems it was left half way before

darioush (commented):
I really like some of the simplifications this change affords.

- Could we make any improvements to the unmarshalling situation?
- Could we avoid making the module type by having RegisterModule take Key, Address, and Contract for example?

PR #434, opened 2023-01-12
Generalized upgrades rb

ceyonur:
## Why this should be merged

Removes static precompile types in PrecompileUpgrade struct to reduce manual code addition for new precompiles. Also flattens the PrecompileUpgrade struct as it is treated in the codebase. 

This PR also covers some other cleanups/refactors:
* Removes big `stateful_precompile_test.go` and splits them into related-precompile files under `core/contractstatefultest` test package
* Uses Precompile Config Keys to fetch/determine if a precompile config is activated, rather than the contract address.
* Bunch of renaming
* Removes unneccessary `String` from precompile interface

## How this works
Introduces a precompile module interface and manager that can register module types that can contain address and a config key.
 
Introduces custom unmarshaler/marshaler to ChainConfig and PrecompileUpgrade to deal with json maps from keys to precompile configs. 

## How this was tested

Added new unit tests for ChainConfig/PrecompileUpgrade unmarshaler/marshalers. Existing unit tests should cover other changes.


accounts/abi/bind/precompile_config_template.go
```diff
@@ -14,14 +14,14 @@ const tmplSourcePrecompileConfigGo = `
 // For testing take a look at other precompile tests in core/stateful_precompile_test.go and config_test.go in other precompile folders.
 
 /* General guidelines for precompile development:
-1- Read the comment and set a suitable contract address in precompile/params.go. E.g:
-	{{.Contract.Type}}Address = common.HexToAddress("ASUITABLEHEXADDRESS")
-2- Set gas costs in contract.go
+1- Read the comment and set a suitable contract address in generated contract.go. E.g:
+	ContractAddress = common.HexToAddress("ASUITABLEHEXADDRESS")
+2- Set gas costs in generated contract.go
 3- It is recommended to only modify code in the highlighted areas marked with "CUSTOM CODE STARTS HERE". Modifying code outside of these areas should be done with caution and with a deep understanding of how these changes may impact the EVM.
```
anusha-ctrl, line 20: One thing might be helpful is to move set gas costs to one of the later steps, or right after step 3. Most people don't know how many reads or writes they'll do until after they implement their precompile function.

accounts/abi/bind/precompile_contract_template.go
```diff
@@ -77,14 +79,16 @@ const (
 var (
 	_ = errors.New
 	_ = big.NewInt
-	_ = strings.NewReader
-	_ = fmt.Printf
-	_ = json.Unmarshal
 )
```
anusha-ctrl, line 82: Nice

core/vm/contracts.go
```diff
@@ -158,12 +158,13 @@ func init() {
...
-			panic(fmt.Errorf("precompile address collides with existing native address: %s", k))
+	for _, module := range precompile.RegisteredModules() {
+		address := module.Address()
+		if _, ok := PrecompileAllNativeAddresses[address]; ok {
+			panic(fmt.Errorf("precompile address collides with existing native address: %s", address))
 		}
-		if k == constants.BlackholeAddr {
-			panic(fmt.Errorf("cannot use address %s for stateful precompile - overlaps with blackhole address", k))
+		if address == constants.BlackholeAddr {
+			panic(fmt.Errorf("cannot use address %s for stateful precompile - overlaps with blackhole address", address))
```
aaronbuchwald, line 167: Should we add a check that none of the registered modules have a conflicting address with each other?

core/vm/contractstatefultests/allow_list_test.go
```diff
@@ -0,0 +1,284 @@
+// (c) 2019-2020, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package contractstatefultests
```
aaronbuchwald, line 4: Can we rename this something shorter?

core/vm/contractstatefultests/allow_list_test.go
```diff
@@ -0,0 +1,284 @@
...
+	"github.com/ava-labs/subnet-evm/core/state"
+	"github.com/ava-labs/subnet-evm/precompile"
+	"github.com/ava-labs/subnet-evm/precompile/allowlist"
+	"github.com/ava-labs/subnet-evm/vmerrs"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/stretchr/testify/require"
+)
+
+func TestAllowListRun(t *testing.T) {
+	type test struct {
```
aaronbuchwald, line 20: Can we share the `test` type across all of the different precompile tests?

core/vm/evm.go
```diff
@@ -508,7 +508,7 @@ func (evm *EVM) create(caller ContractRef, codeAndHash *codeAndHash, gas uint64,
 		return nil, common.Address{}, 0, vmerrs.ErrContractAddressCollision
 	}
 	// If the allow list is enabled, check that [evm.TxContext.Origin] has permission to deploy a contract.
-	if evm.chainRules.IsPrecompileEnabled(precompile.ContractDeployerAllowListAddress) {
+	if evm.chainRules.IsPrecompileEnabled(deployerallowlist.ContractAddress) {
```
aaronbuchwald, line 511: It seems like we are using the name `deployerallowlist.ContractAddress` here but `feemanager.ConfigKey` in `core/tx_pool.go`. Could we standardize this?

params/precompile_modules.go
```diff
@@ -0,0 +1,43 @@
...
+// {YourPrecompile}Address       = common.HexToAddress("0x03000000000000000000000000000000000000??")
+func init() {
+	// Order matters here.
+	precompile.RegisterModule(deployerallowlist.ContractDeployerAllowListConfig{})
+	precompile.RegisterModule(nativeminter.ContractNativeMinterConfig{})
+	precompile.RegisterModule(txallowlist.TxAllowListConfig{})
+	precompile.RegisterModule(feemanager.FeeManagerConfig{})
+	precompile.RegisterModule(rewardmanager.RewardManagerConfig{})
+	// ADD YOUR PRECOMPILE HERE
+	// precompile.RegisterModule({yourPrecompilePackage}.{YourPrecompile}Config{})
```
aaronbuchwald, line 42: How would a user register a module in the address space `0x03...` ? Can we add support for this with a different function that starts in a different range?

accounts/abi/bind/precompile_config_template.go
ceyonur: done

core/vm/contracts.go
ceyonur: we do that here
https://github.com/ava-labs/subnet-evm/blob/6039731200a77a27ea66e872cf86a7e594e403d2/precompile/stateful_precompile_module.go#L70-L74

params/precompile_modules.go
ceyonur: Contract addresses are defined in their relative packages (not registered through this). The `StatefulPrecompileModule` interface has `Address()` function which returns those constant addresses in the packages.

If we prefer to register those addresses here then we would need to change how we create singleton contracts (should take contract address as param), but certainly possible.

core/vm/evm.go
ceyonur: hmm I think AWM stuff needs this to be a contract address for predicates (rather than the config keys). We should probably revert this to use contract addresses.

core/vm/contractstatefultests/allow_list_test.go
ceyonur: done

core/vm/contractstatefultests/allow_list_test.go
ceyonur: moved them to external tests/ package

core/vm/evm.go
ceyonur: reverted to use contract addresses

params/precompile_config.go
```diff
@@ -276,22 +255,23 @@ func (c *ChainConfig) EnabledStatefulPrecompiles(blockTimestamp *big.Int) []prec
 // - during block processing to update the state before processing the given block.
 func (c *ChainConfig) ConfigurePrecompiles(parentTimestamp *big.Int, blockContext precompile.BlockContext, statedb precompile.StateDB) error {
 	blockTimestamp := blockContext.Timestamp()
-	for _, address := range precompile.UsedAddresses { // Note: configure precompiles in a deterministic order.
-		for _, config := range c.getActivatingPrecompileConfigs(parentTimestamp, blockTimestamp, address, c.PrecompileUpgrades) {
+	for _, module := range precompile.RegisteredModules() { // Note: configure precompiles in a deterministic order.
```
darioush, line 258:
```suggestion
	for _, module := range precompile.RegisteredModules() { // Note: precompiles are configured in a deterministic order.
```

darioush:
Looks good, left a few minor nits.

precompile/utils.go
```diff
@@ -12,8 +13,26 @@ import (
...
+
 var functionSignatureRegex = regexp.MustCompile(`[\w]+\(((([\w]+)?)|((([\w]+),)+([\w]+)))\)`)
 
+// AddressRange represents a continuous range of addresses
+type AddressRange struct {
+	Start common.Address
+	End   common.Address
+}
+
+// Contains returns true iff [addr] is contained within the (inclusive)
```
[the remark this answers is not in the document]
ceyonur: done

precompile/txallowlist/config.go
```diff
@@ -45,17 +48,17 @@ func NewDisableTxAllowListConfig(blockTimestamp *big.Int) *TxAllowListConfig {
 }
 
 // Address returns the address of the contract deployer allow list.
-func (c *TxAllowListConfig) Address() common.Address {
-	return precompile.TxAllowListAddress
+func (c TxAllowListConfig) Address() common.Address {
```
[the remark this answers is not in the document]
ceyonur: reorganized them

precompile/nativeminter/contract.go
```diff
@@ -83,8 +85,8 @@ func mintNativeCoin(accessibleState precompile.PrecompileAccessibleState, caller
 	}
 
 	stateDB := accessibleState.GetStateDB()
-	// Verify that the caller is in the allow list
-	callerStatus := allowlist.GetAllowListStatus(stateDB, precompile.ContractNativeMinterAddress, caller)
+	// Verify that the caller is in the allow list and therefore has the right to modify it
```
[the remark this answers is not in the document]
ceyonur: yep, done

accounts/abi/bind/precompile_contract_template.go
```diff
@@ -20,23 +20,23 @@ type tmplPrecompileContract struct {
...
 // Code generated
-// This file is a generated precompile contract with stubbed abstract functions.
+// This file is a generated precompile contract config with stubbed abstract functions.
 // The file is generated by a template. Please inspect every code and comment in this file before use.
 
 // There are some must-be-done changes waiting in the file. Each area requiring you to add your code is marked with CUSTOM CODE to make them easy to find and modify.
 // Additionally there are other files you need to edit to activate your precompile.
 // These areas are highlighted with comments "ADD YOUR PRECOMPILE HERE".
-// For testing take a look at other precompile tests in core/stateful_precompile_test.go
+// For testing take a look at other precompile tests in core/stateful_precompile_test.go and config_test.go in other precompile folders.
```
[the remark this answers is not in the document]
ceyonur: fixed

internal/ethapi/api.go
```diff
@@ -233,7 +234,7 @@ func (s *TxPoolAPI) Inspect() map[string]map[string]map[string]string {
 	pending, queue := s.b.TxPoolContent()
 
 	// Define a formatter to flatten a transaction into a string
-	var format = func(tx *types.Transaction) string {
+	format := func(tx *types.Transaction) string {
```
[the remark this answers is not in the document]
ceyonur: reverted

internal/ethapi/api.go
```diff
@@ -1981,11 +1988,11 @@ func (s *TransactionAPI) Resend(ctx context.Context, sendArgs TransactionArgs, g
 	matchTx := sendArgs.toTransaction()
 
 	// Before replacing the old transaction, ensure the _new_ transaction fee is reasonable.
-	var price = matchTx.GasPrice()
```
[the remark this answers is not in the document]
ceyonur: reverted

params/config.go
```diff
@@ -157,9 +157,84 @@ type ChainConfig struct {
...
+	NetworkUpgrades                        // Config for timestamps that enable avalanche network upgrades
+	Precompiles     ChainConfigPrecompiles `json:"-"` // Config for enabling precompiles from genesis. JSON encode/decode will be handled by the custom marshaler/unmarshaler.
+	UpgradeConfig   `json:"-"`             // Config specified in upgradeBytes (avalanche network upgrades or enable/disabling precompiles). Skip encoding/decoding directly into ChainConfig.
+}
+
+type ChainConfigPrecompiles map[string]precompile.StatefulPrecompileConfig
+
+// The Precompiles field is a map of precompile module keys to their
+// configuration.
+func (ccp *ChainConfigPrecompiles) UnmarshalJSON(data []byte) error {
```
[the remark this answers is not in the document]
ceyonur: done

tests/statefulprecompiles/tx_allowlist_test.go
```diff
@@ -0,0 +1,199 @@
+// (c) 2019-2020, Ava Labs, Inc. All rights reserved.
```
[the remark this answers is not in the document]
ceyonur: like tests/statefulprecompiles/{precompilename}/test.go? or directly tests/{precompilename}/test.go?

params/precompile_modules.go
```diff
@@ -0,0 +1,43 @@
...
+// in the future.
+// ContractDeployerAllowListAddress = common.HexToAddress("0x0200000000000000000000000000000000000000")
+// ContractNativeMinterAddress      = common.HexToAddress("0x0200000000000000000000000000000000000001")
+// TxAllowListAddress               = common.HexToAddress("0x0200000000000000000000000000000000000002")
+// FeeManagerAddress         				= common.HexToAddress("0x0200000000000000000000000000000000000003")
+// RewardManagerAddress             = common.HexToAddress("0x0200000000000000000000000000000000000004")
+// ADD YOUR PRECOMPILE HERE
+// {YourPrecompile}Address       = common.HexToAddress("0x03000000000000000000000000000000000000??")
+func init() {
+	// Order matters here.
```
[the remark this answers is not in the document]
aaronbuchwald: As a result of not needing the order to matter, I believe that we can do:

```go
precompile.RegisterModule(NewModule())
```

in each individual precompile package and they will register themselves.

Then we can change this file to ensure that each precompile gets imported and has an opportunity to call `init()` without requiring that it does so in a deterministic order.

I think keeping that here would make perfect sense and we should be able to have the code in this file look like:

```go
import _ "github.com/ava-labs/subnet-evm/precompile/deployerallowlist"
```

for each individual precompile right? This is the strategy used with the tracer registration, where each tracer registers itself https://github.com/ava-labs/subnet-evm/blob/master/eth/tracers/native/4byte.go#L42

precompile/allowlist/config_test.go
```diff
@@ -0,0 +1,97 @@
...
+func TestVerifyAllowlistConfig(t *testing.T) {
+	admins := []common.Address{{1}}
+	enableds := []common.Address{{2}}
+	tests := []struct {
+		name          string
+		config        AllowListConfig
+		expectedError string
+	}{
+		{
+			name:          "invalid allow list config in tx allowlist",
```
aaronbuchwald, line 22: Can we remove tx allowlist from the names of these tests?

accounts/abi/bind/precompile_config_template.go
```diff
@@ -32,25 +32,29 @@ Typically, custom codes are required in only those areas.
 
 package {{.Package}}
 
+{{$contract := .Contract}}
```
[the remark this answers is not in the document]
ceyonur: we need it when conditionally importing allowlist

params/config_test.go
```diff
@@ -136,3 +142,62 @@ func TestCheckCompatible(t *testing.T) {
...
+	require.True(contractNativeMinterConf.Equal(testContractNativeMinterConfig))
+
+	// Marshal and unmarshal again and check that the result is the same
+	marshaled, err := json.Marshal(c)
+	require.NoError(err)
+	c2 := ChainConfig{}
+	err = json.Unmarshal(marshaled, &c2)
+	require.NoError(err)
+	require.Equal(c, c2)
+}
```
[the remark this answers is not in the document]
ceyonur: they won't be the same as `EIP150Hash` is not pointer and will be defaulted to empty hash in marshalling like :
`"eip150Hash":"0x0000000000000000000000000000000000000000000000000000000000000000"`
plus ordering will not be the same

params/precompile_config.go
```diff
@@ -61,153 +66,127 @@ func (p *PrecompileUpgrade) getByAddress(address common.Address) (precompile.Sta
...
+
+	// verify genesis precompiles
+	for key, config := range c.Precompiles {
+		if err := config.Verify(); err != nil {
+			return err
 		}
-		if !hasKey {
-			return fmt.Errorf("empty precompile upgrade at index %d", i)
+		// check the genesis chain config for any enabled upgrade
+		lastUpgradeMap[key] = lastUpgradeData{
```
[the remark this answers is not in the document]
ceyonur: we did not have it before, still have no checks for that. but we should be checking if the timestamp is nil or not to decide it was disabled. I think @darioush added a check for that here https://github.com/ava-labs/subnet-evm/pull/474/files

params/precompile_modules.go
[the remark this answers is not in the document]
ceyonur: As much as I want to move these to their specific packages, it won't be very reliable and debugable. The reason is we also import them in evm packages (for fee manager and reward manager). So this means order of imports would matter here which could easily break in other packages.

plugin/evm/vm_upgrade_bytes_test.go
```diff
@@ -28,7 +28,7 @@ func TestVMUpgradeBytesPrecompile(t *testing.T) {
 	upgradeConfig := &params.UpgradeConfig{
 		PrecompileUpgrades: []params.PrecompileUpgrade{
 			{
-				TxAllowListConfig: txallowlist.NewTxAllowListConfig(big.NewInt(enableAllowListTimestamp.Unix()), testEthAddrs[0:1], nil),
+				StatefulPrecompileConfig: txallowlist.NewTxAllowListConfig(big.NewInt(enableAllowListTimestamp.Unix()), testEthAddrs[0:1], nil),
```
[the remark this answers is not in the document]
ceyonur: my linter works really weird, it gives error for this file but not for the other. I removed it from here, if Ci linter gives a warning I can add keys to every other place.

tests/statefulprecompiles/allow_list_test.go
```diff
@@ -0,0 +1,284 @@
...
+		suppliedGas uint64
+		readOnly    bool
+
+		expectedRes []byte
+		expectedErr string
+
+		config *allowlist.AllowListConfig
+
+		assertState func(t *testing.T, state *state.StateDB)
+	}
```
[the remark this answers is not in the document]
ceyonur: AllowListConfig does not implement StatefulPrecompileConfig.
Currently allow_list_test tests it's own functionality and other precompiles do not need to test them in their files. 
We can complicate these UTs a little bit to use same test suite and run same tests for every precompile. It would be some repetitive work and would test the same code multiple times. Currently I removed those repeated tests from other files and only kept them in contract/tx allow lists since they do not have any other tests.

accounts/abi/bind/precompile_config_template.go
```diff
@@ -14,14 +14,14 @@ const tmplSourcePrecompileConfigGo = `
...
+	ContractAddress = common.HexToAddress("ASUITABLEHEXADDRESS")
+2- It is recommended to only modify code in the highlighted areas marked with "CUSTOM CODE STARTS HERE". Modifying code outside of these areas should be done with caution and with a deep understanding of how these changes may impact the EVM.
 Typically, custom codes are required in only those areas.
-4- Add your upgradable config in params/precompile_config.go
-5- Add your precompile upgrade in params/config.go
-6- Add your config unit test in {generatedpkg}/config_test.go
+3- Set gas costs in generated contract.go
+4- Register your precompile module in params/precompile_modules.go
+5- Add your config unit tests under generated package config_test.go
+6- Add your contract unit tests under core/vm/contractstatefultests/{precompilename}_test.go
```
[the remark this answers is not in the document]
ceyonur: done

accounts/abi/bind/precompile_contract_template.go
```diff
@@ -20,23 +20,23 @@ type tmplPrecompileContract struct {
...
+	ContractAddress = common.HexToAddress("ASUITABLEHEXADDRESS")
+2- Set gas costs in generated contract.go
 3- It is recommended to only modify code in the highlighted areas marked with "CUSTOM CODE STARTS HERE". Modifying code outside of these areas should be done with caution and with a deep understanding of how these changes may impact the EVM.
 Typically, custom codes are required in only those areas.
-4- Add your upgradable config in params/precompile_config.go
-5- Add your precompile upgrade in params/config.go
-6- Add your config unit test in {generatedpkg}/config_test.go
+4- Register your precompile module in params/precompile_modules.go
+5- Add your config unit tests under generated package config_test.go
+6- Add your contract unit tests under core/vm/contractstatefultests/{precompilename}_test.go
```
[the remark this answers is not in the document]
ceyonur: updated

accounts/abi/bind/precompile_contract_template.go
```diff
@@ -142,26 +142,26 @@ func init() {
...
 	if err != nil {
 		panic(err)
 	}
 }
 
 {{if .Contract.AllowList}}
 // Get{{.Contract.Type}}AllowListStatus returns the role of [address] for the {{.Contract.Type}} list.
-func Get{{.Contract.Type}}AllowListStatus(stateDB precompile.StateDB, address common.Address) precompile.AllowListRole {
-	return precompile.GetAllowListStatus(stateDB, {{.Contract.Type}}Address, address)
+func Get{{.Contract.Type}}AllowListStatus(stateDB precompile.StateDB, address common.Address) allowlist.AllowListRole {
```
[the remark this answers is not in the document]
ceyonur: changed

params/config.go
```diff
@@ -157,9 +157,85 @@ type ChainConfig struct {
...
 
-	NetworkUpgrades              // Config for timestamps that enable avalanche network upgrades
-	PrecompileUpgrade            // Config for enabling precompiles from genesis
-	UpgradeConfig     `json:"-"` // Config specified in upgradeBytes (avalanche network upgrades or enable/disabling precompiles). Skip encoding/decoding directly into ChainConfig.
+	NetworkUpgrades                        // Config for timestamps that enable avalanche network upgrades
+	Precompiles     ChainConfigPrecompiles `json:"-"` // Config for enabling precompiles from genesis. JSON encode/decode will be handled by the custom marshaler/unmarshaler.
+	UpgradeConfig   `json:"-"`             // Config specified in upgradeBytes (avalanche network upgrades or enable/disabling precompiles). Skip encoding/decoding directly into ChainConfig.
+}
+
+type ChainConfigPrecompiles map[string]precompile.StatefulPrecompileConfig
```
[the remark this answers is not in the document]
ceyonur: moved

params/config_test.go
```diff
@@ -136,3 +142,62 @@ func TestCheckCompatible(t *testing.T) {
...
+	testRewardManagerConfig := rewardmanager.NewRewardManagerConfig(
+		big.NewInt(1671542573),
+		[]common.Address{common.HexToAddress("0x8db97C7cEcE249c2b98bDC0226Cc4C2A57BF52FC")},
+		nil,
+		&rewardmanager.InitialRewardConfig{
+			AllowFeeRecipients: true,
+		})
+	require.True(rewardManagerConf.Equal(testRewardManagerConfig))
+
+	contractNativeMinterConf := c.Precompiles[nativeminter.ConfigKey]
```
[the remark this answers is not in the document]
ceyonur: done

params/precompile_config.go
```diff
@@ -4,54 +4,59 @@
...
 // as a network upgrade.
 type PrecompileUpgrade struct {
-	ContractDeployerAllowListConfig *deployerallowlist.ContractDeployerAllowListConfig `json:"contractDeployerAllowListConfig,omitempty"` // Config for the contract deployer allow list precompile
-	ContractNativeMinterConfig      *nativeminter.ContractNativeMinterConfig           `json:"contractNativeMinterConfig,omitempty"`      // Config for the native minter precompile
-	TxAllowListConfig               *txallowlist.TxAllowListConfig                     `json:"txAllowListConfig,omitempty"`               // Config for the tx allow list precompile
-	FeeManagerConfig                *feemanager.FeeManagerConfig                       `json:"feeManagerConfig,omitempty"`                // Config for the fee manager precompile
-	RewardManagerConfig             *rewardmanager.RewardManagerConfig                 `json:"rewardManagerConfig,omitempty"`             // Config for the reward manager precompile
-	// ADD YOUR PRECOMPILE HERE
-	// {YourPrecompile}Config  *precompile.{YourPrecompile}Config `json:"{yourPrecompile}Config,omitempty"`
+	precompile.StatefulPrecompileConfig
```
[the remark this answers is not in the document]
ceyonur: done

params/precompile_config.go
```diff
@@ -4,54 +4,59 @@
...
+// based on the key. Keys are defined in each precompile module, and registered in
+// params/precompile_modules.go.
+// Ex: {"feeManagerConfig": {...}} where "feeManagerConfig" is the key
+func (u *PrecompileUpgrade) UnmarshalJSON(data []byte) error {
+	raw := make(map[string]json.RawMessage)
+	if err := json.Unmarshal(data, &raw); err != nil {
+		return err
+	}
+	if len(raw) != 1 {
+		return errMultipleKeys
```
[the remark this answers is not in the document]
ceyonur: I think the err message "PrecompileUpgrade must have exactly one key"  but added empty key err

params/precompile_config.go
```diff
@@ -61,153 +66,127 @@ func (p *PrecompileUpgrade) getByAddress(address common.Address) (precompile.Sta
...
+	// next range over upgrades to verify correct use of disabled and blockTimestamps.
+	var lastBlockTimestamp *big.Int
+	for i, upgrade := range c.PrecompileUpgrades {
+		key := upgrade.Key()
+
+		lastUpgrade, ok := lastUpgradeMap[key]
 		var (
-			lastUpgraded *big.Int
 			disabled     bool
+			lastUpgraded *big.Int
```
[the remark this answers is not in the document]
ceyonur: changed them

params/precompile_config.go
```diff
@@ -61,153 +66,127 @@ func (p *PrecompileUpgrade) getByAddress(address common.Address) (precompile.Sta
...
+		// Note: It is OK for multiple configs of different keys to specify the same timestamp.
+		if lastBlockTimestamp != nil && upgradeTimestamp.Cmp(lastBlockTimestamp) < 0 {
+			return fmt.Errorf("PrecompileUpgrades[%d] config timestamp (%v) < previous timestamp (%v)", i, upgradeTimestamp, lastBlockTimestamp)
+		}
 
-			if err := config.Verify(); err != nil {
-				return err
-			}
+		if disabled == upgrade.IsDisabled() {
+			return fmt.Errorf("PrecompileUpgrades[%d] disable should be [%v]", i, !disabled)
```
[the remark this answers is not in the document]
ceyonur: very good idea!

params/precompile_config.go
```diff
@@ -61,153 +66,127 @@ func (p *PrecompileUpgrade) getByAddress(address common.Address) (precompile.Sta
...
+	if !ok {
+		return configs
+	}
+
+	key := module.Key()
+
 	// First check the embedded [upgrade] for precompiles configured
 	// in the genesis chain config.
-	if config, ok := c.PrecompileUpgrade.getByAddress(address); ok {
+	if config, ok := c.Precompiles[key]; ok {
```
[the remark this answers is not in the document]
ceyonur: yep good idea

params/precompile_config.go
```diff
@@ -276,22 +255,23 @@ func (c *ChainConfig) EnabledStatefulPrecompiles(blockTimestamp *big.Int) []prec
 // - during block processing to update the state before processing the given block.
 func (c *ChainConfig) ConfigurePrecompiles(parentTimestamp *big.Int, blockContext precompile.BlockContext, statedb precompile.StateDB) error {
 	blockTimestamp := blockContext.Timestamp()
-	for _, address := range precompile.UsedAddresses { // Note: configure precompiles in a deterministic order.
-		for _, config := range c.getActivatingPrecompileConfigs(parentTimestamp, blockTimestamp, address, c.PrecompileUpgrades) {
+	for _, module := range precompile.RegisteredModules() { // Note: precompiles are configured in a deterministic order.
```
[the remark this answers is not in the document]
ceyonur: updated

precompile/stateful_precompile_module.go
```diff
@@ -0,0 +1,99 @@
...
+	for _, stm := range registeredModules {
+		if stm.Address() == address {
+			return stm, true
+		}
+	}
+
+	return nil, false
+}
+
+func GetPrecompileModule(name string) (StatefulPrecompileModule, bool) {
```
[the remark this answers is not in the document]
ceyonur: changed to key

precompile/allowlist/config_test.go
ceyonur: removed

precompile/stateful_precompile_module.go
```diff
@@ -0,0 +1,99 @@
...
+
+	return nil
+}
+
+func GetPrecompileModuleByAddress(address common.Address) (StatefulPrecompileModule, bool) {
+	for _, stm := range registeredModules {
+		if stm.Address() == address {
+			return stm, true
+		}
+	}
```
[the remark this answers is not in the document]
ceyonur: hmm when I added this we were using keys for lookups in active precompiles, but since we now switched to use addresses I think we should index them by addresses

params/precompile_modules.go
aaronbuchwald: I see, I think this is a better approach as it currently is.

tests/statefulprecompiles/allow_list_test.go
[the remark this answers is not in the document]
aaronbuchwald: I think it would be worth it to duplicate the tests to ensure that the behavior stays the same for the allow list across each one.

Since this is very important to ensure the behavior stays consistent, I think it's worthwhile. Should this be a follow up PR or GitHub issue to follow up with?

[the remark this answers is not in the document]
ceyonur: I prefer a follow up not to block this further. Created:https://github.com/ava-labs/subnet-evm/issues/497

core/genesis_test.go
```diff
@@ -270,8 +271,7 @@ func TestPrecompileActivationAfterHeaderBlock(t *testing.T) {
 	contractDeployerConfig := deployerallowlist.NewContractDeployerAllowListConfig(big.NewInt(51), nil, nil)
 	activatedGenesis.Config.UpgradeConfig.PrecompileUpgrades = []params.PrecompileUpgrade{
 		{
-			// Enable ContractDeployerAllowList at timestamp 50
-			ContractDeployerAllowListConfig: contractDeployerConfig,
+			StatefulPrecompileConfig: contractDeployerConfig,
```
darioush, line 274: not sure I like the name "Stateful" that much, what value does that prefix add?

params/precompile_config.go
```diff
@@ -61,153 +71,135 @@ func (p *PrecompileUpgrade) getByAddress(address common.Address) (precompile.Sta
...
+// GetActivePrecompileConfig returns the most recent precompile config corresponding to [address].
 // If none have occurred, returns nil.
-func (c *ChainConfig) getActivePrecompileConfig(blockTimestamp *big.Int, address common.Address, upgrades []PrecompileUpgrade) precompile.StatefulPrecompileConfig {
-	configs := c.getActivatingPrecompileConfigs(nil, blockTimestamp, address, upgrades)
+func (c *ChainConfig) GetActivePrecompileConfig(address common.Address, blockTimestamp *big.Int) precompile.StatefulPrecompileConfig {
+	configs := c.getActivatingPrecompileConfigs(address, nil, blockTimestamp, c.PrecompileUpgrades)
 	if len(configs) == 0 {
 		return nil
 	}
 	return configs[len(configs)-1] // return the most recent config
```
darioush, line 160: I wonder if there is a bit of room for more efficiency here. We can also look as a followup task.

params/precompile_modules.go
```diff
@@ -0,0 +1,46 @@
...
+func init() {
+	// Order is important here.
+	// RegisterModule registers a precompile in the order it is registered.
+	// The order of registration is important because it determines the configuration order
+	// in the state.
+	precompile.RegisterModule(deployerallowlist.NewModule())
+	precompile.RegisterModule(nativeminter.NewModule())
+	precompile.RegisterModule(txallowlist.NewModule())
+	precompile.RegisterModule(feemanager.NewModule())
+	precompile.RegisterModule(rewardmanager.NewModule())
```
darioush, line 43: I think it would be nice if the registration of duplicate addresses failed somehow, I ran into that issue on the other PR and there was no panic / error otherwise.

plugin/evm/vm_upgrade_bytes_test.go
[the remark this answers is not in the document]
darioush: I think the linter enforces for external packages.

core/genesis_test.go
ceyonur: changed it to Config

params/precompile_modules.go
ceyonur: added errors and panics

plugin/evm/vm_upgrade_bytes_test.go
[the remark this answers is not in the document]
ceyonur: yea makes sense

internal/ethapi/api.go
```diff
@@ -626,12 +627,19 @@ func (api *BlockChainAPI) ChainId() *hexutil.Big {
...
 	}
-	return s.b.ChainConfig().GetActivePrecompileUpgrade(blockTimestamp)
+	res := make(params.ChainConfigPrecompiles)
+	for _, module := range modules.RegisteredModules() {
+		if config := s.b.ChainConfig().GetActivePrecompileConfig(module.Address, blockTimestamp); config != nil && !config.IsDisabled() {
+			res[module.ConfigKey] = config
+		}
+	}
+
+	return res
```
aaronbuchwald, line 642: Can we move this back to being a function on `*params.ChainConfig` instead of requiring this logic here for marshaling?

precompile/contract/mock_interfaces.go
```diff
@@ -59,24 +57,7 @@ func (m *mockAccessibleState) CallFromPrecompile(caller common.Address, addr com
...
-	}
-}
-
-func (m *mockChainConfig) GetFeeConfig() commontype.FeeConfig { return m.feeConfig }
-
-func (m *mockChainConfig) AllowedFeeRecipients() bool { return m.allowedFeeRecipients }
-
-type noopStatefulPrecompileConfig struct {
-}
+type noopStatefulPrecompileConfig struct{}
```
aaronbuchwald, line 60: Can we get rid of this duplicate?

precompile/modules/registerer.go
```diff
@@ -0,0 +1,115 @@
...
+	if index >= len(data) {
+		data = append(data, stm)
+		return data
+	}
+	// shift the slice to the right and leave a space for the new element
+	data = append(data[:index+1], data[index:]...)
+	// Insert the new element.
+	data[index] = stm
+	return data
+}
```
aaronbuchwald, line 115: Can we switch to using `sort` for simplicity https://pkg.go.dev/sort#Sort ? I don't think performance should be a concern here.

precompile/registry/registry.go
```diff
@@ -0,0 +1,38 @@
...
+// For forks of subnet-evm, users should start at 0x0300000000000000000000000000000000000000 to ensure
+// that their own modifications do not conflict with stateful precompiles that may be added to subnet-evm
+// in the future.
+// ContractDeployerAllowListAddress = common.HexToAddress("0x0200000000000000000000000000000000000000")
+// ContractNativeMinterAddress      = common.HexToAddress("0x0200000000000000000000000000000000000001")
+// TxAllowListAddress               = common.HexToAddress("0x0200000000000000000000000000000000000002")
+// FeeManagerAddress                = common.HexToAddress("0x0200000000000000000000000000000000000003")
+// RewardManagerAddress             = common.HexToAddress("0x0200000000000000000000000000000000000004")
+// ADD YOUR PRECOMPILE HERE
+// {YourPrecompile}Address          = common.HexToAddress("0x03000000000000000000000000000000000000??")
```
aaronbuchwald, line 38: Should we remove this part of the comment?

precompile/modules/module.go
```diff
@@ -0,0 +1,20 @@
...
+
+type Module struct {
+	// ConfigKey is the key used in json config files to specify this precompile config.
+	ConfigKey string
+	// Address returns the address where the stateful precompile is accessible.
+	Address common.Address
+	// Contract returns a thread-safe singleton that can be used as the StatefulPrecompiledContract when
+	// this config is enabled.
+	Contract contract.StatefulPrecompiledContract
+	contract.Configurator
```
aaronbuchwald, line 19: Can we add a comment for `contract.Configurator`? Also, is there a reason that we embed `contract.Configurator` ?

precompile/modules/module.go
```diff
@@ -0,0 +1,20 @@
...
+// See the file LICENSE for licensing terms.
+
+package modules
+
+import (
+	"github.com/ava-labs/subnet-evm/precompile/contract"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+type Module struct {
```
aaronbuchwald, line 11: What's the rationale for making this a struct instead of an interface?

precompile/contracts/txallowlist/module.go
```diff
@@ -0,0 +1,47 @@
...
+var ContractAddress = common.HexToAddress("0x0200000000000000000000000000000000000002")
+
+var Module = modules.Module{
+	ConfigKey:    ConfigKey,
+	Address:      ContractAddress,
+	Contract:     TxAllowListPrecompile,
+	Configurator: &configuror{},
+}
+
+type configuror struct{}
```
aaronbuchwald, line 28: can we change this to be `configurator` to match the interface name and add a declaration at the top to enforce a type check within this file that it matches the interface?

aaronbuchwald: Same comment for the rest of the modules

precompile/contracts/rewardmanager/contract.go
```diff
@@ -45,34 +45,31 @@ var (
...
 )
 
-func init() {
-	parsed, err := abi.JSON(strings.NewReader(RewardManagerRawABI))
-	if err != nil {
-		panic(err)
-	}
-	RewardManagerABI = parsed
-	RewardManagerPrecompile, err = createRewardManagerPrecompile(precompile.RewardManagerAddress)
+func parseABI(rawABI string) abi.ABI {
```
aaronbuchwald, line 55: Can we put this into a utility, so that it's not duplicated across precompile contracts?

precompile/contracts/rewardmanager/contract.go
```diff
@@ -45,34 +45,31 @@ var (
 	//go:embed contract.abi
 	RewardManagerRawABI string
 
-	RewardManagerABI        abi.ABI                                // will be initialized by init function
-	RewardManagerPrecompile precompile.StatefulPrecompiledContract // will be initialized by init function
+	RewardManagerABI        = parseABI(RewardManagerRawABI)
+	RewardManagerPrecompile = createRewardManagerPrecompile() // will be initialized by init function
```
aaronbuchwald, line 49: can we move these calls into `init()` so that it's more clear they will be called in order?

precompile/contracts/rewardmanager/contract.go
```diff
@@ -13,8 +13,8 @@ import (
 
 	"github.com/ava-labs/subnet-evm/accounts/abi"
 	"github.com/ava-labs/subnet-evm/constants"
-	"github.com/ava-labs/subnet-evm/precompile"
 	"github.com/ava-labs/subnet-evm/precompile/allowlist"
+	"github.com/ava-labs/subnet-evm/precompile/contract"
```
aaronbuchwald, line 17: can we remove the import `"embed" ` below this?

core/state_processor.go
```diff
@@ -168,3 +170,50 @@ func ApplyTransaction(config *params.ChainConfig, bc ChainContext, author *commo
...
+// to apply the necessary state transitions for the upgrade.
+// This function is called:
+// - within genesis setup to configure the starting state for precompiles enabled at genesis,
+// - during block processing to update the state before processing the given block.
+// - during block producing to apply the precompile upgrades before producing the block.
+func ApplyPrecompileActivations(c *params.ChainConfig, parentTimestamp *big.Int, blockContext contract.BlockContext, statedb *state.StateDB) error {
+	blockTimestamp := blockContext.Timestamp()
+	// Note: RegisteredModules returns precompiles sorted by module addresses.
+	// This is important because we want to configure precompiles in the same order
+	// so that the state is deterministic.
```
aaronbuchwald, line 185:
```suggestion
	// Note: RegisteredModules returns precompiles sorted by module addresses.
	// This ensures that the order we call Configure for each precompile is consistent.
	// This ensures even if precompiles read/write state other than their own they will observe
	// an identical global state in a deterministic order when as they are configured.
```

params/config_test.go
```diff
@@ -136,3 +142,64 @@ func TestCheckCompatible(t *testing.T) {
...
+        "allowFeeRecipients": true
+      }
+    },
+    "contractNativeMinterConfig": {
+      "blockTimestamp": 0,
+      "adminAddresses": [
+        "0x8db97C7cEcE249c2b98bDC0226Cc4C2A57BF52FC"
+      ]
+    }
+  }`)
```
aaronbuchwald, line 181: Can we fix the indenting of the JSON here?

params/precompile_upgrade.go
```diff
@@ -0,0 +1,261 @@
...
+	raw := make(map[string]json.RawMessage)
+	if err := json.Unmarshal(data, &raw); err != nil {
+		return err
+	}
+	if len(raw) == 0 {
+		return errNoKey
+	}
+	if len(raw) > 1 {
+		return errMultipleKeys
+	}
```
aaronbuchwald, line 44: Can we update this to a single formatted error that requires the length of raw is 1 and returns the actual number if anything different is found?

precompile/contracts/deployerallowlist/config.go
```diff
@@ -0,0 +1,58 @@
...
+func (c *Config) Equal(cfg config.Config) bool {
+	// typecast before comparison
+	other, ok := (cfg).(*Config)
+	if !ok {
+		return false
+	}
+	return c.Uprade.Equal(&other.Uprade) && c.Config.Equal(&other.Config)
+}
+
+func (c *Config) Verify() error { return c.Config.Verify() }
```
aaronbuchwald, line 58: Should we remove this since `c.Config` is embedded?

aaronbuchwald: Marking as resolved since it seems reasonable to keep this explicit as most other precompile configs have further verification to do.

precompile/contracts/deployerallowlist/module.go
```diff
@@ -0,0 +1,47 @@
...
+var ContractAddress = common.HexToAddress("0x0200000000000000000000000000000000000000")
+
+var Module = modules.Module{
+	ConfigKey:    ConfigKey,
+	Address:      ContractAddress,
+	Contract:     ContractDeployerAllowListPrecompile,
+	Configurator: &configuror{},
+}
+
+type configuror struct{}
```
aaronbuchwald, line 28: nit: can we change this to `configurator` ?

precompile/contracts/feemanager/contract_test.go
```diff
@@ -0,0 +1,273 @@
...
+			input: func() []byte {
+				input, err := PackSetFeeConfig(testFeeConfig)
+				require.NoError(t, err)
+
+				return input
+			},
+			suppliedGas: SetFeeConfigGasCost - 1,
+			readOnly:    false,
+			expectedErr: vmerrs.ErrOutOfGas.Error(),
+		},
```
aaronbuchwald, line 237: Note: this drops two allow list tests from the original.

We should see if we can turn the allow list test suite into a a test suite that can be imported and used by all of the precompiles that use the allow list.

precompile/contracts/rewardmanager/config.go
```diff
@@ -0,0 +1,120 @@
...
+
+// Equal returns true if [cfg] is a [*RewardManagerConfig] and it has been configured identical to [c].
+func (c *Config) Equal(cfg config.Config) bool {
+	// typecast before comparison
+	other, ok := (cfg).(*Config)
+	if !ok {
+		return false
+	}
+	// modify this boolean accordingly with your custom RewardManagerConfig, to check if [other] and the current [c] are equal
+	// if RewardManagerConfig contains only UpgradeableConfig and precompile.AllowListConfig you can skip modifying it.
```
aaronbuchwald, line 109: Should we remove this comment? I see it was in the previous version of the code, but it seems to indicate that the precompile is meant to be modified or borrows from the language of the precompilegen.

precompile/contracts/rewardmanager/module.go
```diff
@@ -0,0 +1,63 @@
...
+	return &Config{}
+}
+
+// Configure configures [state] with the desired admins based on [cfg].
+func (*configuror) Configure(chainConfig contract.ChainConfig, cfg config.Config, state contract.StateDB, _ contract.BlockContext) error {
+	config, ok := cfg.(*Config)
+	if !ok {
+		return fmt.Errorf("incorrect config %T: %v", config, config)
+	}
+	// TODO: should we move this to the end and return it for consistency with other precompiles?
```
aaronbuchwald, line 46: Do you just mean consistency in that `AllowListConfig.Configure(...)` is usually called at the end as opposed to the beginning?

We can, I don't think it's required to do so.

precompile/contracts/txallowlist/contract_test.go
```diff
@@ -0,0 +1,214 @@
...
+				input, err := allowlist.PackModifyAllowList(adminAddr, allowlist.EnabledRole)
+				require.NoError(t, err)
+
+				return input
+			},
+			suppliedGas: allowlist.ModifyAllowListGasCost - 1,
+			readOnly:    false,
+			expectedErr: vmerrs.ErrOutOfGas.Error(),
+		},
+		"read allow listwith no role": {
```
aaronbuchwald, line 148:
```suggestion
		"read allow list with no role": {
```

aaronbuchwald (approved):
LGTM - left some nits, but think we can merge this into `precompile-improvements-main` and resolve from there to make review easier.

internal/ethapi/api.go
ceyonur: moved

precompile/contract/mock_interfaces.go
ceyonur: removed

precompile/modules/registerer.go
ceyonur: done

precompile/registry/registry.go
ceyonur: I think we should collect those addresses in a place for an easy lookup.

precompile/modules/module.go
ceyonur: ConfigKey, Address and Contract are all constants/singleton variables. Module is defined as a static object and should be accessed as one. Modules do not have different implementations but have only different values. So I think struct would fit better in that case. Creating a new instance for Module interfaces seems a little off to me as they supposed to be definitions rather than implementations. 

Plus we can remove all constant accesses from packages (like addresses config keys) and export only modules. Currently we export all these constants.

precompile/contracts/txallowlist/module.go
ceyonur: done

precompile/contracts/rewardmanager/contract.go
ceyonur: yea good call, moved

precompile/contracts/rewardmanager/contract.go
ceyonur: We're already using `init` in module files. Using `init` here would register the module with a nil contract.

precompile/contracts/rewardmanager/contract.go
ceyonur: it's a native golang package moved it to top.

params/config_test.go
ceyonur: done

params/precompile_upgrade.go
ceyonur: done

precompile/contracts/deployerallowlist/module.go
ceyonur: changed

precompile/contracts/feemanager/contract_test.go
ceyonur: created #497

precompile/contracts/rewardmanager/config.go
ceyonur: removed

precompile/contracts/rewardmanager/module.go
ceyonur: moved

precompile/contracts/rewardmanager/module.go
```diff
@@ -0,0 +1,64 @@
...
+	} else if chainConfig.AllowedFeeRecipients() {
+		// configure the RewardManager according to chainConfig
+		EnableAllowFeeRecipients(state)
+	} else {
+		// chainConfig does not have any reward address
+		// if chainConfig does not enable fee recipients
+		// default to disabling rewards
+		DisableFeeRewards(state)
+	}
+	if err := config.Config.Configure(state, ContractAddress); err != nil {
```
darioush, line 60: nit: `return config.Config.Configure(state, ContractAddress)`

precompile/contracts/nativeminter/config.go
```diff
@@ -0,0 +1,98 @@
...
+	for addr, amount := range c.InitialMint {
+		if amount == nil {
+			return fmt.Errorf("initial mint cannot contain nil amount for address %s", addr)
+		}
+		bigIntAmount := (*big.Int)(amount)
+		if bigIntAmount.Sign() < 1 {
+			return fmt.Errorf("initial mint cannot contain invalid amount %v for address %s", bigIntAmount, addr)
+		}
+	}
+	return nil
```
darioush, line 97: nit: we could verify initial mint first then return `c.Config.Verify()` (subjective) (similar with Equal)

precompile/contracts/nativeminter/config.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/ava-labs/subnet-evm/precompile/allowlist"
+	"github.com/ava-labs/subnet-evm/precompile/config"
+	"github.com/ava-labs/subnet-evm/utils"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/math"
+)
+
+var _ config.Config = &Config{}
+
+// Config wraps [AllowListConfig] and uses it to implement the StatefulPrecompileConfig
```
darioush, line 19: this comment should be updated

darioush (commented):
Getting close. Would be great if we could reduce the repetition in the tests somehow.

precompile/contracts/feemanager/contract_test.go
aaronbuchwald: My bad, I forgot you already created this issue. Thanks for being on top of this

precompile/modules/module.go
aaronbuchwald: Makes sense to me. Thanks for the explanation

precompile/contracts/feemanager/contract_test.go
```diff
@@ -0,0 +1,273 @@
...
+			preCondition: func(t *testing.T, state *state.StateDB) {
+				err := StoreFeeConfig(state, testFeeConfig, contract.NewMockBlockContext(big.NewInt(6), 0))
+				require.NoError(t, err)
+			},
+			input: func() []byte {
+				return PackGetFeeConfigInput()
+			},
+			suppliedGas: GetFeeConfigGasCost,
+			readOnly:    true,
+			expectedRes: func() []byte {
```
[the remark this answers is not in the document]
darioush: will deal with this later. :)

precompile/contracts/rewardmanager/config.go
```diff
@@ -0,0 +1,119 @@
...
+func (Config) Key() string { return ConfigKey }
+
+func (c *Config) Verify() error {
+	if err := c.Config.Verify(); err != nil {
+		return err
+	}
+	if c.InitialRewardConfig != nil {
+		return c.InitialRewardConfig.Verify()
+	}
+	return nil
```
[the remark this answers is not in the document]
ceyonur: I think this is resolved already

precompile/contracts/deployerallowlist/contract_test.go
```diff
@@ -0,0 +1,216 @@
...
+			expectedRes: []byte{},
+			assertState: func(t *testing.T, state *state.StateDB) {
+				res := GetContractDeployerAllowListStatus(state, noRoleAddr)
+				require.Equal(t, allowlist.EnabledRole, res)
+			},
+		},
+		"set no role": {
+			caller: adminAddr,
+			input: func() []byte {
+				input, err := allowlist.PackModifyAllowList(adminAddr, allowlist.EnabledRole)
```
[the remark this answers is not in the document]
ceyonur: fixed

precompile/contracts/deployerallowlist/contract_test.go
```diff
@@ -0,0 +1,216 @@
...
+				require.NoError(t, err)
+
+				return input
+			},
+			suppliedGas: allowlist.ModifyAllowListGasCost,
+			readOnly:    false,
+			expectedErr: allowlist.ErrCannotModifyAllowList.Error(),
+		},
+		"set deployer from non-admin": {
+			caller: noRoleAddr,
```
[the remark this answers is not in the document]
ceyonur: added

precompile/contracts/deployerallowlist/contract_test.go
```diff
@@ -0,0 +1,216 @@
...
+		},
+		"read allow list admin role": {
+			caller: adminAddr,
+			input: func() []byte {
+				return allowlist.PackReadAllowList(noRoleAddr)
+			},
+			suppliedGas: allowlist.ReadAllowListGasCost,
+			readOnly:    false,
+			expectedRes: common.Hash(allowlist.EnabledRole).Bytes(),
+			assertState: nil,
```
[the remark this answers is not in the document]
ceyonur: 👍

precompile/contracts/feemanager/config.go
```diff
@@ -0,0 +1,79 @@
...
+// Config wraps [AllowListConfig] and uses it to implement the StatefulPrecompileConfig
+// interface while adding in the FeeManager specific precompile address.
+type Config struct {
+	allowlist.Config // Config for the fee config manager allow list
+	config.Uprade
+	InitialFeeConfig *commontype.FeeConfig `json:"initialFeeConfig,omitempty"` // initial fee config to be immediately activated
+}
+
+// NewConfig returns a config for a network upgrade at [blockTimestamp] that enables
+// FeeManager with the given [admins] and [enableds] as members of the allowlist with [initialConfig] as initial fee config if specified.
```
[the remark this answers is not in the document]
ceyonur: done

precompile/contracts/feemanager/contract_test.go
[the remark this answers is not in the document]
ceyonur: hm I think we should keep them as bytes, but we can accept t when actually calling the function.

precompile/contracts/feemanager/contract_test.go
```diff
@@ -0,0 +1,273 @@
...
+			assertState: func(t *testing.T, state *state.StateDB) {
+				feeConfig := GetStoredFeeConfig(state)
+				require.Equal(t, testFeeConfig, feeConfig)
+				lastChangedAt := GetFeeConfigLastChangedAt(state)
+				require.EqualValues(t, testBlockNumber, lastChangedAt)
+			},
+		},
+		"get fee config from non-enabled address": {
+			caller: noRoleAddr,
+			preCondition: func(t *testing.T, state *state.StateDB) {
```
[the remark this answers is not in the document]
ceyonur: yup makes sense

precompile/contracts/rewardmanager/module.go
ceyonur: done

PR #438, opened 2023-01-18
Implement gossiping to proposers

tedim52:
## Why this should be merged
- Currently, when transactions are submitted to the transaction pool, `subnet-evm` will
    - signal the consensus engine to call “BuildBlock” by sending a “PendingTxs” signal
    - gossip the new transactions submitted to the block
- Currently, snowman++ will select certain validators to act as proposers for a given window
    - thus, only a few validators(the proposers) will actually call “BuildBlock” when signaled
- Insight:
    - gossiping new transactions to validators that won’t be able to propose new blocks is wasted traffic / bandwidth
    - we should gossip new transactions only to the proposers at a given window

## How this works
- Exposes necessary interfaces to gossip to specific nodes
- Changes block builder to gossip transactions directly to proposers if `TargetedProposerGossipEnabled` is true

## How this was tested
- A version of this PR was tested on coreth by spinning up local avalanche-network and verifying that gossip was being sent to proposers via prints statements

### DIFF
--- a/peer/client.go
+++ b/peer/client.go
@@ -7,7 +7,8 @@ import (
 	"errors"
 
 	"github.com/ava-labs/avalanchego/ids"
-
+	"github.com/ava-labs/avalanchego/utils/set"
+	
 	"github.com/ava-labs/avalanchego/version"
 )
 
--- a/peer/client.go
+++ b/peer/client.go
@@ -32,6 +32,8 @@ type NetworkClient interface {
 	// Gossip sends given gossip message to peers
 	Gossip(gossip []byte) error
 
+	GossipSpecific(gossip []byte, nodeIDs set.Set[ids.NodeID]) error
+
 	// TrackBandwidth should be called for each valid request with the bandwidth
 	// (length of response divided by request time), and with 0 if the response is invalid.
 	TrackBandwidth(nodeID ids.NodeID, bandwidth float64)
@@ -84,6 +86,10 @@ func (c *client) Request(nodeID ids.NodeID, request []byte) ([]byte, error) {
 
 func (c *client) Gossip(gossip []byte) error {
 	return c.network.Gossip(gossip)
+}
+
+func (c *client) GossipSpecific(gossip []byte, nodeIDs set.Set[ids.NodeID]) error {
+	return c.network.GossipSpecific(gossip, nodeIDs)
 }
 
 func (c *client) TrackBandwidth(nodeID ids.NodeID, bandwidth float64) {
--- a/peer/client.go
+++ b/peer/client.go
@@ -33,7 +33,8 @@ type NetworkClient interface {
 	// Gossip sends given gossip message to peers
 	Gossip(gossip []byte) error
 
-	GossipSpecific(gossip []byte, nodeIDs set.Set[ids.NodeID]) error
+	// GossipSpecific sends given gossip message to peers specified by [nodeIDs]
+	GossipSpecific(nodeIDs set.Set[ids.NodeID], gossip []byte) error
 
 	// TrackBandwidth should be called for each valid request with the bandwidth
 	// (length of response divided by request time), and with 0 if the response is invalid.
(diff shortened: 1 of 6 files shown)


peer/client.go
```diff
@@ -32,6 +33,8 @@ type NetworkClient interface {
 	// Gossip sends given gossip message to peers
 	Gossip(gossip []byte) error
 
+	GossipSpecific(gossip []byte, nodeIDs set.Set[ids.NodeID]) error
```
aaronbuchwald, line 36: Can we change the order of these parameters to match the style of the other networking functions?

plugin/evm/gossiper.go
```diff
@@ -40,6 +41,8 @@ const (
 type Gossiper interface {
 	// GossipTxs sends AppGossip message containing the given [txs]
 	GossipTxs(txs []*types.Transaction) error
+
+	GossipTxsToNodes(txs []*types.Transaction, nodeIDs set.Set[ids.NodeID]) error
```
aaronbuchwald, line 45: same comment here on changing the params order

plugin/evm/block_builder.go
```diff
@@ -169,11 +169,17 @@ func (b *blockBuilder) awaitSubmittedTxs() {
 					// Give time for this node to build a block before attempting to
 					// gossip
 					time.Sleep(waitBlockTime)
-					// [GossipTxs] will block unless [gossiper.txsToGossipChan] (an
-					// unbuffered channel) is listened on
-					if err := b.gossiper.GossipTxs(ethTxsEvent.Txs); err != nil {
+
+					// Retrieve nodes to gossip to
+					proposers, err := b.ctx.ProposerRetriever.GetCurrentProposers(context.TODO()) // UNSURE OF WHAT CONTEXT TO USE HERE
```
aaronbuchwald, line 174: `context.TODO()` is correct here since it indicates that we should in fact fill this in with a context later, but would need to refactor the code to make the correct context available to do so.

You can delete the comment for now and we have the `context.TODO()` to search for and replace with the correct context when we are ready to thread through all of the context's as needed.

plugin/evm/block_builder.go
```diff
@@ -169,11 +169,17 @@ func (b *blockBuilder) awaitSubmittedTxs() {
 					// Give time for this node to build a block before attempting to
 					// gossip
 					time.Sleep(waitBlockTime)
-					// [GossipTxs] will block unless [gossiper.txsToGossipChan] (an
-					// unbuffered channel) is listened on
-					if err := b.gossiper.GossipTxs(ethTxsEvent.Txs); err != nil {
+
+					// Retrieve nodes to gossip to
```
aaronbuchwald, line 173: For the first iteration, we should add a config parameter in `plugin/evm/config.go` that determines whether subnet-evm uses the existing gossip algorithm or targeted gossip.

plugin/evm/block_builder.go
```diff
@@ -169,11 +169,17 @@ func (b *blockBuilder) awaitSubmittedTxs() {
...
+					proposers, err := b.ctx.ProposerRetriever.GetCurrentProposers(context.TODO()) // UNSURE OF WHAT CONTEXT TO USE HERE
+					if err != nil {
+						log.Warn("failed to retrieve list of proposers", "err", err)
+					}
```
darioush, line 177: let's be careful here to not call `GossipTxsToNodes` with `proposers = nil` or guard against it explicitly in `GossipTxsToNodes`.

tedim52: Opted for guarding against it in `GossipTxsToNodes`

plugin/evm/block_builder.go
```diff
@@ -172,23 +180,31 @@ func (b *blockBuilder) awaitSubmittedTxs() {
...
+						}
+						var nextHeight uint64 = 0 // Can set this either from SetPreference or some of the other calls... VM specific
+						proposers, err := b.proposerRetriever.Proposers(context.TODO(), nextHeight, pChainHeight)
 						if err != nil {
 							log.Warn("failed to retrieve list of proposers", "err", err)
 						}
+						proposerSet := set.NewSet[ids.NodeID](len(proposers))
+						for _, proposer := range proposers {
+							proposerSet.Add(proposer)
+						}
```
darioush, line 200: nit: `proposerSet.Add(proposers...)`

plugin/evm/block_builder.go
```diff
@@ -36,9 +40,17 @@ type blockBuilder struct {
 	ctx         *snow.Context
 	chainConfig *params.ChainConfig
 
+	config Config
```
aaronbuchwald, line 43: can we create a separate block builder config that only includes the necessary parameters for the block builder instead of passing through the VM's whole config?

aaronbuchwald:
Closing due to going a different route of implementing this on a new shared networking component. Thanks for your contribution putting this up 🙌

PR #451, opened 2023-01-20
adding code upgrader precompile

friskyfoxdk:
## Why this should be merged

The CodeUpgrader precompile allows networks to patch a smart contract via a network upgrade. This can be a very powerful tool, but also could cause errors if used incorrectly, due to state mismatches in the new code (similar to the risks in other code upgrade patterns like OpenZeppelin Upgrades). Use with care and only after ample testing.

Note that due to the risks associated with a tool like this, there is no externally callable method on the precompile. It only takes action via the configure options one time when activated via a network upgrade. This is by design.

*** NOTE ***

This is not yet intended to merge, though it is technically ready. This is meant to spur a discussion about whether or not this tool should be included in the base subnet-evm repo, or if it should be instead kept in a subnet-specific fork.

The reasoning is that this tool may prove useful for many other subnets to have in their toolbelts, and therefore would make sense in the core repo.

## How this works

The function is actually quite simple. When the configured timestamp is reached, the precompile will automatically take the code present on the 'source' contract (which should already have been deployed to the network before the upgrade) and copies it to the 'target' contract, effectively 'upgrading' the code of the target contract to match the code of the 'source' contract.

## How this was tested

This was tested by deploying a simple ERC20 contract to the blockchain that was missing any method to update the cap. A new contract was then deployed which had a method to update the cap.

A network upgrade was configured with the first contract as the target and the new contract as the source. Once the network upgrade triggered, the code from the source contract was copied to the target contract, and the method to update the cap was now callable.


friskyfoxdk:
Also note, even though this precompile has no 'callable' methods, I still had to specify an address for it (0x0000...005). It would be nice if there was a way to make precompiles that only take action during network upgrades, and therefore have a simpler interface that doesn't require an address.

aaronbuchwald:
This seems like a reasonable feature request, but there's some fear about how this might impact explorers and security for users that expect to be interacting with a particular contract that was originally deployed or was already audited. Additionally, if a contract was already audited performing admin state updates such as this could allow a contract to reach an unreachable state that could cause problems and break the assumptions of any audits that had been completed on the original code.

If we are going to enable this, we would want to create an alternative way to do so that does not require a callable precompile as you pointed out. We would most likely put together another form of upgrade that applies changes to the state such as:

- balance updates for airdrops
- state changes to contracts such as this
- potentially deploying contracts to specific addresses

friskyfoxdk:
Interesting idea about the deploying to certain addresses. 

I agree about the need for a precompile that is not callable for instances like these. What would the next steps be for this?

aaronbuchwald:
> Interesting idea about the deploying to certain addresses.
> 
> I agree about the need for a precompile that is not callable for instances like these. What would the next steps be for this?

The starting point would be to see how the `PrecompileUpgrade` in `NetworkUpgrades` is used. We would most likely create an additional field in `NetworkUpgrades` that would handle these state transitions and perform the updates in the same places.

Currently we use `CheckConfigurePrecompiles` to handle the state transition that occurs whenever a precompile activates and requires an update to the state. We could use the exact same mechanism here and change the name to be more general.

Since we already call `CheckConfigurePrecompiles` in the miner, where we build new blocks, and in verification where we verify blocks, I believe that would be the only change to the code necessary.

We would then need to write thorough unit tests that to capture all of the new behaviors made possible by these generic state updates. If you want to write a design doc for these changes, I'd be happy to review it.

aaronbuchwald:
Closing in favor of continuing the conversation on https://github.com/ava-labs/subnet-evm/pull/490

Issue #444, opened 2023-01-19
Create messages database/backend that the validator is willing to sign


aaronbuchwald:
Implement a WarpBackend to be used by the precompile `onAccept` and the signature request handler:

```
type WarpBackend interface {
	AddMessage(ctx context.Context, unsignedMessage *teleporter.UnsignedMessage) error
	GetSignature(ctx context.Context, messageHash ids.ID) ([]byte, error)
}
```

PR #452, opened 2023-01-20
Warp backend interface and implementation

minghinmatthewlam:
## Why this should be merged
This PR adds a `WarpBackend` interface which will be used by the warp precompile to add accepted messages, and also get signatures for the signature handler. Issue #444 

## How this works
- Adds a `WarpBackend` interface 
- Supports `AddMessage` for warp precompile to add messages to the db
- Supports `GetSignature` for signature handler to request a message signature
- Implemented the backend with `WarpMessagesDB`
- Signature caching as part of the backend


plugin/evm/warp_backend.go
```diff
@@ -0,0 +1,94 @@
...
+	return db, nil
+}
+
+func (w *WarpMessagesDB) AddMessage(ctx context.Context, unsignedMessage *teleporter.UnsignedMessage) error {
+	messageHash, err := ids.ToID(unsignedMessage.Bytes())
+	if err != nil {
+		return fmt.Errorf("failed to add message with key %s to warp database: %w", messageHash.String(), err)
+	}
+
+	w.Put(messageHash[:], unsignedMessage.Bytes())
```
darioush, line 68: should this error be returned?

plugin/evm/warp_backend.go
```diff
@@ -0,0 +1,94 @@
...
+	return nil
+}
+
+func (w *WarpMessagesDB) GetSignature(ctx context.Context, messageHash ids.ID) ([]byte, error) {
+	if sig, ok := w.signatureCache.Get(messageHash[:]); ok {
+		return sig.([]byte), nil
+	} else {
```
darioush, line 75: nit: unindent/remove else

plugin/evm/warp_backend.go
```diff
@@ -0,0 +1,94 @@
...
+type WarpBackend interface {
+	// AddMessage is called in the precompile OnAccept, to add warp messages into the database.
+	AddMessage(ctx context.Context, unsignedMessage *teleporter.UnsignedMessage) error
+
+	// GetSignature returns the signature of the requested message hash.
+	GetSignature(ctx context.Context, messageHash ids.ID) ([]byte, error)
+}
+
+// WarpMessagesDB implements WarpBackend, keeping track of warp messages, and generating message signatures.
+type WarpMessagesDB struct {
```
darioush, line 40: should we un-export this struct?

plugin/evm/warp_backend.go
```diff
@@ -0,0 +1,94 @@
...
+	db := &WarpMessagesDB{
+		Database:       prefixdb.New(dbPrefix, vmDB),
+		snowCtx:        snowCtx,
+		signatureCache: signatureCache,
+	}
+
+	return db, nil
+}
+
+func (w *WarpMessagesDB) AddMessage(ctx context.Context, unsignedMessage *teleporter.UnsignedMessage) error {
```
darioush, line 62: should we take `[]byte` instead of `*teleporter.UnsignedMessage`?

plugin/evm/warp_backend.go
```diff
@@ -0,0 +1,94 @@
...
+	messageHash, err := ids.ToID(unsignedMessage.Bytes())
+	if err != nil {
+		return fmt.Errorf("failed to add message with key %s to warp database: %w", messageHash.String(), err)
```
darioush, line 65: is it useful to include `messageHash` in this error if `err` is non nil? (won't it always be an empty hash?)

plugin/evm/warp_backend.go
darioush: hum. On second thought this is okay since we use `teleporter` in the other method.

plugin/evm/warp_backend.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func (w *WarpMessagesDB) AddMessage(ctx context.Context, unsignedMessage *teleporter.UnsignedMessage) error {
+	messageHashBytes := hashing.ComputeHash256(unsignedMessage.Bytes())
+	messageHash, err := ids.ToID(messageHashBytes)
+	if err != nil {
+		return fmt.Errorf("failed to add message with key %s to warp database: %w", messageHash.String(), err)
+	}
+
+	return w.Put(messageHash[:], unsignedMessage.Bytes())
```
aaronbuchwald, line 71: Would it make sense to sign the message here and place the signature in the database instead of keeping the whole message? We only need to fetch the signature so no need to store the full message.

Main question to me is the performance tradeoff.

minghinmatthewlam: yup makes sense, made the change and added comment.

plugin/evm/warp_backend.go
```diff
@@ -0,0 +1,82 @@
...
+	// AddMessage is called in the precompile OnAccept, to add warp messages into the database.
+	AddMessage(ctx context.Context, unsignedMessage *teleporter.UnsignedMessage) error
+
+	// GetSignature returns the signature of the requested message hash.
+	GetSignature(ctx context.Context, messageHash ids.ID) ([]byte, error)
+}
+
+// warpBackend implements WarpBackend, keeps track of warp messages, and generates message signatures.
+type warpBackend struct {
+	database.Database
```
aaronbuchwald, line 38: nit: can we make this a field instead of embedding it so that it is more clear

plugin/evm/warp_backend.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/snow"
+	"github.com/ava-labs/avalanchego/utils/hashing"
+	"github.com/ava-labs/avalanchego/vms/platformvm/teleporter"
+)
+
+var (
+	_ WarpBackend = &warpBackend{}
+
+	dbPrefix = []byte("warp")
```
aaronbuchwald, line 23: let's remove this prefix in favor of taking in a database that we expect to already be prefixed as needed

plugin/evm/warp_backend.go
```diff
@@ -0,0 +1,82 @@
...
+func NewWarpBackend(snowCtx *snow.Context, vmDB *versiondb.Database, signatureCacheSize int) WarpBackend {
+	return &warpBackend{
+		Database:       prefixdb.New(dbPrefix, vmDB),
+		snowCtx:        snowCtx,
+		signatureCache: &cache.LRU{Size: signatureCacheSize},
+	}
+}
+
+func (w *warpBackend) AddMessage(ctx context.Context, unsignedMessage *teleporter.UnsignedMessage) error {
+	messageHashBytes := hashing.ComputeHash256(unsignedMessage.Bytes())
```
aaronbuchwald, line 53: nit: can we use `hashing.ComputeHash256Array` since this will already return a `[32]byte` type ?

aaronbuchwald: Then we will not need to call `ids.ToID`

plugin/evm/warp_backend.go
aaronbuchwald: Once this is removed, we should make this var block into a single line since there will only be one variable being declared here.

plugin/evm/warp_backend.go
```diff
@@ -0,0 +1,82 @@
...
+	signature, err := w.snowCtx.TeleporterSigner.Sign(unsignedMessage)
+	if err != nil {
+		return fmt.Errorf("failed to sign warp message %s: %w", messageHash.String(), err)
+	}
+
+	return w.Put(messageHash[:], signature)
+}
+
+func (w *warpBackend) GetSignature(ctx context.Context, messageHash ids.ID) ([]byte, error) {
+	// Attempt to get the signature from cache before calling the db.
```
aaronbuchwald, line 70:
```suggestion
```

aaronbuchwald: I don't think this comment is necessary since this piece of code should be self-explanatory

plugin/evm/warp_backend.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	// We generate the signature here and only save the signature in the db.
+	// It is left to smart contracts built on top of Warp to save messages if required.
+	signature, err := w.snowCtx.TeleporterSigner.Sign(unsignedMessage)
+	if err != nil {
+		return fmt.Errorf("failed to sign warp message %s: %w", messageHash.String(), err)
+	}
+
+	return w.Put(messageHash[:], signature)
```
aaronbuchwald, line 66: Can we add the signature to the cache as well and update the comment to that effect?

plugin/evm/warp_backend.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+// warpBackend implements WarpBackend, keeps track of warp messages, and generates message signatures.
+type warpBackend struct {
+	database.Database
+	snowCtx        *snow.Context
+	signatureCache *cache.LRU
+}
+
+// NewWarpBackend creates a new warpBackend, and initializes the signature cache and message tracking database.
```
aaronbuchwald, line 43:
```suggestion
// NewWarpBackend creates a new WarpBackend, and initializes the signature cache and message tracking database.
```

plugin/evm/warp_backend.go
```diff
@@ -0,0 +1,82 @@
...
+var (
+	_ WarpBackend = &warpBackend{}
+
+	dbPrefix = []byte("warp")
+)
+
+// WarpBackend keeps track of messages that are accepted by the warp precompiles and add them into a database.
+// The backend is also used to query for warp message signatures by the signature request handler.
+type WarpBackend interface {
+	// AddMessage is called in the precompile OnAccept, to add warp messages into the database.
```
aaronbuchwald, line 29:
```suggestion
	// AddMessage signs [unsignedMessage] and adds it to the warp backend database
```

plugin/evm/warp_backend.go
```diff
@@ -0,0 +1,82 @@
...
+	AddMessage(ctx context.Context, unsignedMessage *teleporter.UnsignedMessage) error
+
+	// GetSignature returns the signature of the requested message hash.
+	GetSignature(ctx context.Context, messageHash ids.ID) ([]byte, error)
```
aaronbuchwald, line 33: nit:
```suggestion
	// GetSignature returns the signature of the requested messageID.
	GetSignature(ctx context.Context, messageID ids.ID) ([]byte, error)
```

we typically use `ID` in avalanchego and subnet-evm instead of hash, so we should make this change throughout the file

plugin/evm/warp_backend.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/ava-labs/avalanchego/vms/platformvm/teleporter"
+)
+
+var (
+	_ WarpBackend = &warpBackend{}
+
+	dbPrefix = []byte("warp")
+)
+
+// WarpBackend keeps track of messages that are accepted by the warp precompiles and add them into a database.
```
aaronbuchwald, line 26:
```suggestion
// WarpBackend tracks signature eligible warp messages and provides an interface to fetch them.
```

plugin/evm/warp_backend.go
```diff
@@ -0,0 +1,82 @@
+// (c) 2019-2020, Ava Labs, Inc. All rights reserved.
```
aaronbuchwald, line 1: Can we move this into its own package `plugin/evm/warp` and add unit tests?

plugin/evm/warp_backend.go
ceyonur: hmm would this create a problem if node shutsdown then BLS key changes and boots again? I don't know how this will be used but this should not assume that BLS keys will never change on the node.

plugin/evm/warp_backend.go
minghinmatthewlam: changed to use `hashing.ComputeHash256Array` which returns the `Hash256` type. When putting in the cache i cast it `ids.ID(messageID`, because during testing when putting `Hash256` as key, i would not cache hit when getting a `ids.ID` type of same message bytes.

plugin/evm/warp/backend.go
```diff
@@ -0,0 +1,76 @@
...
+	w.signatureCache.Put(ids.ID(messageID), signature)
+	return nil
+}
+
+func (w *warpBackend) GetSignature(ctx context.Context, messageID ids.ID) ([]byte, error) {
+	if sig, ok := w.signatureCache.Get(messageID); ok {
+		return sig.([]byte), nil
+	}
+
+	signature, err := w.db.Get(messageID[:])
```
minghinmatthewlam, line 69: Was thinking when this case happens, since if we have the messageID in the db, the signature is likely in the cache as well. I think it happens when the cache size starts getting full and we evict signatures from the cache, then we can call the db for the old signatures.

plugin/evm/warp_backend.go
minghinmatthewlam: hmm ye good question, I'm not sure if during the node's shutdown, whether the backend/cache would be wiped. So that after it comes back up with a different BLS key, it'd start from scratch.

ceyonur: cache would be wiped bu db wont be. since we started to write signatures to db they will persist between restarts, and also in the case where BLS key changes. So any old signature will be available for the node even if the node would not be able to sign the bytes due to key change.

plugin/evm/warp/backend.go
```diff
@@ -0,0 +1,76 @@
...
+type warpBackend struct {
+	db             database.Database
+	snowCtx        *snow.Context
+	signatureCache *cache.LRU
+}
+
+// NewWarpBackend creates a new WarpBackend, and initializes the signature cache and message tracking database.
+func NewWarpBackend(snowCtx *snow.Context, db database.Database, signatureCacheSize int) WarpBackend {
+	return &warpBackend{
+		db:             db,
```
ceyonur, line 40: should we use a prefixed db here?

minghinmatthewlam: we opted to pass in a `database.Database`, assuming the passed in db will be prefixed. https://github.com/ava-labs/subnet-evm/pull/452#discussion_r1084524857

plugin/evm/warp/backend_test.go
```diff
@@ -0,0 +1,115 @@
...
+func TestInterfaceStructOneToOne(t *testing.T) {
+	// checks struct provides at least the methods signatures in the interface
+	var _ WarpBackend = (*warpBackend)(nil)
+	// checks interface and struct have the same number of methods
+	backendType := reflect.TypeOf(&warpBackend{})
+	BackendType := reflect.TypeOf((*WarpBackend)(nil)).Elem()
+	if backendType.NumMethod() != BackendType.NumMethod() {
+		t.Fatalf("no 1 to 1 compliance between struct methods (%v) and interface methods (%v)", backendType.NumMethod(), BackendType.NumMethod())
+	}
+}
```
darioush, line 39: not sure we need this test.

plugin/evm/warp/backend_test.go
```diff
@@ -0,0 +1,115 @@
...
+	require.NoError(t, err)
+
+	// Try getting a signature for a message that was not added.
+	messageID := hashing.ComputeHash256Array(unsignedMsg.Bytes())
+	_, err = be.GetSignature(context.Background(), messageID)
+	require.Error(t, err)
+	require.True(t, *called)
+}
+
+func TestCacheTypes(t *testing.T) {
```
darioush, line 89: can we remove this test it doesn't test any of our code?

plugin/evm/warp/backend.go
minghinmatthewlam: That's true, I'd expect it to be more likely that a recently added message will need to be accessed from cache, so might be better to not add an old entry back into the cache.

plugin/evm/warp_backend.go
michaelkaplan13: Yeah this is definitely a valid concern. I think there are few considerations here with respect to storing the signature rather than the message (or message hash):
1. An honest node should never return an invalid signature. This could happen if we store signatures and then the BLS key changes. If we store signatures in the DB we'll need to be careful to invalidate them when we go to use them after changing a BLS key.
2. It's not the end of the world if an honest node is unable to return a signature for an older message after it's BLS key has changed. We only need 67% of nodes to sign for any given message.
3. Ideally, we'd avoid re-signing messages unnecessarily. 

Considering all of that, I think the cleanest option is to store a hash of each message, where the hash is what will be signed to generate the BLS signature of a message. We can still cache signatures that are generated since the cache will be cleared the restart if the BLS key changes. And we have the benefit of not needing to store arbitrary size messages to disk.

I think the simplicity of that approach makes it more desirable than storing the signature with some identifier of which key was used to generate the signature, and then checking that the key is still the current key when returning a signature from the DB.

Curious to hear other's thoughts though (cc @aaronbuchwald, @ceyonur, @minghinmatthewlam)

minghinmatthewlam: I like the approach in terms of the most recently added messages will already have their signatures saved in the cache, and won't need to call on the db to resign the hash to generate a signature. Also, this would be simpler than having some identifier to check whether the saved signature in the db matches the current key of the node.

I agree with point 2, one node changing BLS key and stale db signatures we'd still only need 67%, but considering different sized subnets/nodes with big % stake in subnets, might be good to avoid this case.

ceyonur: It makes much sense to use hashes for signatures rather than the full messages. In that case I think we would not need to store anything in the DB but just in the cache, right?

plugin/evm/warp/backend.go
michaelkaplan13: The more common case (compared to cache eviction) to consider here is the restart of the node. The cache will be empty, but it is likely for a relayer to query for signatures of recent messages still, and we'd still want to cache those results rather than only caching signatures on initial insertion into the DB. So I think it's fine to insert it back into the cache here still :+1:

bernard-avalabs: That makes sense to me.

One possible thing to consider: As it stands, a malicious relayer could purposely query for old entries as a way to pollute the cache. There is a small amplification factor to this attack since a node has to not only serve the request from the DB, but may also have to serve a later request from the DB (that otherwise would have been served from the cache) due to a cache miss.

I don't think we need to worry about this, and the solution is to rate limit requests from each relayer rather than to change our caching approach.

plugin/evm/warp/backend.go
```diff
@@ -0,0 +1,76 @@
...
+	if err := w.db.Put(messageID[:], signature); err != nil {
+		return fmt.Errorf("failed to put warp signature in db: %w", err)
+	}
+
+	w.signatureCache.Put(ids.ID(messageID), signature)
+	return nil
+}
+
+func (w *warpBackend) GetSignature(ctx context.Context, messageID ids.ID) ([]byte, error) {
+	if sig, ok := w.signatureCache.Get(messageID); ok {
```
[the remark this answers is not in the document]
minghinmatthewlam: Discussed in standup and separately with @michaelkaplan13 , since our use of the cache and db are simple PUT and GET calls and not updating the values, no atomicity between the db and cache should suffice our uses.

michaelkaplan13:
Wanted to note a further cleanup option we could implement here that I discussed with @aaronbuchwald yesterday:

- We want to persist the message hashes that willing to be signed for some period of time so that a node will be able to sign recent messages even after restarting.
- We don't need persist those message hashes forever because one, we don't need validators to sign messages that have already been delivered to the destination, and two, messages should always be able to be resent on chain in the unlikely event that they get dropped by >33% of validator stake before being delivered.

It would be great then if this database only stored message hashes for a period of time and automatically deleted them afterwards so as to not have this database grow unbounded in size. We could implement this by keeping 2 indices in the database: one storing just the message hashes, and the other mapping a prefix with a timestamp to the message hash. Then every so often we could iterate over the prefix-timestamp keys and delete any message hashes more than 7 days old. 

I don't think we need to implement this in this PR necessarily but wanted to get everyone's thoughts
(cc @aaronbuchwald @minghinmatthewlam @bernard-avalabs)

aaronbuchwald:
> Wanted to note a further cleanup option we could implement here that I discussed with @aaronbuchwald yesterday:
> 
> * We want to persist the message hashes that willing to be signed for some period of time so that a node will be able to sign recent messages even after restarting.
> * We don't need persist those message hashes forever because one, we don't need validators to sign messages that have already been delivered to the destination, and two, messages should always be able to be resent on chain in the unlikely event that they get dropped by >33% of validator stake before being delivered.
> 
> It would be great then if this database only stored message hashes for a period of time and automatically deleted them afterwards so as to not have this database grow unbounded in size. We could implement this by keeping 2 indices in the database: one storing just the message hashes, and the other mapping a prefix with a timestamp to the message hash. Then every so often we could iterate over the prefix-timestamp keys and delete any message hashes more than 7 days old.
> 
> I don't think we need to implement this in this PR necessarily but wanted to get everyone's thoughts (cc @aaronbuchwald @minghinmatthewlam @bernard-avalabs)

Agreed this is the right approach in the long term. I don't think this needs to be a hard requirement for this to get merged or even v1 of this work, but doing it this way the first time it's in use would save us a database migration process later, so it would be good to include.

plugin/evm/warp/backend.go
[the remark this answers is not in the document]
aaronbuchwald: Yup both the cache and db implementations are thread safe.

plugin/evm/warp/backend_test.go
aaronbuchwald: Let's remove this

plugin/evm/warp/backend_test.go
```diff
@@ -0,0 +1,78 @@
...
+	var _ WarpBackend = (*warpBackend)(nil)
+	// checks interface and struct have the same number of methods
+	backendType := reflect.TypeOf(&warpBackend{})
+	BackendType := reflect.TypeOf((*WarpBackend)(nil)).Elem()
+	if backendType.NumMethod() != BackendType.NumMethod() {
+		t.Fatalf("no 1 to 1 compliance between struct methods (%v) and interface methods (%v)", backendType.NumMethod(), BackendType.NumMethod())
+	}
+}
+
+func TestWarpBackend_ValidMessage(t *testing.T) {
```
aaronbuchwald, line 37: Can we remove the `WarpBackend` prefix from the test names?

plugin/evm/warp/backend_test.go
```diff
@@ -0,0 +1,78 @@
...
+	messageID := hashing.ComputeHash256Array(unsignedMsg.Bytes())
+	signature, err := be.GetSignature(context.Background(), messageID)
+	require.NoError(t, err)
+
+	expectedSig, err := snowCtx.TeleporterSigner.Sign(unsignedMsg)
+	require.NoError(t, err)
+	require.Equal(t, expectedSig, signature)
+}
+
+func TestWarpBackend_InvalidMessage(t *testing.T) {
```
aaronbuchwald, line 60: Can we add one more test case with a cache of size 0?

aaronbuchwald (commented):
LGTM after addressing the last few comments

michaelkaplan13:
> > Wanted to note a further cleanup option we could implement here that I discussed with @aaronbuchwald yesterday:
> > 
> > * We want to persist the message hashes that willing to be signed for some period of time so that a node will be able to sign recent messages even after restarting.
> > * We don't need persist those message hashes forever because one, we don't need validators to sign messages that have already been delivered to the destination, and two, messages should always be able to be resent on chain in the unlikely event that they get dropped by >33% of validator stake before being delivered.
> > 
> > It would be great then if this database only stored message hashes for a period of time and automatically deleted them afterwards so as to not have this database grow unbounded in size. We could implement this by keeping 2 indices in the database: one storing just the message hashes, and the other mapping a prefix with a timestamp to the message hash. Then every so often we could iterate over the prefix-timestamp keys and delete any message hashes more than 7 days old.
> > I don't think we need to implement this in this PR necessarily but wanted to get everyone's thoughts (cc @aaronbuchwald @minghinmatthewlam @bernard-avalabs)
> 
> Agreed this is the right approach in the long term. I don't think this needs to be a hard requirement for this to get merged or even v1 of this work, but doing it this way the first time it's in use would save us a database migration process later, so it would be good to include.

Also should call out that it will require reference counting the message hashes, since they could occur multiple times and we don't want to delete a hash from a week ago if the same hash was just re-submitted.

Definitely agree on waiting for a separate PR to implement this. :+1:

bernard-avalabs:
> It would be great then if this database only stored message hashes for a period of time and automatically deleted them afterwards so as to not have this database grow unbounded in size. We could implement this by keeping 2 indices in the database: one storing just the message hashes, and the other mapping a prefix with a timestamp to the message hash. Then every so often we could iterate over the prefix-timestamp keys and delete any message hashes more than 7 days old.

If we don't need fine-grained pruning of expired entries, a generational pruning approach may be cheaper. Example: we have a separate new and old "table". Every 7 days, the new table becomes the old table, and the previous old table is dropped/cleared since all of its entries should have expired.  Lookups will require checking both the new and old table.

However, I'm not sure if dropping a table is any faster than iterating through and deleting expired entries in a prefix database.

plugin/evm/warp/backend.go
```diff
@@ -46,17 +46,16 @@ func NewWarpBackend(snowCtx *snow.Context, db database.Database, signatureCacheS
 func (w *warpBackend) AddMessage(ctx context.Context, unsignedMessage *teleporter.UnsignedMessage) error {
 	messageID := hashing.ComputeHash256Array(unsignedMessage.Bytes())
 
-	// We generate the signature here and only save the signature in the db and cache.
-	// It is left to smart contracts built on top of Warp to save messages if required.
+	// We save the message instead of signature for db, in case for bls key changes.
```
darioush, line 49: if bls keys may change, (when) is it OK to cache the signature?

plugin/evm/warp/backend.go
```diff
@@ -0,0 +1,85 @@
+// (c) 2019-2020, Ava Labs, Inc. All rights reserved.
```
darioush, line 1: nit: can we use 2023 for new files?

plugin/evm/warp/backend.go
minghinmatthewlam: This is related to [this comment](https://github.com/ava-labs/subnet-evm/pull/452#discussion_r1084669771). Signatures should be okay to cache while normally running for adding messages and getting signatures, but when the node goes offline and possibly changes its bls key, the cache would be wiped. Then after restart, the cache gets repopulated with signatures with the updated bls key.

michaelkaplan13 (approved):
LGTM. Thanks for creating the ticket to add periodic clean up of old messages in the future also.

plugin/evm/warp/backend.go
aaronbuchwald: yes, the assumption is that the BLS key will only change after a restart.

plugin/evm/warp/backend_test.go
```diff
@@ -0,0 +1,88 @@
...
+)
+
+func TestAddAndGetValidMessage(t *testing.T) {
+	db := memdb.New()
+
+	snowCtx := snow.DefaultContextTest()
+	sk, err := bls.NewSecretKey()
+	require.NoError(t, err)
+	snowCtx.TeleporterSigner = teleporter.NewSigner(sk, sourceChainID)
+	be := NewWarpBackend(snowCtx, db, 500)
```
aaronbuchwald, line 32: can we change `be` to `backend` ?

minghinmatthewlam:
Linking [warp backend cleanup](https://github.com/ava-labs/subnet-evm/issues/471) as a follow up from this PR. Addressing this comment: https://github.com/ava-labs/subnet-evm/pull/452#issuecomment-1407081652

plugin/evm/warp/backend.go
darioush: could we capture this in a comment in code? otherwise, LGTM

darioush (approved):
Great work!

PR #453, opened 2023-01-20
Awm precompile interface

### LINKED ISSUE #442: Pre-post precompile interface changes including VerifyWithContext + BuildWithContext placement
- [x] Add predicate and onAccept functions to the precompile interface
- [x] Add the ability to get the original ordering of the storage slots for a given precompile address in the access list
- [x] Change the onAccept function signature to take in all of the log data as individual arguments
- [x] Add VerifyWithContext, BuildWithContext, and ShouldVerifyWithContext to VM

cam-schultz:
## Why this should be merged
Adds precompile predicate changes needed to support AWM in subnet-evm. Fixes #442 

## How this works
- Adds `Predicate` and `OnAccept` methods to the precompile interface
- Stores and makes accessible the original storage slot order for a given precompile address in the access list
- Adds `BuildBlockWithContext`, `ShouldVerifyWithContext`, and `VerifyWithContext` to the VM

## How this was tested
CI


accounts/abi/bind/precompile_template.go
```diff
@@ -215,6 +215,24 @@ func (c *{{.Contract.Type}}Config) Verify() error {
...
+// Predicate optionally returns a function to enforce as a predicate for a transaction to be valid
+// if the access list of the transaction includes a tuple that references the precompile address.
+// Returns nil here to indicate that this precompile does not enforce a predicate.
+func (c *{{.Contract.Type}}Config) Predicate() PredicateFunc {
+	// CUSTOM CODE STARTS HERE
+	// Add your own custom predicate code for {{.Contract.Type}}Config here
+	// and return an optional function if your precompile enforces a predicate.
+	return nil
+}
+// OnAccept optionally returns a function to perform on any log with the precompile address.
```
darioush, line 227: nit: add wspace

core/state/statedb.go
```diff
@@ -674,6 +679,15 @@ func (db *StateDB) ForEachStorage(addr common.Address, cb func(key, value common
 	return nil
 }
 
+// copyPredicateStorageSlots creates a deep copy of the provided predicateStorageSlots map.
+func copyPredicateStorageSlots(predicateStorageSlots map[common.Address][]common.Hash) map[common.Address][]common.Hash {
+	res := make(map[common.Address][]common.Hash)
+	for address, slots := range predicateStorageSlots {
+		res[address] = slots
+	}
```
darioush, line 687: should we do a deep copy of the hashes?

core/state/statedb.go
```diff
@@ -1060,12 +1075,20 @@ func (s *StateDB) PrepareAccessList(sender common.Address, dst *common.Address,
 	for _, addr := range precompiles {
 		s.AddAddressToAccessList(addr)
 	}
+
+	// Note: If an address is specified multiple times the access list, only the
```
darioush, line 1079:
```suggestion
	// Note: If an address is specified multiple times in the access list, only the
```

plugin/evm/block.go
```diff
@@ -101,15 +102,53 @@ func (b *Block) syntacticVerify() error {
...
-func (b *Block) verify(writes bool) error {
+// ShouldVerifyWithContext implements the block.WithVerifyContext interface
+// TODO: Cache the result such that if called multiple times for the same block,
+// we do not need to recompute the value.
+func (b *Block) ShouldVerifyWithContext(context.Context) (bool, error) {
+	precompileConfigs := b.vm.currentRules().Precompiles
+	for _, tx := range b.ethBlock.Transactions() {
+		for _, accessTuple := range tx.AccessList() {
+			if _, ok := precompileConfigs[accessTuple.Address]; ok {
+				log.Debug("Block should be verified with proposer VM context.")
```
darioush, line 118: could we log some information about the block here (eg hash, number)?

plugin/evm/block.go
```diff
@@ -101,15 +102,53 @@ func (b *Block) syntacticVerify() error {
...
+				return true, nil
+			}
+		}
+	}
+	log.Debug("Block does not require proposer VM context for verification.")
```
darioush, line 123: same comment as above.

plugin/evm/block.go
```diff
@@ -101,15 +102,53 @@ func (b *Block) syntacticVerify() error {
...
+
+	return false, nil
+}
+
+// VerifyWithContext implements the block.WithVerifyContext interface
+func (b *Block) VerifyWithContext(ctx context.Context, proposerVMBlockCtx *block.Context) error {
+	if proposerVMBlockCtx != nil {
+		log.Debug("Verifying block with context", "pChainBlockHeight", proposerVMBlockCtx.PChainHeight)
+	} else {
+		log.Debug("VerifyWithContext called with nil context")
```
darioush, line 133: same comment about the logs in this fn

plugin/evm/vm.go
```diff
@@ -620,7 +622,33 @@ func (vm *VM) Shutdown(context.Context) error {
...
+func (vm *VM) buildBlock(ctx context.Context) (snowman.Block, error) {
+	log.Debug("Trying to build block without context.")
+	return vm.buildBlockWithContext(ctx, nil)
+}
+
+func (vm *VM) buildBlockWithContext(ctx context.Context, proposerVMBlockCtx *block.Context) (snowman.Block, error) {
+	if proposerVMBlockCtx != nil {
+		log.Debug("Building block with context", "pChainBlockHeight", proposerVMBlockCtx.PChainHeight)
+	} else {
+		log.Debug("Attempting to build block with nil context")
```
darioush, line 634: nit: could we make the strings for the 2 debug logs more similar?

plugin/evm/vm.go
```diff
@@ -652,6 +680,37 @@ func (vm *VM) buildBlock(context.Context) (snowman.Block, error) {
...
+				continue
+			}
+
+			predicate := precompileConfig.Predicate()
+			if predicate == nil {
+				continue
+			}
+
+			if err := predicate(vm.ctx, proposerVMBlockCtx, accessTuple.StorageKeys); err != nil {
+				log.Warn("Transaction predicate verification failed.", "txId", tx.Hash(), "precompileAddress", accessTuple.Address.Hex())
```
darioush, line 705: could we use `Debug` instead? we should avoid emitting warn logs where an end user could cause lots of warn logs by submitting invalid txs.

plugin/evm/vm.go
```diff
@@ -620,7 +622,33 @@ func (vm *VM) Shutdown(context.Context) error {
...
+
+	// Get the pending transactions from the pool.
+	pending := vm.txPool.Pending(true)
+
+	// Verify any transaction predicates with the given block context.
+	// Remove any transactions from the pool from addresses that submitted transactions with invalid predicates.
+	for sender, txs := range pending {
+		if invalidIndex, err := vm.checkTransactionPredicates(txs, proposerVMBlockCtx); err != nil {
+			log.Debug("Removing transactions from sender of transaction with invalid predicate.", "sender", sender.Hex())
+			delete(pending, sender)
```
darioush, line 645: why is this necessary? (doesn't seem we reference pending again)

plugin/evm/vm.go
```diff
@@ -620,7 +622,33 @@ func (vm *VM) Shutdown(context.Context) error {
...
+			for i := invalidIndex; i < len(txs); i++ {
+				vm.txPool.RemoveTx(txs[i].Hash())
+			}
```
darioush, line 648: can we add a comment why these txs are removed?

plugin/evm/block.go
```diff
@@ -101,15 +102,53 @@ func (b *Block) syntacticVerify() error {
...
 	}
 
+	// If the chain is not yet bootstrapped, we do not need to verify the transaction predicates
+	// because the block is already included in the chain, so we know that predicates must have
+	// been valid at the time the block was accepted.
+	if b.vm.bootstrapped {
+		if _, err := b.vm.checkTransactionPredicates(b.ethBlock.Transactions(), proposerVMBlockCtx); err != nil {
+			return fmt.Errorf("predicate transaction verification failed: %w", err)
+		}
+	}
```
ceyonur, line 150: Hm IMHO we should be checking them while bootstrapping if possible. Wouldn't this cause a node to invalidly bootstrap to the tip even if a predicate configuration changes or missing (I assume predicates depend on the configs)?

plugin/evm/vm.go
```diff
@@ -642,7 +670,7 @@ func (vm *VM) buildBlock(context.Context) (snowman.Block, error) {
 	// We call verify without writes here to avoid generating a reference
 	// to the blk state root in the triedb when we are going to call verify
 	// again from the consensus engine with writes enabled.
-	if err := blk.verify(false /*=writes*/); err != nil {
+	if err := blk.verify(proposerVMBlockCtx, false /*=writes*/); err != nil {
```
ceyonur, line 673: AFAIK in case this block activates a precompile with a predicate, they won't be checked since `currentRules` in vm will not activate the precompile for this block (as it won't accepted yet). This is fine as we also do the same thing for precompiles that would change the validity of txs/the block by precompiles activated at the same block. E.g fee config manager and coinbase would depend on the state of the previous block rather than the generated one. But we should clearly document this expected behaviour with both comments and the function signatures. 

cc @aaronbuchwald

core/state/statedb.go
```diff
@@ -111,6 +111,10 @@ type StateDB struct {
 	// Per-transaction access list
 	accessList *accessList
 
+	// Ordered storage slots to be used in predicate verification as set in the tx access list.
+	// Only set in PrepareAccessList, and un-modified through execution.
+	predicateStorageSlots map[common.Address][]common.Hash
```
aaronbuchwald, line 116: Since this is intended to be used as a byte array, can we change the type to `map[common.Address][]byte` ?

core/state/statedb.go
aaronbuchwald: As per the comment above, I think we should change the value type in this map to a byte slice and then agree on copying the bytes here.

plugin/evm/vm.go
```diff
@@ -652,6 +680,37 @@ func (vm *VM) buildBlock(context.Context) (snowman.Block, error) {
...
+	precompileConfigs := vm.currentRules().Precompiles
+	for i, tx := range txs {
+		for _, accessTuple := range tx.AccessList() {
+			var (
+				precompileConfig   precompile.StatefulPrecompileConfig
+				isPrecompileAccess bool
+			)
+			if precompileConfig, isPrecompileAccess = precompileConfigs[accessTuple.Address]; !isPrecompileAccess {
+				continue
+			}
```
ceyonur, line 697:
```suggestion
			precompileConfig, isPrecompileAccess := precompileConfigs[accessTuple.Address]
			if !isPrecompileAccess {
				continue
			}
```

core/vm/interface.go
```diff
@@ -77,6 +77,7 @@ type StateDB interface {
 	// AddSlotToAccessList adds the given (address,slot) to the access list. This operation is safe to perform
 	// even if the feature/fork is not active yet
 	AddSlotToAccessList(addr common.Address, slot common.Hash)
+	GetPredicateStorageSlots(address common.Address) (bool, []common.Hash)
```
aaronbuchwald, line 80: style nit: the order of these return types should be switched

core/state/statedb.go
```diff
@@ -1102,3 +1125,13 @@ func (s *StateDB) AddressInAccessList(addr common.Address) bool {
...
 }
+
+// GetPredicateStorageSlots returns whether or not the given address was included in the optional
+// access list of the transaction, and the storage slots that were associated with that address
+// if it was included. The storage slots are returned in the same order as they appeared in the transaction.
+// These are the same storage slots that are used to verify any transaction
+// predicates for transactions with access list addresses that match a precompile address.
+func (s *StateDB) GetPredicateStorageSlots(address common.Address) (bool, []common.Hash) {
+	storageSlots, exists := s.predicateStorageSlots[address]
+	return exists, storageSlots
```
aaronbuchwald, line 1136: i don't think we should extract these from the map and then return them in reverse order

precompile/stateful_precompile_config.go
```diff
@@ -7,22 +7,34 @@ import (
 	"fmt"
 	"math/big"
 
+	"github.com/ava-labs/avalanchego/snow"
+	"github.com/ava-labs/avalanchego/snow/engine/snowman/block"
 	"github.com/ethereum/go-ethereum/common"
 )
 
+// PredicateFunc is the function type for validating that an access list tuple touching a precompile follows the predicate
+type PredicateFunc func(chainContext *snow.Context, blockContext *block.Context, storageSlots []common.Hash) error
```
aaronbuchwald, line 16: Would it make sense to update the type of storage slots to a byte slice as well?

plugin/evm/block.go
```diff
@@ -101,15 +102,53 @@ func (b *Block) syntacticVerify() error {
 }
 
 // Verify implements the snowman.Block interface
+// Since VerifyWithContext is implemented, we never expect this to be called.
```
aaronbuchwald, line 105: @StephenButtolph we expect this to be called prior to the proposervm activating right?

```suggestion
// Since VerifyWithContext is implemented, Verify should not be called after the proposervm activates
```

plugin/evm/vm.go
```diff
@@ -620,7 +622,33 @@ func (vm *VM) Shutdown(context.Context) error {
 }
 
 // buildBlock builds a block to be wrapped by ChainState
-func (vm *VM) buildBlock(context.Context) (snowman.Block, error) {
+func (vm *VM) buildBlock(ctx context.Context) (snowman.Block, error) {
+	log.Debug("Trying to build block without context.")
```
aaronbuchwald, line 626:
```suggestion
	log.Debug("Building block without context.")
```

plugin/evm/vm.go
aaronbuchwald: This is a duplicate log since this is called from `buildBlock`. Can we remove the log in `buildBlock` and only keep the logs in this function?

```suggestion
		log.Debug("Building block without context")
```

plugin/evm/vm.go
aaronbuchwald: I believe these transactions should be passed in to `vm.miner.GenerateBlock(txs)` as an argument after filtering out any transactions that fail the predicate.

plugin/evm/vm.go
```diff
@@ -652,6 +680,37 @@ func (vm *VM) buildBlock(context.Context) (snowman.Block, error) {
 	return blk, nil
 }
 
+// checkTransactionPredicates checks the stateful precompile predicates of any of the given
+// transactions that reference a stateful precompile address in their access list.
+// Returns [len(txs), nil] if and only if all referenced predicates are met.
+// Otherwise, returns the index of the first transaction that was invalid and the predicate error.
+func (vm *VM) checkTransactionPredicates(txs types.Transactions, proposerVMBlockCtx *block.Context) (int, error) {
+	precompileConfigs := vm.currentRules().Precompiles
```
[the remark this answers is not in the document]
aaronbuchwald: Can we move the predicate verification function into a standalone function in the `core/` package that takes all of the arguments it needs including `params.Rules` ?

Then we can use that in block verification where we already get `params.Rules` for the block.

For building a block, can we add a new function to the tx pool with an added parameter to enforce predicates? The current pending function takes in a boolean argument whether or not to enforce tips, so we can follow that same style and add another function that enforces predicates as well. Once we have that in the tx pool, the miner can switch from calling `Pending(true)` to `PendingWithPredicates(true, rules)`, which will enforce the predicates for all of the transactions returned to the miner.

core/state/statedb.go
cam-schultz: Changed this and all uses (including the type of `PredicateFunc`) to `[]byte`. Also added a helper in `utils/bytes.go` to serialize a `[]common.Hash` to `[]byte`. Let me know if there's a better place to put that function.

core/state/statedb.go
cam-schultz: Fixed this to make a deep copy.

plugin/evm/vm.go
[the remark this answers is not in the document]
cam-schultz: I moved `checkTransactionPredicates` to a standalone function in `core/state` that takes as parameters `params.Rules` and `*snow.Context`. I also added `txPool.PendingWithPredicates` that calls this new function. `PendingWithPredicates` is called from `buildBlockWithContext`, but does not use the return value in this case. Is this what you had in mind? Or were you envisioning `PendingWithPredicates` would be called in `worker.commitNewWork` (currently calls `txPool.Pending`)? If so, we'll need to give `worker` access to the required rules and contexts.

plugin/evm/vm.go
cam-schultz: The call to `PendingWithPredicates` removes txs that fail the predicate from the pending pool. 

> I believe these transactions should be passed in to vm.miner.GenerateBlock(txs) as an argument after filtering out any transactions that fail the predicate.

Not sure what change you are suggesting here; `GenerateBlock` does not accept any parameters.

plugin/evm/vm.go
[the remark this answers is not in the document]
cam-schultz: I modified this function to take the rules + snow context as parameters, and moved the function to `package state`, following Aaron's suggestions [here](https://github.com/ava-labs/subnet-evm/pull/453#discussion_r1084241603)

plugin/evm/block.go
cam-schultz: @michaelkaplan13 or @aaronbuchwald can comment more on this, but for AWM a major advantage of doing signature verification in the precompile predicate (rather than as part of precompile execution) is that we don't need to check the predicate during bootstrapping. If the transaction exists in the blockchain, then we know it passed the predicate in the past, so there is no need to re-verify the AWM signature.

core/state/statedb.go
```diff
@@ -66,6 +70,34 @@ func (n *proofList) Delete(key []byte) error {
...
+			}
+
+			predicate := precompileConfig.Predicate()
+			if predicate == nil {
+				continue
+			}
+
+			if err := predicate(snowCtx, proposerVMBlockCtx, utils.HashSliceToBytes(accessTuple.StorageKeys)); err != nil {
+				log.Debug("Transaction predicate verification failed.", "txId", tx.Hash(), "precompileAddress", accessTuple.Address.Hex())
+				return i, err
```
bernard-avalabs, line 93: My understanding is that we only need to return the index of the first transaction that has a predicate error as the caller will remove the failing transaction and all of the remaining transactions from the transaction pool. We want to remove all of the remaining transactions because, although not explicitly specified in the interface, the transactions are actually from the same origin account and sorted by the nonce. 

It might be helpful to add a comment to explain this property of the transactions array. Otherwise, it is not clear why can't process the next transaction after a failed predicate execution.

core/state/statedb.go
```diff
@@ -66,6 +70,34 @@ func (n *proofList) Delete(key []byte) error {
 	panic("not supported")
 }
 
+// CheckTransactionPredicates checks the stateful precompile predicates of any of the given
+// transactions that reference a stateful precompile address in their access list.
+// Returns [len(txs), nil] if and only if all referenced predicates are met.
+// Otherwise, returns the index of the first transaction that was invalid and the predicate error.
+func CheckTransactionPredicates(rules params.Rules, snowCtx *snow.Context, txs types.Transactions, proposerVMBlockCtx *block.Context) (int, error) {
```
ceyonur, line 77: I don't think statedb is the best place for this. We increase the cross-dependency with params package and it's a cyclic-import hell there. Can we move this to it's own file under direct `core/` folder? 
Do you have another preferred folder in core package? @aaronbuchwald

cam-schultz: Moved to `core/predicate_check.go`

plugin/evm/vm.go
```diff
@@ -623,31 +622,18 @@ func (vm *VM) Shutdown(context.Context) error {
...
-	for sender, txs := range pending {
-		if invalidIndex, err := vm.checkTransactionPredicates(txs, proposerVMBlockCtx); err != nil {
-			log.Debug("Removing transactions from sender of transaction with invalid predicate.", "sender", sender.Hex())
-			delete(pending, sender)
-			for i := invalidIndex; i < len(txs); i++ {
-				vm.txPool.RemoveTx(txs[i].Hash())
-			}
-		}
-	}
+	vm.txPool.PendingWithPredicates(true, vm.currentRules(), vm.ctx, proposerVMBlockCtx)
```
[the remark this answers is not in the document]
cam-schultz: Renamed to `CheckPredicate`

core/state/statedb.go
cam-schultz: And renamed to `CheckPredicateForSenderTxs` and added doc comments to clarify proper usage.

core/state/statedb.go
cam-schultz: Added comments to clarify these points.

core/tx_pool.go
```diff
@@ -610,6 +612,31 @@ func (pool *TxPool) Pending(enforceTips bool) map[common.Address]types.Transacti
...
+func (pool *TxPool) CheckPredicates(enforcePredicates bool, rules params.Rules, snowCtx *snow.Context, proposerVMBlockCtx *block.Context) {
+	pool.mu.Lock()
+	defer pool.mu.Unlock()
+	for addr, list := range pool.pending {
+		txs := list.Flatten()
+
+		// If the miner requests predicate enforcement, we remove all transactions after and including the index of the first failed predicate
+		if enforcePredicates && !pool.locals.contains(addr) {
+			if invalidIndex, err := CheckPredicatesForSenderTxs(rules, snowCtx, txs, proposerVMBlockCtx); err != nil {
+				log.Debug("Removing transactions from sender of transaction with invalid predicate.", "sender", addr.Hex())
```
darioush, line 631: could we log something about the invalid tx?

plugin/evm/block.go
```diff
@@ -101,15 +103,54 @@ func (b *Block) syntacticVerify() error {
...
-func (b *Block) verify(writes bool) error {
+// ShouldVerifyWithContext implements the block.WithVerifyContext interface
+// TODO: Cache the result such that if called multiple times for the same block,
+// we do not need to recompute the value.
+func (b *Block) ShouldVerifyWithContext(context.Context) (bool, error) {
+	precompileConfigs := b.vm.currentRules().Precompiles
+	for _, tx := range b.ethBlock.Transactions() {
+		for _, accessTuple := range tx.AccessList() {
+			if _, ok := precompileConfigs[accessTuple.Address]; ok {
+				log.Debug(fmt.Sprintf("Block %s (%s) at height %d should be verified with proposer VM context.", b.ID().Hex(), b.ID(), b.Height()))
```
darioush, line 119: could we switch to using the comma delaminated style of logging instead of `fmt.Sprintf` for consistency? this also makes the logs searchable in observability

core/state/statedb.go
michaelkaplan13: This sounds good to me, just wanted to call out that we'll need to handle stripping off leading (and or trailing?) 0 bytes when using the storage slots in the warp precompile still. The actual value will an RLP encoded list of Teleporter messages, padded to have a length that's a multiple of 32.

plugin/evm/block.go
michaelkaplan13: Yeah that's my understanding as well. 

Specifically for AWM, the predicate depends on the block context that includes the current proposer VM P-chain block height, and that block height is used to look up the state of the validator set from the P-chain. Looking up that state from the P-chain at a specific height is an O(n) operation where n is the number of blocks back from the current tip of the chain, so its not feasible to verify an old predicate when bootstrapping because the P-chain height could be years old.

But as @cam-schultz said, a transaction can only be included in the chain if its predicate is valid at the time the block is verified, so when bootstrapping previously accepted blocks we know that the predicate must have been valid already. The predicates are defined by the precompiles and could depend on config changes there, but any config change affecting the precompile/predicate behavior would be considered a required update/hard fork.

plugin/evm/block.go
michaelkaplan13: AFAIK, `Verify` will still be called even after the proposervm activation if `ShouldVerifyWithContext` returns false for a given block (which will be the case for blocks that don't depend on any context). 

This allows optimizations because the verification result is known to never change for a block if it does not depend on a context, but that is not the case for blocks that do depend on a context because the verification result may change if the context changes.

utils/bytes.go
```diff
@@ -16,3 +18,12 @@ func IncrOne(bytes []byte) {
 		}
 	}
 }
+
+// HashSliceToBytes serializes a []common.Hash into a []byte
+func HashSliceToBytes(hashes []common.Hash) []byte {
```
michaelkaplan13, line 23: Should probably add a unit test for this.

plugin/evm/vm.go
aaronbuchwald: I am suggesting to add a parameter to `GenerateBlock` so that it takes transactions as a parameter instead of calling `Pending(true)` inside of the miner.

core/state/statedb.go
aaronbuchwald: That's perfect and good call out.

core/predicate_check.go
```diff
@@ -0,0 +1,41 @@
+package core
```
aaronbuchwald, line 1: nit: add license

core/predicate_check.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+// CheckPredicatesForSenderTxs checks the stateful precompile predicates of any of the given
+// transactions that reference a stateful precompile address in their access list.
+// The parameter [txs] represents a flattened, nonce-ordered list of transactions originating from the same sender
+// Returns [len(txs), nil] if and only if all referenced predicates are met.
+// Otherwise, returns the index of the first transaction that was invalid and the predicate error.
+// In the failure case, all transactions after and including the index that failed the predicate should be considered invalid,
+// since the input [txs] are nonce-ordered and from the same sender.
+func CheckPredicatesForSenderTxs(rules params.Rules, snowCtx *snow.Context, txs types.Transactions, proposerVMBlockCtx *block.Context) (int, error) {
```
aaronbuchwald, line 19: nit: can we change the order of parameters to `rules, snowCtx, proposerVMBlockCtx, txs` this seems more cohesive to me

core/state/statedb.go
```diff
@@ -674,6 +680,17 @@ func (db *StateDB) ForEachStorage(addr common.Address, cb func(key, value common
 	return nil
 }
 
+// copyPredicateStorageSlots creates a deep copy of the provided predicateStorageSlots map.
+func copyPredicateStorageSlots(predicateStorageSlots map[common.Address][]byte) map[common.Address][]byte {
+	res := make(map[common.Address][]byte)
+	for address, slots := range predicateStorageSlots {
+		s := make([]byte, len(slots))
+		copy(s, slots)
+		res[address] = s
```
aaronbuchwald, line 689: Can we use `common.CopyBytes` here?

plugin/evm/vm.go
cam-schultz: Ah gotcha, thanks for clarifying. Made this change now.

plugin/evm/vm.go
cam-schultz: I added a comment explaining this to `Block.verify`, but didn't alter any function signatures. Do you have any suggestions on how best to do so to document this? Where is the precompile activation check done?

utils/bytes_test.go
```diff
@@ -0,0 +1,70 @@
...
+				common.BytesToHash([]byte{4, 5, 6}),
+			},
+			expected: []byte{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 2, 3, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 4, 5, 6},
+		},
+	} {
+		t.Run(name, func(t *testing.T) {
+			output := HashSliceToBytes(test.input)
+			fmt.Println(test.input)
+			fmt.Println(output)
+			fmt.Println(test.expected)
```
darioush, line 66: nit: could we drop the `Println` from the tests?

cam-schultz: Oops, corrected now.

core/state/statedb.go
```diff
@@ -1060,12 +1076,21 @@ func (s *StateDB) PrepareAccessList(sender common.Address, dst *common.Address,
 	for _, addr := range precompiles {
 		s.AddAddressToAccessList(addr)
 	}
+
+	// Note: If an address is specified multiple times in the access list, only the
+	// last storage slots provided for it are used in predicates.
```
aaronbuchwald, line 1081: Calling this out because this is a very weird invariant that could be error prone moving forward.

core/predicate_check.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/ethereum/go-ethereum/log"
+)
+
+// CheckPredicatesForSenderTxs checks the stateful precompile predicates of any of the given
+// transactions that reference a stateful precompile address in their access list.
+// The parameter [txs] represents a flattened, nonce-ordered list of transactions originating from the same sender
+// Returns [len(txs), nil] if and only if all referenced predicates are met.
+// Otherwise, returns the index of the first transaction that was invalid and the predicate error.
+// In the failure case, all transactions after and including the index that failed the predicate should be considered invalid,
+// since the input [txs] are nonce-ordered and from the same sender.
```
aaronbuchwald, line 21: There is nothing in this function that requires the nonces to be in order from the same address, could we remove this comment?

It can be used this way because it will return the index of the first transaction that is invalid, but this is up to the caller, so I don't think the comment is necessary.

core/predicate_check.go
```diff
@@ -0,0 +1,44 @@
...
+				continue
+			}
+
+			predicate := precompileConfig.Predicate()
+			if predicate == nil {
+				continue
+			}
+
+			if err := predicate(snowCtx, proposerVMBlockCtx, utils.HashSliceToBytes(accessTuple.StorageKeys)); err != nil {
+				log.Debug("Transaction predicate verification failed.", "txId", tx.Hash(), "precompileAddress", accessTuple.Address.Hex())
```
aaronbuchwald, line 37:
```suggestion
				log.Debug("Transaction predicate verification failed.", "txHash", tx.Hash(), "precompileAddress", accessTuple.Address.Hex())
```

we typically use `txHash` for geth code. We also would use `txID` instead of `txId` in this repo.

precompile/stateful_precompile_config.go
```diff
@@ -33,9 +45,23 @@ type StatefulPrecompileConfig interface {
...
+	// for any access list tuple whose address matches the address of the precompile.
+	// This allows the precompile to enforce a predicate on the transaction itself for it to be considered valid
+	// to be included in a block.
+	Predicate() PredicateFunc
+
+	// OnAccept returns an optional function which is called when a block gets accepted on each log whose
+	// address matches the address of the precompile.
+	// This can be used to perform precompile specific logic on acceptance, so that the precompile can emit
+	// events and perform logic on those events only after the block has been accepted.
+	OnAccept() OnAcceptFunc
```
minghinmatthewlam, line 63: where in the code do we call `OnAccept()`? Also I'm thinking how we'd have access to the warp backend to call `AddMessage()` within `OnAccept()` given the current parameters.

core/predicate_check.go
bernard-avalabs: I believe I suggested adding this comment previously. It wasn't clear to me why this function returned an index to the first failed transaction until I looked at PendingWithPredicates.

I agree that it is more appropriate for the comment to be at the caller. It looks like the requirement is already explained in the comments above PendingWithPredicates (I'm not sure if those comments were there when I made my original suggestion).

core/tx_pool.go
```diff
@@ -610,6 +612,38 @@ func (pool *TxPool) Pending(enforceTips bool) map[common.Address]types.Transacti
...
+func (pool *TxPool) PendingWithPredicates(enforcePredicates bool, rules params.Rules, snowCtx *snow.Context, proposerVMBlockCtx *block.Context) map[common.Address]types.Transactions {
+	pool.mu.Lock()
+	defer pool.mu.Unlock()
+
+	pending := make(map[common.Address]types.Transactions)
+	for addr, list := range pool.pending {
+		txs := list.Flatten()
+
+		// If the miner requests predicate enforcement, we remove all transactions after and including the index of the first failed predicate
+		if enforcePredicates && !pool.locals.contains(addr) {
```
aaronbuchwald, line 632: We cannot exempt locals here, this would lead to building an invalid block.

aaronbuchwald: We should also enforce tips here as well as enforcing the predicates rather than only enforcing the predicates.

For `enforceTips` we can exempt locals because the transaction is not invalid it is just paying a tip below what the miner has stated as its minimum tip for inclusion in the block.

plugin/evm/vm.go
```diff
@@ -620,8 +621,21 @@ func (vm *VM) Shutdown(context.Context) error {
...
+
+func (vm *VM) buildBlockWithContext(ctx context.Context, proposerVMBlockCtx *block.Context) (snowman.Block, error) {
+	if proposerVMBlockCtx != nil {
+		log.Debug("Building block with context", "pChainBlockHeight", proposerVMBlockCtx.PChainHeight)
+	} else {
+		log.Debug("Building block without context")
+	}
+
+	// Verify any transaction predicates with the given block context.
+	pendingTxs := vm.txPool.PendingWithPredicates(true, vm.currentRules(), vm.ctx, proposerVMBlockCtx)
```
aaronbuchwald, line 636: Using `vm.currentRules()` will use the rules for the current head block of the chain, which will in most cases be different from the block that is about to be built. Since the block being built may have a different timestamp, a different network upgrade could have activated and the ruleset could change.

As a result, we should make sure to calculate the correct `params.Rules` for the actual block we are building or else we may apply an incorrect rule set here, which could halt the chain.

We may need to revert part of the change here and call `PendingWithPredicates` inside of `miner.GenerateBlock` and pass in the P-Chain context instead since that is where the newly created block is created and assigned a block number and timestamp.

plugin/evm/block.go
```diff
@@ -101,15 +103,57 @@ func (b *Block) syntacticVerify() error {
...
+		for _, accessTuple := range tx.AccessList() {
+			if _, ok := precompileConfigs[accessTuple.Address]; ok {
+				log.Debug("Should verify block with proposer VM context", "block", b.ID(), "height", b.Height())
+				return true, nil
+			}
+		}
+	}
+	log.Debug("Block does not require proposer VM context for verification.", "block", b.ID(), "height", b.Height())
+
+	return false, nil
```
aaronbuchwald, line 126: If `ShouldVerifyWithContext` returns false, then wouldn't the amount of work done in `VerifyWithContext` be the same?

It would just iterate the transactions and their access lists and skip all of them since no access list would name a predicate.

What's the point of using `ShouldVerifyWithContext` like this instead of just having this return true and doing the verification every time in that case?

What's the intended use case of separating `ShouldVerifyWithContext` ?

plugin/evm/block.go
```diff
@@ -101,15 +103,57 @@ func (b *Block) syntacticVerify() error {
...
+
+// VerifyWithContext implements the block.WithVerifyContext interface
+func (b *Block) VerifyWithContext(ctx context.Context, proposerVMBlockCtx *block.Context) error {
+	if proposerVMBlockCtx != nil {
+		log.Debug("Verifying block with context", "block", b.ID(), "height", b.Height())
+	} else {
+		log.Debug("Verifying block without context", "block", b.ID(), "height", b.Height())
+	}
+	return b.verify(proposerVMBlockCtx, true)
+}
```
aaronbuchwald, line 137: Since `VerifyWithContext` may be called multiple times (change from previous behavior where `Verify` was only called once on a block), this breaks some assumptions in the way we currently handle snapshots and the triedb.

One way to handle this would be to encode the P-Chain block height in the `Extra` data of the block and make this a requirement after some network upgrade.

The goal of this would be to ensure that there cannot be two different valid proposervm blocks containing the same EVM block.

plugin/evm/block.go
```diff
@@ -101,15 +103,57 @@ func (b *Block) syntacticVerify() error {
...
+
+// Verify the block, including checking precompile predicates
+// Note that if this block activates a precompile with a predicate, that predicate is not checked
+// in this block. The VM doesn't activate the precompile until the block is accepted.
```
aaronbuchwald, line 141: This comment is incorrect. When a precompile activates, it is immediately callable within that block.

If we do not enforce the predicate where the precompile has already activated, this creates an attack vector where in the block that the precompile activates the predicate will not be checked correctly.

This is why we need to use the rule set for the actual block and cannot use `vm.currentRules()`.

plugin/evm/block.go
```diff
@@ -101,15 +103,57 @@ func (b *Block) syntacticVerify() error {
...
 	if err := b.syntacticVerify(); err != nil {
 		return fmt.Errorf("syntactic block verification failed: %w", err)
 	}
 
+	// If the chain is not yet bootstrapped, we do not need to verify the transaction predicates
+	// because the block is already included in the chain, so we know that predicates must have
+	// been valid at the time the block was accepted.
+	if b.vm.bootstrapped {
+		rules := b.vm.chainConfig.AvalancheRules(b.ethBlock.Number(), b.ethBlock.Timestamp())
+		if _, err := core.CheckPredicatesForSenderTxs(rules, b.vm.ctx, proposerVMBlockCtx, b.ethBlock.Transactions()); err != nil {
```
aaronbuchwald, line 152: What's the intended behavior if `proposerVMBlockCtx` is nil? We should call this out more clearly.

plugin/evm/block.go
aaronbuchwald: I believe there could still be multiple proposervm headers with the same internal EVM block if it included a different proposer.

plugin/evm/block.go
cam-schultz: It's up to the predicate to handle this case. `VerifyWithContext` is guaranteed to never be called with a `nil` context, *but* `Verify` may be called even if `ShouldVerifyWithContext` would have returned true. This will be the case if the proposer VM is not yet fully activated.

I'll add a comment at the definition of `PredicateFunc` explaining this.

plugin/evm/vm.go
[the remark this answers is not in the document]
aaronbuchwald: Yes, the miner has the block number and timestamp so it can create the rules and we will need to pass in the relevant context.

core/tx_pool.go
cam-schultz: Fixed.

cam-schultz: Fixed.

plugin/evm/vm.go
cam-schultz: I modified `miner.GenerateBlock` to take the `snowCtx` and `proposerVMBlockCtx`, which are eventually passed to the predicates before the block is built. We also use the new block's rules when checking the predicate.

plugin/evm/block.go
cam-schultz: Removed this comment.

plugin/evm/block.go
```diff
@@ -101,15 +103,55 @@ func (b *Block) syntacticVerify() error {
...
-	return b.verify(true)
+	return b.verify(nil, true)
 }
 
-func (b *Block) verify(writes bool) error {
+// ShouldVerifyWithContext implements the block.WithVerifyContext interface
+// TODO: Cache the result such that if called multiple times for the same block,
+// we do not need to recompute the value.
+func (b *Block) ShouldVerifyWithContext(context.Context) (bool, error) {
+	precompileConfigs := b.vm.currentRules().Precompiles
```
darioush, line 115: Should this refer to the parent block or this block instead?

plugin/evm/block.go
```diff
@@ -101,15 +103,55 @@ func (b *Block) syntacticVerify() error {
...
+				return true, nil
+			}
+		}
+	}
+	log.Debug("Block does not require proposer VM context for verification.", "block", b.ID(), "height", b.Height())
+
+	return false, nil
+}
+
+// VerifyWithContext implements the block.WithVerifyContext interface
```
darioush, line 129: could we add a type assertion for this eg,`var _ block.WithVerifyContext = &Block{}`

plugin/evm/block.go
cam-schultz: Amended this function to check if the block is currently processing, and if so, only verify the predicates and return.

plugin/evm/block.go
cam-schultz: Done.

plugin/evm/block.go
cam-schultz: Yes, good catch. Fixed now.

plugin/evm/block.go
```diff
@@ -101,15 +105,64 @@ func (b *Block) syntacticVerify() error {
...
+func (b *Block) VerifyWithContext(ctx context.Context, proposerVMBlockCtx *block.Context) error {
+	if proposerVMBlockCtx != nil {
+		log.Debug("Verifying block with context", "block", b.ID(), "height", b.Height())
+	} else {
+		log.Debug("Verifying block without context", "block", b.ID(), "height", b.Height())
+	}
+	// VerifyWithContext may be called multiple times on the same block with multiple contexts, but the engine will only call Accept or Reject once.
+	// As such, we need to ensure that we only verify the block once, so if the block is currently processing, we only check the predicates and return.
+	if b.status == choices.Processing {
```
darioush, line 140: I am not sure this is correct, since:
(a) I am not sure that the block's status will not be Processing the 1st time Verify is called,
(b) even if it is not, we will still want to check predicates the 1st time.

miner/miner.go
```diff
@@ -62,8 +64,8 @@ func (miner *Miner) SetEtherbase(addr common.Address) {
 	miner.worker.setEtherbase(addr)
 }
 
-func (miner *Miner) GenerateBlock() (*types.Block, error) {
-	return miner.worker.commitNewWork()
+func (miner *Miner) GenerateBlock(snowCtx *snow.Context, proposerVMBlockCtx *block.Context) (*types.Block, error) {
```
ceyonur, line 67: I think these two always contexts go together, maybe we should consider creating a new Predicate context and wrap these two.

precompile/stateful_precompile_config.go
```diff
@@ -33,9 +46,23 @@ type StatefulPrecompileConfig interface {
...
```
ceyonur, line 64: It feels more natural for these two functions (`Predicate` and `OnAccept`) to be a part of Contract type rather than of Config type. But it might complicate things and break so much stuff. We can keep this as it is and consider moving to contract type as an improvement

core/predicate_check.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/ava-labs/avalanchego/snow/engine/snowman/block"
+	"github.com/ava-labs/subnet-evm/core/types"
+	"github.com/ava-labs/subnet-evm/params"
+	"github.com/ava-labs/subnet-evm/utils"
+	"github.com/ethereum/go-ethereum/log"
+)
+
+// CheckPredicatesForSenderTxs checks the stateful precompile predicates of any of the given
+// transactions that reference a stateful precompile address in their access list.
+// The parameter [txs] represents a flattened, nonce-ordered list of transactions originating from the same sender
```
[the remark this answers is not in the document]
cam-schultz: There are two use cases for this function:
1) Check all transactions from a particular address, and return the index of the first transaction that fails a predicate. All other transactions from that address are truncated by the caller
2) Verify an entire batch of transactions, not necessarily from the same address. If any fail the predicate, then the batch is invalid. This is the behavior in the call to this function from `VerifyWithContext`.

I agree that the comment on this function only reflects the first case. I will update it accordingly. From this function's perspective, what the caller chooses to do with the return value is of no concern.

core/tx_pool.go
```diff
@@ -610,6 +612,52 @@ func (pool *TxPool) Pending(enforceTips bool) map[common.Address]types.Transacti
...
+	defer pool.mu.Unlock()
+
+	pending := make(map[common.Address]types.Transactions)
+	for addr, list := range pool.pending {
+		txs := list.Flatten()
+
+		// If the miner requests tip enforcement, cap the lists now
+		if enforceTips {
+			for i, tx := range txs {
+				if tx.EffectiveGasTipIntCmp(pool.gasPrice, pool.priced.urgent.baseFee) < 0 {
```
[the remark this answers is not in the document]
cam-schultz: See @aaronbuchwald 's comment here: https://github.com/ava-labs/subnet-evm/pull/453#discussion_r1093975501

miner/miner.go
cam-schultz: Good call I added `PredicateContext` to package `precompile` so that it can be used in `core` and in `PredicateFunc` in `precompile`.

core/tx_pool.go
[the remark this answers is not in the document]
ceyonur: I think he commented that for `enforcePredicates`, and he also mentioned we can exempt locals for `enforceTips` in https://github.com/ava-labs/subnet-evm/pull/453#discussion_r1093976364

`enforceTips` does not tell anything about skipping predicate checks.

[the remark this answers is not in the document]
cam-schultz: Ah you're right. I'll add that check in.

core/tx_pool.go
```diff
@@ -610,6 +612,52 @@ func (pool *TxPool) Pending(enforceTips bool) map[common.Address]types.Transacti
...
+				}
+			}
+		}
+
+		// If the miner requests predicate enforcement, we remove all transactions after and including the index of the first failed predicate
+		if enforcePredicates {
+			if invalidIndex, err := CheckPredicatesForSenderTxs(rules, snowCtx, proposerVMBlockCtx, txs); err != nil {
+				log.Debug("Removing transactions from sender of transaction with invalid predicate.", "sender", addr.Hex(), "failedTx", txs[invalidIndex].Hash())
+				for i := invalidIndex; i < len(txs); i++ {
+					pool.RemoveTx(txs[i].Hash())
```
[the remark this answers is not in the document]
cam-schultz: Incorporated your suggestion to separate tx filtering and removal. I left `CheckPredicatesForSenderTxs` in core, as it takes a `types.Transactions` parameter, so we cannot move it upstream.

aaronbuchwald:
The e2e tests are failing because there is no AvalancheGo release to be installed for this branch. Created a ticket to update CI, so that it can depend on an AvalancheGo branch/commit: https://github.com/ava-labs/subnet-evm/issues/500

precompile/stateful_precompile_config.go
```diff
@@ -7,22 +7,41 @@ import (
...
+	SnowCtx            *snow.Context
+	ProposerVMBlockCtx *block.Context
+}
+
+// PredicateFunc is the function type for validating that an access list tuple touching a precompile follows the predicate
+// It is possible [blockContext] to be nil if the proposer VM is not fully activated, so predicates should explicitly check this.
+type PredicateFunc func(predicateContext *PredicateContext, storageSlots []byte) error
+
+// OnAcceptFunc is called on any log produced in a block where the address matches the precompile address
+type OnAcceptFunc func(txHash common.Hash, logIndex int, topics []common.Hash, logData []byte) error
```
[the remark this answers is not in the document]
aaronbuchwald: This creates a cyclical dependency for now. After your precompile improvements changes are merged, we should definitely update this to use the actual log.

plugin/evm/block.go
```diff
@@ -101,15 +106,72 @@ func (b *Block) syntacticVerify() error {
 }
 
 // Verify implements the snowman.Block interface
+// Since VerifyWithContext is implemented, we never expect this to be called.
```
michaelkaplan13, line 109: I think we should either remove or clarify this comment since `Verify` is still expected to be called in certain cases.

plugin/evm/vm.go
michaelkaplan13: Looping back here, do we need to add a special case to make sure that no warp messages should be allowed to be received in the activation block of the warp precompile? Doesn't need to be in this PR, but it is something we should track if so.

plugin/evm/block.go
cam-schultz: Good call. Removed.

core/predicate_check.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/ava-labs/subnet-evm/params"
+	"github.com/ava-labs/subnet-evm/precompile"
+	"github.com/ava-labs/subnet-evm/utils"
+	"github.com/ethereum/go-ethereum/log"
+)
+
+// CheckPredicatesForSenderTxs checks the stateful precompile predicates of any of the given
+// transactions that reference a stateful precompile address in their access list.
+// The parameter [txs] represents a flattened list of transactions. It is up to the caller to decide the ordering of
+// the transactions in [txs] in relation to the return value of this function.
```
aaronbuchwald, line 17: Can we clarify this ? This is hard to understand

plugin/evm/vm.go
aaronbuchwald: Yes this is a bug as implemented and will not be merged in its current form.

This should use the rule set for the block instead of `vm.currentRules()`.

aaronbuchwald:
Closing in favor of https://github.com/ava-labs/subnet-evm/pull/508

Issue #445, opened 2023-01-19
Add signature request handler to the VM


aaronbuchwald:
Add signature request handler using `WarpBackend` specified by https://github.com/ava-labs/subnet-evm/issues/444.

The signature request handler should handle a new `AppRequest` message and respond with the requested signature if available.

aaronbuchwald:
Closed by #459

PR #459, opened 2023-01-26
Signature Request Handler

minghinmatthewlam:
## Why this should be merged
This PR adds a `SignatureRequestHandler` that handles `SignatureRequest`, which is an AppRequest used to request for a warp message's signature. Issue #445 

## How this works
- New message types of `SignatureRequest` and `SignatureResponse`
- Sync handler constructs a new `SignatureRequestHandler` that takes in a `WarpBackend` to handle signatures
- Signature request handler requests the warp backend for the corresponding message's signature, and uses the codec to marshal the signature response to return.

## How this was tested
- Unit tests for both signature request and response marshals.
- Unit test for signature request handler to check for expected signature response.
- Handler stats to check expected code calls for corresponding inputs.


plugin/evm/message/signature_request_test.go
```diff
@@ -0,0 +1,62 @@
+// (c) 2022, Ava Labs, Inc. All rights reserved.
```
darioush, line 1: nit: probably should start using 2023 going forward

plugin/evm/message/signature_request_test.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+// TestMarshalSignatureResponse asserts that the structure or serialization logic hasn't changed, primarily to
+// ensure compatibility with the network.
+func TestMarshalSignatureResponse(t *testing.T) {
+	var signature [bls.SignatureLen]byte
+	sig, err := hex.DecodeString("0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef")
+	if err != nil {
+		t.Error("failed to decode string to hex")
+	}
```
darioush, line 46: nit: 
```suggestion
	require.NoError(err, "failed to decode string to hex")
```

sync/handlers/handler.go
```diff
@@ -48,6 +49,9 @@ func NewSyncHandler(
 		stateTrieLeafsRequestHandler: NewLeafsRequestHandler(evmTrieDB, provider, networkCodec, stats),
 		blockRequestHandler:          NewBlockRequestHandler(provider, networkCodec, stats),
 		codeRequestHandler:           NewCodeRequestHandler(evmTrieDB.DiskDB(), networkCodec, stats),
+
+		// TODO: need to pass in the warp backend from the subnet-evm.
+		signatureRequestHandler: NewSignatureRequestHandler(nil, networkCodec, stats),
```
darioush, line 54: we shouldn't add this to the `sync/handlers` package, may need some refactoring.

sync/handlers/signature_request.go
```diff
@@ -0,0 +1,67 @@
...
+	startTime := time.Now()
+	s.stats.IncSignatureRequest()
+
+	// Always report signature request time
+	defer func() {
+		s.stats.UpdateSignatureRequestTime(time.Since(startTime))
+	}()
+
+	var signature [bls.SignatureLen]byte
+	sig, err := s.backend.GetSignature(ctx, signatureRequest.MessageID)
```
darioush, line 50: should `GetSignature` return `[bls.SignatureLen]byte` instead?

sync/handlers/signature_request_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+			// If the expected response is empty, assert that the handler returns an empty response and return early.
+			if len(expectedResponse) == 0 {
+				assert.Len(t, responseBytes, 0, "expected response to be empty")
+				return
+			}
+			var response message.SignatureResponse
+			if _, err = message.Codec.Unmarshal(responseBytes, &response); err != nil {
+				t.Fatal("error unmarshalling SignatureResponse", err)
+			}
```
darioush, line 92: let's switch this test `require` (feel free to update the other files as well, for consistency)

sync/handlers/handler.go
aaronbuchwald: if we pass `nil` in here, then if a peer sends a signature request, this node will panic with a nil pointer error.

Since we're merging un-activated code in, we need to gate it so that it cannot actually be hit and cause an issue. Can we create a `NoopSignatureRequestHandler` and use that here instead until we are ready to activate this?

It would also be alright to actually construct the warp backend and pass it in here since there will not be any messages in it and it will just end up dropping each request, but using a noop handler will be even better because it will not cause the request handler to hit the database.

aaronbuchwald: Agreed we should refactor this package out now that we are putting a non-sync request handler in here. Can we move this out to a top-level `handlers` package for now?

I think we may want to refactor the networking code more, but just moving it out from here is good for this PR.

handlers/signature_request.go
```diff
@@ -0,0 +1,78 @@
...
+import (
+	"context"
+	"time"
+
+	"github.com/ava-labs/avalanchego/codec"
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/utils/crypto/bls"
+	"github.com/ava-labs/subnet-evm/plugin/evm/message"
+	"github.com/ava-labs/subnet-evm/plugin/evm/warp"
+	"github.com/ava-labs/subnet-evm/sync/handlers/stats"
```
aaronbuchwald, line 15: we should not need to import `sync/handlers/stats` here. Can we refactor this?

handlers/signature_request.go
```diff
@@ -0,0 +1,78 @@
...
+	startTime := time.Now()
+	s.stats.IncSignatureRequest()
+
+	// Always report signature request time
+	defer func() {
+		s.stats.UpdateSignatureRequestTime(time.Since(startTime))
+	}()
+
+	var signature [bls.SignatureLen]byte
+	sig, err := s.backend.GetSignature(ctx, signatureRequest.MessageID)
```
aaronbuchwald, line 55: I believe there's another comment saying the same thing, but can we ensure that `GetSignature` returns the type that will be needed by the caller.

sync/handlers/handler.go
```diff
@@ -48,6 +51,9 @@ func NewSyncHandler(
 		stateTrieLeafsRequestHandler: NewLeafsRequestHandler(evmTrieDB, provider, networkCodec, stats),
 		blockRequestHandler:          NewBlockRequestHandler(provider, networkCodec, stats),
 		codeRequestHandler:           NewCodeRequestHandler(evmTrieDB.DiskDB(), networkCodec, stats),
+
+		// TODO: need to pass in the warp backend from the subnet-evm to create signature request handler.
```
aaronbuchwald, line 55:
```suggestion
		// TODO: initialize actual signature request handler when warp is ready
```

sync/handlers/handler.go
```diff
@@ -35,6 +37,7 @@ type syncHandler struct {
 	stateTrieLeafsRequestHandler *LeafsRequestHandler
 	blockRequestHandler          *BlockRequestHandler
 	codeRequestHandler           *CodeRequestHandler
+	signatureRequestHandler      handlers.SignatureRequestHandler
```
aaronbuchwald, line 40: We should move the handler out of `sync/handlers/handler.go` not only the `SignatureRequestHandler`. Sync refers to state sync and this is now a more general network request handler.

handlers/signature_request_test.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/ava-labs/avalanchego/utils/hashing"
+	"github.com/ava-labs/avalanchego/vms/platformvm/teleporter"
+	"github.com/ava-labs/subnet-evm/plugin/evm/message"
+	"github.com/ava-labs/subnet-evm/plugin/evm/warp"
+	"github.com/ava-labs/subnet-evm/sync/handlers/stats"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestSignatureHandler(t *testing.T) {
```
aaronbuchwald, line 24: Could we add a unit test to `plugin/evm/vm_test.go` as well that verifies sending a request with this code is dropped correctly and does not result in a panic as a regression for the previous version that left the warp backend unpopulated? We should add a TODO to modify the test when we actually activate warp so that the expected behavior changes from dropping without a panic or error to responding correctly.

This should be a good starting point for what that unit test should look like: https://github.com/ava-labs/subnet-evm/blob/master/plugin/evm/vm_test.go#L2995.

The test should:

- Set up the VM
- Create a signature request for a non-existent signature
- Call `vm.AppRequest(...)` with the message and ensure that it does not return an error or panic
- Include a TODO to update the test when the signature request handler should be activated

plugin/evm/warp/backend.go
```diff
@@ -11,11 +11,16 @@ import (
...
+	"github.com/ava-labs/avalanchego/utils/crypto/bls"
 	"github.com/ava-labs/avalanchego/utils/hashing"
 	"github.com/ava-labs/avalanchego/vms/platformvm/teleporter"
 )
 
-var _ WarpBackend = &warpBackend{}
+var (
+	_ WarpBackend = &warpBackend{}
+
+	emptySignature = [bls.SignatureLen]byte{}
```
bernard-avalabs, line 22: I assume emptySignature is an invalid signature and can never be mistaken for a valid one. 

(For my own understanding): Is it easier/cleaner for the caller to handle an empty invalid signature instead of a nil?

sync/handlers/handler.go
minghinmatthewlam: Refactored out the `SignatureRequestHandler` and created a `NetworkHandler` which should include sync and non-sync related handlers. Also made corresponding stats.

plugin/evm/warp/backend.go
aaronbuchwald: Can we remove this to make the code more obviously correct?

This will only be returned when there is a non-nil error.

handlers/signature_request_test.go
aaronbuchwald: Bumping this

minghinmatthewlam: just pushed the `vm_test` but figuring out fixing tree formats

plugin/evm/vm_test.go
```diff
@@ -813,10 +813,11 @@ func TestSetPreferenceRace(t *testing.T) {
...
 // from another VM (which have a common ancestor under the finalized
 // frontier).
-//   A
-//  / \
-// B   C
 //
+//	 A
+//	/ \
+//
+// B   C
```
aaronbuchwald, line 820: Can you revert the formatting changes to the existing graphs?

sync/handlers/signature_request.go
minghinmatthewlam: updated return value

plugin/evm/warp/backend.go
minghinmatthewlam: re @bernard-avalabs , not sure if it's cleaner or necessarily easier to handle the empty signature case, but more so changed because we want `warpBackend.GetSignature` to return the expected signature type, and with that we can't return nil as type `[bls.SignatureLen]byte`

aaronbuchwald (approved):

handlers/warp/signature_request_test.go
```diff
@@ -0,0 +1,99 @@
...
+	}{
+		"normal": {
+			setup: func() (request message.SignatureRequest, expectedResponse []byte) {
+				return message.SignatureRequest{
+					MessageID: messageID,
+				}, signature[:]
+			},
+			verifyStats: func(t *testing.T, stats *stats.MockHandlerStats) {
+				assert.EqualValues(t, 1, mockHandlerStats.SignatureRequestCount)
+				assert.EqualValues(t, 1, mockHandlerStats.SignatureRequestHit)
```
michaelkaplan13, line 57: For completeness, we should also check that `mockHandlerStats.SignatureRequestMiss` is 0 here, and that `mockHandlerStats.SignatureRequestHit` is 0 in the test case below.

michaelkaplan13 (commented):
One nit on a unit test, but otherwise LGTM. 

Thanks to the platform team for all of the thorough reviews prior to mine.

handlers/warp/signature_request_test.go
minghinmatthewlam: agreed, updated the unit tests to include.

handlers/warp/signature_request_test.go
```diff
@@ -0,0 +1,101 @@
...
+		"unknown": {
+			setup: func() (request message.SignatureRequest, expectedResponse []byte) {
+				return message.SignatureRequest{
+					MessageID: unknownMessageID,
+				}, nil
+			},
+			verifyStats: func(t *testing.T, stats *stats.MockHandlerStats) {
+				assert.EqualValues(t, 1, mockHandlerStats.SignatureRequestCount)
+				assert.EqualValues(t, 1, mockHandlerStats.SignatureRequestMiss)
+				assert.EqualValues(t, 1, mockHandlerStats.SignatureRequestHit)
```
michaelkaplan13, line 71: This should be 0, not 1, right?

minghinmatthewlam: thanks for catching, also realized was wondering how tests passed, turns out need to add extra `verifyStats` check for the nil case as it wasn't getting hit before.

michaelkaplan13 (approved):

darioush:
this is LGTM, some minor nits here https://github.com/ava-labs/subnet-evm/pull/501/files

PR #461, opened 2023-01-26
nil timestamp in genesis should be upgradeable

darioush:
## Why this should be merged
`nil` timestamp does not enable the precompile at genesis, so it should be upgradeable.

## How this works
Checks for `nil` timestamp explicitly

## How this was tested
UT


holisticode (approved):

params/precompile_config.go
```diff
@@ -129,7 +129,7 @@ func (c *ChainConfig) verifyPrecompileUpgrades() error {
 			disabled     bool
 		)
 		// check the genesis chain config for any enabled upgrade
-		if config, ok := c.PrecompileUpgrade.getByKey(key); ok {
+		if config, ok := c.PrecompileUpgrade.getByKey(key); ok && config.Timestamp() != nil {
```
aaronbuchwald, line 132: Should this return an error if there is a nil block timestamp rather than skipping?

aaronbuchwald (approved):

params/precompile_config_test.go
```diff
@@ -59,6 +59,25 @@ func TestVerifyWithChainConfig(t *testing.T) {
...
 }
 
+func TestVerifyWithChainConfigAtNilTimestamp(t *testing.T) {
+	admins := []common.Address{{1}}
+	baseConfig := *SubnetEVMDefaultChainConfig
+	config := &baseConfig
+	config.PrecompileUpgrade = PrecompileUpgrade{
+		// this does NOT enable the precompile, so it should be upgradeable.
+		TxAllowListConfig: precompile.NewTxAllowListConfig(nil, nil, nil),
+	}
```
ceyonur, line 69: Should we add `require.False(t, config.IsTxAllowList(common.Big0))`after this, just in case?

ceyonur (approved):

aaronbuchwald (approved):

PR #467, opened 2023-01-29
Fixes #466 Updates rjeczalik/notify 0.9.2 -> 0.9.3 because it solves a deprecation on MacOS Ventura

### LINKED ISSUE #466: Deprecation-Warning when building on MacOS Ventura (notify-library), PR attached
With MacOS Ventura, a deprecation warning appears when building with [github.com/rjeczalik/notify.git](https://github.com/rjeczalik/notify/issues/216) 0.9.2

0.9.3 already contains the fix for that.  I am going to submit PR as well.

Description: https://github.com/rjeczalik/notify/issues/212

itinance:
Fixes #466 

This PR upgrades the library [https://github.com/rjeczalik/notify](https://github.com/rjeczalik/notify) from 0.9.2 to 0.9.3, which fixes MacOS deprecation and eliminates deprecated `io/ioutils`.

Description for the fix of notify-library: https://github.com/rjeczalik/notify/issues/212

## Why this should be merged

- It fixes deprecates on MacOS 13 and added necessary changes for future MacOS-versions
- it gets rid of now-deprecated Go-library `io/ioutils`

## How this works

The fix for 0.9.2 can be found here: https://github.com/rjeczalik/notify/pull/215, adding some necessary changes for future-MacOS versions.
It also includes https://github.com/rjeczalik/notify/pull/214, which gets rid of now-deprecated Go-library `io/ioutils`

## How this was tested

- unit tests can be found at https://github.com/rjeczalik/notify
- I compiled a EVM-subnet with and without this commit on MacOS Ventura after cleaning the Go Cache, and can confirm that the deprecate-warning disappears with this


itinance:
signed and `go mod tidy`

darioush:
Thanks for your PR.
It seems upstream has taken a slightly different path: https://github.com/ethereum/go-ethereum/pull/26176
Once the next version of go-ethereum is released, these changes will reflected in subnet-evm soon.

itinance:
I see. Pls feel free to close this PR then if we want to go with next upstream-update instead.

PR #473, opened 2023-01-31
add nil checks

ceyonur:
## Why this should be merged

Prevents panic in fee config

## How this works

adds nil checks 

## How this was tested

added UT


darioush (approved):

aaronbuchwald (approved):
LGTM

PR #475, opened 2023-02-01
Get signature endpoint

### LINKED ISSUE #446: Create a new API endpoint and client (see plugin/evm/service.go and plugin/evm/client.go) to request and gather signatures (this can be replaced by a relayer that plugs into VM level App messaging)

cam-schultz:
## Why this should be merged
Adds an endpoint `warp_getSignature` that returns the signature for a given message ID from the warp backend. This will likely be used primarily for testing purposes, but could act as an alternative to VM AppRequests.

Fixes #446 

## How this works
Adds a new `WarpAPI` and method `GetSignature`, and a corresponding client `WarpClient` and method.

## How this was tested
CI tests. Manual testing via CURL and a standalone app that uses `WarpClient`


plugin/evm/service.go
```diff
@@ -43,3 +45,17 @@ func (api *SnowmanAPI) IssueBlock(ctx context.Context) error {
 	api.vm.builder.signalTxsReady()
 	return nil
 }
+
+func (api *SnowmanAPI) GetSignature(ctx context.Context, signatureRequest *message.SignatureRequest) (*message.SignatureResponse, error) {
```
cam-schultz, line 49: I used the methods in [plugin/evm/admin](https://github.com/ava-labs/subnet-evm/blob/master/plugin/evm/admin.go) as examples that accept and return JSON. I also [this](https://github.com/ava-labs/subnet-evm/blob/master/rpc/service.go#L125) comment, but was unable to locate the referenced documentation.

plugin/evm/vm.go
```diff
@@ -208,6 +209,9 @@ type VM struct {
 	// State sync server and client
 	StateSyncServer
 	StateSyncClient
+
+	// AWM backend
+	backend warp.WarpBackend
```
cam-schultz, line 214: This is not currently initialized.

plugin/evm/client.go
```diff
@@ -72,3 +75,9 @@ func (c *client) GetVMConfig(ctx context.Context) (*Config, error) {
 	err := c.requester.SendRequest(ctx, "admin.getVMConfig", struct{}{}, res)
 	return res.Config, err
 }
+
+func (c *client) GetSignature(ctx context.Context, signatureRequest message.SignatureRequest) (*[bls.SignatureLen]byte, error) {
+	res := &message.SignatureResponse{}
+	err := c.requester.SendRequest(ctx, "snowman.getSignature", &signatureRequest, res)
+	return &res.Signature, err
```
aaronbuchwald, line 82: It seems like this is using the same client struct to hit a different API, but you would need to create two different clients one for the admin API and one in order to use the `GetSignature` API.

If we are creating a new client, I don't think it should be the case that one of the functions does not work and you have to initialize it differently.

Can we split this out into an admin client and a separate `GetSignature` or `Warp` API ?

plugin/evm/service.go
aaronbuchwald: Can we make this a separate service and add it under the extension `/warp` ?

plugin/evm/vm.go
aaronbuchwald: Can we initialize this?

plugin/evm/client.go
aaronbuchwald: Can we also change this to `warp.getSignature` instead of `snowman` ?

aaronbuchwald:
We should test that this endpoint correctly returns an error both via a CURL request and through the implemented `client.go`.

Looking at the implementation, I don't think it will right now because it uses `message.SignatureRequest` which does not include JSON tags.

plugin/evm/warp_service.go
```diff
@@ -0,0 +1,27 @@
...
+package evm
+
+import (
+	"context"
+
+	"github.com/ava-labs/subnet-evm/plugin/evm/message"
+	"github.com/ethereum/go-ethereum/log"
+)
+
+// SnowmanAPI introduces snowman specific functionality to the evm
```
minghinmatthewlam, line 13: nit: update comment for WarpAPI

plugin/evm/warp_client.go
```diff
@@ -0,0 +1,38 @@
...
+
+// Client implementation for interacting with EVM [chain]
+type warpClient struct {
+	requester rpc.EndpointRequester
+}
+
+// NewClient returns a Client for interacting with EVM [chain]
+func NewWarpClient(uri, chain string) WarpClient {
+	return &warpClient{
+		requester: rpc.NewEndpointRequester(fmt.Sprintf("%s/ext/bc/%s/warp", uri, chain)),
```
minghinmatthewlam, line 30: nit: where is this endpoint from? Do we need the same chainID twice in the endpoint or does one work as well?

cam-schultz:
I tested this endpoint both with CURL and the warp client struct. I modeled the API after `internal/ethapi/api.go`, which has imposes following properties:
- The method syntax is `warp_getSignature`, delimited by an underscore
- The endpoint is `/rpc`
- The parameters are positional, and the `ids.ID` parameter of `GetSignature` should be `cb58` encoded

plugin/evm/warp_client.go
cam-schultz: The implementation has been updated since you left this comment, but the endpoint is the same as the eth rpc API, i.e. `/rpc`. Methods are denoted by `warp_methodName` in the `method` field.

plugin/evm/warp_client.go
```diff
@@ -0,0 +1,57 @@
...
+// warpClient implementation for interacting with EVM [chain]
+type warpClient struct {
+	client *rpc.Client
+}
+
+// NewWarpClient returns a WarpClient for interacting with EVM [chain]
+func NewWarpClient(uri, chain string) (WarpClient, error) {
+	client, err := rpc.Dial(fmt.Sprintf("%s/ext/bc/%s/rpc", uri, chain))
+	if err != nil {
+		log.Error("failed to dial client")
```
aaronbuchwald, line 34: Can we remove all of the logs from this file in favor of returning wrapped errors with the necessary details to the caller?

plugin/evm/warp_service.go
```diff
@@ -0,0 +1,29 @@
...
+
+// WarpAPI introduces snowman specific functionality to the evm
+type WarpAPI struct{ vm *VM }
+
+// GetSignature returns the BLS signature associated with a messageID. In the raw request, [messageID] should be cb58 encoded
+func (api *WarpAPI) GetSignature(ctx context.Context, messageID ids.ID) (*message.SignatureResponse, error) {
+	signature, err := api.vm.backend.GetSignature(ctx, messageID)
+	if err != nil {
+		log.Debug("Unknown warp signature requested", "messageID", messageID)
+		return nil, nil
```
aaronbuchwald, line 22: This should return an error and not emit a log

plugin/evm/warp_service.go
```diff
@@ -0,0 +1,29 @@
...
+	}
+
+	response := message.SignatureResponse{
+		Signature: signature,
+	}
+	return &response, nil
```
aaronbuchwald, line 28: Can we change the return type here so that it will be human readable?

This currently gives the JSON response:

```json
{
    "jsonrpc": "2.0",
    "id": 1,
    "result": {
        "Signature": [
            65,
            87,
            66,
            71,
            153,
            164,
            180,
            109,
            4,
            214,
            33,
            162,
            250,
            6,
            9,
            215,
            58,
            247,
            80,
            1,
            123,
            239,
            57,
            185,
            243,
            98,
            51,
            76,
            236,
            227,
            127,
            177,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0,
            0
        ]
    }
}
```

To get this I modified the code here to be:

```go
// GetSignature returns the BLS signature associated with a messageID. In the raw request, [messageID] should be cb58 encoded
func (api *WarpAPI) GetSignature(ctx context.Context, messageID ids.ID) (*message.SignatureResponse, error) {
	// signature, err := api.vm.backend.GetSignature(ctx, messageID)
	// if err != nil {
	// 	log.Debug("Unknown warp signature requested", "messageID", messageID)
	// 	return nil, nil
	// }

	var signature [96]byte
	copy(signature[:], messageID[:])

	response := message.SignatureResponse{
		Signature: signature,
	}
	return &response, nil
}
```

This should return a hex format. I would recommend wrapping the bytes with `hexutil.Bytes` as it's used in some of the other eth RPC services to wrap bytes.

cam-schultz: Fixed. Confirmed that the response is in hex format.

darioush:
https://github.com/ava-labs/subnet-evm/pull/502/files nits

aaronbuchwald:
Closing in favor of #507

PR #481, opened 2023-02-03
remove current rules

ceyonur:
## Why this should be merged

Removes unnecessary function that would cause misunderstanding.

## How this works

Removes unnecessary function

## How this was tested

A successful compilation should cover this.


aaronbuchwald (approved):
LGTM

darioush (approved):

PR #482, opened 2023-02-03
suggested alternative to modifying TxPool

darioush:


miner/worker.go
```diff
@@ -213,6 +216,32 @@ func (w *worker) commitNewWork(predicateContext *precompile.PredicateContext) (*
 	return w.commit(env)
 }
 
+type txsBySender map[common.Address]types.Transactions
```
aaronbuchwald, line 219: Do we need this type alias?

PR #485, opened 2023-02-03
Warp precompile

minghinmatthewlam:


contract-examples/contracts/IWarpMessenger.sol
```diff
@@ -0,0 +1,43 @@
...
+    // getVerifiedWarpMessage parses the message in the predicate storage slots as a Warp Message.
+    // This message is then delivered on chain by performing evm.Call with the Warp precompile as the caller,
+    // the destinationAddress as the receiver.
+    // The full message and a boolean indicating if the operation executed successfully is returned to the caller.
+    function getVerifiedWarpMessage(uint256 messageIndex)
+        external
+        returns (WarpMessage calldata message, bool success);
+
+    function getBlockchainID() external view returns (bytes32 blockchainID);
+}
```
aaronbuchwald, line 43: nit: add new line

core/stateful_precompile_test.go
```diff
@@ -1286,3 +1289,175 @@ func TestRewardManagerRun(t *testing.T) {
...
+			input: func() []byte {
+				input, err := precompile.PackGetBlockchainID()
+				require.NoError(t, err)
+
+				return input
+			},
+			suppliedGas: precompile.GetBlockchainIDGasCost,
+			readOnly:    true,
+			expectedErr: vmerrs.ErrWriteProtection.Error(),
+		},
```
aaronbuchwald, line 1373: Is there any reason not to allow this function to be marked as `view`? I think this should be allowed in a read only state.

contract-examples/contracts/IWarpMessenger.sol
```diff
@@ -0,0 +1,43 @@
...
+    function sendWarpMessage(
+        bytes32 destinationChainID,
+        bytes32 destinationAddress,
+        bytes calldata payload
+    ) external;
+
+    // getVerifiedWarpMessage parses the message in the predicate storage slots as a Warp Message.
+    // This message is then delivered on chain by performing evm.Call with the Warp precompile as the caller,
+    // the destinationAddress as the receiver.
+    // The full message and a boolean indicating if the operation executed successfully is returned to the caller.
```
aaronbuchwald, line 37: can we update this comment to reflect that this no longer performs a call operation but just returns the message directly to the caller?

precompile/warp_messenger.go
```diff
@@ -0,0 +1,425 @@
...
+	for _, messageBytes := range messagesBytes {
+		message, err := teleporter.ParseMessage(messageBytes)
+		if err != nil {
+			return err
+		}
+
+		// TODO: Should we add a special chain ID that is allowed as the "anycast" chain ID? Just need to think through if there are any security implications.
+		if message.DestinationChainID != predicateContext.SnowCtx.ChainID {
+			return ErrWrongChainID
+		}
```
aaronbuchwald, line 190: I think the answer is yes. There are definitely security implications since this will need to be checked in many cases.

We may have a better and less error prone UI if we create a separate function to verify an AnyCast message.

precompile/warp_messenger.go
```diff
@@ -0,0 +1,425 @@
...
+
+		err = message.Signature.Verify(
+			context.Background(),
+			&message.UnsignedMessage,
+			predicateContext.SnowCtx.ValidatorState,
+			predicateContext.ProposerVMBlockCtx.PChainHeight,
+			WarpQuorumNumerator,
+			WarpQuorumDenominator)
```
aaronbuchwald, line 198: Do we want these values to be parameterizable and set in the WarpPrecompile config?

We could also provide default values in that case and add to the `Verify` function to validate that the parameters are valid.

precompile/warp_messenger.go
```diff
@@ -0,0 +1,425 @@
...
+		if err != nil {
+			return err
+		}
+		log.Info("Warp message verification passed.")
+	}
+
+	log.Info("Warp precompile predicate passed.")
```
aaronbuchwald, line 205: We should either remove these logs or change them to debug.

precompile/warp_messenger.go
michaelkaplan13: This is carry-over comment from my initial implementation I believe.

For the current design, we should remove this check on the destination chain ID entirely. The value of the destination chain ID is returned as part of the message from `getVerifiedWarpMessage()`, and it is left up to the the Teleporter contract (or other dApps) to verify its value. We added `getBlockchainID` to allow dApps to check if a message was sent to this specific chain from within solidity.

precompile/warp_messenger.go
```diff
@@ -0,0 +1,425 @@
...
+	// You can use unpacked [messageIndex] variable in your code
+	inputIndex, err := UnpackGetVerifiedWarpMessageInput(input)
+	if err != nil {
+		return nil, remainingGas, err
+	}
+
+	storageSlots, exists := accessibleState.GetStateDB().GetPredicateStorageSlots(WarpMessengerAddress)
+	if !exists || storageSlots == nil {
+		return nil, remainingGas, ErrMissingStorageSlots
+	}
```
aaronbuchwald, line 290: what's the case where exists is true and storage slots is nil? If this is necessary, then we should fix the behavior in `GetPredicateStorageSlots` so that it's not required.

precompile/warp_messenger.go
```diff
@@ -0,0 +1,425 @@
...
+
+	// Parse the raw message to be processed.
+	signedMessage := signedMessages[messageIndex]
+	message, err := teleporter.ParseMessage(signedMessage)
+	if err != nil {
+		return nil, remainingGas, err
+	}
+
+	// Charge gas per validator included in the aggregate signature
```
michaelkaplan13, line 314: @aaronbuchwald correct me if I'm wrong here, but we'll likely want to remove the gas charge per signature aggregated from the precompile. Instead, Aaron mentioned handling it along with the intrinsic gas of the transaction. The public key aggregation and signature verification occurs whether or not the transaction actually calls `getVerifiedWarpMessage`, so charging here doesn't prevent that DOS vector.

precompile/warp_messenger.go
```diff
@@ -0,0 +1,425 @@
...
+	if readOnly {
+		return nil, remainingGas, vmerrs.ErrWriteProtection
+	}
+	// attempts to unpack [input] into the arguments to the GetVerifiedWarpMessageInput.
+	// Assumes that [input] does not include selector
+	// You can use unpacked [messageIndex] variable in your code
+	inputIndex, err := UnpackGetVerifiedWarpMessageInput(input)
+	if err != nil {
+		return nil, remainingGas, err
+	}
```
aaronbuchwald, line 285: Since this includes a variable length input and we are going to parse it, which will be an O(n) operation, we need to charge an appropriate amount of gas to account for that prior to unpacking/parsing.

We will also need to charge gas for the decoding of the messages as this is currently implemented. One alternative approach that may make this more performant is to parse the messages once in the `Predicate`, charge inside of `IntrinsicGas` for that computation before the transaction execution begins, and that way when we call this function, we do not need to re-do the parsing and charge the O(n) cost again.

precompile/warp_messenger.go
minghinmatthewlam: As part of warp, shouldn't we provide the check that the warp message gets delivered onto the right chain, instead as part of the teleporter contract?

minghinmatthewlam: I guess in terms of extensibility, if we don't have this check, supposedly a dapp developer could send warp message to not only the specified destination chain if needed.

core/stateful_precompile_test.go
minghinmatthewlam: updated `getBlockchainID` and `getVerifiedWarpMessage` to be marked as view and allowed in a read only

contract-examples/contracts/IWarpMessenger.sol
minghinmatthewlam: updated comment

precompile/warp_messenger.go
minghinmatthewlam: removed

precompile/warp_messenger.go
minghinmatthewlam: there shouldn't be a case with `GetPredicateStorageSlots` so removed the extra `storageSlots == nil` check

precompile/warp_messenger.go
```diff
@@ -0,0 +1,425 @@
...
+	)
+}
+
+func getVerifiedWarpMessage(accessibleState PrecompileAccessibleState, caller common.Address, addr common.Address, input []byte, suppliedGas uint64, readOnly bool) (ret []byte, remainingGas uint64, err error) {
+	if remainingGas, err = deductGas(suppliedGas, GetVerifiedWarpMessageGasCost); err != nil {
+		return nil, 0, err
+	}
+	if readOnly {
+		return nil, remainingGas, vmerrs.ErrWriteProtection
+	}
```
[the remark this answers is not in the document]
minghinmatthewlam: updated to `view` and allowed for `readOnly`

precompile/warp_messenger.go
minghinmatthewlam: Added an `InitialWarpConfig` that should parameterize the quorum numerator and denominator values, checking for invalid cases, and has a default of 67% threshold.

precompile/warp_messenger.go
```diff
@@ -0,0 +1,444 @@
...
+	QuorumNumerator   uint64 `json:"quorumNumerator,omitempty"`
+	QuorumDenominator uint64 `json:"quorumDenominator,omitempty"`
+}
+
+func (i *InitialWarpConfig) Verify() error {
+	switch {
+	case i.QuorumDenominator == uint64(0):
+		return ErrInvalidQuorumDenominator
+	case i.QuorumNumerator > i.QuorumDenominator:
+		return ErrGreaterQuorumNumerator
```
aaronbuchwald, line 96: Should we add stricter requirements here? Min 1/2 or 2/3? What params are we comfortable with here cc @michaelkaplan13

precompile/warp_messenger.go
```diff
@@ -0,0 +1,444 @@
...
+type SendWarpMessageInput struct {
+	DestinationChainID [32]byte
+	DestinationAddress [32]byte
+	Payload            []byte
+}
+
+type InitialWarpConfig struct {
+	QuorumNumerator   uint64 `json:"quorumNumerator,omitempty"`
+	QuorumDenominator uint64 `json:"quorumDenominator,omitempty"`
+}
```
aaronbuchwald, line 89: The prefix `Initial` has been used when we want to take an initial action like mint funds or add specific addresses to the allow list. Can we remove this and include these parameters directly in `WarpConfig` ?

precompile/warp_messenger.go
```diff
@@ -0,0 +1,444 @@
...
+		return false
+	}
+
+	return i.QuorumNumerator == other.QuorumNumerator && i.QuorumDenominator == other.QuorumDenominator
+}
+
+func (i *InitialWarpConfig) Configure(state StateDB) {
+	WarpQuorumNumerator = i.QuorumNumerator
+	WarpQuorumDenominator = i.QuorumDenominator
+}
```
aaronbuchwald, line 113: This is incorrect. We cannot just set global variables when `Configure` is called. `Configure` is called while verifying a block, which is before it gets accepted and there may be multiple different blocks in processing that could trigger `Configure` getting called.

If we need to use this in the future, we should store it in the state in the same way that we do for the other precompiles when they save config information.

minghinmatthewlam:
Converting to draft, and eventually closing in favor of a warp precompile branch with the refactored precompile structure

precompile/warp_messenger.go
michaelkaplan13: I think leaving this up for a subnet to define the minimum threshold themselves is probably the best route. We can make clear the default is >2/3 of stake. It would be reasonable to go all the way down to >1/3 IMO, since that would still be able to withstand 1/3 byzantine stake on the other subnet. 

Ultimately though, I don't think there is any reason for us to impose a number here, other than <= 1 since that would be invalid.

Issue #483, opened 2023-02-03
Stateful precompile tutorial setup error


kevinz917:
**Describe the bug**
Experienced 2 bugs by following the tutorial here:
https://docs.avax.network/subnets/hello-world-precompile-tutorial#step-0-generating-the-precompile

The first error happens through downloading the repo. It gives a "SSH permission not found" error
![image](https://user-images.githubusercontent.com/40219813/216677725-da6ee42f-10d0-4667-9414-e86b517ab2b9.png)

The second error shows after following this step
![image](https://user-images.githubusercontent.com/40219813/216677417-7c6a988f-6875-4314-acbb-0961f8844604.png)

Error logs
![image](https://user-images.githubusercontent.com/40219813/216677460-1b63c09e-bb7e-4407-b0fc-1ed624b7bddb.png)

**Operating System**
Macbook M1 on macOS Montery

**Additional context**

aaronbuchwald:
Thanks for putting this up. This is a result of the CGO flags that must be set when using the blst library that is imported by AvalancheGo. When building subnet-evm, we need to set this flag to ensure that it builds correctly:

https://github.com/ava-labs/subnet-evm/blob/master/scripts/constants.sh#L44

If you're running into this error, you can fix it by setting this flag when running it.

```bash
CGO_CFLAGS="-O -D__BLST_PORTABLE__" go run ./cmd/precompilegen/main.go <rest of the flags here>
```

We will work on fixing this so that you don't need to populate a variable like this in order to get it to work.

aaronbuchwald:
Hey Kevin, I made a PR to create a new script so that you do not need to set the `CGO_FLAGS` manually.

The subnet-evm PR to add the new script is here: https://github.com/ava-labs/subnet-evm/pull/489 and there is an accompanying docs PR that updates the PrecompileGen tutorial here if you want to take a look: https://github.com/ava-labs/avalanche-docs/pull/1196.

Our docs repo auto-generates a preview URL, so you can check out the relevant changed docs here: https://7dda7439.avalanche-docs.pages.dev/subnets/hello-world-precompile-tutorial#step-0-generating-the-precompile. To test it, you will just need to checkout the related branch of subnet-evm called `precompile-gen-script`.

So the first step would change to:

```bash
cd $GOPATH/src/github.com/ava-labs/subnet-evm
git checkout precompile-gen-script
git checkout -b hello-world-stateful-precompile
```

Please let me know if this fixes the issue for you.

PR #489, opened 2023-02-05
Add create precompile script to set CGO flag before running precompilegen script

aaronbuchwald:
## Why this should be merged

This PR should be merged in order to ensure that the CGO flags required by the blst library imported in AvalancheGo are set as required.

This PR addresses: https://github.com/ava-labs/subnet-evm/issues/483

## How this works

This PR creates a new script `./scripts/create_precompile.sh` which sets the required constants and then runs the same command that `go run ...` command that was previously required to generate a precompile stub.

## How this was tested

This was tested by completing step 0 of the PrecompileGen tutorial with one modification included in this PR: https://github.com/ava-labs/avalanche-docs/pull/1196

## How is this documented

Change to PrecompileGen tutorial is documented here: https://github.com/ava-labs/avalanche-docs/pull/1196

### DIFF
--- a/scripts/generate_precompile.sh
+++ b/scripts/generate_precompile.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+set -e
+
+# This script generates a Stateful Precompile stub based off of a Solidity ABI file.
+# It first sets the necessary CGO_FLAGs for the BLST library used in AvalancheGo and
+# then runs PrecompileGen.
+if ! [[ "$0" =~ scripts/generate_precompile.sh ]]; then
+  echo "must be run from repository root, but got $0"
+  exit 255
+fi
+
+# Load the versions
+SUBNET_EVM_PATH=$(
+  cd "$(dirname "${BASH_SOURCE[0]}")"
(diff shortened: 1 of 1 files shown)


scripts/create_precompile.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+set -e
+
+# This script starts N nodes (TODO N instead of 5) and waits for ctrl-c to shutdown the process group of AvalancheGo processes
+# Uses data directory to store all AvalancheGo data neatly in one location with minimal config overhead
```
ceyonur, line 5: can we fix comment?

ceyonur:
can we also change the script name to `generate_precompile.sh` to align it with the precompilegen?

aaronbuchwald:
> can we also change the script name to `generate_precompile.sh` to align it with the precompilegen?

Updated here and in the docs PR, good call

darioush (approved):

ceyonur (approved):

anusha-ctrl (approved):

PR #490, opened 2023-02-06
State upgrades

friskyfoxdk:
## Why this should be merged

After discussion with @aaronbuchwald relating to https://github.com/ava-labs/subnet-evm/pull/451, we determined that it would be useful to have a new state upgrade system get built to complement the precompile system.

A State Upgrade is similar to a precompile in that it can do things that aren't usually possible via smart contracts alone. Because these actions are so powerful, it makes the most sense for them to be applied atomically during a network upgrade, rather than having an ongoing mechanism in place to 'call' them. This reduces the security risks and the chances of accidental mistakes.

This system would differ from the Precompile system in the following ways:

1. State Upgrades happen atomically, and as such do not have a notion of being enabled or disabled. Rather, they are executed at a given timestamp.
2. Because State Upgrades are not continuously enabled, there is no need for a callable address or methods.

This system will cover 4 basic state upgrades, as follows:

1. Modifying the code of a given contract.
2. Modifying the value of a storage slot for a given address.
3. Modifying the balance of a given account.
4. Deploying a contract to a given address.

## How this works

State Upgrades can be configured to execute at a given timestamp, in the upgrade.json file, very similarly to Precompiles. When the block is mined where that timestamp has passed, the State Upgrade that is configured will be executed.

## How this was tested

There need to be tests written for each type of state upgrade. Currently the only test is for code upgrading.

## How is this documented

There needs to be more documentation added for this feature, currently only the code documents the feature.

### DIFF
--- a/contract-examples/contracts/ExampleCodeUpgradeAfter.sol
+++ b/contract-examples/contracts/ExampleCodeUpgradeAfter.sol
@@ -1,0 +1,19 @@
+//SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+// ExampleContractAfter shows how the CodeUpgrader precompile can be used to upgrade the code of a smart contract.
+contract ExampleCodeUpgraderAfter {
+    // Make a 6 gap variable.
+    uint256[7] private ______gap;
+
+    // Cap is now at slot 0x7.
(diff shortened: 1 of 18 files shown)


friskyfoxdk:
Closing this in favor of https://github.com/ava-labs/subnet-evm/pull/549

PR #493, opened 2023-02-06
Break cyclic dependency between params and precompile packages

aaronbuchwald:
Creating this draft PR to discuss potential re-design alongside the precompiles as modules change and break up the cyclic dependency between `params/` and `precompile/` packages.

This branch has removed some precompile packages completely, so it is not intended to be merged.


aaronbuchwald:
Closing because this was only opened as an example of how to break the cyclic dependency with the precompile refactor.

PR #495, opened 2023-02-07
Signature handler refactor

aaronbuchwald:
## Why this should be merged

This PR re-organizes the stats and handlers and reduces the duplicated stats code.

## How this works

This PR uses the embedded stats and handlers rather than duplicating function calls to do many passthroughs to the underlying stats.

## How this was tested

This PR is tested in CI

## How is this documented

no documentation necessary for this change

### DIFF
--- a/handlers/warp/stats/stats.go
+++ b/handlers/warp/stats/stats.go
@@ -1,0 +1,95 @@
+// (c) 2023, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package stats
+
+import (
+	"sync"
+	"time"
+
+	"github.com/ava-labs/subnet-evm/metrics"
+)
+
+var (
+	_ SignatureRequestHandlerStats = (*handlerStats)(nil)
+	_ SignatureRequestHandlerStats = (*MockSignatureRequestHandlerStats)(nil)
+)
+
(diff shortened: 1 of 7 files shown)


handlers/warp/stats/stats.go
```diff
@@ -0,0 +1,95 @@
...
+
+import (
+	"sync"
+	"time"
+
+	"github.com/ava-labs/subnet-evm/metrics"
+)
+
+var (
+	_ SignatureRequestHandlerStats = (*handlerStats)(nil)
```
minghinmatthewlam, line 14: Trying to figure out difference between `(*handlerStats)(nil)` and `&handlerStats{}`, is it the difference between nil and a default struct value?

Should we always use `(*struct)(nil)` i'm thinking for `MockHandlerStats` etc

minghinmatthewlam (approved):

handlers/warp/stats/stats.go
aaronbuchwald: They're both used to ensure that the struct implements the interface.

I've switched to using `(*type)(nil)`. It doesn't make much of a difference and tbh we should be consistent, but currently use both in different places.

Here's one article: https://splice.com/blog/golang-verify-type-implements-interface-compile-time/.

PR #498, opened 2023-02-07
build with avago v1.9.8

cam-schultz:
## Why this should be merged
Fixes build conflicts with Avalanchego v1.9.8

## How this works
Provides a type specialization on the LRU cache used by the gossiper and the warp backend

## How this was tested
CI

### DIFF
--- a/plugin/evm/gossiper.go
+++ b/plugin/evm/gossiper.go
@@ -62,7 +62,7 @@ type pushGossiper struct {
 
 	// [recentTxs] prevent us from over-gossiping the
 	// same transaction in a short period of time.
-	recentTxs *cache.LRU
+	recentTxs *cache.LRU[common.Hash, interface{}]
 
 	codec  codec.Manager
 	signer types.Signer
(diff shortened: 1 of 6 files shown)


plugin/evm/gossiper.go
```diff
@@ -62,7 +62,7 @@ type pushGossiper struct {
 
 	// [recentTxs] prevent us from over-gossiping the
 	// same transaction in a short period of time.
-	recentTxs *cache.LRU
+	recentTxs *cache.LRU[common.Hash, interface{}]
```
darioush, line 65: @aaronbuchwald not sure if you prefer `struct{}` here instead of `interface{}` (then we can put struct{}{} instead of nil).

aaronbuchwald: Ya let's use `struct{}{}` here

aaronbuchwald (approved):
LGTM

minghinmatthewlam (approved):

darioush (approved):
https://github.com/ava-labs/coreth-internal/pull/1144#pullrequestreview-1288375684 equivalent PR to coreth

PR #506, opened 2023-02-09
Warp precompile on accept

minghinmatthewlam:
## Why this should be merged
Plumbing for vm backend to the precompiles, so far being the warp backend used for the warp precompile's `OnAccept` call to add messages. Issue #447 related.

## How this works
- Create a new precompile backend that implements warp backend
- Created a `NoopBackend` while warp is not initialized to pass down
- For all updated test signatures will pass `nil` for backend, which then gets changed into a `noopBackend`
- Warp precompile `onAccept` checks transaction log topics, then uses warp backend to add message.

## How is this documented
- comments

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -304,10 +304,6 @@ func NewBlockChain(db ethdb.Database, cacheConfig *CacheConfig, chainConfig *par
 	if cacheConfig == nil {
 		return nil, errCacheConfigNotSpecified
 	}
-	// TODO: This is added to handle the nil backends passed in from test, when warp is initialized we should remove
-	if backend == nil {
-		backend = precompile.NewNoopBackend()
-	}
 	bodyCache, _ := lru.New(bodyCacheLimit)
 	receiptsCache, _ := lru.New(receiptsCacheLimit)
 	blockCache, _ := lru.New(blockCacheLimit)
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -48,7 +48,7 @@ import (
 	"github.com/ava-labs/subnet-evm/ethdb"
 	"github.com/ava-labs/subnet-evm/metrics"
 	"github.com/ava-labs/subnet-evm/params"
-	"github.com/ava-labs/subnet-evm/plugin/evm"
+	"github.com/ava-labs/subnet-evm/precompile"
 	"github.com/ava-labs/subnet-evm/trie"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/event"
@@ -248,7 +248,7 @@ type BlockChain struct {
 	prefetcher Prefetcher // Block state prefetcher interface
 	processor  Processor  // Block transaction processor interface
 	vmConfig   vm.Config
-	vmBackend  evm.Backend
+	vmBackend  precompile.Backend
 
 	badBlocks *lru.Cache // Bad block cache
 
@@ -300,10 +300,7 @@ type BlockChain struct {
 // NewBlockChain returns a fully initialised block chain using information
 // available in the database. It initialises the default Ethereum Validator and
 // Processor.
-func NewBlockChain(
-	db ethdb.Database, cacheConfig *CacheConfig, chainConfig *params.ChainConfig, engine consensus.Engine,
-	vmConfig vm.Config, lastAcceptedHash common.Hash, backend evm.Backend,
-) (*BlockChain, error) {
+func NewBlockChain(db ethdb.Database, cacheConfig *CacheConfig, chainConfig *params.ChainConfig, engine consensus.Engine, vmConfig vm.Config, lastAcceptedHash common.Hash, backend precompile.Backend) (*BlockChain, error) {
 	if cacheConfig == nil {
 		return nil, errCacheConfigNotSpecified
 	}
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -304,6 +304,10 @@ func NewBlockChain(db ethdb.Database, cacheConfig *CacheConfig, chainConfig *par
 	if cacheConfig == nil {
 		return nil, errCacheConfigNotSpecified
 	}
+	// TODO: This is added to handle the nil backends passed in from test, when warp is initialized we should remove
+	if backend == nil {
+		backend = precompile.NewNoopBackend()
+	}
 	bodyCache, _ := lru.New(bodyCacheLimit)
 	receiptsCache, _ := lru.New(receiptsCacheLimit)
 	blockCache, _ := lru.New(blockCacheLimit)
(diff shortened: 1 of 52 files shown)


minghinmatthewlam:
test file changes are due to function signature change for `NewBlockchain`, which now pass `nil` in the test for the backend, and then has a check in `core/blockchain.go` to create a `NewNoopBackend` to handle nil backends.

precompile/backend.go
```diff
@@ -0,0 +1,33 @@
...
+	return &noopBackend{}
+}
+
+func (n noopBackend) AddMessage(ctx context.Context, unsignedMessage *teleporter.UnsignedMessage) error {
+	return nil
+}
+
+func (n noopBackend) GetSignature(ctx context.Context, messageHash ids.ID) ([]byte, error) {
+	return nil, nil
+}
```
aaronbuchwald, line 33: Can we remove this? I believe we only need `AddMessage` here

precompile/backend.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/vms/platformvm/teleporter"
+	"github.com/ava-labs/subnet-evm/plugin/evm/warp"
+)
+
+var _ Backend = (*noopBackend)(nil)
+
+// Backend defines the interface for precompiles to interact with vm backends.
+type Backend interface {
+	warp.WarpBackend
```
aaronbuchwald, line 18: Can we update this to just `AddMessage` instead of embedding `warp.WarpBackend` since we don't need the full interface.

precompile/warp_messenger.go
```diff
@@ -252,7 +255,27 @@ func (c *WarpMessengerConfig) verifyPredicate(predicateContext *PredicateContext
 // OnAccept optionally returns a function to perform on any log with the precompile address.
 // If enabled, this will be called after the block is accepted to perform post-accept computation.
 func (c *WarpMessengerConfig) OnAccept() OnAcceptFunc {
-	return nil
+	return c.onAccept
+}
+
+func (c *WarpMessengerConfig) onAccept(backend Backend, txHash common.Hash, logIndex int, topics []common.Hash, logData []byte) error {
```
aaronbuchwald, line 261: Note: any error returned from here will be treated as fatal.

precompile/warp_messenger.go
```diff
@@ -252,7 +255,27 @@ func (c *WarpMessengerConfig) verifyPredicate(predicateContext *PredicateContext
...
+	if backend == nil {
+		return ErrMissingPrecompileBackend
+	}
+
+	if len(topics) != 3 || topics[0] != common.HexToHash(SubmitMessageEventID) {
+		return ErrInvalidLogTopics
+	}
```
aaronbuchwald, line 268: Can we split this up so we get a more specific error if something goes wrong and we hit this error?

We should return a formatted error with the number of topics if it's incorrect and the actual hash value of the topic at index 0 if it does not match the expected one.

plugin/evm/vm.go
```diff
@@ -277,6 +279,7 @@ func (vm *VM) Initialize(
 	vm.db = versiondb.New(baseDB)
 	vm.acceptedBlockDB = prefixdb.New(acceptedPrefix, vm.db)
 	vm.metadataDB = prefixdb.New(metadataPrefix, vm.db)
+	vm.precompileBackend = precompile.NewNoopBackend()
```
aaronbuchwald, line 282: Can we pass the actual backend here instead of `NewNoopBackend` ?

I think we should create a mock backend that we can use for testing instead of a noop backend. The difference here would be that a mock backend would allow us to track what operations are performed on it for validation during tests, whereas a no-op backend is just no-ops.

We can specify the actual one instead of a noop backend here, because the warp precompile itself will not be merged until it is ready for production.

core/blockchain.go
```diff
@@ -1053,6 +1057,33 @@ func (bc *BlockChain) Accept(block *types.Block) error {
 	bc.addAcceptorQueue(block)
 	acceptedBlockGasUsedCounter.Inc(int64(block.GasUsed()))
 	acceptedTxsCounter.Inc(int64(len(block.Transactions())))
+	return bc.handlePrecompilePostAccept(block)
+}
+
+// handlePrecompilePostAccept handles executing the post-accept functions of any active precompiles
+// on the logs contained in the block.
+func (bc *BlockChain) handlePrecompilePostAccept(block *types.Block) error {
```
aaronbuchwald, line 1065: Out of scope for this PR, but we need to ensure that this provides the atomicity required by the precompiles and works correctly with how we handle a potential shutdown interrupting this operation, such that we resume and call accept on the same block if the VM did not finish marking it as accepted.

core/blockchain.go
```diff
@@ -298,13 +300,14 @@ type BlockChain struct {
...
-	vmConfig vm.Config, lastAcceptedHash common.Hash,
-) (*BlockChain, error) {
+func NewBlockChain(db ethdb.Database, cacheConfig *CacheConfig, chainConfig *params.ChainConfig, engine consensus.Engine, vmConfig vm.Config, lastAcceptedHash common.Hash, backend precompile.Backend) (*BlockChain, error) {
 	if cacheConfig == nil {
 		return nil, errCacheConfigNotSpecified
 	}
+	// TODO: This is added to handle the nil backends passed in from test, when warp is initialized we should remove
+	if backend == nil {
+		backend = precompile.NewNoopBackend()
+	}
```
aaronbuchwald, line 310: Can we remove this now?

aaronbuchwald: We should define an invariant for when this needs to be populated vs. when it can be nil.

For existing tests that do not touch warp, it should be safe to leave this as nil.

Additionally, if we are re-processing blocks that have already been accepted (in which case we will not call Accept on them and the precompile's `onAccept` functionality), it should also be safe for this to be left nil.

precompile/warp_messenger.go
minghinmatthewlam: updated to separate errors

precompile/backend.go
minghinmatthewlam: updated

plugin/evm/vm.go
minghinmatthewlam: Passing in initialized warp backend to make `precompileBackend`, and created a mock backend with stats for tracking. 

Looking to see how/where I can have a test that creates a new blockchain with `MockBackend`, submits transaction to warp precompile to emit log, blockchain accepts block and calls `handlePrecompilePostAccept`, warp precompile `OnAccept` gets called, check stats for backend.

minghinmatthewlam:
Closing PR for post refactored precompile branch

PR #507, opened 2023-02-09
Get signature endpoint: alternative PR with different packaging

darioush:
## Why this should be merged
Makes warp packages symmetrical with sync and separates stats
We can merge this instead of https://github.com/ava-labs/subnet-evm/pull/475
(basing directly on master to reduce confusion)

## How this works
This branch has 3 commits at the end, each of which I consider an improvement to 475 in file & package structure.
Open to other ideas too

## How this was tested
UT


warp/service/warp_client.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/ava-labs/avalanchego/utils/cb58"
+	"github.com/ava-labs/subnet-evm/plugin/evm/message"
+	"github.com/ava-labs/subnet-evm/rpc"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+)
+
+var _ WarpClient = (*warpClient)(nil)
+
+type WarpClient interface {
+	GetSignature(ctx context.Context, signatureRequest message.SignatureRequest) (*hexutil.Bytes, error)
```
aaronbuchwald, line 19:
```suggestion
	GetSignature(ctx context.Context, messageID ids.ID) ([]byte, error)
```

We should not import the message package at all for this.

warp/service/warp_client.go
```diff
@@ -0,0 +1,51 @@
...
+		client: client,
+	}, nil
+}
+
+// GetSignature requests the BLS signature associated with a messageID
+func (c *warpClient) GetSignature(ctx context.Context, signatureRequest message.SignatureRequest) (*hexutil.Bytes, error) {
+	req, err := cb58.Encode(signatureRequest.MessageID[:])
+	if err != nil {
+		return nil, fmt.Errorf("failed to base58 encode the request. messageID: %s, error: %w", signatureRequest.MessageID, err)
+	}
```
aaronbuchwald, line 43: We can get rid of this after the type change above

warp/service/warp_client.go
```diff
@@ -0,0 +1,51 @@
...
+
+	var res SignatureResponse
+	err = c.client.CallContext(ctx, &res, "warp_getSignature", req)
+	if err != nil {
+		return nil, fmt.Errorf("call to warp_getSignature failed. err: %w", err)
+	}
+	return &res.Signature, err
```
aaronbuchwald, line 50: We should wrap the returned `[]byte` type with `hexutil.Bytes` internal to this function so that it gets marshaled to hex output, but still returns the type `[]byte`.

warp/service/warp_service.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+// WarpAPI introduces snowman specific functionality to the evm
+type WarpAPI struct {
+	Backend WarpBackend
+}
+
+// GetSignature returns the BLS signature associated with a messageID. In the raw request, [messageID] should be cb58 encoded
+func (api *WarpAPI) GetSignature(ctx context.Context, messageID ids.ID) (*SignatureResponse, error) {
+	signature, err := api.Backend.GetSignature(ctx, messageID)
```
aaronbuchwald, line 25: We should get rid of `SignatureResponse` type and just return `[]byte` . If we wrap the `[]byte` type internally with the type `hexutil.Bytes` it should marshal it to a correct hex output instead of returning byte by byte.

warp/service/warp_client.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+// GetSignature requests the BLS signature associated with a messageID
+func (c *warpClient) GetSignature(ctx context.Context, signatureRequest message.SignatureRequest) (*hexutil.Bytes, error) {
+	req, err := cb58.Encode(signatureRequest.MessageID[:])
+	if err != nil {
+		return nil, fmt.Errorf("failed to base58 encode the request. messageID: %s, error: %w", signatureRequest.MessageID, err)
+	}
+
+	var res SignatureResponse
```
aaronbuchwald, line 45: We should be able to change this type to just `[]byte` after making the changes inside of `warp_service.go`

warp/service/backend.go
```diff
@@ -1,7 +1,7 @@
 // (c) 2023, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE for licensing terms.
 
-package warp
+package service
```
aaronbuchwald, line 4: We usually use `service` as the name of an API service, so I think putting the actual Warp Backend here is an outlier.

Could we either move this to a the warp package directly inside of `warp/` or move both the warp backend and the service/client to the same `warp/` package, so they're all just in warp rather than `warpService` ? I'm leaning towards the latter.

warp/warp_service.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/ethereum/go-ethereum/common/hexutil"
+)
+
+// WarpAPI introduces snowman specific functionality to the evm
+type WarpAPI struct {
+	Backend WarpBackend
+}
+
+// GetSignature returns the BLS signature associated with a messageID.
+func (api *WarpAPI) GetSignature(ctx context.Context, messageID ids.ID) (hexutil.Bytes, error) {
```
cam-schultz, line 20: Just confirming that the `params` argument in the JSON request can be hex encoded with this signature. Similarly, can you confirm that returning `hexutil.Bytes` directly gives a hex encoded output?

warp/warp_client.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/subnet-evm/rpc"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+)
+
+var _ WarpClient = (*warpClient)(nil)
+
+type WarpClient interface {
+	GetSignature(ctx context.Context, signatureRequest ids.ID) ([]byte, error)
```
cam-schultz, line 18: nit: can we name the second argument `messageID`?

cam-schultz (commented):
Can you confirm that the moved files are in the expected location in the diff? I ask because https://github.com/ava-labs/subnet-evm/pull/475 at one point had erroneously moved some of these files due to a stale rebase off of https://github.com/ava-labs/subnet-evm/pull/459.

darioush:
@cam-schultz The files are where I think they should be (this PR is against master)
Encoding/Decoding will use the MarshalJSON/UnmarshalJSON methods, which is implemented for hexutil.Bytes (to hex encode with a 0x prefix) and for ids.ID as such:

```go
func (id *ID) UnmarshalJSON(b []byte) error {
	str := string(b)
	if str == nullStr { // If "null", do nothing
		return nil
	} else if len(str) < 2 {
		return errMissingQuotes
	}

	lastIndex := len(str) - 1
	if str[0] != '"' || str[lastIndex] != '"' {
		return errMissingQuotes
	}

	// Parse CB58 formatted string to bytes
	bytes, err := cb58.Decode(str[1:lastIndex])
	if err != nil {
		return fmt.Errorf("couldn't decode ID to bytes: %w", err)
	}
	*id, err = ToID(bytes)
	return err
}
```

warp/warp_service.go
aaronbuchwald: Testing locally with modified code to get a response:

```bash
curl --location --request POST 'http://127.0.0.1:9652/ext/bc/2ioeFdBb3qMEfnTQc9kd3CqXBmTmxqh96GyCY49b8X3UmZx5fZ/rpc' \
--header 'Content-Type: application/json' \
--data-raw '{
    "jsonrpc": "2.0",
    "method": "warp_getSignature",
    "params": [
        "11111111111111111111111111111111LpoYY"
    ],
    "id": 1
}'
```

Output:
```json
{
    "jsonrpc": "2.0",
    "id": 1,
    "result": "0x0000000000000000000000000000000000000000000000000000000000000000"
}
```

with a different response output:
```json
{
    "jsonrpc": "2.0",
    "id": 1,
    "result": "0x6d0d36349f470af56409c02b53d5dfcc1a78071961f456321d9c707489d7116f"
}
```

aaronbuchwald (approved):
LGTM

warp/warp_service.go
cam-schultz: Does `c.client.CallContext(ctx, &res, "warp_getSignature", messageID[:])` in `warpClient.GetSignature` encode the bytes as cb58? I had done that manually in https://github.com/ava-labs/subnet-evm/pull/475, but did not test if it did that cb58 coding automatically.

warp/handlers/stats.go
```diff
@@ -1,7 +1,7 @@
 // (c) 2023, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE for licensing terms.
 
-package stats
```
minghinmatthewlam, line 4: wondering if we should keep the packaging to `stats` for consistency with the sync stats.

warp/warp_service.go
darioush: Yes it needs to be `messageID` so the type has the json encoding attached.

cam-schultz (approved):
LGTM 👍

PR #508, opened 2023-02-10
Cleanup check precompile predicates

aaronbuchwald:
## Why this should be merged

This PR adds some cleanup to the awm precompile interface branch


aaronbuchwald:
Marking this as blocked on https://github.com/ava-labs/subnet-evm/pull/389

plugin/evm/block.go
```diff
@@ -157,17 +160,21 @@ func (b *Block) verify(proposerVMBlockCtx *block.Context, writes bool) error {
 		return fmt.Errorf("syntactic block verification failed: %w", err)
 	}
 
-	// If the chain is not yet bootstrapped, we do not need to verify the transaction predicates
-	// because the block is already included in the chain, so we know that predicates must have
-	// been valid at the time the block was accepted.
+	// Only enforce predicates if the chain has already bootstrapped.
+	// If the chain is still bootstrapping, we can assume that all blocks we are verifying have
+	// been accepted by then network and the predicate was validated by the network when the
```
cam-schultz, line 165:
```suggestion
	// been accepted by the network and the predicate was validated by the network when the
```

core/state/statedb.go
```diff
@@ -1102,3 +1127,13 @@ func (s *StateDB) AddressInAccessList(addr common.Address) bool {
 func (s *StateDB) SlotInAccessList(addr common.Address, slot common.Hash) (addressPresent bool, slotPresent bool) {
 	return s.accessList.Contains(addr, slot)
 }
+
+// GetPredicateStorageSlots returns the storage slots associated with a given address, and whether or not
+// that address was included in the optional access list of the transaction.
+// The storage slots are returned in the same order as they appeared in the transaction.
+// These are the same storage slots that are used to verify any transaction
+// predicates for transactions with access list addresses that match a precompile address.
+func (s *StateDB) GetPredicateStorageSlots(address common.Address) ([]byte, bool) {
```
michaelkaplan13, line 1136: I think you called this out someplace else. But could `GetPredicateStorageSlots` just return a `[]byte` and `nil` if the address wasn't included in the access list?

I guess the only behavior difference is that the case when the address was not included at all and the case when the address was included with an empty storage slots array would be indistinguishable. I'm not sure whether there is a use case for a precompile predicate to need to be invoked with `nil` data, but probably best to leave in the option for that either way, so I'd be in favor of keeping this as is.

miner/worker.go
```diff
@@ -212,6 +216,28 @@ func (w *worker) commitNewWork() (*types.Block, error) {
 	return w.commit(env)
 }
 
+// enforcePredicates takes a set of pending transactions (grouped by sender, and ordered by nonce) and returns a
+// subset of those transactions (grouped by sender) that satisfy predicateContext.
+// Any transaction sent by a given sender that is received after a transaction that does not
+// satisfy predicateContext is removed from the TxPool and is not included in the return value.
+func (w *worker) enforcePredicates(rules params.Rules, predicateContext *precompile.PredicateContext, pending map[common.Address]types.Transactions) map[common.Address]types.Transactions {
```
michaelkaplan13, line 223: Nit: (Optional) Not sure what your normal linting rules are, but I think it'd be easier to read if this function signature was split across multiple lines since its 188 characters.

plugin/evm/block.go
```diff
@@ -102,14 +107,77 @@ func (b *Block) syntacticVerify() error {
...
+		predicateCtx := precompile.PredicateContext{
+			SnowCtx:            b.vm.ctx,
+			ProposerVMBlockCtx: proposerVMBlockCtx,
+		}
+
+		for _, tx := range b.ethBlock.Transactions() {
+			if err := core.CheckPredicates(rules, &predicateCtx, tx); err != nil {
+				return err
+			}
+		}
```
michaelkaplan13, line 178: Looks like this exact code block is duplicated in `verify` and `VerifyWithContext`. May want to consider moving it to a helper to avoid the duplication.

michaelkaplan13 (approved):
3 nits, but otherwise LGTM.

darioush:
some suggestions: https://github.com/ava-labs/subnet-evm/pull/511/files

core/state/statedb.go
aaronbuchwald: Good callout. I agree it would be useful to use `exists` to track whether the precompile address was included and make it so that you can distinguish between including an empty array and not including it at all.

We should also make sure to do a pass over all of invariants to ensure:
1. They are clearly commented in the code
2. We are all on the same page and understand them
3. Write clear documentation for them

precompile/stateful_precompile_config.go
```diff
@@ -7,22 +7,41 @@ import (
...
 
+// PredicateContext provides context to stateful precompile predicates
+type PredicateContext struct {
+	SnowCtx            *snow.Context
+	ProposerVMBlockCtx *block.Context
+}
+
+// PredicateFunc is the function type for validating that an access list tuple touching a precompile follows the predicate
+// It is possible [blockContext] to be nil if the proposer VM is not fully activated, so predicates should explicitly check this.
+type PredicateFunc func(predicateContext *PredicateContext, storageSlots []byte) error
```
aaronbuchwald, line 23: TODO: switch to making this an optional interface and exclude from template

aaronbuchwald: Same comment for `OnAccept`

aaronbuchwald:
Migrated outstanding TODOs to this comment: https://github.com/ava-labs/subnet-evm/pull/524#issuecomment-1435417171

Closing in favor of #524

Issue #494, opened 2023-02-06
Flaky Test: TestUpdateGasPriceInitializesPrice


aaronbuchwald:
**Describe the bug**
The following test flakes sometimes in CI: https://github.com/ava-labs/subnet-evm/actions/runs/4106650918/jobs/7085148883#step:6:55

**Expected behavior**
The test should pass consistently.

**Screenshots**
<img width="1348" alt="Screen Shot 2023-02-06 at 15 38 33" src="https://user-images.githubusercontent.com/24684335/217111525-70fb35b3-8482-464d-977f-ccd8b7ba61aa.png">

aaronbuchwald:
The bug is here: https://github.com/ava-labs/subnet-evm/blob/master/plugin/evm/gasprice_update_test.go#L41-L57

The `attemptWait` function starts a goroutine to wait for a waitgroup to finish and then close a channel. However, this creates a race condition because that goroutine may finish after the passed in ticker has already gone off to signal that the test has timed out and should fail

aaronbuchwald:
Fixed by #514

PR #514, opened 2023-02-16
Remove unnecessary gasprice updater logic and tests

aaronbuchwald:
## Why this should be merged

This PR fixes the unnecessary logic in gas price updater and addresses https://github.com/ava-labs/subnet-evm/issues/494.

## How this works

This PR removes the gas price updater implementation and tests. This code was a janky holdover from coreth where we needed to update the gas price at some point in the future on startup.

In Subnet-EVM, we set the gas price and minimum fee based off of the genesis fee config and it can additionally be updated via the fee config manager.

This PR also changes the default tx pool price limit from 1 to 0. This is not a change in practice since the gas price updater was already doing this. There is no need to enforce a minimum tip by default and a node can enforce a minimum tip by setting it via config.

## How this was tested

CI

## How is this documented

This change does not impact any APIs or configs, so no documentation is necessary for this PR.


plugin/evm/block_builder_test.go
```diff
@@ -14,6 +14,24 @@ import (
 	"github.com/ava-labs/avalanchego/snow"
 )
 
+func attemptAwait(t *testing.T, wg *sync.WaitGroup, delay time.Duration) {
```
aaronbuchwald, line 17: migrate from deleted test file

core/tx_pool.go
```diff
@@ -1441,10 +1442,6 @@ func (pool *TxPool) reset(oldHead, newHead *types.Header) {
 
 	// when we reset txPool we should explicitly check if fee struct for min base fee has changed
 	// so that we can correctly drop txs with < minBaseFee from tx pool.
-	// TODO: this should be checking IsSubnetEVM since we also support minimumFee for SubnetEVM
```
aaronbuchwald, line 1444: We now require `SubnetEVMTimestamp` at genesis: https://github.com/ava-labs/subnet-evm/pull/419/files

darioush (approved):

ceyonur (approved):

PR #518, opened 2023-02-17
precompile: test refactoring approach 2

darioush:


ceyonur (commented):
I don't think we need to manually add AllowList tests to each test map. I feel like the first approach (base branch) is much easier to follow.

darioush:
Yep I prefer the base branch too. Closing.

PR #520, opened 2023-02-17
Rename configs: alternative

darioush:


precompile/config/config.go
```diff
@@ -2,7 +2,7 @@
 // See the file LICENSE for licensing terms.
 
 // Defines the stateless interface for unmarshalling an arbitrary config of a precompile
-package config
+package precompileConfig
```
ceyonur, line 5: I think if we want to change the package name, we should also change the folder name

ceyonur: I'm not %100 sure about Go best practices but I feel like package name should be `precompileconfig`

precompile/allowlist/config.go
```diff
@@ -10,15 +10,15 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 )
 
-// Config specifies the initial set of addresses with Admin or Enabled roles.
-type Config struct {
+// AllowList specifies the initial set of addresses with Admin or Enabled roles.
+type AllowList struct {
```
ceyonur, line 14: I'd prefer this to be `AllowListConfig`, since `AllowList` is more like the collection of these contract functions.

darioush, line 14: we could also leave this as Config and instead of embedding it in other configs name it AllowList there.

ceyonur: hm not sure how this would work with embedded jsons. Is there a way to embed jsons without embedding structs? (and definetly not with a custom marshaler again :D)

darioush: yeah, I went with your above suggestion instead.

precompile/config/config.go
aaronbuchwald: Agree it should be all lowercase. Generally, I think multi-word package names are discouraged. I'd prefer to leave as config and import under the alias `precompileConfig`, but maybe that's too much manual work.

darioush: I tried that before but I think the lowercase option is better (not a fan of aliasing), feel free to have a look at the previous commit and lmk what you think @aaronbuchwald .

This package is going to be a bit hard to fit with the standard rules.

aaronbuchwald (approved):

Issue #407, opened 2022-12-09
Update goreleaser github action to not mark "rc"s as latest release


aaronbuchwald:
This ticket is to modify the goreleaser github action to ensure that if a new tag is created with "rc" in the name, it will not be tagged as the latest release automatically.

ceyonur:
Any update on this?

aaronbuchwald:
This is in the backlog, I haven't looked into it yet.

aaronbuchwald:
Fixed by #521

PR #521, opened 2023-02-17
Update github actions to ignore rcs

aaronbuchwald:
## Why this should be merged

This PR changes GH Actions to only run on tags that follow semantic versions and ignores rcs.

Addresses https://github.com/ava-labs/subnet-evm/issues/407

## How this works

This updates from matching all tags to matching semantic versions.

## How this was tested

Needs to be merged to test that rcs no longer trigger GH Actions to run and mark as latest release

## How is this documented

No documentation necessary for this change.

### DIFF
--- a/.github/workflows/lint-tests-release.yml
+++ b/.github/workflows/lint-tests-release.yml
@@ -5,7 +5,7 @@ on:
     branches:
       - master
     tags:
-      - "*"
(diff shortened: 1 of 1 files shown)


darioush (approved):

aaronbuchwald:
Going to merge this without two reviews since it's a small change, and I'm going to push another rc.

PR #523, opened 2023-02-17
move blackhole check to module registerer

ceyonur:
## Why this should be merged

Precompile address check should be done during module registration. The previous check was in `core/vm` package `init()`. Precompile Registration also happens on `init` function on precompile packages. This means that if `core/vm` is imported before `precompile` package, it could actually skip checking if the registered address equals to the blackhole address. 

## How this works

Moves black hole address check from contracts init to `RegisterPrecompile`.

## How this was tested

Added unit test

### DIFF
--- a/precompile/modules/registerer_test.go
+++ b/precompile/modules/registerer_test.go
@@ -44,11 +44,16 @@ func TestInsertSortedByAddress(t *testing.T) {
 	require.Equal(t, []Module{module0, module1, module2, module3}, data)
 }
 
-func TestRegisterModule(t *testing.T) {
-	// test that blackhole address is not allowed
-	moduleBlackhole := Module{
+func TestRegisterModuleInvalidAddresses(t *testing.T) {
+	// Test the blockhole address cannot be registered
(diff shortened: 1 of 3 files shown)


precompile/modules/registerer_test.go
```diff
@@ -42,3 +43,12 @@ func TestRegisterModule(t *testing.T) {
 	data = insertSortedByAddress(data, module2)
 	require.Equal(t, []Module{module0, module1, module2, module3}, data)
 }
+
+func TestRegisterModule(t *testing.T) {
```
aaronbuchwald, line 47: Could we also add a test case for a non-reserved address?

aaronbuchwald (approved):
LGTM - left one comment for an additional test case

precompile/modules/registerer_test.go
aaronbuchwald: Pushed a commit for this

aaronbuchwald (approved):

darioush (approved):

PR #524, opened 2023-02-17
Precompile pre post handling

aaronbuchwald:
## Why this should be merged

This PR adds precompile predicate enforcement and onAccept handling that will be used by the warp and shared memory precompiles.

## How this works

This PR adds optional `Accepter` and `Predicater` interfaces, which are to be optionally implemented by precompile contracts (contract not config or module).

These optional interfaces are NOT intended to be used for custom precompiles outside of this repo and the comments in this PR reflect that. Making his an optional interface removes the need to include it in the precompile gen template and makes it so that a user of the precompilegen template does not even need to see or think about the `Accepter` or `Predicater` interface at all.

The predicate is enforced on the relevant access list when calling `VerifyWithContext` or `BuildWithContext`.

The Accept handling is performed during block acceptance within `plugin/evm/block.go`. This is a change from a previous implementation to simplify passing in the backend, so that it does not require changing as much code.

Additionally, the shared memory precompile will need to perform shared memory operations to be committed atomically alongside updating the block that is marked as last accepted.

I spent some time trying to make this logic work within `core/blockchain.go`, but found that it will be easier to maintain this within `plugin/evm/block.go`.

## How this was tested

This was tested via CI and should be a no-op change since no precompiles currently use the added optional interfaces.

Now that we have the module registration logic, we can implement a test case for this that includes its own implementation of a precompile that uses a predicate and on accept to test this.

## How is this documented

This is not intended for end users to touch, so there is not currently an accompanying documentation PR. We should document this in the future, but deferring that for now.

### DIFF
--- a/core/predicate_check.go
+++ b/core/predicate_check.go
@@ -1,0 +1,40 @@
+// (c) 2023, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package core
+
+import (
+	"fmt"
+
+	"github.com/ava-labs/subnet-evm/core/types"
+	"github.com/ava-labs/subnet-evm/params"
+	"github.com/ava-labs/subnet-evm/precompile/contract"
+	"github.com/ava-labs/subnet-evm/precompile/modules"
+	"github.com/ava-labs/subnet-evm/utils"
+)
+
+// CheckPredicates checks that all precompile predicates are satisfied within the current [predicateContext] for [tx]
+func CheckPredicates(rules params.Rules, predicateContext *contract.PredicateContext, tx *types.Transaction) error {
+	precompileConfigs := rules.ActivePrecompiles
+	for _, accessTuple := range tx.AccessList() {
+		_, isPrecompile := precompileConfigs[accessTuple.Address]
+		if !isPrecompile {
+			continue
+		}
+
+		module, ok := modules.GetPrecompileModuleByAddress(accessTuple.Address)
+		if !ok {
+			return fmt.Errorf("accessed precompile config under address %s with no registered module", accessTuple.Address)
+		}
+		predicater, ok := module.Contract.(contract.Predicater)
+		if !ok {
+			continue
+		}
+
+		if err := predicater.VerifyPredicate(predicateContext, utils.HashSliceToBytes(accessTuple.StorageKeys)); err != nil {
+			return fmt.Errorf("predicate %s failed verification for tx %s: %w", accessTuple.Address, tx.Hash(), err)
+		}
+	}
+
+	return nil
+}
--- a/core/predicate_check.go
+++ b/core/predicate_check.go
@@ -24,7 +24,7 @@ func CheckPredicates(rules params.Rules, predicateContext *contract.PredicateCon
 
 		module, ok := modules.GetPrecompileModuleByAddress(accessTuple.Address)
 		if !ok {
-			return fmt.Errorf("accessed precompile config under address %s with no registered module", accessTuple.Address)
+			return fmt.Errorf("predicate accessed precompile config under address %s with no registered module", accessTuple.Address)
 		}
 		predicater, ok := module.Contract.(contract.Predicater)
 		if !ok {
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -39,6 +39,7 @@ import (
 	"github.com/ava-labs/subnet-evm/core/types"
 	"github.com/ava-labs/subnet-evm/metrics"
 	"github.com/ava-labs/subnet-evm/trie"
+	"github.com/ava-labs/subnet-evm/utils"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/log"
@@ -110,6 +111,9 @@ type StateDB struct {
 
 	// Per-transaction access list
 	accessList *accessList
+	// Ordered storage slots to be used in predicate verification as set in the tx access list.
+	// Only set in PrepareAccessList, and un-modified through execution.
+	predicateStorageSlots map[common.Address][]byte
 
 	// Journal of state modifications. This is the backbone of
 	// Snapshot and RevertToSnapshot.
@@ -156,17 +160,18 @@ func NewWithSnapshot(root common.Hash, db Database, snap snapshot.Snapshot) (*St
 		return nil, err
 	}
 	sdb := &StateDB{
-		db:                  db,
-		trie:                tr,
-		originalRoot:        root,
-		stateObjects:        make(map[common.Address]*stateObject),
-		stateObjectsPending: make(map[common.Address]struct{}),
-		stateObjectsDirty:   make(map[common.Address]struct{}),
-		logs:                make(map[common.Hash][]*types.Log),
-		preimages:           make(map[common.Hash][]byte),
-		journal:             newJournal(),
-		accessList:          newAccessList(),
-		hasher:              crypto.NewKeccakState(),
+		db:                    db,
+		trie:                  tr,
+		originalRoot:          root,
+		stateObjects:          make(map[common.Address]*stateObject),
+		stateObjectsPending:   make(map[common.Address]struct{}),
+		stateObjectsDirty:     make(map[common.Address]struct{}),
+		logs:                  make(map[common.Hash][]*types.Log),
+		preimages:             make(map[common.Hash][]byte),
+		journal:               newJournal(),
+		predicateStorageSlots: make(map[common.Address][]byte),
+		accessList:            newAccessList(),
+		hasher:                crypto.NewKeccakState(),
 	}
 	if snap != nil {
 		if snap.Root() != root {
@@ -674,6 +679,15 @@ func (db *StateDB) ForEachStorage(addr common.Address, cb func(key, value common
 	return nil
 }
 
+// copyPredicateStorageSlots creates a deep copy of the provided predicateStorageSlots map.
+func copyPredicateStorageSlots(predicateStorageSlots map[common.Address][]byte) map[common.Address][]byte {
+	res := make(map[common.Address][]byte)
+	for address, slots := range predicateStorageSlots {
+		res[address] = common.CopyBytes(slots)
+	}
+	return res
+}
+
 // Copy creates a deep, independent copy of the state.
 // Snapshots of the copied state cannot be applied to the copy.
 func (s *StateDB) Copy() *StateDB {
@@ -740,6 +754,7 @@ func (s *StateDB) Copy() *StateDB {
 	// However, it doesn't cost us much to copy an empty list, so we do it anyway
 	// to not blow up if we ever decide copy it in the middle of a transaction
 	state.accessList = s.accessList.Copy()
+	state.predicateStorageSlots = copyPredicateStorageSlots(s.predicateStorageSlots)
 
 	// If there's a prefetcher running, make an inactive copy of it that can
 	// only access data but does not actively preload (since the user will not
@@ -1060,11 +1075,16 @@ func (s *StateDB) PrepareAccessList(sender common.Address, dst *common.Address, 
 	for _, addr := range precompiles {
 		s.AddAddressToAccessList(addr)
 	}
+
+	// Note: If an address is specified multiple times in the access list, only the
+	// last storage slots provided for it are used in predicates.
+	s.predicateStorageSlots = make(map[common.Address][]byte)
 	for _, el := range list {
 		s.AddAddressToAccessList(el.Address)
 		for _, key := range el.StorageKeys {
 			s.AddSlotToAccessList(el.Address, key)
 		}
+		s.predicateStorageSlots[el.Address] = utils.HashSliceToBytes(el.StorageKeys)
 	}
 }
 
@@ -1101,4 +1121,14 @@ func (s *StateDB) AddressInAccessList(addr common.Address) bool {
 // SlotInAccessList returns true if the given (address, slot)-tuple is in the access list.
 func (s *StateDB) SlotInAccessList(addr common.Address, slot common.Hash) (addressPresent bool, slotPresent bool) {
 	return s.accessList.Contains(addr, slot)
+}
+
+// GetPredicateStorageSlots returns the storage slots associated with a given address, and whether or not
+// that address was included in the optional access list of the transaction.
+// The storage slots are returned in the same order as they appeared in the transaction.
+// These are the same storage slots that are used to verify any transaction
+// predicates for transactions with access list addresses that match a precompile address.
+func (s *StateDB) GetPredicateStorageSlots(address common.Address) ([]byte, bool) {
+	storageSlots, exists := s.predicateStorageSlots[address]
+	return storageSlots, exists
 }
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -44,6 +44,7 @@ import (
 	"github.com/ava-labs/subnet-evm/core/state"
 	"github.com/ava-labs/subnet-evm/core/types"
 	"github.com/ava-labs/subnet-evm/params"
+	"github.com/ava-labs/subnet-evm/precompile/contract"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/event"
 	"github.com/ethereum/go-ethereum/log"
@@ -112,7 +113,7 @@ func (w *worker) setEtherbase(addr common.Address) {
 }
 
 // commitNewWork generates several new sealing tasks based on the parent block.
-func (w *worker) commitNewWork() (*types.Block, error) {
+func (w *worker) commitNewWork(predicateContext *contract.PredicateContext) (*types.Block, error) {
 	w.mu.RLock()
 	defer w.mu.RUnlock()
 
@@ -192,8 +193,11 @@ func (w *worker) commitNewWork() (*types.Block, error) {
 		return nil, err
 	}
 
-	// Fill the block with all available pending transactions.
+	// Get the pending txs from TxPool
 	pending := w.eth.TxPool().Pending(true)
+	// Filter out transactions that don't satisfy predicateContext and remove them from TxPool
+	rules := w.chainConfig.AvalancheRules(header.Number, new(big.Int).SetUint64(header.Time))
+	pending = w.enforcePredicates(rules, predicateContext, pending)
 
 	// Split the pending transactions into locals and remotes
 	localTxs := make(map[common.Address]types.Transactions)
@@ -390,4 +394,26 @@ func totalFees(block *types.Block, receipts []*types.Receipt) *big.Float {
 		feesWei.Add(feesWei, new(big.Int).Mul(new(big.Int).SetUint64(receipts[i].GasUsed), tx.GasPrice()))
 	}
 	return new(big.Float).Quo(new(big.Float).SetInt(feesWei), new(big.Float).SetInt(big.NewInt(params.Ether)))
+}
+
+// enforcePredicates takes a set of pending transactions (grouped by sender, and ordered by nonce) and returns a
+// subset of those transactions (grouped by sender) that satisfy predicateContext.
+// Any transaction sent by a given sender that is received after a transaction that does not
+// satisfy predicateContext is removed from the TxPool and is not included in the return value.
+func (w *worker) enforcePredicates(rules params.Rules, predicateContext *contract.PredicateContext, pending map[common.Address]types.Transactions) map[common.Address]types.Transactions {
+	result := make(map[common.Address]types.Transactions, len(pending))
+	for addr, txs := range pending {
+		for i, tx := range txs {
+			if err := core.CheckPredicates(rules, predicateContext, tx); err != nil {
+				log.Debug("Transaction predicate failed verification in miner", "sender", addr, "err", err)
+				w.eth.TxPool().RemoveTx(tx.Hash()) // RemoveTx will move all subsequent transactions back to the future queue
+				txs = txs[:i]                      // Cut off any transactions past the failed predicate
+				break
+			}
+		}
+		if len(txs) > 0 {
+			result[addr] = txs
+		}
+	}
+	return result
 }
--- a/plugin/evm/block.go
+++ b/plugin/evm/block.go
@@ -12,10 +12,19 @@ import (
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/rlp"
 
+	"github.com/ava-labs/subnet-evm/core"
 	"github.com/ava-labs/subnet-evm/core/types"
+	"github.com/ava-labs/subnet-evm/precompile/contract"
 
 	"github.com/ava-labs/avalanchego/ids"
 	"github.com/ava-labs/avalanchego/snow/choices"
+	"github.com/ava-labs/avalanchego/snow/consensus/snowman"
+	"github.com/ava-labs/avalanchego/snow/engine/snowman/block"
+)
+
+var (
+	_ snowman.Block           = (*Block)(nil)
+	_ block.WithVerifyContext = (*Block)(nil)
 )
 
 // Block implements the snowman.Block interface
@@ -102,15 +111,82 @@ func (b *Block) syntacticVerify() error {
 
 // Verify implements the snowman.Block interface
 func (b *Block) Verify(context.Context) error {
-	return b.verify(true)
+	return b.verify(nil, true)
 }
 
-func (b *Block) verify(writes bool) error {
+// ShouldVerifyWithContext implements the block.WithVerifyContext interface
+func (b *Block) ShouldVerifyWithContext(context.Context) (bool, error) {
+	precompileConfigs := b.vm.chainConfig.AvalancheRules(b.ethBlock.Number(), b.ethBlock.Timestamp()).ActivePrecompiles
+
+	// Check if any of the transactions in the block list a precompile address in the access list.
+	for _, tx := range b.ethBlock.Transactions() {
+		for _, accessTuple := range tx.AccessList() {
+			if _, ok := precompileConfigs[accessTuple.Address]; ok {
+				log.Debug("Block verification requires proposerVM context", "block", b.ID(), "height", b.Height())
+				return true, nil
+			}
+		}
+	}
+
+	log.Debug("Block verification does not require proposerVM context", "block", b.ID(), "height", b.Height())
+	return false, nil
+}
+
+// VerifyWithContext implements the block.WithVerifyContext interface
+func (b *Block) VerifyWithContext(ctx context.Context, proposerVMBlockCtx *block.Context) error {
+	if proposerVMBlockCtx != nil {
+		log.Debug("Verifying block with context", "block", b.ID(), "height", b.Height())
+	} else {
+		log.Debug("Verifying block without context", "block", b.ID(), "height", b.Height())
+	}
+
+	return b.verify(proposerVMBlockCtx, true)
+}
+
+// Verify the block is valid.
+// Enforces that the predicates are valid within [proposerVMBlockCtx].
+// Writes the block to disk iff writes=true
+func (b *Block) verify(proposerVMBlockCtx *block.Context, writes bool) error {
 	if err := b.syntacticVerify(); err != nil {
 		return fmt.Errorf("syntactic block verification failed: %w", err)
 	}
 
+	// Only enforce predicates if the chain has already bootstrapped.
+	// If the chain is still bootstrapping, we can assume that all blocks we are verifying have
+	// been accepted by the network (so the predicate was validated by the network when the
+	// block was originally verified).
+	if b.vm.bootstrapped {
+		if err := b.verifyPredicates(proposerVMBlockCtx); err != nil {
+			return fmt.Errorf("failed to verify predicates: %w", err)
+		}
+	}
+
+	// The engine may call VerifyWithContext multiple times on the same block with different contexts.
+	// Since the engine will only call Accept/Reject once, we should only call InsertBlockManual once.
+	// Additionally, if a block is already in processing, then it has already passed verification and
+	// at this point we have checked the predicates are still valid in the different context so we
+	// can return nil.
+	if b.vm.State.IsProcessing(b.id) {
+		return nil
+	}
+
 	return b.vm.blockChain.InsertBlockManual(b.ethBlock, writes)
+}
+
+// verifyPredicates verifies the predicates in the block are valid according to proposerVMBlockCtx.
+func (b *Block) verifyPredicates(proposerVMBlockCtx *block.Context) error {
+	rules := b.vm.chainConfig.AvalancheRules(b.ethBlock.Number(), b.ethBlock.Timestamp())
+	predicateCtx := &contract.PredicateContext{
+		SnowCtx:            b.vm.ctx,
+		ProposerVMBlockCtx: proposerVMBlockCtx,
+	}
+
+	for _, tx := range b.ethBlock.Transactions() {
+		if err := core.CheckPredicates(rules, predicateCtx, tx); err != nil {
+			return err
+		}
+	}
+	return nil
 }
 
 // Bytes implements the snowman.Block interface
--- a/plugin/evm/block.go
+++ b/plugin/evm/block.go
@@ -13,8 +13,10 @@ import (
 	"github.com/ethereum/go-ethereum/rlp"
 
 	"github.com/ava-labs/subnet-evm/core"
+	"github.com/ava-labs/subnet-evm/core/rawdb"
 	"github.com/ava-labs/subnet-evm/core/types"
 	"github.com/ava-labs/subnet-evm/precompile/contract"
+	"github.com/ava-labs/subnet-evm/precompile/modules"
 
 	"github.com/ava-labs/avalanchego/ids"
 	"github.com/ava-labs/avalanchego/snow/choices"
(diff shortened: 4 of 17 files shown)


aaronbuchwald:
TODO: clearly document new invariants and update precompilegen tutorial with any necessary changes.

Now that this uses optional interfaces for accepter and predicater, this should not require any changes on top of what's already in the precompile refactor.

core/predicate_check.go
```diff
@@ -0,0 +1,40 @@
...
+		_, isPrecompile := precompileConfigs[accessTuple.Address]
+		if !isPrecompile {
+			continue
+		}
+
+		module, ok := modules.GetPrecompileModuleByAddress(accessTuple.Address)
+		if !ok {
+			return fmt.Errorf("predicate accessed precompile config under address %s with no registered module", accessTuple.Address)
+		}
+		predicater, ok := module.Contract.(contract.Predicater)
```
minghinmatthewlam, line 29: How come we don't require `Predicater` implemented as part of `StatefulPrecompiledContract`? Is this so that some precompiles don't need to have a predicate function? I tend to like the predicate included in `StatefulPrecompiledContract` and not have the following `if !ok {continue}` check

plugin/evm/block.go
```diff
@@ -51,13 +62,55 @@ func (b *Block) Accept(context.Context) error {
...
+			if !ok {
+				continue
+			}
+
+			module, ok := modules.GetPrecompileModuleByAddress(log.Address)
+			if !ok {
+				return fmt.Errorf("accepter accessed precompile config under address %s with no registered module", log.Address)
+			}
+
+			accepter, ok := module.Contract.(contract.Accepter)
```
minghinmatthewlam, line 100: same question as above for predicater

miner/worker.go
```diff
@@ -391,3 +395,25 @@ func totalFees(block *types.Block, receipts []*types.Receipt) *big.Float {
...
+// enforcePredicates takes a set of pending transactions (grouped by sender, and ordered by nonce) and returns a
+// subset of those transactions (grouped by sender) that satisfy predicateContext.
+// Any transaction sent by a given sender that is received after a transaction that does not
+// satisfy predicateContext is removed from the TxPool and is not included in the return value.
+func (w *worker) enforcePredicates(rules params.Rules, predicateContext *contract.PredicateContext, pending map[common.Address]types.Transactions) map[common.Address]types.Transactions {
+	result := make(map[common.Address]types.Transactions, len(pending))
+	for addr, txs := range pending {
+		for i, tx := range txs {
+			if err := core.CheckPredicates(rules, predicateContext, tx); err != nil {
+				log.Debug("Transaction predicate failed verification in miner", "sender", addr, "err", err)
```
minghinmatthewlam, line 408: does logging the transaction here help? I guess it might be included for `core.CheckPrediates` error already

miner/worker.go
```diff
@@ -391,3 +395,25 @@ func totalFees(block *types.Block, receipts []*types.Receipt) *big.Float {
...
+				w.eth.TxPool().RemoveTx(tx.Hash()) // RemoveTx will move all subsequent transactions back to the future queue
```
darioush, line 413: can we clarify this comment?

miner/worker.go
```diff
@@ -391,3 +395,25 @@ func totalFees(block *types.Block, receipts []*types.Receipt) *big.Float {
 	}
 	return new(big.Float).Quo(new(big.Float).SetInt(feesWei), new(big.Float).SetInt(big.NewInt(params.Ether)))
 }
+
+// enforcePredicates takes a set of pending transactions (grouped by sender, and ordered by nonce) and returns a
```
darioush, line 399: can we wrap this comment better?

plugin/evm/block.go
```diff
@@ -102,17 +155,84 @@ func (b *Block) syntacticVerify() error {
...
+	} else {
+		log.Debug("Verifying block without context", "block", b.ID(), "height", b.Height())
+	}
+
+	return b.verify(proposerVMBlockCtx, true)
+}
+
+// Verify the block is valid.
+// Enforces that the predicates are valid within [proposerVMBlockCtx].
+// Writes the block to disk iff writes=true
```
darioush, line 192:
```suggestion
// Pins the block's state transition to memory in statedb iff writes=true
```

plugin/evm/vm.go
```diff
@@ -46,6 +46,7 @@ import (
 	_ "github.com/ava-labs/subnet-evm/eth/tracers/native"
 
 	// Force-load precompiles to trigger registration
+	"github.com/ava-labs/subnet-evm/precompile/contract"
```
darioush, line 49: could we move the comment so it stays with `registry`?

plugin/evm/block.go
```diff
@@ -102,17 +155,84 @@ func (b *Block) syntacticVerify() error {
...
+// verifyPredicates verifies the predicates in the block are valid according to proposerVMBlockCtx.
+func (b *Block) verifyPredicates(proposerVMBlockCtx *block.Context) error {
+	rules := b.vm.chainConfig.AvalancheRules(b.ethBlock.Number(), b.ethBlock.Timestamp())
+	predicateCtx := &contract.PredicateContext{
+		SnowCtx:            b.vm.ctx,
+		ProposerVMBlockCtx: proposerVMBlockCtx,
+	}
+
+	for _, tx := range b.ethBlock.Transactions() {
+		if err := core.CheckPredicates(rules, predicateCtx, tx); err != nil {
```
minghinmatthewlam, line 229: similar to the other predicate checks, should we log for failure with the transaction?

precompile/contract/interfaces.go
```diff
@@ -19,6 +20,29 @@ type StatefulPrecompiledContract interface {
...
+type Predicater interface {
+	VerifyPredicate(predicateContext *PredicateContext, storageSlots []byte) error
+}
+
+// Accepter is an optional interface for StatefulPrecompiledContracts to implement.
+// If implemented, Accept will be called for every log with the address of the precompile when the block is accepted.
+// WARNING: this is not intended to be used for custom precompiles. Backwards compatibility with custom precompiles that
+// use the Accepter interface will not be supported.
+type Accepter interface {
+	Accept(txHash common.Hash, logIndex int, topics []common.Hash, logData []byte) error
```
cam-schultz, line 43: Are we still planning on passing in a unified Log struct? This was originally called out [here](https://github.com/ava-labs/subnet-evm/pull/453/files#r1098366475).

aaronbuchwald: No, the refactor will not make this possible.

core/predicate_check.go
aaronbuchwald: I prefer this so that it does not need to be implemented/filled out by every precompile even when it's only applicable to a few. Removing it from the standard template also makes it less likely that a custom precompile implementor will shoot themselves in the foot with this imo.

Open to changing it back if people feel strongly, but I prefer the optional interfaces here.

plugin/evm/block.go
aaronbuchwald: Same comment as above

miner/worker.go
aaronbuchwald: The tx hash was returned in one error case, but not the other, so making sure the tx hash is returned in both cases.

plugin/evm/block.go
aaronbuchwald: The consensus engine will log a debug error for blocks that fail verification, so no need for any further logging here.

This is not the case in the miner where we may drop a transaction that fails verification and still successfully return a newly built block to the engine, such that it would not log any error.

miner/worker.go
aaronbuchwald: What would you suggest? Is it just a question of what the future queue within the mempool is that's missing from the comment?

core/predicate_check.go
minghinmatthewlam: I tend to prefer previous change but don't have strong preference. Think it helps with consistency amongst precompiles, and also think without prior context, it'd be hard to tell reading through the code that `StatefulPrecompiledContract` can implement the `Predicater`.

minghinmatthewlam: I do agree that making the relationship between contract and predicater more hidden could prevent some incorrect predicate uses, so yea no strong preference.

core/state/statedb.go
```diff
@@ -1060,11 +1075,16 @@ func (s *StateDB) PrepareAccessList(sender common.Address, dst *common.Address,
 	for _, addr := range precompiles {
 		s.AddAddressToAccessList(addr)
 	}
+
+	// Note: If an address is specified multiple times in the access list, only the
+	// last storage slots provided for it are used in predicates.
```
aaronbuchwald, line 1080: We should change this invariant to make it easier to reason about

utils/bytes.go
```diff
@@ -16,3 +18,12 @@ func IncrOne(bytes []byte) {
 		}
 	}
 }
+
+// HashSliceToBytes serializes a []common.Hash into a []byte
+func HashSliceToBytes(hashes []common.Hash) []byte {
+	bytes := make([]byte, common.HashLength*len(hashes))
+	for i, hash := range hashes {
+		copy(bytes[i*common.HashLength:(i+1)*common.HashLength], hash.Bytes())
+	}
```
aaronbuchwald, line 27: nit: we should clean this up

aaronbuchwald:
Marking as DO NOT MERGE until after the v0.4.10 release with the precompile refactor.

core/state/statedb.go
```diff
@@ -1060,11 +1075,19 @@ func (s *StateDB) PrepareAccessList(sender common.Address, dst *common.Address,
...
+	// During predicate verification, we enforce that a precompile address can only
+	// have one access tuple per address to avoid a situation where we verify multiple
+	// predicates and only expose the data from the last one.
+	s.predicateStorageSlots = make(map[common.Address][]byte)
```
darioush, line 1084: could we move this logic to a separate function eg `preparePredicateStorageSlots` or similar? Also we discussed switching to logic to consolidate all instances (although existing logic is fine too IMO)

miner/worker.go
darioush: Yes, and also what does "subsequent" refer to in this context

precompile/contract/interfaces.go
```diff
@@ -19,6 +20,29 @@ type StatefulPrecompiledContract interface {
...
+// PredicateContext provides context to stateful precompile predicates
+type PredicateContext struct {
+	SnowCtx            *snow.Context
+	ProposerVMBlockCtx *block.Context
+}
+
+// Predicater is an optional interface for StatefulPrecompiledContracts to implement.
+// If implemented, the predicate will be enforced on every transaction in a block, prior to the block's execution.
+// If VerifyPredicate returns an error, the block will fail verification with no further processing.
+// WARNING: this is not intended to be used for custom precompiles. Backwards compatibility with custom precompiles that
```
darioush, line 32: could we clarify what "Custom precompiles" means in this context?

darioush (approved):

cam-schultz (approved):

miner/worker.go
aaronbuchwald: A subsequent nonce

core/state/statedb.go
aaronbuchwald: Yup

precompile/contract/interfaces.go
aaronbuchwald: What would you suggest? We've always called precompiles outside of Subnet-EVM, so not sure what else we should call them to clarify here.

miner/worker.go
aaronbuchwald: Updated to specify a subsequent nonce, I think if someone needs additional context on the future queue it's reasonable to check `RemoveTx` implementation in the transaction pool rather than elaborate in the comment here.

core/predicate_check.go
```diff
@@ -0,0 +1,49 @@
...
+		module, ok := modules.GetPrecompileModuleByAddress(address)
+		if !ok {
+			return fmt.Errorf("predicate accessed precompile config under address %s with no registered module for tx %s", address, tx.Hash())
+		}
+		predicater, ok := module.Contract.(contract.Predicater)
+		if !ok {
+			continue
+		}
+		// Return an error if we've already checked a predicate for this address
+		if _, ok := precompileAddressChecks[address]; ok {
```
minghinmatthewlam, line 38: just want to confirm that for a transaction, each precompile address' predicate should only appear in one access tuple. As in, no multiple access tuples of one transaction can specify the same precompile address.

aaronbuchwald: Yup that's correct. This is an intentional change from the previous code based on a conversation with @darioush to avoid a potentially weird situation where we verify all of the predicates and only expose the data from the last one.

It seemed pretty jank to overwrite the predicate storage slots with whatever the last item in the access list was, so this is intended to remove that edge case.

minghinmatthewlam: that makes sense, I like the change from overwriting last item.

plugin/evm/block.go
```diff
@@ -102,17 +155,84 @@ func (b *Block) syntacticVerify() error {
...
+	// been accepted by the network (so the predicate was validated by the network when the
+	// block was originally verified).
+	if b.vm.bootstrapped {
+		if err := b.verifyPredicates(proposerVMBlockCtx); err != nil {
+			return fmt.Errorf("failed to verify predicates: %w", err)
+		}
+	}
+
+	// The engine may call VerifyWithContext multiple times on the same block with different contexts.
+	// Since the engine will only call Accept/Reject once, we should only call InsertBlockManual once.
```
minghinmatthewlam, line 209: How is it enforced to only call Accept/Reject once?

utils/bytes_test.go
```diff
@@ -0,0 +1,65 @@
...
+	} {
+		t.Run(name, func(t *testing.T) {
+			output := common.CopyBytes(test.input)
+			IncrOne(output)
+			assert.Equal(t, output, test.expected)
+		})
+	}
+}
+
+func TestHashSliceToBytes(t *testing.T) {
```
minghinmatthewlam, line 40: nit: not sure if worth to add an easy empty hash slice

minghinmatthewlam (approved):

plugin/evm/block.go
```diff
@@ -51,13 +62,55 @@ func (b *Block) Accept(context.Context) error {
 	if err := vm.blockChain.Accept(b.ethBlock); err != nil {
 		return fmt.Errorf("chain could not accept %s: %w", b.ID(), err)
 	}
+	// Call Accept for relevant precompile logs. This should apply DB operations to the VM's versionDB
+	// to be committed atomically with marking this block as accepted.
+	if err := b.handlePrecompileAccept(); err != nil {
```
minghinmatthewlam, line 67: just a question, is there a convention for custom error handling in `subnet-evm`? Or is it just looking through the calls of that function and seeing if it already returns descriptive errors?

plugin/evm/block.go
aaronbuchwald: This is an invariant maintained by the AvalancheGo consensus engine.

aaronbuchwald: See https://docs.avax.network/subnets/create-a-vm-blobvm#introduction for more details.

aaronbuchwald: See also the comments on the interface in AvalancheGo: https://github.com/ava-labs/avalanchego/blob/master/snow/consensus/snowman/block.go

plugin/evm/block.go
aaronbuchwald: There's not a standard convention, just generally try to provide descriptive errors that will be helpful to users/node operators/developers in debugging.

Definitely something we could improve on and it's nice to have descriptive errors before you need them, especially for security critical code.

utils/bytes_test.go
aaronbuchwald: Adding a case for an empty slice, good call

miner/worker.go
```diff
@@ -391,3 +395,34 @@ func totalFees(block *types.Block, receipts []*types.Receipt) *big.Float {
...
+func (w *worker) enforcePredicates(
+	rules params.Rules,
+	predicateContext *contract.PredicateContext,
+	pending map[common.Address]types.Transactions,
+) map[common.Address]types.Transactions {
+	result := make(map[common.Address]types.Transactions, len(pending))
+	for addr, txs := range pending {
+		for i, tx := range txs {
+			if err := core.CheckPredicates(rules, predicateContext, tx); err != nil {
+				log.Debug("Transaction predicate failed verification in miner", "sender", addr, "err", err)
```
anusha-ctrl, line 414: Is it possible to add a test for this logic. It looks right, but it would be good to test

miner/worker.go
```diff
@@ -391,3 +395,34 @@ func totalFees(block *types.Block, receipts []*types.Receipt) *big.Float {
...
+				// and move all transactions from the same address with a subsequent nonce back to the
+				// future queue of the transaction pool.
+				w.eth.TxPool().RemoveTx(tx.Hash())
+				txs = txs[:i] // Cut off any transactions past the failed predicate in the return value
+				break
+			}
+		}
+		if len(txs) > 0 {
+			result[addr] = txs
+		}
```
ceyonur, line 425: I believe a break would suffice if we use append-like approach for results. like
```go
      result[addr] = append(result[addr], tx)
```

plugin/evm/block.go
ceyonur, line 100: same command regarding to needlesly traverse in logs if no precompiles are eligible for `Accept()`

plugin/evm/block.go
```diff
@@ -102,17 +155,84 @@ func (b *Block) syntacticVerify() error {
 
 // Verify implements the snowman.Block interface
 func (b *Block) Verify(context.Context) error {
-	return b.verify(true)
+	return b.verify(nil, true)
+}
+
+// ShouldVerifyWithContext implements the block.WithVerifyContext interface
+func (b *Block) ShouldVerifyWithContext(context.Context) (bool, error) {
```
ceyonur, line 162: where do we use this?

plugin/evm/block.go
```diff
@@ -102,17 +155,84 @@ func (b *Block) syntacticVerify() error {
...
+	}
+
+	// The engine may call VerifyWithContext multiple times on the same block with different contexts.
+	// Since the engine will only call Accept/Reject once, we should only call InsertBlockManual once.
+	// Additionally, if a block is already in processing, then it has already passed verification and
+	// at this point we have checked the predicates are still valid in the different context so we
+	// can return nil.
+	if b.vm.State.IsProcessing(b.id) {
+		return nil
+	}
```
ceyonur, line 215: I did not get why we need this. This was not needed before, and it seems to me that it's not directly related to predicate checks, right? So why are we adding this here?

plugin/evm/vm.go
```diff
@@ -639,9 +641,22 @@ func (vm *VM) Shutdown(context.Context) error {
...
+		log.Debug("Building block with context", "pChainBlockHeight", proposerVMBlockCtx.PChainHeight)
+	} else {
+		log.Debug("Building block without context")
+	}
+	predicateCtx := contract.PredicateContext{
+		SnowCtx:            vm.ctx,
+		ProposerVMBlockCtx: proposerVMBlockCtx,
+	}
+
+	block, err := vm.miner.GenerateBlock(&predicateCtx)
```
ceyonur, line 659: seems to me that we build with context regardlessly

plugin/evm/block.go
darioush: This is directly called by the engine to determine which Verify fn should be called

plugin/evm/block.go
darioush: this is now needed because VerifyWithContext may be called multiple times with different proposervm contexts.
however, we should still Verify the EVM block only once.

precompile/contract/interfaces.go
darioush:
```
 // WARNING: If you are implementing a custom precompile, beware that subnet-evm
 // will not maintain backwards compatibility of this interface and your code should not
// rely on this. Designed for use only by precompiles that are shipped as part of subnet-evm.
```

miner/worker.go
```diff
@@ -192,8 +193,11 @@ func (w *worker) commitNewWork() (*types.Block, error) {
 		return nil, err
 	}
 
-	// Fill the block with all available pending transactions.
+	// Get the pending txs from TxPool
 	pending := w.eth.TxPool().Pending(true)
+	// Filter out transactions that don't satisfy predicateContext and remove them from TxPool
+	rules := w.chainConfig.AvalancheRules(header.Number, new(big.Int).SetUint64(header.Time))
+	pending = w.enforcePredicates(rules, predicateContext, pending)
```
darioush, line 200: should we add a nil check on predicateContext here to avoid needing to reason about the behavior of other functions with a nil predicateContext?

precompile/contract/interfaces.go
darioush, line 43: From my understanding, Accept should also take *SnowCtx so it can gain access to the shared memory.

precompile/contract/interfaces.go
aaronbuchwald: That's way better, thanks for the suggestion!

plugin/evm/vm.go
aaronbuchwald: The `proposerVMBlockCtx` may be nil if the engine calls `BuildBlock` instead of `BuildBlockWithContext`.

miner/worker.go
aaronbuchwald: It's valid for the engine to leave the predicateContext nil, so it's up to the active precompiles to determine whether or not a nil predicateContext is valid.

precompile/contract/interfaces.go
```diff
@@ -47,6 +71,7 @@ type StateDB interface {
 	Exist(common.Address) bool
 
 	AddLog(addr common.Address, topics []common.Hash, data []byte, blockNumber uint64)
+	GetPredicateStorageSlots(address common.Address) ([]byte, bool)
```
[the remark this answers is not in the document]
aaronbuchwald: We turn the access list of `[]common.Hash` into ` []byte` when calling `PrepareAccessList` here: https://github.com/ava-labs/subnet-evm/blob/master/core/state_transition.go#L346.

Precompiles can then use `GetPredicateStorageSlots` to get the predicate bytes during their execution to see what was validated by the predicate. This will be used by the warp and shared memory precompiles to check the verified messages and imported UTXOs respectively.

plugin/evm/block.go
```diff
@@ -102,17 +155,84 @@ func (b *Block) syntacticVerify() error {
...
+}
+
+// ShouldVerifyWithContext implements the block.WithVerifyContext interface
+func (b *Block) ShouldVerifyWithContext(context.Context) (bool, error) {
+	precompileConfigs := b.vm.chainConfig.AvalancheRules(b.ethBlock.Number(), b.ethBlock.Timestamp()).ActivePrecompiles
+
+	// Check if any of the transactions in the block list a precompile address in the access list.
+	for _, tx := range b.ethBlock.Transactions() {
+		for _, accessTuple := range tx.AccessList() {
+			if _, ok := precompileConfigs[accessTuple.Address]; ok {
```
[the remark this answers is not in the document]
aaronbuchwald: Ya I'll update this to just the precompiles that have a predicate. Will actually change this so that `rules` has a separate field for the active precompiles that include a predicate, so that this is easier to access.

core/predicate_check.go
```diff
@@ -0,0 +1,49 @@
...
+			return fmt.Errorf("predicate %s failed verification for tx %s: specified %s in access list multiple times", address, tx.Hash(), address)
+		}
+		precompileAddressChecks[address] = struct{}{}
+
+		if err := predicater.VerifyPredicate(predicateContext, utils.HashSliceToBytes(accessTuple.StorageKeys)); err != nil {
```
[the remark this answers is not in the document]
aaronbuchwald: We only need to error in that case if it's required by the precompile. I'll add a comment for this.

precompile/contract/interfaces.go
aaronbuchwald: Yes, I'd prefer to modify this as needed when the precompile that uses it goes in because I want to add a `Backend` interface that provides all of the necessary functionality instead of adding one-offs for everything.

core/state/statedb.go
```diff
@@ -156,17 +160,18 @@ func NewWithSnapshot(root common.Hash, db Database, snap snapshot.Snapshot) (*St
...
+		db:                    db,
+		trie:                  tr,
+		originalRoot:          root,
+		stateObjects:          make(map[common.Address]*stateObject),
+		stateObjectsPending:   make(map[common.Address]struct{}),
+		stateObjectsDirty:     make(map[common.Address]struct{}),
+		logs:                  make(map[common.Hash][]*types.Log),
+		preimages:             make(map[common.Hash][]byte),
+		journal:               newJournal(),
+		predicateStorageSlots: make(map[common.Address][]byte),
```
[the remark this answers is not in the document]
aaronbuchwald: I'd prefer to keep this as is, so that it's handled the same as `accessList` which is populated on the line below and also set in `PrepareAccessList` before it's used.

miner/worker.go
aaronbuchwald: Do you mean putting this line within the for loop iterating transactions and calling append for each valid transaction?

plugin/evm/block.go
aaronbuchwald: Updated

miner/worker.go
ceyonur: yep

core/predicate_check.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/ava-labs/subnet-evm/utils"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// CheckPredicates checks that all precompile predicates are satisfied within the current [predicateContext] for [tx]
+func CheckPredicates(rules params.Rules, predicateContext *contract.PredicateContext, tx *types.Transaction) error {
+	precompileConfigs := rules.ActivePrecompiles
+	// Track addresses that we've performed a predicate check for
+	precompileAddressChecks := make(map[common.Address]struct{})
+	for _, accessTuple := range tx.AccessList() {
```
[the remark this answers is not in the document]
ceyonur: did you push that?

miner/worker.go
aaronbuchwald: Updated to separate out two different contexts one with the ProposerVM wrapper so that it can be set only if needed by the ProposerVM, and one that does not require it so that it can be used within shared memory without impacting the optional interfaces `ShouldVerifyWithContext` and `VerifyWithContext`

miner/worker.go
aaronbuchwald: I'd prefer to do this once here

plugin/evm/shared_memory_writer.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/ava-labs/subnet-evm/precompile/precompileconfig"
+)
+
+var _ precompileconfig.SharedMemoryWriter = &sharedMemoryWriter{}
+
+type sharedMemoryWriter struct {
+	requests map[ids.ID]*atomic.Requests
+}
+
+func NewSharedMemoryWriter() *sharedMemoryWriter {
```
aaronbuchwald, line 18: We should move this to its own package

core/predicate_check.go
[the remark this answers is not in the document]
aaronbuchwald: Separated out precompiles that require a general predicate and that require a predicate on the proposervm header, so also adding the requirement that the `ProposerVMBlockCtx` is non nil

plugin/evm/shared_memory_writer.go
aaronbuchwald: This seems unnecessary with how small it is at this point.

core/state/statedb.go
```diff
@@ -1060,12 +1075,26 @@ func (s *StateDB) PrepareAccessList(sender common.Address, dst *common.Address,
...
 	}
+	s.preparePredicateStorageSlots(list)
+}
+
+// preparePredicateStorageSlots populates the predicateStorageSlots field from the transaction's access list
+// Note: if an address is specified multiple times in the access list, only the last storage slots provided
+// for it are used in predicates.
+// During predicate verification, we require that a precompile address is only specififed in the access list
+// once to avoid a situation where we verify multiple predicate and only expose data from the last one.
+func (s *StateDB) preparePredicateStorageSlots(list types.AccessList) {
```
darioush, line 1093: since we already have the active precompiles as input to `PrepareAccessList` can we filter on the precompile addresses here to avoid making a full copy of the access list?

plugin/evm/block.go
```diff
@@ -102,17 +166,95 @@ func (b *Block) syntacticVerify() error {
...
+		PrecompilePredicateContext: precompileconfig.PrecompilePredicateContext{
+			SnowCtx: b.vm.ctx,
+		},
+		ProposerVMBlockCtx: proposerVMBlockCtx,
+	}, true)
 }
 
-func (b *Block) verify(writes bool) error {
+// Verify the block is valid.
+// Enforces that the predicates are valid within [proposerVMBlockCtx].
```
darioush, line 217: should this comment be updated?

plugin/evm/block.go
```diff
@@ -102,17 +166,95 @@ func (b *Block) syntacticVerify() error {
...
+	// at this point we have checked the predicates are still valid in the different context so we
+	// can return nil.
+	if b.vm.State.IsProcessing(b.id) {
+		return nil
+	}
+
 	return b.vm.blockChain.InsertBlockManual(b.ethBlock, writes)
 }
 
+// verifyPredicates verifies the predicates in the block are valid according to proposerVMBlockCtx.
```
darioush, line 246: should this comment be updated?

.gitignore
```diff
@@ -50,5 +50,4 @@ cmd/simulator/.simulator/*
 # goreleaser
 dist/
 
-# generator rpc file for e2e tests
-contract-examples/dynamic_rpc.json
\ No newline at end of file
+node_modules
```
darioush, line 53: do we need / want this change?
I think it is from one of my PRs that was merged into this PR.
I'm ok to keep the change, just want to make sure it's intentional.

precompile/precompileconfig/config.go
```diff
@@ -25,3 +31,60 @@ type Config interface {
 	// Verify is called on startup and an error is treated as fatal. Configure can assume the Config has passed verification.
 	Verify() error
 }
+
+// PredicateContext is the context passed in to the PrecompilePredicater interface.
```
darioush, line 35: comment should be updated to match struct name (or we can keep the shorter names)

precompile/precompileconfig/config.go
```diff
@@ -25,3 +31,60 @@ type Config interface {
...
+type PrecompilePredicateContext struct {
+	SnowCtx *snow.Context
+}
+
+// Predicater is an optional interface for StatefulPrecompileContracts to implement.
```
darioush, line 40: comment should be updated to match struct name

.gitignore
aaronbuchwald: Ya this was merged from yours. It could have been separated out, but I kept it in because it seems like a good change.

darioush (approved):

.gitignore
ceyonur, line 53: hm this is already specified under `contract-examples/.gitignore`. did we specifically need to ignore this in root directory?

params/config.go
```diff
@@ -574,6 +587,15 @@ type Rules struct {
...
 	ActivePrecompiles map[common.Address]precompileconfig.Config
+	// PrecompilePredicates maps addresses to stateful precompile predicate functions
+	// that are enabled for this rule set.
+	PredicatePrecompiles map[common.Address]precompileconfig.PrecompilePredicater
+	// ProposerPredicates maps addresses to stateful precompile predicate functions
+	// that are enabled for this rule set and require access to the ProposerVM wrapper.
+	ProposerPredicates map[common.Address]precompileconfig.ProposerPredicater
+	// AccepterPrecompiles map addresses to stateful precompile accepter functions
+	// that are enabled for this rule set.
+	AccepterPrecompiles map[common.Address]precompileconfig.Accepter
```
ceyonur, line 598: noice

params/config.go
```diff
@@ -610,9 +632,21 @@ func (c *ChainConfig) AvalancheRules(blockNum, blockTimestamp *big.Int) Rules {
...
 			rules.ActivePrecompiles[module.Address] = config
+			if precompilePredicate, ok := config.(precompileconfig.PrecompilePredicater); ok {
+				rules.PredicatePrecompiles[module.Address] = precompilePredicate
+			}
+			if proposerPredicate, ok := config.(precompileconfig.ProposerPredicater); ok {
+				rules.ProposerPredicates[module.Address] = proposerPredicate
+			}
+			if precompileAccepter, ok := config.(precompileconfig.Accepter); ok {
+				rules.AccepterPrecompiles[module.Address] = precompileAccepter
+			}
```
ceyonur, line 649: noicer

plugin/evm/block.go
```diff
@@ -62,47 +61,59 @@ func (b *Block) Accept(context.Context) error {
...
+	// changes to the vm's versionDB.
+	return vm.ctx.SharedMemory.Apply(sharedMemoryWriter.requests, vdbBatch)
 }
 
 // handlePrecompileAccept calls Accept on any logs generated with an active precompile address that implements
 // contract.Accepter
 // This function assumes that the Accept function will ONLY operate on state maintained in the VM's versiondb.
 // This ensures that any DB operations are performed atomically with marking the block as accepted.
-func (b *Block) handlePrecompileAccept() error {
+func (b *Block) handlePrecompileAccept(sharedMemoryWriter *sharedMemoryWriter) error {
```
ceyonur, line 92: can we update comment to explain why shared memory is required here?

plugin/evm/block.go
```diff
@@ -184,13 +205,18 @@ func (b *Block) VerifyWithContext(ctx context.Context, proposerVMBlockCtx *block
 		log.Debug("Verifying block without context", "block", b.ID(), "height", b.Height())
 	}
 
-	return b.verify(proposerVMBlockCtx, true)
+	return b.verify(&precompileconfig.ProposerPredicateContext{
+		PrecompilePredicateContext: precompileconfig.PrecompilePredicateContext{
+			SnowCtx: b.vm.ctx,
+		},
+		ProposerVMBlockCtx: proposerVMBlockCtx,
+	}, true)
```
ceyonur, line 213: can we split new predicate context to a variable and then use `b.verify(predicateContext, true)` for readability? we can also create a function for `NewProposerPredicateContext` which would take `ctx` and `proporservmblockctx` as inputs (not really sure how custom we would want this to be though)

precompile/precompileconfig/config.go
```diff
@@ -25,3 +31,60 @@ type Config interface {
...
 }
+
+// PrecompilePredicateContext is the context passed in to the PrecompilePredicater interface.
+type PrecompilePredicateContext struct {
+	SnowCtx *snow.Context
+}
+
+// PrecompilePredicater is an optional interface for StatefulPrecompileContracts to implement.
+// If implemented, the predicate will be enforced on every transaction in a block, prior to
+// the blcok's execution.
```
ceyonur, line 42:
```suggestion
// the block's execution.
```

precompile/precompileconfig/config.go
```diff
@@ -25,3 +31,60 @@ type Config interface {
...
+
+// ProposerPredicater is an optional interface for StatefulPrecompiledContracts to implement.
+// If implemented, the predicate will be enforced on every transaction in a block, prior to
+// the block's execution.
+// If VerifyPredicate returns an error, the block will fail verification with no further processing.
+// Note: ProposerVMBlockCtx is guaranteed to be non-nil.
+// WARNING: If you are implementing a custom precompile, beware that subnet-evm
+// will not maintain backwards compatibility of this interface and your code should not
+// rely on this. Designed for use only by precompiles that ship with subnet-evm.
+type ProposerPredicater interface {
```
ceyonur, line 67: I'm a bit confused about the difference between `ProposerPredicate` vs `PrecompilePredicate`. will precompiles implement both of these interfaces?

params/config.go
aaronbuchwald: 🤝

precompile/precompileconfig/config.go
aaronbuchwald: Adding a comment for this. Precompiles should implement `ProposerPredicate` iff they depend on the ProposerVM Block Context ie. Warp.

plugin/evm/block.go
aaronbuchwald: Going to leave this as is, since I don't think this would really improve readability.

plugin/evm/block.go
aaronbuchwald: Updated

ceyonur (approved):

PR #528, opened 2023-02-21
Rename new config

ceyonur:
## Why this should be merged

We had 2 `NewConfig` in precompiles, renamed the one in configurator to `MakeConfig`. Also used built-in `new` to create a new instance of Configs.

## How this was tested

e2e and UTs should cover this

## How is this documented

It will be documented in docs repo with a separate PR for all changes in #389 PR.


darioush:
I slightly prefer NewConfig, the change to use `new()` is nice though.
Alternatively if the only use of NewConfig is to immediately deserialize json, should we consider a fn such as `FromJson([]byte) config.Config` instead?

ceyonur:
The `NewConfig`  in configurator can be confused with `NewConfig(timestamp, admins,...)` in config. We could add that `ParseJSON/FromJSON` to the interface but I really don't want to put complicated unmarshalling/marshalling stuff in the precompile packages.

darioush (approved):

anusha-ctrl:
lgtm, i like the `new()`. If there is still some doubts on naming, maybe `NewPrecompileConfig()` or `CreateConfig()`

anusha-ctrl (approved):

aaronbuchwald (approved):

PR #537, opened 2023-02-23
Shared memory precompiles (wip)

darioush:
The shared memory precompile allows the caller to import and export the native token (avax) to other blockchains on the same subnet using the shared memory.
Additionally, it is possible to transfer between ANTs (as defined on the X-Chain) and ERC20 contracts (as deployed on subnet-evm) (wip)


contract-examples/contracts/ERC20SharedMemory.sol
```diff
@@ -0,0 +1,69 @@
...
+  address constant SHARED_MEMORY_ADDRESS = 0x0200000000000000000000000000000000000005;
+  ISharedMemory sharedMemory = ISharedMemory(SHARED_MEMORY_ADDRESS);
+  ERC20SharedMemory private erc20;
+
+  function setUp() public {
+    erc20 = new ERC20SharedMemory(100);
+  }
+
+  function test_exportAVAX(
+      uint value, bytes32 destinationChainID, address addr) public {
```
richardpringle, line 40: Should avoid using the term `value` when not meant in a generic context.
```suggestion
      uint amount, bytes32 destinationChainID, address addr) public {
```

contract-examples/test/shared_memory_export.ts
```diff
@@ -0,0 +1,98 @@
...
+      to: testContract.address,
+      value: amount,
+    })
+    let receipt = await tx.wait()
+    expect(receipt.status == 1).to.be.true
+
+    // ExportAVAX
+    // Note we export AVAX to testContract.address, which is the contract we
+    // just deployed. This is because the import test will also deploy a
+    // contract from the same account with the same nonce.
```
richardpringle, line 76: I know it's implied, but you probably want to specify 
```suggestion
    // contract from the same account with the same nonce on blockchain-b.
```

or maybe, "the other blockchain" if you want to be more generic.

contract-examples/test/shared_memory_export.ts
```diff
@@ -0,0 +1,98 @@
...
+    let foundLog = txReceipt.logs.find(
+      (log: Event, _: any, __: any) => 
+        log.address === SHARED_MEMORY_ADDRESS &&
+        log.topics.length === 2 && // TODO: review the indexed vs. non-indexed log data
+        // TODO: get the string from the contract abi
+        log.topics[0] === ethers.utils.id("ExportAVAX(uint64,bytes32,uint64,uint64,address[])") &&
+        log.topics[1] == blockchainIDB // destination
+    )
+    // TODO: consider verifying more about the logs
+    expect(foundLog).to.exist;
```
richardpringle, line 93: What we want is `vm.expectEmit` from [here](https://book.getfoundry.sh/forge/cheatcodes)... 
It might be better to have something like this as an end-result
```ts
test("should export AVAX via contract", [
  {
    method: "step_fundTest",
    value: ethers.utils.parseUnits("1", "gwei"),
  },
  "step_exportAVAXFromAToB",
])
```
with 
```solidity
bytes32 constant BLOCKCHAIN_A_ID = // ...
bytes32 constant BLOCKCHAIN_B_ID = // ...

contract ERC20SharedMemoryTest is ExtendedDSTest {
  event ExportAVAX(uint64 amount, bytes32 indexed destinationChainID, uint64 locktime, uint64 threshold, address[] addrs);
  // ...
  function step_exportAVAXFromAToB() {
    address testAddress = address(this);

    address[] memory addrs = new address[](1);
    addrs[0] = testAddress;

    unit amount = testAddress.balance;
    sharedMemory.exportAVAX{value: amount}(BLOCKCHAIN_B_ID, 0, 1, addrs);
    
    // new variable from ExtendedDSTest
    vm.expectEmit(false, true, false, true);
    emit ExportAVAX(amount, BLOCKCHAIN_B_ID, /* .. */, addrs);

    assertEq(testAddress.balance, 0);
  }
}
```
Where the `test` function in TS checks that the logs that were emitted from the `sharedMemory` precompile match the `ExportAVAX` event from the test-contract.

It wouldn't be a trivial change, but I don't think it would be overly complicated either.

Thoughts?

richardpringle: Also... is there any way to tell what chain-id a smart contract is being executed on from the smart-contract itself?

richardpringle (changes requested):
Partial review

PR #538, opened 2023-02-23
fix config template for precompilegen

darioush:
## Why this should be merged
Removes duplicate structs from config.go

## How this was tested
Regenerated precompile for shared memory abi


ceyonur (approved):
Good catch! 👍

aaronbuchwald (approved):

PR #539, opened 2023-02-23
move runner and export it

ceyonur:
## Why this should be merged

Makes it possible for external programs to import the main runner function from subnet-evm like

in another repo's  `plugin/main.go`:
```go
// (c) 2019-2020, Ava Labs, Inc. All rights reserved.

package main

import (
	"github.com/ava-labs/subnet-evm/plugin/runner"

	_ "github.com/ava-labs/precompilevm/helloworld"
)

func main() {
	runner.Run()
}

```

## How this works

moves main function to a separate exported package

## How this was tested

Locally tested

## How is this documented

Will add a new repo for example and document it.


plugin/runner/runner.go
```diff
@@ -0,0 +1,34 @@
...
+
+	"github.com/ava-labs/avalanchego/utils/logging"
+	"github.com/ava-labs/avalanchego/utils/ulimit"
+	"github.com/ava-labs/avalanchego/version"
+	"github.com/ava-labs/avalanchego/vms/rpcchainvm"
+
+	"github.com/ava-labs/subnet-evm/plugin/evm"
+)
+
+func Run() {
```
darioush, line 19: should we add a string argument so the external repo can add some text to the PrintVersion message?

darioush (approved):

anusha-ctrl (approved):
lgtm

plugin/runner/runner.go
aaronbuchwald: Agreed it would be nice to allow an external repo to optionally specify some additional output to be included in `PrintVersion()`.

ceyonur: added

aaronbuchwald:
Adding the DO NOT MERGE label as @ceyonur and I talked offline that we want to spend a little more time thinking about the best way to do this before putting out an example that we would expect the community to start using.

PR #548, opened 2023-02-27
Bump minimum golang version to v1.20.1

aaronbuchwald:
## Why this should be merged

Bumps the minimum golang version to v1.20.1.

## How this works

As above.

## How this was tested

CI

## How is this documented

Updated README

### DIFF
--- a/.github/workflows/lint-tests-release.yml
+++ b/.github/workflows/lint-tests-release.yml
@@ -28,7 +28,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        go: ["1.18"]
+        go: ["1.20"]
         os: [ubuntu-20.04]
     steps:
       - uses: actions/checkout@v3
@@ -55,7 +55,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v3
         with:
-          go-version: 1.18
+          go-version: "1.20"
       - name: Use Node.js
         uses: actions/setup-node@v3
         with:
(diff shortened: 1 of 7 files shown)


.github/workflows/lint-tests-release.yml
```diff
@@ -85,7 +85,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v3
         with:
-          go-version: 1.18
+          go-version: 1.20
```
patrick-ogrady, line 88: Quotes

.goreleaser.yml
```diff
@@ -8,9 +8,7 @@ builds:
     ldflags: -X github.com/ava-labs/subnet-evm/plugin/evm.Version=v{{.Version}}
     goos:
       - linux
-      - darwin
```
patrick-ogrady, line 11: I think you can revert these

.github/workflows/lint-tests-release.yml
```diff
@@ -108,7 +108,7 @@ jobs:
         with:
           distribution: goreleaser
           version: latest
-          args: release --rm-dist
+          args: release --clean --debug
```
darioush, line 111: should this be reverted?

aaronbuchwald: `--rm-dist` is deprecated in favor of `--clean`: https://goreleaser.com/deprecations/#-rm-dist.

I added `--debug` in thinking it may be helpful since we do have issues with goreleaser sometimes, but didn't get much from the debug output, so will revert that part of this change.

aaronbuchwald: Updated

darioush (approved):

PR #549, opened 2023-02-28
state modifications as network upgrade

darioush:
## Why this should be merged
This PR implements support for performing state modifications as a network upgrade.
Node operators may specify modifications to state including:
- Balance changes
- Setting of a given storage slot to a given value
- Changing the code at an address

as a network upgrade via `upgradeBytes` (aka `upgrades.json`)

## How this works
Follows a similar pattern to the precompile activations, without modifying the runtime with a callable contract address.

## How this was tested
Unit tests
Tested the balance change portion on a local network.

## How is this documented
[docs PR](https://github.com/ava-labs/avalanche-docs/pull/1257)


core/state_processor.go
```diff
@@ -218,3 +225,21 @@ func ApplyPrecompileActivations(c *params.ChainConfig, parentTimestamp *big.Int,
 	}
 	return nil
 }
+
+// ApplyStateUpgrades checks if any of the state upgrades specified by the chain config are activated by the block
+// transition from [parentTimestamp] to the timestamp set in [header]. If this is the case, it calls [Configure]
+// to apply the necessary state transitions for the upgrade.
+// This function is called:
+// - during block processing to update the state before processing the given block.
+// - during block producing to apply the precompile upgrades before producing the block.
```
darioush, line 234: -> state

stateupgrade/interfaces.go
```diff
@@ -0,0 +1,35 @@
...
+package stateupgrade
+
+import (
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// StateDB is the interface for accessing EVM state in state upgrades
+type StateDB interface {
```
darioush, line 13: we could also move stateupgrade under precompile or reuse interfaces there but I think it is slightly better to keep them separate

params/state_upgrade.go
```diff
@@ -0,0 +1,44 @@
...
+	// remove the storage key
+	SetStorage map[common.Address]map[common.Hash]common.Hash `json:"setStorage,omitempty"`
+
+	// Sets the code of the specified contract to the given value
+	SetCode map[common.Address][]byte `json:"setCode,omitempty"`
+
+	// Deploys contracts with the specified creation bytecode to the
+	// specified addresses, instead of the normal rules for deriving
+	// the address of a created contract.
+	DeployContractTo []ContractDeploy `json:"deployContractTo,omitempty"`
```
ceyonur, line 31: should we use same notation for `alloc` in genesis? https://github.com/ava-labs/subnet-evm/blob/master/core/genesis.go#L102 
like:
```
type StateUpgradeAccount struct {
	Code       []byte                      `json:"code,omitempty"`
	Storage    map[common.Hash]common.Hash `json:"storage,omitempty"`
	Balance    *big.Int                    `json:"balance" gencodec:"required"`
	Nonce      uint64                      `json:"nonce,omitempty"` 
}
```

and 
``` 
type StateUpgrade struct {
  BlockTimestamp *big.Int `json:"blockTimestamp"`
  StateUpgradeAlloc map[common.Address]GenesisAccount `json:"alloc"`
}
```

darioush: This is also possible, I'd say this would be more inline with setting the balance to the specified amount instead of modifying the balance by the specified amount. Thoughts, @aaronbuchwald ?

darioush: Changed it to something along those lines.
- Not sure we want to expose modifying nonce
- Kept BalanceChange semantics instead of setting balance.

params/state_upgrade.go
```diff
@@ -42,7 +42,27 @@ func (s *StateUpgrade) Equal(other *StateUpgrade) bool {
 type stateUpgradeAccountMarshaling struct {
 	Code          hexutil.Bytes
 	Storage       map[storageJSON]storageJSON
-	BalanceChange *math.HexOrDecimal256
+	BalanceChange *SignedHexOrDecimal256
+}
+
+type SignedHexOrDecimal256 math.HexOrDecimal256
```
darioush, line 48: Alternatively we can split this to `AddBalance` and `SubBalance`, but I think this is fine.

stateupgrade/state_upgrade.go
```diff
@@ -0,0 +1,46 @@
...
+	if upgrade.BalanceChange != nil {
+		state.AddBalance(account, upgrade.BalanceChange)
+	}
+	if upgrade.Code != nil {
+		if state.GetNonce(account) == 0 {
+			// If the nonce is 0, we will set it to a non-zero value
+			// so the account is not considered empty.
+			state.SetNonce(account, 1)
+		}
+		state.SetCode(account, upgrade.Code)
```
friskyfoxdk, line 40: Is the approach to specify the bytecode in the upgrade.json file when configuring a code state upgrade?

darioush: @friskyfoxdk yes that is the thought process, you can specify it like:
```
"stateUpgrades":[{
	"blockTimestamp":1677610800,
	"accounts":{
	      "0xaddr": { "code": "0xbytecode" }
	}
```

or similar.

core/genesis.go
```diff
@@ -322,6 +316,19 @@ func (g *Genesis) ToBlock(db ethdb.Database) *types.Block {
...
+	blockContext := types.NewBlockWithHeader(head)
+	err = ApplyPrecompileActivations(g.Config, nil, blockContext, statedb)
+	if err != nil {
+		panic(fmt.Sprintf("unable to configure precompiles in genesis block: %v", err))
+	}
+	// Configure any state upgrades that should activate in the genesis.
+	err = ApplyStateUpgrades(g.Config, nil, blockContext, statedb)
+	if err != nil {
+		panic(fmt.Sprintf("unable to configure state upgrades in genesis block: %v", err))
+	}
```
aaronbuchwald, line 330: What's the rationale for moving this here? Is this so that `ApplyStateUpgrades` occurs after setting the genesis allocation?

aaronbuchwald: Would it make sense to make it a requirement that state upgrades can only be enabled via `upgrade.json` and they cannot be used in the genesis since they should never be required there?

core/state_processor.go
```diff
@@ -70,20 +71,25 @@ func NewStateProcessor(config *params.ChainConfig, bc *BlockChain, engine consen
...
-	err := ApplyPrecompileActivations(p.config, new(big.Int).SetUint64(parent.Time), block, statedb)
-	if err != nil {
-		log.Error("failed to configure precompiles processing block", "hash", block.Hash(), "number", block.NumberU64(), "timestamp", block.Time(), "err", err)
+	if err := ApplyPrecompileActivations(p.config, parentTimestamp, block, statedb); err != nil {
+		log.Error("failed to configure precompiles processing block", "hash", blockHash, "number", blockNumber, "timestamp", block.Time(), "err", err)
+		return nil, nil, 0, err
+	}
+	// Configure any state upgrades that should go into effect during this block.
+	if err := ApplyStateUpgrades(p.config, parentTimestamp, block, statedb); err != nil {
+		log.Error("failed to configure state upgrades processing block", "hash", blockHash, "number", blockNumber, "timestamp", block.Time(), "err", err)
```
aaronbuchwald, line 92: Can we combine these into a single function?

miner/worker.go
```diff
@@ -186,11 +186,18 @@ func (w *worker) commitNewWork() (*types.Block, error) {
...
 	// Configure any stateful precompiles that should go into effect during this block.
-	err = core.ApplyPrecompileActivations(w.chainConfig, new(big.Int).SetUint64(parent.Time()), types.NewBlockWithHeader(header), env.state)
+	blockContext := types.NewBlockWithHeader(header)
+	err = core.ApplyPrecompileActivations(w.chainConfig, new(big.Int).SetUint64(parent.Time()), blockContext, env.state)
 	if err != nil {
 		log.Error("failed to configure precompiles mining new block", "parent", parent.Hash(), "number", header.Number, "timestamp", header.Time, "err", err)
 		return nil, err
 	}
+	// Configure any state upgrades that should go into effect during this block.
+	err = core.ApplyStateUpgrades(w.chainConfig, new(big.Int).SetUint64(parent.Time()), blockContext, env.state)
```
aaronbuchwald, line 196: Can we use `bigTimestamp` here instead of creating it twice in each case?

params/config.go
```diff
@@ -220,6 +220,9 @@ type UpgradeConfig struct {
 	// forks must be present or upgradeBytes will be rejected.
 	NetworkUpgrades *NetworkUpgrades `json:"networkUpgrades,omitempty"`
 
+	// Config for modifying state as a network upgrade.
+	StateUpgrades []StateUpgrade `json:"stateUpgrades,omitempty"`
```
aaronbuchwald, line 224: can we make this a `map[uint64]StateUpgrade` so that StateUpgrades do not need to specify a timestamp.

core/genesis.go
darioush: We can do that, your 1st comment is correct. If we do that (disallow state upgrades in genesis), then we cannot combine the functions, which would you prefer?

params/config.go
darioush: We could do this if you strongly preferred, it would require changing the code and adding some sort logic which would be different than what we have for precompiles, I'd slightly prefer to keep the symmetry.

core/state_processor.go
```diff
@@ -218,3 +217,29 @@ func ApplyPrecompileActivations(c *params.ChainConfig, parentTimestamp *big.Int,
 	}
 	return nil
 }
+
+// applyStateUpgrades checks if any of the state upgrades specified by the chain config are activated by the block
+// transition from [parentTimestamp] to the timestamp set in [header]. If this is the case, it calls [Configure]
+// to apply the necessary state transitions for the upgrade.
+func applyStateUpgrades(c *params.ChainConfig, parentTimestamp *big.Int, blockContext stateupgrade.BlockContext, statedb *state.StateDB) error {
```
aaronbuchwald, line 224: TODO: move timestamp out of state upgrade itself

params/config.go
aaronbuchwald: Good point, will leave this as is.

stateupgrade/state_upgrade.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"math/big"
+
+	"github.com/ava-labs/subnet-evm/params"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// Configure applies the state upgrade to the state.
+func Configure(stateUpgrade *params.StateUpgrade, state StateDB) error {
```
ceyonur, line 14: why this is not a method for `StateUpgrade`? to keep this same with precompile upgrades?

stateupgrade/state_upgrade.go
```diff
@@ -0,0 +1,45 @@
...
+
+	if upgrade.BalanceChange != nil {
+		state.AddBalance(account, (*big.Int)(upgrade.BalanceChange))
+	}
+	if len(upgrade.Code) != 0 {
+		// if the nonce is 0, set the nonce to 1 as we would when deploying a contract at
+		// the address.
+		if state.GetNonce(account) == 0 {
+			state.SetNonce(account, 1)
+		}
```
ceyonur, line 38: I think we only set nonce if `EIP 158` is activated
https://github.com/ava-labs/subnet-evm/blob/master/core/vm/evm.go#L526

Not sure if we should add this here though.

params/precompile_upgrade.go
```diff
@@ -149,18 +148,33 @@ func (c *ChainConfig) verifyPrecompileUpgrades() error {
...
 }
 
-// GetActivePrecompileConfig returns the most recent precompile config corresponding to [address].
+// verifyStateUpgrades checks [c.StateUpgrades] is well formed:
+// - the specified blockTimestamps must monotonically increase
+func (c *ChainConfig) verifyStateUpgrades() error {
+	var previousUpgradeTimestamp *big.Int
+	for i, upgrade := range c.StateUpgrades {
+		upgradeTimestamp := upgrade.BlockTimestamp
+		// Verify specified timestamps are strictly monotonically increasing.
```
ceyonur, line 157: should we check prohibited addreses here in case we don't override any of them (especially precompiles)? I'd say if we want this to be a `alloc`-like mechanism in genesis, then we can keep these pretty much limitless. but still might worth considering to limit those.

stateupgrade/state_upgrade.go
darioush: Yes. I don't think we want to import the execution functionality from `params`, otherwise we may encounter the same import loop we saw with precompiles originally

stateupgrade/state_upgrade.go
darioush: Yes we should probably preserve this

stateupgrade/state_upgrade.go
ceyonur: > Yes. I don't think we want to import the execution functionality from `params`, otherwise we may encounter the same import loop we saw with precompiles originally

we would not need if we move it to params :D

stateupgrade/state_upgrade.go
```diff
@@ -0,0 +1,45 @@
+// (c) 2023 Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package stateupgrade
```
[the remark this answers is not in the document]
aaronbuchwald: I agree that this would make more sense under a general upgrades package containing both state upgrades and precompiles. I'd like to avoid putting it directly under precompiles since it is not a precompile.

stateupgrade/state_upgrade.go
aaronbuchwald: Ya let's add this check for symmetry with `create`

params/precompile_upgrade.go
aaronbuchwald: I think we can leave this unchecked and document this as an emergency developer tool.

stateupgrade/state_upgrade.go
darioush: Added.

stateupgrade/state_upgrade.go
friskyfoxdk: Yes, that should be fine, thanks for confirming.

aaronbuchwald (approved):

stateupgrade/state_upgrade.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	if upgrade.BalanceChange != nil {
+		state.AddBalance(account, (*big.Int)(upgrade.BalanceChange))
+	}
+	if len(upgrade.Code) != 0 {
+		// if the nonce is 0, set the nonce to 1 as we would when deploying a contract at
+		// the address.
+		if isEIP158 && state.GetNonce(account) == 0 {
+			state.SetNonce(account, 1)
```
anusha-ctrl, line 38: are we setting nonce = 1 to prevent account deletion according to EIP158?

anusha-ctrl: If we set the code, why do we need lines 37-38?

anusha-ctrl: Let me know if I am misunderstanding something. https://github.com/ethereum/EIPs/blob/master/EIPS/eip-158.md

precompile/contract/interfaces.go
```diff
@@ -34,14 +34,11 @@ type StateDB interface {
 	GetState(common.Address, common.Hash) common.Hash
 	SetState(common.Address, common.Hash, common.Hash)
 
-	SetCode(common.Address, []byte)
-
 	SetNonce(common.Address, uint64)
```
anusha-ctrl, line 37: why did we remove these functions from precompile?

anusha-ctrl:
do we need to modify genesis as well?

precompile/contract/interfaces.go
darioush: They were not used, so we should keep the interfaces narrower when possible.
I believe it was used prior to the refactoring.

stateupgrade/state_upgrade.go
darioush: this code was added for symmetry with https://github.com/ava-labs/subnet-evm/blob/master/core/vm/evm.go?rgh-link-date=2023-03-01T10%3A25%3A28Z#L526

darioush:
@aaronbuchwald explicitly wanted to leave state modification out of genesis (since you can already control genesis via the alloc).
I am not strongly opinionated on whether we should apply the state modification upgrade at genesis.

stateupgrade/state_upgrade.go
anusha-ctrl: Right [that](https://github.com/ava-labs/subnet-evm/blob/master/core/vm/evm.go?rgh-link-date=2023-03-01T10%3A25%3A28Z#L526) makes sense to increase the nonce after creation since it is considered ["empty".](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-158.md) However, we set the account code to some codebytes which makes it not "empty". Not sure why we are also increasing nonce as well

precompile/contract/interfaces.go
anusha-ctrl: Ok, makes sense but what if future precompiles want to use these functions? I'm ok either way

anusha-ctrl:
Ah yes,`GenesisAlloc` solves this thanks! 👍

stateupgrade/state_upgrade.go
darioush: true. at the end, a contract deployed to a new address with create will have nonce = 1 and if we deploy a new contract via the state upgrade I think it is better to also have nonce = 1

anusha-ctrl (approved):

params/precompile_upgrade.go
```diff
@@ -149,18 +149,33 @@ func (c *ChainConfig) verifyPrecompileUpgrades() error {
...
+		if previousUpgradeTimestamp != nil && upgradeTimestamp.Cmp(previousUpgradeTimestamp) <= 0 {
```
ceyonur, line 159: if we decided not to use state upgrades at genesis, should we check if timestamp != 0? cc @aaronbuchwald

stateupgrade/state_upgrade.go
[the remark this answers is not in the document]
ceyonur: what's our final decision with this?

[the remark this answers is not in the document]
aaronbuchwald: I'll defer to @darioush on this

params/precompile_upgrade.go
aaronbuchwald: Ya I think we should add this check since we are not allowing them in the genesis

darioush:
@ceyonur We decided to consider revisiting the packaging structure in a later PR.

PR #551, opened 2023-03-01
Set minimum golang version to go1.19.6

patrick-ogrady:
## Why this should be merged
go1.20.1 is not stable yet (https://github.com/golang/go/issues/58798). In the meantime, we'll stay on go1.19.6.

related: #548

### DIFF
--- a/.github/workflows/lint-tests-release.yml
+++ b/.github/workflows/lint-tests-release.yml
@@ -28,7 +28,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        go: ["1.20"]
+        go: ["1.19"]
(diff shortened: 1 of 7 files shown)


.github/workflows/lint-tests-release.yml
```diff
@@ -24,17 +24,13 @@ jobs:
           args: --timeout 3m
 
   unit_test:
-    name: Golang Unit Tests v${{ matrix.go }} (${{ matrix.os }})
```
patrick-ogrady, line 27: If there is only one golang version we are testing, we shouldn't use a matrix.

aaronbuchwald (approved):

darioush (approved):

PR #554, opened 2023-03-02
Add ChainConfig JSON Unmarshaller

ceyonur:
## Why this should be merged

Adds unmarshaller to chain config json

## How this works

Adds custom `UnmarshalJSON` function 

## How this was tested

Added unit test

## How is this documented

Existing API doc already covers this.


params/config.go
```diff
@@ -664,3 +617,69 @@ func (c *ChainConfig) GetFeeConfig() commontype.FeeConfig {
...
+	var u upgrades
+	if err := json.Unmarshal(input, &u); err != nil {
+		return err
+	}
+	cu.ChainConfig = cc
+	cu.UpgradeConfig = u.UpgradeConfig
+	return nil
+}
+
+func (c *ChainConfig) ToWithUpgradesJSON() *ChainConfigWithUpgradesJSON {
```
darioush, line 680: can we add a comment about the purpose of this function?

params/config_test.go
```diff
@@ -244,4 +260,9 @@ func TestChainConfigMarshalWithUpgrades(t *testing.T) {
 	require.NoError(t, err)
 	expectedJSON := `{"chainId":1,"feeConfig":{"gasLimit":8000000,"targetBlockRate":2,"minBaseFee":25000000000,"targetGas":15000000,"baseFeeChangeDenominator":36,"minBlockGasCost":0,"maxBlockGasCost":1000000,"blockGasCostStep":200000},"homesteadBlock":0,"eip150Block":0,"eip150Hash":"0x0000000000000000000000000000000000000000000000000000000000000000","eip155Block":0,"eip158Block":0,"byzantiumBlock":0,"constantinopleBlock":0,"petersburgBlock":0,"istanbulBlock":0,"muirGlacierBlock":0,"subnetEVMTimestamp":0,"upgrades":{"precompileUpgrades":[{"txAllowListConfig":{"blockTimestamp":100}}]}}`
 	require.JSONEq(t, expectedJSON, string(result))
```
darioush, line 262: nit: maybe we should remove this JSONEq check?

darioush (approved):

params/config.go
ceyonur: added

params/config_test.go
ceyonur: Well, I think it wouldn't hurt to keep this as an extra test.

aaronbuchwald (approved):

PR #561, opened 2023-03-03
Fix Dockerfile and add Build Image to GH Actions

aaronbuchwald:
## Why this should be merged

This PR fixes the Dockerfile for Subnet-EVM.

## How this works

This removes the specific version of git included in the Dockerfile which now results in a downgrade when using the new golang v1.19.6 buster image.

## How this was tested

This was tested locally by using the script `./scripts/build_image.sh`

## How is this documented

No documentation change necessary.


scripts/constants.sh
```diff
@@ -26,7 +26,7 @@ fi
 
 echo "Using branch: ${CURRENT_BRANCH}"
 
-BUILD_IMAGE_ID=${BUILD_IMAGE_ID:-"$AVALANCHEGO_VERSION-$SUBNET_EVM_COMMIT_ID"}
+BUILD_IMAGE_ID=${BUILD_IMAGE_ID:-"$AVALANCHEGO_VERSION-Subnet-EVM-$CURRENT_BRANCH"}
```
hexfusion, line 29:
```suggestion
BUILD_IMAGE_ID=${BUILD_IMAGE_ID:-"${AVALANCHEGO_VERSION}-Subnet-EVM-${CURRENT_BRANCH}"}
```

.github/workflows/ci.yml
```diff
@@ -43,6 +43,16 @@ jobs:
       - run: ./scripts/coverage.sh
         shell: bash
 
+  build_image:
+    name: Build Docker Image
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v3
+      - run: echo ${{ secrets.docker_pass }} | docker login --username ${{ secrets.docker_username }} --password-stdin
```
aaronbuchwald, line 51: This uses the same setup as AvalancheGo, ran into issues with the docker login GitHub action where I wasn't able to get it working: https://github.com/docker/login-action

darioush: should we keep the echo? not sure where these logs end up

anusha-ctrl (approved):

PR #562, opened 2023-03-03
remove stderr pipelining

ceyonur:
## Why this should be merged

We don't need to pipe eth logger to stderr as we have a proper logger served by rpcchainvm. 

## How this works

Removes special casing for stderr

## How this was tested

Locally tested


aaronbuchwald (approved):
LGTM

darioush (approved):

PR #572, opened 2023-03-15
Single warp message per tx (Teleporter demo only)

michaelkaplan13:
Changes to account for only allowing a single Warp message to be delivered per transaction rather than multiple.

The access list was previously an RLP encoded array of messages, and now it is just a single serialized message. A single byte with value of 0x01 is prepended to the message before we pad it with zeros to account for the case when the message serialized starts with a 0 byte.

Note: This PR is only meant for the `teleporter-demo` branch, not to be merged to `master`


precompile/contracts/warpmessenger/contract.go
```diff
@@ -211,22 +206,10 @@ func getBlockchainID(accessibleState contract.AccessibleState, caller common.Add
...
-	if err != nil {
-		return big.NewInt(0), err
-	}
-	unpacked := *abi.ConvertType(res[0], new(*big.Int)).(**big.Int)
-	return unpacked, nil
-}
-
-// PackGetVerifiedWarpMessage packs [messageIndex] of type *big.Int into the appropriate arguments for getVerifiedWarpMessage.
-// the packed bytes include selector (first 4 func signature bytes).
+// PackGetVerifiedWarpMessage packs the include selector (first 4 func signature bytes).
```
cam-schultz, line 209: Can you clarify this comment?

cam-schultz (approved):
LGTM. Just some minor nits.

precompile/contracts/warpmessenger/contract.go
michaelkaplan13: Went one better and removed this function all together since it wasn't used anywhere. :sweat_smile: Same with `PackGetBlockchainID()`.

PR #573, opened 2023-03-16
Warp preparation

aaronbuchwald:
## Why this should be merged

This PR adds small changes in preparation for the warp precompile to reduce the warp PR's size.

## How this works

- Updates hash slice encoding/decoding and add a unit test
- Changes the behavior of precompile predicates so that it does not require the proposervm header if a proposer predicate is enabled (proposervm will not always pass in a context before activation, so it is up to the precompile predicate to handle nil)
- Add Warp Message Writer interface into the `AcceptContext`
- Remove `ctx` from the warp backend interface

## How this was tested

CI

## How is this documented

No documentation changes required for this change.

### DIFF
--- a/warp/backend.go
+++ b/warp/backend.go
@@ -4,7 +4,6 @@
 package warp
 
 import (
-	"context"
 	"fmt"
 
 	"github.com/ava-labs/avalanchego/cache"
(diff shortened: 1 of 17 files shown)


darioush (approved):

warp/backend.go
```diff
@@ -65,7 +64,7 @@ func (w *warpBackend) AddMessage(ctx context.Context, unsignedMessage *avalanche
 	return nil
 }
 
-func (w *warpBackend) GetSignature(ctx context.Context, messageID ids.ID) ([bls.SignatureLen]byte, error) {
+func (w *warpBackend) GetSignature(messageID ids.ID) ([bls.SignatureLen]byte, error) {
 	if sig, ok := w.signatureCache.Get(messageID); ok {
```
anusha-ctrl, line 68: just curious, why did we remove `ctx`

anusha-ctrl (approved):

warp/backend.go
aaronbuchwald: We do not currently use it anywhere and in the warp precompile PR, we will need to call `AddMessage` from the precompile `Accept` handler, which doesn't have a context available, so it seemed cleaner to me to get rid of it unless we need to use it.

PR #574, opened 2023-03-16
release tickers on shutdown

darioush:
## Why this should be merged
Fewer leaked goroutines in stopped chains (eg in some unit tests).

## How this works
This releases tickers when the awaitEthTxGossip function returns.

## How this was tested
UTs

## How is this documented
No change.

### DIFF
--- a/plugin/evm/gossiper.go
+++ b/plugin/evm/gossiper.go
@@ -235,8 +235,6 @@ func (n *pushGossiper) queuePriorityRegossipTxs() types.Transactions {
 func (n *pushGossiper) awaitEthTxGossip() {
 	n.shutdownWg.Add(1)
 	go n.ctx.Log.RecoverAndPanic(func() {
-		defer n.shutdownWg.Done()
-
 		var (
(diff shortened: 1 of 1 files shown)


dhrubabasu (approved):

plugin/evm/gossiper.go
```diff
@@ -242,6 +242,11 @@ func (n *pushGossiper) awaitEthTxGossip() {
 			regossipTicker         = time.NewTicker(n.config.RegossipFrequency.Duration)
 			priorityRegossipTicker = time.NewTicker(n.config.PriorityRegossipFrequency.Duration)
 		)
+		defer func() {
+			gossipTicker.Stop()
+			regossipTicker.Stop()
+			priorityRegossipTicker.Stop()
+		}()
```
StephenButtolph, line 249: nit: can we only have one defer? I think we can move the `n.shutdownWg.Done()` into this.

StephenButtolph (approved):

ceyonur (approved):

PR #577, opened 2023-03-16
adds tests for custom abi fns

darioush:
## Why this should be merged
Adds test coverage for fns added for precompile handling in the abi package

## How this works
Tests some assumptions

## How this was tested
CI

## How is this documented
No change

### DIFF
--- a/accounts/abi/abi_extra_test.go
+++ b/accounts/abi/abi_extra_test.go
@@ -1,0 +1,45 @@
+// (c) 2023, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package abi
+
+import (
+	"bytes"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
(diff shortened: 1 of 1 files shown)


accounts/abi/abi_extra_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+func TestUnpackInputIntoInterface(t *testing.T) {
+	abi, err := JSON(strings.NewReader(TEST_ABI))
+	require.NoError(t, err)
+
+	input := []byte("hello")
+	data, err := abi.Pack("receive", input)
+	require.NoError(t, err)
+
+	var v []byte
```
aaronbuchwald, line 26: Can we add a test case for an interface in the same style as for `UnpackIntoInterface` ?https://github.com/ava-labs/subnet-evm/blob/master/accounts/abi/abi_test.go#L775

aaronbuchwald (approved):

PR #578, opened 2023-03-17
simplify functionSignatureRegex

darioush:
Simplifies the function signature regex (removes unneeded [] and () and combines +? to *)

### DIFF
--- a/precompile/contract/utils.go
+++ b/precompile/contract/utils.go
@@ -20,7 +20,7 @@ const (
 	ReadGasCostPerSlot  = 5_000
 )
 
-var functionSignatureRegex = regexp.MustCompile(`[\w]+\(((([\w]+)?)|((([\w]+),)+([\w]+)))\)`)
+var functionSignatureRegex = regexp.MustCompile(`\w+\((\w*|(\w+,)+\w+)\)`)
 
 // CalculateFunctionSelector returns the 4 byte function selector that results from [functionSignature]
 // Ex. the function setBalance(addr address, balance uint256) should be passed in as the string:


precompile/contract/utils.go
```diff
@@ -20,7 +20,7 @@ const (
 	ReadGasCostPerSlot  = 5_000
 )
 
-var functionSignatureRegex = regexp.MustCompile(`[\w]+\(((([\w]+)?)|((([\w]+),)+([\w]+)))\)`)
+var functionSignatureRegex = regexp.MustCompile(`\w+\((\w*|(\w+,)+\w+)\)`)
```
aaronbuchwald, line 23: Fun with ChatGPT:

```
In Solidity, a function signature is the first 4 bytes of the Keccak-256 hash of the function's canonical signature. The canonical signature includes the function name and parameter types enclosed in parentheses, separated by commas.

Here's a regular expression that matches a Solidity function signature:

ruby
Copy code
/^[a-zA-Z_]\w*\((\w+(,\s*\w+)*)?\)$/
This regular expression matches a string that starts with a letter or underscore, followed by zero or more letters, underscores, or digits, representing the function name. The function name is followed by an opening parenthesis, zero or more parameter types separated by commas and optional spaces, and a closing parenthesis.

For example, this regular expression matches function signatures like:

scss
Copy code
myFunction()
myFunction(uint256,string)
_transfer(address,address,uint256)
Note that this regular expression does not include the return types, as they are not part of the function signature in Solidity.
```

aaronbuchwald (approved):

precompile/contract/utils.go
ceyonur: resim

ceyonur (approved):

PR #579, opened 2023-03-17
Update load test to use ANR for network setup

aaronbuchwald:
## Why this should be merged

This PR splits out the e2e test setup using the ANR that will be used for the warp e2e tests. This is intended to reduce the size of the actual Warp PR.

## How this works

This PR copies the style of https://github.com/ava-labs/hypersdk/blob/warp-messaging/examples/tokenvm/tests/e2e/e2e_test.go to use the ANR to set up a subnet with BLS keys registered on the P-Chain.

It adds the following
- thin wrapper for ANR to minimize how much of the ANR interface is exposed to the test logic itself
- adds a minimal wrapper of the ANR as a new command to set up an empty default network, a network with two subnets, and stop the network runner

## How this was tested

Tested locally to ensure the load test still works and the wrapped ANR commands work correclty.

## How is this documented

This PR should add documentation to the README explaining how to use the new command before getting merged.


aaronbuchwald:
This PR is blocked on https://github.com/ava-labs/avalanche-network-runner/pull/467, which will be used to create new subnets with validators that have registered their BLS keys on the P-Chain.

tests/utils/runner/runner.go
```diff
@@ -0,0 +1,295 @@
...
+	"github.com/ava-labs/avalanche-network-runner/rpcpb"
+	runner_server "github.com/ava-labs/avalanche-network-runner/server"
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/utils/logging"
+	"github.com/ava-labs/avalanchego/utils/wrappers"
+	"github.com/ava-labs/subnet-evm/plugin/evm"
+	"github.com/ethereum/go-ethereum/log"
+)
+
+type Subnet struct {
```
aaronbuchwald, line 22: This file wraps the ANR with a struct that can be used to setup a network and execute tests without needing to know the details of the backend setup.

Ideally, this would be replaced with an interface that can use ANR, an existing network, a k8s backend, etc.

tests/utils/runner/runner.go
```diff
@@ -0,0 +1,377 @@
...
+			"throttler-outbound-validator-alloc-size":"107374182"
+		}`,
+	}
+	// If AVALANCHEGO_BUILD_PATH is populated, override location set by GOPATH
+	if envBuildPath, exists := os.LookupEnv("AVALANCHEGO_BUILD_PATH"); exists {
+		defaultConfig.AvalancheGoExecPath = fmt.Sprintf("%s/avalanchego", envBuildPath)
+		defaultConfig.PluginDir = fmt.Sprintf("%s/plugins", envBuildPath)
+	}
+	return defaultConfig
+}
```
ceyonur, line 79: can we make this a var?

tests/utils/runner/runner.go
```diff
@@ -0,0 +1,377 @@
...
+func (n *NetworkManager) startServer(ctx context.Context) (<-chan struct{}, error) {
+	done := make(chan struct{})
+	zapServerLog, err := n.logFactory.Make("server")
+	if err != nil {
+		return nil, fmt.Errorf("failed to make server log: %w", err)
+	}
+
+	n.anrServer, err = runner_server.New(
+		runner_server.Config{
+			Port:                ":12352",
```
ceyonur, line 109: should these be a part of config?

tests/utils/runner/runner.go
```diff
@@ -0,0 +1,377 @@
...
+func (n *NetworkManager) initClient() error {
+	if n.anrClient != nil {
+		return nil
+	}
+
+	return n.startClient()
+}
+
+// init starts the ANR server and client if they are not yet populated
+func (n *NetworkManager) init() error {
```
ceyonur, line 185: can we rename this to `start()` (or similar) to avoid any possible confusion with built-in `init` function

tests/utils/runner/runner.go
ceyonur, line 1: can we rename the file to network_manager?

tests/utils/runner/runner.go
```diff
@@ -0,0 +1,377 @@
...
+	// "start" is async, so wait some time for cluster health
+	log.Info("waiting for all VMs to report healthy", "VMID", evm.ID)
+	for {
+		v, err := n.anrClient.Health(ctx)
+		log.Info("Pinged CLI Health", "result", v, "err", err)
+		if err != nil {
+			time.Sleep(1 * time.Second)
+			continue
+		}
+		// TODO: clean this up
```
anusha-ctrl, line 240: this logic seems weird. if err != nil we will continuously wait, when would we enter the second code block?

tests/utils/runner/runner.go
```diff
@@ -0,0 +1,377 @@
...
+		GlobalNodeConfig: `{
+			"log-display-level":"info",
+			"proposervm-use-current-height":true,
+			"throttler-inbound-validator-alloc-size":"107374182",
+			"throttler-inbound-node-max-processing-msgs":"100000",
+			"throttler-inbound-bandwidth-refill-rate":"1073741824",
+			"throttler-inbound-bandwidth-max-burst-size":"1073741824",
+			"throttler-inbound-cpu-validator-alloc":"100000",
+			"throttler-inbound-disk-validator-alloc":"10737418240000",
+			"throttler-outbound-validator-alloc-size":"107374182"
```
aaronbuchwald, line 70: remove unneeded params

tests/utils/runner/runner.go
aaronbuchwald: Updated

tests/utils/runner/runner.go
aaronbuchwald: It currently uses this function so that it can expand environment variables when it's called. We could do the same in `init()`, but since we use environment variables here, I'd prefer to expand them when the function is called.

tests/utils/runner/runner.go
aaronbuchwald: Ya I think it's fine for the port to be a part of the config. Also, this is using the avalanche-network-runner code, and I think we should use the current interface rather than try to push style changes to that dependency unless we're doing a larger re-write.

Overall, I added this here to try and keep the use of ANR as well-separated from our actual tests as possible.

tests/utils/runner/runner.go
aaronbuchwald: I don't think this should be easily confused with the built-in `init()` function and this follows the convention used in avalanchego to use `init()` to ensure that it is set up correctly before accessing the client/server on the network manager itself.

darioush (approved):

PR #587, opened 2023-03-21
Add ability to specify commit of avalanchego in ci

### LINKED ISSUE #500: Update E2E Tests to Optionally run with AvalancheGo branch/commit instead of release
**Context and scope**
Currently the e2e tests that run in CI depend on an AvalancheGo release, which is installed in a specific location to be used in the e2e tests here: https://github.com/ava-labs/subnet-evm/blob/master/.github/workflows/lint-tests-release.yml#L66-L68

We should update this so that if a commit or branch name is specified in `scripts/versions.sh` here: https://github.com/ava-labs/subnet-evm/blob/master/scripts/versions.sh#L6 then it will checkout the relevant commit/branch of AvalancheGo and build it in order to run the e2e tests.

This will improve the CI workflow when there are breaking changes made to AvalancheGo that we would like to test in Subnet-EVM.

richardpringle:
## Why this should be merged

Resolves #500

## How this works

use the `AVALANCHE_VERSION` environment variable to specify a commit or branch name (or tag without a prebuilt binary).

## How this was tested

TODO: test with github actions

## How is this documented

documentation for the whole process needs to be added

### DIFF
--- a/scripts/install_avalanchego_release.sh
+++ b/scripts/install_avalanchego_release.sh
@@ -27,20 +27,64 @@ if [[ ${GOOS} == "darwin" ]]; then
   AVAGO_DOWNLOAD_PATH=${BASEDIR}/avalanchego-macos-${VERSION}.zip
 fi
 
-AVALANCHEGO_BUILD_PATH=${AVALANCHEGO_BUILD_PATH:-${BASEDIR}/avalanchego-${VERSION}}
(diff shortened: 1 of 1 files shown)


richardpringle:
Branch name is no good for the `Build Docker Image` step, closing in favour of #588

PR #589, opened 2023-03-22
Add ability to specify commit of avalanchego in ci

### LINKED ISSUE #500: Update E2E Tests to Optionally run with AvalancheGo branch/commit instead of release

richardpringle:
## Why this should be merged

Resolves #500

## How this works

use the `AVALANCHE_VERSION` environment variable to specify a commit or branch name (or tag without a prebuilt binary).

## How this was tested

tested here for a `commit`:
https://github.com/ava-labs/subnet-evm/actions/runs/4512439095/jobs/7945887581

Screenshot 2023-03-24 at 10 54 24 AM

and here for a `branch`:
https://github.com/ava-labs/subnet-evm/actions/runs/4512412729/jobs/7945824005

Screenshot 2023-03-24 at 10 56 36 AM

## How is this documented

documentation for the whole process needs to be added

### DIFF
--- a/scripts/install_avalanchego_release.sh
+++ b/scripts/install_avalanchego_release.sh
@@ -11,41 +11,118 @@ source "$SUBNET_EVM_PATH"/scripts/versions.sh
 # Load the constants
 source "$SUBNET_EVM_PATH"/scripts/constants.sh
 
-VERSION=$AVALANCHEGO_VERSION
-
 ############################
 # download avalanchego
 # https://github.com/ava-labs/avalanchego/releases
 GOARCH=$(go env GOARCH)
 GOOS=$(go env GOOS)
-BASEDIR=${BASE_DIR:-"/tmp/avalanchego-release"}
+BASEDIR=${BASEDIR:-"/tmp/avalanchego-release"}
+AVALANCHEGO_BUILD_PATH=${AVALANCHEGO_BUILD_PATH:-${BASEDIR}/avalanchego}
+
 mkdir -p ${BASEDIR}
-AVAGO_DOWNLOAD_URL=https://github.com/ava-labs/avalanchego/releases/download/${VERSION}/avalanchego-linux-${GOARCH}-${VERSION}.tar.gz
-AVAGO_DOWNLOAD_PATH=${BASEDIR}/avalanchego-linux-${GOARCH}-${VERSION}.tar.gz
+
+AVAGO_DOWNLOAD_URL=https://github.com/ava-labs/avalanchego/releases/download/${AVALANCHEGO_VERSION}/avalanchego-linux-${GOARCH}-${AVALANCHEGO_VERSION}.tar.gz
+AVAGO_DOWNLOAD_PATH=${BASEDIR}/avalanchego-linux-${GOARCH}-${AVALANCHEGO_VERSION}.tar.gz
+
 if [[ ${GOOS} == "darwin" ]]; then
-  AVAGO_DOWNLOAD_URL=https://github.com/ava-labs/avalanchego/releases/download/${VERSION}/avalanchego-macos-${VERSION}.zip
-  AVAGO_DOWNLOAD_PATH=${BASEDIR}/avalanchego-macos-${VERSION}.zip
+  AVAGO_DOWNLOAD_URL=https://github.com/ava-labs/avalanchego/releases/download/${AVALANCHEGO_VERSION}/avalanchego-macos-${AVALANCHEGO_VERSION}.zip
+  AVAGO_DOWNLOAD_PATH=${BASEDIR}/avalanchego-macos-${AVALANCHEGO_VERSION}.zip
 fi
 
-AVALANCHEGO_BUILD_PATH=${AVALANCHEGO_BUILD_PATH:-${BASEDIR}/avalanchego-${VERSION}}
-mkdir -p $AVALANCHEGO_BUILD_PATH
+BUILD_DIR=${AVALANCHEGO_BUILD_PATH}-${AVALANCHEGO_VERSION}
 
-if [[ ! -f ${AVAGO_DOWNLOAD_PATH} ]]; then
-  echo "downloading avalanchego ${VERSION} at ${AVAGO_DOWNLOAD_URL} to ${AVAGO_DOWNLOAD_PATH}"
-  curl -L ${AVAGO_DOWNLOAD_URL} -o ${AVAGO_DOWNLOAD_PATH}
-fi
-echo "extracting downloaded avalanchego to ${AVALANCHEGO_BUILD_PATH}"
-if [[ ${GOOS} == "linux" ]]; then
-  mkdir -p ${AVALANCHEGO_BUILD_PATH} && tar xzvf ${AVAGO_DOWNLOAD_PATH} --directory ${AVALANCHEGO_BUILD_PATH} --strip-components 1
-elif [[ ${GOOS} == "darwin" ]]; then
-  unzip ${AVAGO_DOWNLOAD_PATH} -d ${AVALANCHEGO_BUILD_PATH}
-  mv ${AVALANCHEGO_BUILD_PATH}/build/* ${AVALANCHEGO_BUILD_PATH}
-  rm -rf ${AVALANCHEGO_BUILD_PATH}/build/
+extract_archive() {
+  mkdir -p ${BUILD_DIR}
+
+  if [[ ${AVAGO_DOWNLOAD_PATH} == *.tar.gz ]]; then
+    tar xzvf ${AVAGO_DOWNLOAD_PATH} --directory ${BUILD_DIR} --strip-components 1
+  elif [[ ${AVAGO_DOWNLOAD_PATH} == *.zip ]]; then
+    unzip ${AVAGO_DOWNLOAD_PATH} -d ${BUILD_DIR}
+    mv ${BUILD_DIR}/build/* ${BUILD_DIR}
+    rm -rf ${BUILD_DIR}/build/
+  fi
+}
+
+# first check if we already have the archive
+if [[ -f ${AVAGO_DOWNLOAD_PATH} ]]; then
+  # if the download path already exists, extract and exit
+  echo "found avalanchego ${AVALANCHEGO_VERSION} at ${AVAGO_DOWNLOAD_PATH}"
+
+  extract_archive
+else
+  # try to download the archive if it exists
+  if curl -s --head --request GET ${AVAGO_DOWNLOAD_URL} | grep "302" > /dev/null; then
+    echo "${AVAGO_DOWNLOAD_URL} found"
+    echo "downloading to ${AVAGO_DOWNLOAD_PATH}"
+    curl -L ${AVAGO_DOWNLOAD_URL} -o ${AVAGO_DOWNLOAD_PATH}
+
+    extract_archive
+  else
+    # else the version is a git commitish (or it's invalid)
+    GIT_CLONE_URL=https://github.com/ava-labs/avalanchego.git
+    GIT_CLONE_PATH=${BASEDIR}/avalanchego-repo/
+    
+    # check to see if the repo already exists, if not clone it 
+    if [[ ! -d ${GIT_CLONE_PATH} ]]; then
+      echo "cloning ${GIT_CLONE_URL} to ${GIT_CLONE_PATH}"
+      git clone --no-checkout ${GIT_CLONE_URL} ${GIT_CLONE_PATH}
+    fi
+
+    # check to see if the commitish exists in the repo
+    WORKDIR=$(pwd)
+
+    cd ${GIT_CLONE_PATH}
+
+    git fetch
+
+    echo "checking out ${AVALANCHEGO_VERSION}"
+
+    set +e
+    # try to checkout the branch
+    git checkout origin/${AVALANCHEGO_VERSION} > /dev/null 2>&1
+    CHECKOUT_STATUS=$?
+    set -e
+
+    # if it's not a branch, try to checkout the commit 
+    if [[ $CHECKOUT_STATUS -ne 0 ]]; then
+      set +e
+      git checkout ${AVALANCHEGO_VERSION} > /dev/null 2>&1
+      CHECKOUT_STATUS=$?
+      set -e
+
+      if [[ $CHECKOUT_STATUS -ne 0 ]]; then
+        echo
+        echo "'${VERSION}' is not a valid release tag, commit hash, or branch name"
+        exit 1
+      fi
+    fi
+
+    COMMIT=$(git rev-parse HEAD)
+
+    # use the commit hash instead of the branch name or tag
+    BUILD_DIR=${AVALANCHEGO_BUILD_PATH}-${COMMIT}
+
+    # if the build-directory doesn't exist, build avalanchego
+    if [[ ! -d ${BUILD_DIR} ]]; then    
+      echo "building avalanchego ${COMMIT} to ${BUILD_DIR}"
+      ./scripts/build.sh
+      mkdir -p ${BUILD_DIR}
+
+      mv ${GIT_CLONE_PATH}/build/* ${BUILD_DIR}/
+    fi
+
+    cd $WORKDIR
+  fi
 fi
 
 AVALANCHEGO_PATH=${AVALANCHEGO_BUILD_PATH}/avalanchego
 AVALANCHEGO_PLUGIN_DIR=${AVALANCHEGO_BUILD_PATH}/plugins
 
-echo "Installed AvalancheGo release ${VERSION}"
+mkdir -p ${AVALANCHEGO_BUILD_PATH}
+    
+cp ${BUILD_DIR}/avalanchego ${AVALANCHEGO_PATH}
+
+
+echo "Installed AvalancheGo release ${AVALANCHEGO_VERSION}"
 echo "AvalancheGo Path: ${AVALANCHEGO_PATH}"
 echo "Plugin Dir: ${AVALANCHEGO_PLUGIN_DIR}"
--- a/scripts/versions.sh
+++ b/scripts/versions.sh
@@ -3,7 +3,8 @@
 # Set up the versions to be used - populate ENV variables only if they are not already populated
 SUBNET_EVM_VERSION=${SUBNET_EVM_VERSION:-'v0.4.12'}
 # Don't export them as they're used in the context of other calls
-AVALANCHEGO_VERSION=${AVALANCHE_VERSION:-'v1.9.11'}
+AVALANCHE_VERSION=${AVALANCHE_VERSION:-'v1.9.11'}
+AVALANCHEGO_VERSION=${AVALANCHEGO_VERSION:-$AVALANCHE_VERSION}
 GINKGO_VERSION=${GINKGO_VERSION:-'v2.2.0'}
 
 # This won't be used, but it's here to make code syncs easier


scripts/install_avalanchego_release.sh
```diff
@@ -27,20 +27,64 @@ if [[ ${GOOS} == "darwin" ]]; then
...
+    mkdir -p $GIT_CLONE_PATH
+
+    WORKDIR=$PWD
+    cd $GIT_CLONE_PATH
+
+    # if the git repo already exists, fetch, otherwise clone
+    if [[ -d .git ]]; then
+      git fetch
+    else
+      git clone ${GIT_CLONE_URL} .
```
darioush, line 53: should we make this a shallow clone of the repository and avoid fetching the entire repository if the dir exists?

scripts/install_avalanchego_release.sh
```diff
@@ -27,20 +27,64 @@ if [[ ${GOOS} == "darwin" ]]; then
...
 fi
 
-AVALANCHEGO_BUILD_PATH=${AVALANCHEGO_BUILD_PATH:-${BASEDIR}/avalanchego-${VERSION}}
+# TODO: Please read:
+# not sure why the version isn't included in CI currently... open to suggestions on how to handle this better
+AVALANCHEGO_BUILD_PATH=${AVALANCHEGO_BUILD_PATH-${VERSION}:-${BASEDIR}/avalanchego-${VERSION}}
 mkdir -p $AVALANCHEGO_BUILD_PATH
 
 if [[ ! -f ${AVAGO_DOWNLOAD_PATH} ]]; then
   echo "downloading avalanchego ${VERSION} at ${AVAGO_DOWNLOAD_URL} to ${AVAGO_DOWNLOAD_PATH}"
```
darioush, line 36: could we move this to the download part of the `if` or update the language so it says `checking for` instead of downloading?

richardpringle: you bet

richardpringle: Screenshot 2023-03-22 at 3 21 00 PM

thoughts

scripts/install_avalanchego_release.sh
```diff
@@ -27,20 +27,66 @@ if [[ ${GOOS} == "darwin" ]]; then
   AVAGO_DOWNLOAD_PATH=${BASEDIR}/avalanchego-macos-${VERSION}.zip
 fi
 
-AVALANCHEGO_BUILD_PATH=${AVALANCHEGO_BUILD_PATH:-${BASEDIR}/avalanchego-${VERSION}}
+# TODO: Please read:
+# not sure why the version isn't included in CI currently... open to suggestions on how to handle this better
+AVALANCHEGO_BUILD_PATH=${AVALANCHEGO_BUILD_PATH-${VERSION}:-${BASEDIR}/avalanchego-${VERSION}}
```
aaronbuchwald, line 32: is the question why do we install this in a directory without the version as a suffix?

richardpringle: Yeah... made me think that it's being used elsewhere.

aaronbuchwald: Currently we use this as a utility in local development to install a release of AvalancheGo, but we use it more frequently in CI to download a specific release that Subnet-EVM depends on to run its e2e tests.

In CI, we populate this environment variable here: https://github.com/ava-labs/subnet-evm/blob/master/.github/workflows/ci.yml#L77 and then pass the same variable in to run our ginkgo e2e tests here: https://github.com/ava-labs/subnet-evm/blob/master/.github/workflows/ci.yml#L83 (we could definitely clean this up a bit).

Just noticed we actually do something else that is weird here which is we use `BASE_DIR` instead of `BASEDIR` in one place in the script (on master) here: https://github.com/ava-labs/subnet-evm/blob/master/scripts/install_avalanchego_release.sh#L21.

Running this script locally now I'm running into the issue that instead of installing in `/tmp/` it's actually installing in the Subnet-EVM directory in the build path: `v1.9.11:-/tmp/avalanchego-release/avalanchego-v1.9.11`.

It looks like this is because the syntax `${var:-<alternative>}` checks if `var` is populated and uses it if so, but otherwise uses `<alternative>`. Since this change is not using a single variable, it seems to just use the whole string `v1.9.11:-/tmp/avalanchego-release/avalanchego-v1.9.11` instead of putting it in `/tmp`.

richardpringle: Okay, take a look now. I left the commit as a fixup and will squash once I get the green light. Either we keep these changes or I can figure out a better way to deal with the branching logic.

scripts/install_avalanchego_release.sh
```diff
@@ -27,22 +27,72 @@ if [[ ${GOOS} == "darwin" ]]; then
   AVAGO_DOWNLOAD_PATH=${BASEDIR}/avalanchego-macos-${VERSION}.zip
 fi
 
-AVALANCHEGO_BUILD_PATH=${AVALANCHEGO_BUILD_PATH:-${BASEDIR}/avalanchego-${VERSION}}
+# TODO: Please read:
+# not sure why the version isn't included in CI currently... open to suggestions on how to handle this better
```
ceyonur, line 31: You mean why it's not being used or why is not working? I don't think we use any `AVALANCHE_VERSION=xxx` as an env var in CI currently. Subnet-EVM (or any other rpcchainVM) cannot work with all AvalancheGO versions. There is a protocl version handshake, and if AvalancheGo bumps that protocol version and older subnet-evm cannot work with it. Similarly a recent Subnet-EVM (with newer protocol version) won't work with an older AvalancheGo version. e2e tests takes around ~5 min. It's probably best to use latest possible avalanchego for Pull request CI (to reduce test time). However we can include "all supported" avalanchego versions in a release-CI.

.github/workflows/ci.yml
```diff
@@ -74,7 +74,7 @@ jobs:
         working-directory: ./contract-examples
       - name: Install AvalancheGo Release
         shell: bash
-        run: BASEDIR=/tmp/e2e-test AVALANCHEGO_BUILD_PATH=/tmp/e2e-test/avalanchego ./scripts/install_avalanchego_release.sh
+        run: BASE_DIR=/tmp/e2e-test AVALANCHEGO_BUILD_PATH=/tmp/e2e-test/avalanchego ./scripts/install_avalanchego_release.sh
```
aaronbuchwald, line 77: Can we align all of these `BASEDIR` and `BASE_DIR` to be the same? The previous bug was that they were ever different as the intended behavior is to allow someone to populate the environment variable and use the default if not populated.

scripts/install_avalanchego_release.sh
```diff
@@ -27,22 +27,72 @@ if [[ ${GOOS} == "darwin" ]]; then
...
+AVALANCHEGO_BUILD_PATH=${AVALANCHEGO_BUILD_PATH:-${BASEDIR}/avalanchego}
+VERSIONED_AVALANCHEGO_BUILD_PATH=${AVALANCHEGO_BUILD_PATH}-${VERSION}
 mkdir -p $AVALANCHEGO_BUILD_PATH
+mkdir -p $VERSIONED_AVALANCHEGO_BUILD_PATH
```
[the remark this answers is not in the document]
aaronbuchwald: I agree with this, I'd prefer to just use a single `AVALANCHEGO_BUILD_PATH` and either build in that location or unpack the downloaded release into that location rather than adding a separate versioned path.

.github/workflows/ci.yml
richardpringle: I don't mind doing that, but it looks like it was done on purpose as the pattern repeats itself in a couple of places, `AVALANCHE_VERSION` vs `AVALANCHE_GO_VERSION` for instance. `BASEDIR` is only meant to be used inside the script.

If I assume that using a "private" variable name like that is justified, I think it would be more pragmatic to check if the user has defined `BASEDIR` instead of `BASE_DIR` and error out with an explanation. I can do the same for `AVALANCHE_GO_VERSION` vs `AVALANCHE_VERSION`

scripts/install_avalanchego_release.sh
richardpringle: >You mean why it's not being used or why is not working?

@ceyonur, thanks for answering! It actually is being used in the last step for the e2e tests in CI. I can actually remove the question now.

aaronbuchwald: This is an interesting idea. As @ceyonur mentioned AvalancheGo and Subnet-EVM can only talk to each other if they are on the same RPCChainVMProtocolVersion.

AvalancheGo documents this for all releases here: https://github.com/ava-labs/avalanchego/blob/master/version/compatibility.json
Subnet-EVM documents this for all releases here: https://github.com/ava-labs/subnet-evm/blob/master/compatibility.json

I think the script (and whats in scope for this PR) should make sure that we can install any tag or commit-ish of AvalancheGo. In the future, your suggestion of moving this to a generally available install script in AvalancheGo and even a supported GitHub action would be really nice too.

As for what we run in CI, we could definitely add a check to CI that runs the Subnet-EVM e2e tests with every version of AvalancheGo that should be supported, which could just use this script (and actually would not depend on this change if we just wanted to run with official AvalancheGo releases). Although, I'd prefer this as a cronjob or manual GitHub action because it seems like overkill to run frequently given both the rpcchainvm server and client are in AvalancheGo, so I think it would make more sense to run this to confirm prior to releases or at a weekly cadence.

scripts/install_avalanchego_release.sh
```diff
@@ -27,22 +27,72 @@ if [[ ${GOOS} == "darwin" ]]; then
...
 
 if [[ ! -f ${AVAGO_DOWNLOAD_PATH} ]]; then
```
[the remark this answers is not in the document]
richardpringle: You're absolutely right for commits and binary-releases, but this doesn't apply for branches, which was specified in the [original issue](https://github.com/ava-labs/subnet-evm/issues/500).

>We should update this so that if a commit or **branch name**

(emphasis mine)

.github/workflows/ci.yml
aaronbuchwald: In this case, I think both should be public environment variables that can be easily overridden by the user and I'd prefer to be consistent in using the same environment variable in the script that the user needs to override rather than renaming in the background.

I think all of the renaming is confusing when looking at the script and wanting to know how to override a variable or where it comes from.

If we want to go back to having public vs. private env variables, I think we should be explicit and follow the convention of using lowercase for private env variables.

scripts/install_avalanchego_release.sh
```diff
@@ -27,22 +27,72 @@ if [[ ${GOOS} == "darwin" ]]; then
...
 fi
-echo "extracting downloaded avalanchego to ${AVALANCHEGO_BUILD_PATH}"
-if [[ ${GOOS} == "linux" ]]; then
-  mkdir -p ${AVALANCHEGO_BUILD_PATH} && tar xzvf ${AVAGO_DOWNLOAD_PATH} --directory ${AVALANCHEGO_BUILD_PATH} --strip-components 1
-elif [[ ${GOOS} == "darwin" ]]; then
-  unzip ${AVAGO_DOWNLOAD_PATH} -d ${AVALANCHEGO_BUILD_PATH}
-  mv ${AVALANCHEGO_BUILD_PATH}/build/* ${AVALANCHEGO_BUILD_PATH}
-  rm -rf ${AVALANCHEGO_BUILD_PATH}/build/
+
+if [[ ! -f ${VERSIONED_AVALANCHEGO_BUILD_PATH}/avalanchego ]]; then
```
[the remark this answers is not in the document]
richardpringle: See my response [here](https://github.com/ava-labs/subnet-evm/pull/589/files#r1146294527)

scripts/install_avalanchego_release.sh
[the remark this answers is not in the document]
richardpringle: @aaronbuchwald, why do you prefer a single build path? 

@ceyonur, the reason I did this was actually to avoid further nesting of logic, making the whole script harder to read. Not the best reason, but a reason, none-the-less.

I'm now thinking that `AVALANCHE_VERSION` is too overloaded and that it would be better to separate things into `AVALANCHE_RELEASE_VERSION`, `AVALANCHE_COMMIT`, and `AVALANCHE_BRANCH` and the logic for each variant is completely different (branches need to pulled in every time).

@aaronbuchwald, I also think it makes a lot more sense to have versioned paths, the same way that `cargo` stores dependencies for instance. I can make changes such that we don't rebuild if an existing commit or release-tagged version already exists, and only rebuild for a specified branch. 

Thoughts?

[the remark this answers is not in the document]
aaronbuchwald: Sounds good to me 👍

scripts/versions.sh
```diff
@@ -3,7 +3,8 @@
 # Set up the versions to be used - populate ENV variables only if they are not already populated
 SUBNET_EVM_VERSION=${SUBNET_EVM_VERSION:-'v0.4.12'}
 # Don't export them as they're used in the context of other calls
-AVALANCHEGO_VERSION=${AVALANCHE_VERSION:-'v1.9.11'}
+AVALANCHE_VERSION=${AVALANCHE_VERSION:-'v1.9.11'}
+AVALANCHEGO_VERSION=${AVALANCHEGO_VERSION:-$AVALANCHE_VERSION}
```
aaronbuchwald, line 7:
```suggestion
AVALANCHEGO_VERSION=${AVALANCHEGO_VERSION:-'v1'.9.11'}
```
These are the same thing, can we make them one variable

scripts/install_avalanchego_release.sh
```diff
@@ -11,41 +11,119 @@ source "$SUBNET_EVM_PATH"/scripts/versions.sh
...
-  curl -L ${AVAGO_DOWNLOAD_URL} -o ${AVAGO_DOWNLOAD_PATH}
-fi
-echo "extracting downloaded avalanchego to ${AVALANCHEGO_BUILD_PATH}"
-if [[ ${GOOS} == "linux" ]]; then
-  mkdir -p ${AVALANCHEGO_BUILD_PATH} && tar xzvf ${AVAGO_DOWNLOAD_PATH} --directory ${AVALANCHEGO_BUILD_PATH} --strip-components 1
-elif [[ ${GOOS} == "darwin" ]]; then
-  unzip ${AVAGO_DOWNLOAD_PATH} -d ${AVALANCHEGO_BUILD_PATH}
-  mv ${AVALANCHEGO_BUILD_PATH}/build/* ${AVALANCHEGO_BUILD_PATH}
-  rm -rf ${AVALANCHEGO_BUILD_PATH}/build/
+extract_archive() {
```
aaronbuchwald, line 35: Nice

scripts/install_avalanchego_release.sh
```diff
@@ -11,41 +11,119 @@ source "$SUBNET_EVM_PATH"/scripts/versions.sh
...
+else
+  # try to download the archive if it exists
+  if curl -s --head --request GET ${AVAGO_DOWNLOAD_URL} | grep "302" > /dev/null; then
+    echo "${AVAGO_DOWNLOAD_URL} found"
+    echo "downloading to ${AVAGO_DOWNLOAD_PATH}"
+    curl -L ${AVAGO_DOWNLOAD_URL} -o ${AVAGO_DOWNLOAD_PATH}
+
+    extract_archive
+  else
+    # else the version is a git commitish (or it's invvalid)
```
aaronbuchwald, line 62:
```suggestion
    # else the version is a git commitish (or it's invalid)
```

scripts/install_avalanchego_release.sh
```diff
@@ -11,41 +11,119 @@ source "$SUBNET_EVM_PATH"/scripts/versions.sh
...
 # download avalanchego
 # https://github.com/ava-labs/avalanchego/releases
 GOARCH=$(go env GOARCH)
 GOOS=$(go env GOOS)
-BASEDIR=${BASE_DIR:-"/tmp/avalanchego-release"}
+BASEDIR=${BASEDIR:-"/tmp/avalanchego-release"}
+
 mkdir -p ${BASEDIR}
+
+VERSION=$AVALANCHEGO_VERSION
```
aaronbuchwald, line 23: Would it be better to use a single variable for this instead of `VERSION` and `AVALANCHEGO_VERSION` ?

scripts/install_avalanchego_release.sh
```diff
@@ -11,41 +11,119 @@ source "$SUBNET_EVM_PATH"/scripts/versions.sh
...
 AVAGO_DOWNLOAD_PATH=${BASEDIR}/avalanchego-linux-${GOARCH}-${VERSION}.tar.gz
+
 if [[ ${GOOS} == "darwin" ]]; then
   AVAGO_DOWNLOAD_URL=https://github.com/ava-labs/avalanchego/releases/download/${VERSION}/avalanchego-macos-${VERSION}.zip
   AVAGO_DOWNLOAD_PATH=${BASEDIR}/avalanchego-macos-${VERSION}.zip
 fi
 
-AVALANCHEGO_BUILD_PATH=${AVALANCHEGO_BUILD_PATH:-${BASEDIR}/avalanchego-${VERSION}}
-mkdir -p $AVALANCHEGO_BUILD_PATH
+BUILD_DIR=${AVALANCHEGO_BUILD_PATH}-${AVALANCHEGO_VERSION}
```
aaronbuchwald, line 33: This seems to fail locally unless you specify `AVALANCHEGO_BUILD_PATH` (this passes CI since it manually populated this env var)

If we are going to use it in `BUILD_DIR` here, then we should update this to ensure that if it's not populated it defaults to a reasonable location like the previous behavior defaulting to use the already populated `BASEDIR`

scripts/install_avalanchego_release.sh
```diff
@@ -11,41 +11,119 @@ source "$SUBNET_EVM_PATH"/scripts/versions.sh
...
+        echo
+        echo "'${VERSION}' is not a valid release tag, commit hash, or branch name"
+        exit 1
+      fi
+    fi
+
+    COMMIT=$(git rev-parse HEAD)
+
+    # use the commit hash instead of the branch name or tag
+    BUILD_DIR=${AVALANCHEGO_BUILD_PATH}-${COMMIT}
```
aaronbuchwald, line 104: Hmm at first I thought that the below code where it builds iff the build dir does not exist may run into issues for branches if there were newly pushed changes, but I see that this ensures that the build dir is based off of the commit, so that's not an issue.

scripts/versions.sh
richardpringle: Yeah, I was a little worried about `AVALANCHE_VERSION` being used in other scripts though. My change makes this backwards compatible, so you can change the other scripts one at a time and make sure you don't break anything. I think that would be be better as a follow up PR so the context is a little more clear. 

So how about I do that immediately after this is merged?

scripts/install_avalanchego_release.sh
richardpringle: 😎  (why I can't do this as a reaction is beyond me)

scripts/install_avalanchego_release.sh
richardpringle: Yeah, I didn't really mean to move this. I kept `VERSION` because it makes the next couple of lines shorter. Happy to change it though

scripts/install_avalanchego_release.sh
richardpringle: Agreed! Will change

scripts/install_avalanchego_release.sh
richardpringle: I'm brilliant, what can I say, 😉

scripts/versions.sh
aaronbuchwald: Sounds good to me

scripts/install_avalanchego_release.sh
```diff
@@ -11,41 +11,118 @@ source "$SUBNET_EVM_PATH"/scripts/versions.sh
...
+    fi
+
+    # check to see if the commitish exists in the repo
+    WORKDIR=$(pwd)
+
+    cd ${GIT_CLONE_PATH}
+
+    git fetch
+
+    echo "checking out ${AVALANCHE_GO_VERSION}"
```
aaronbuchwald, line 78: I think this needs to be `AVALANCHEGO_VERSION` same for the other two places below

aaronbuchwald (approved):
LGTM

scripts/install_avalanchego_release.sh
```diff
@@ -11,41 +11,118 @@ source "$SUBNET_EVM_PATH"/scripts/versions.sh
...
-  mkdir -p ${AVALANCHEGO_BUILD_PATH} && tar xzvf ${AVAGO_DOWNLOAD_PATH} --directory ${AVALANCHEGO_BUILD_PATH} --strip-components 1
-elif [[ ${GOOS} == "darwin" ]]; then
-  unzip ${AVAGO_DOWNLOAD_PATH} -d ${AVALANCHEGO_BUILD_PATH}
-  mv ${AVALANCHEGO_BUILD_PATH}/build/* ${AVALANCHEGO_BUILD_PATH}
-  rm -rf ${AVALANCHEGO_BUILD_PATH}/build/
+extract_archive() {
+  mkdir -p ${BUILD_DIR}
+
+  if [[ ${AVAGO_DOWNLOAD_PATH} == *.tar.gz ]]; then
+    tar xzvf ${AVAGO_DOWNLOAD_PATH} --directory ${BUILD_DIR} --strip-components 1
```
darioush, line 38: should we drop `v`?

scripts/install_avalanchego_release.sh
```diff
@@ -11,41 +11,118 @@ source "$SUBNET_EVM_PATH"/scripts/versions.sh
...
+      echo "cloning ${GIT_CLONE_URL} to ${GIT_CLONE_PATH}"
+      git clone --no-checkout ${GIT_CLONE_URL} ${GIT_CLONE_PATH}
+    fi
+
+    # check to see if the commitish exists in the repo
+    WORKDIR=$(pwd)
+
+    cd ${GIT_CLONE_PATH}
+
+    git fetch
```
darioush, line 76: I still think we shouldn't fetch or clone the entire repository to get a specific branch or commit.

scripts/install_avalanchego_release.sh
richardpringle: I don't think it's worth the change. Every time I make a change, I have to run CI three different times to make sure everything is working as expected (one for tagged version, one for branch, and one for commit). While I know that dropping a "verbose" option won't change any functionality, if I don't run those tests again, the screen shots and linked actions won't actually contain a commit in this branch. 

Long winded way of saying, "if we want to make that change, it should happen in another PR".

scripts/install_avalanchego_release.sh
richardpringle: What do you mean by "fetch or clone the entire repository"? To my knowledge, you can't partially fetch a repository (unless you have git submodules or you're using LFS or something like that). 

From the git docs:
>When no remote is specified, by default the origin remote will be used, unless there’s an upstream branch configured for the current branch.

Regardless, `fetch` will grab everything we need here, then we always checkout a detached-head by prefixing branch names with `origin/<branch_name>` or a commit directly.

darioush (approved):

scripts/install_avalanchego_release.sh
darioush: Yes, however it also fetches other branches that are not needed for this run.
I think this is fine for now, in the longer term I think we should use github actions to [cache](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows) or create [artifacts](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts)

anusha-ctrl (approved):

PR #599, opened 2023-03-29
Empty Signature and Signature Request Handler

minghinmatthewlam:
This PR adds an empty signature response handling when we don't find a message id in the warp backend for a corresponding signature request. 

Replaces use of `hashing.ComputeHash256Array()` with `warp.UnsignedMessage.ID()`


plugin/evm/vm_test.go
```diff
@@ -3181,26 +3181,3 @@ func TestCrossChainMessagestoVM(t *testing.T) {
 	require.NoError(err)
 	require.True(calledSendCrossChainAppResponseFn, "sendCrossChainAppResponseFn was not called")
 }
-
-func TestSignatureRequestsToVM(t *testing.T) {
```
minghinmatthewlam, line 3185: Decided to remove this test as the primary goal of this test previously was in respect to a `NoopSignatureRequestHandler`. Now that the handler is initialized this test won't be helping, and we already check for expected responses in the signature request handler tests.

warp/backend.go
```diff
@@ -65,10 +65,12 @@ func (w *warpBackend) AddMessage(unsignedMessage *avalancheWarp.UnsignedMessage)
 
 	copy(signature[:], sig)
 	w.signatureCache.Put(messageID, signature)
+	log.Debug("Added warp message to backend", "messageID", messageID)
```
aaronbuchwald, line 68:
```suggestion
	log.Debug("Adding warp message to backend", "messageID", messageID)
```

this gets committed in the `versiondb` on Accept, so want to make sure this is not past tense and potentially leading to misleading logs if we ever crash on accept.

plugin/evm/vm_test.go
aaronbuchwald: Can we update this test to check that the VM calls `AppResponse` correctly with a known and unknown `messageID` instead of removing it?

minghinmatthewlam: updated the test for the above.

aaronbuchwald: Can you update this test so that it asserts the expected response is sent by the VM instead of just checking that it can be unmarshalled correctly?

minghinmatthewlam: updated

aaronbuchwald: Nice, this is perfect

aaronbuchwald (approved):

PR #601, opened 2023-04-03
Migrate HardHat tests to DS-tests (Solidity tests)

### LINKED ISSUE #223: Switch from using HardHat to Foundry

richardpringle:
## Why this should be merged
It's the first big step in the direction of solving some of the security vulnerabilities related to using hardhat and other node-modules.

## How this works
Currently, HardHat is still used to run the test, but all the logic is in Solidity

## How this was tested
Example of passing test:
https://github.com/ava-labs/subnet-evm/actions/runs/4600753486/jobs/8127753059?pr=601#step:8:259

## Next Steps

Once all the tests have been migrated over to solidity, we have a couple of options for removing hardhat. We can use `forge` to build all the solidity, then use `ethers` directly to deploy.

EthersJS has the following dependencies
```zsh
root ➜ /com.docker.devenvironments.code/contract-examples (master ✗) $ yarn info ethers dependencies
yarn info v1.22.19
{
  '@adraffy/ens-normalize': '1.9.0',
  '@noble/hashes': '1.1.2',
  '@noble/secp256k1': '1.7.1',
  'aes-js': '4.0.0-beta.3',
  tslib: '2.4.0',
  ws: '8.5.0'
}
```
all of which have no other dependencies
```zsh
root ➜ /com.docker.devenvironments.code/contract-examples (master ✗) $ npm view @adraffy/ens-normalize dependencies
root ➜ /com.docker.devenvironments.code/contract-examples (master ✗) $ npm view @noble/hashes dependencies
root ➜ /com.docker.devenvironments.code/contract-examples (master ✗) $ npm view aes-js dependencies
root ➜ /com.docker.devenvironments.code/contract-examples (master ✗) $ npm view tslib dependencies
root ➜ /com.docker.devenvironments.code/contract-examples (master ✗) $ npm view ws dependencies
root ➜ /com.docker.devenvironments.code/contract-examples (master ✗) $ 
```

We would still be dependent on `NodeJS` and `mocha`. While `mocha` still has quite a few dependencies, it is one of the most used libraries in the `NodeJS` ecosystem and should come up in an `npm audit`...

Speaking of `npm audit`, we should definitely be running that (there are a lot of vulnerabilities in the current dependency tree). 

From there, I think it's worth evaluating whether or not we should remove the JS dependency entirely, and I'm not sure that it is. 

closes #223


contract-examples/test/fee_manager.ts
```diff
@@ -84,12 +86,22 @@ describe("ExampleFeeManager", function () {
...
+  before("setup DS-Test contract", function () {
+    return ethers.getContractFactory("ContractBTest", { signer: owner })
+      .then(factory => factory.deploy())
+      .then(contract => {
+        testContract = contract
+        return contract.deployed().then(() => contract)
+      })
+      .then(contract => contract.setUp())
+      .then(tx => tx.wait())
+  })
```
richardpringle, line 98: This is what the new setup would look like (still using hard-hat)

contract-examples/test/fee_manager.ts
```diff
@@ -84,12 +86,22 @@ describe("ExampleFeeManager", function () {
...
+
   it("should add contract deployer as owner", async function () {
-    const contractOwnerAddr: string = await contract.owner()
-    expect(owner.address).to.be.equal(contractOwnerAddr)
+    await testContract.test_addContractDeployerAsOwner()
```
richardpringle, line 101: New `it` statements look like this.

contract-examples/contracts/ExampleFeeManager.sol
```diff
@@ -84,3 +85,18 @@ contract ExampleFeeManager is AllowList {
...
+    // noop
+  }
+
+  function test_addContractDeployerAsOwner() public {
+    address owner = msg.sender;
+    ExampleFeeManager manager = new ExampleFeeManager();
+
+    assertEq(owner, manager.owner());
+  }
+}
```
richardpringle, line 102: This is what tests should look like now

aaronbuchwald:
Discussed offline, this looks like a good plan to me to:

- Migrate the testing logic to use DSTest
- Deploy/call test contracts from HardHat
- Migrate to deploy/call test contracts from Golang to remove npm/JS

contract-examples/test/fee_manager.ts
```diff
@@ -9,257 +9,89 @@ import {
...
-    let contractRole = await ownerPrecompile.readAllowList(contract.address);
-    expect(contractRole).to.be.equal(ROLES.NONE)
-
-    let enableTx = await ownerPrecompile.setEnabled(contract.address);
-    await enableTx.wait()
-    contractRole = await ownerPrecompile.readAllowList(contract.address);
-    expect(contractRole).to.be.equal(ROLES.ENABLED)
-  });
-
-  it("admin should be able to change fees through contract", async function () {
```
richardpringle, line 117: I deleted this test as it is redundant. The same functionality is covered in `admin should be able to enable change fees` .

(TODO fix the wording in the new test)

contract-examples/test/fee_manager.ts
```diff
@@ -9,257 +9,89 @@ import {
...
+        this.testContract = contract
+        return contract.deployed().then(() => contract)
+      })
+      .then(contract => contract.setUp())
+      .then(tx => Promise.all([feeManagerPromise, tx.wait()]))
+      .then(([feeManager]) => feeManager.setAdmin(this.testContract.address))
+      .then(tx => tx.wait())
   })
 
-  it("admin should be able to enable wagmi fees through contract", async function () {
```
richardpringle, line 143: Enabling wagmi fees or cchain fees calls the same precompile functionality with different values.
Same functionality is covered here: `admin should be able to enable change fees` (new test)

contract-examples/test/fee_manager.ts
```diff
@@ -9,257 +9,89 @@ import {
...
-    res = await contract.getCurrentFeeConfig()
-    expect(res.gasLimit).to.be.equal(WAGMI_FEES.gasLimit)
-    expect(res.minBaseFee).to.be.equal(WAGMI_FEES.minBaseFee)
+  test("contract should not be able to change fee without enabled", "enableWAGMIFeesFailure"); 
 
-    var res = await contract.getFeeConfigLastChangedAt()
-    expect(res).to.be.equal(txRes.blockNumber)
-  })
-
-  it("should let low fee tx to be in mempool", async function () {
```
richardpringle, line 162: same functionality is covered in `should confirm min-fee transaction`

contract-examples/test/fee_manager.ts
```diff
@@ -9,257 +9,89 @@ import {
...
-      maxPriorityFeePerGas: 0
-    });
-    let confirmedTx = await tx.wait()
-    expect(confirmedTx.confirmations).to.be.greaterThanOrEqual(1)
+  test("should confirm min-fee transaction", "minFeeTransaction", {
+    maxFeePerGas: GENESIS_CONFIG.config.feeConfig.minBaseFee,
+    maxPriorityFeePerGas: 0,
   })
 
-  it("should not let low fee tx to be in mempool", async function () {
```
richardpringle, line 178: new test: `should reject a transaction below the minimum`

contract-examples/test/fee_manager.ts
```diff
@@ -9,257 +9,89 @@ import {
...
-      expect(err.toString()).to.include("max fee per gas less than block base fee")
-      return
-    }
-    expect.fail("should have errored")
-  })
+      await this.testContract.lowerMinFeeByOne({ maxFeePerGas }).then(tx => tx.wait());
+    } catch (err) {
+      expect(err.toString()).to.include("max fee per gas less than block base fee");
 
-  it("should be able to get current fee config", async function () {
```
richardpringle, line 204: Test is redundant as functionality is required for all tests that change fees.

contract-examples/test/fee_manager.ts
```diff
@@ -9,257 +9,89 @@ import {
...
-
-    var res = await contract.connect(manager).getCurrentFeeConfig()
-    expect(res.gasLimit).to.be.equal(C_FEES.gasLimit)
+      await this.testContract.lowerMinFeeByOne({ maxFeePerGas: maxFeePerGas + 1 }).then(tx => tx.wait());
 
-    var res = await contract.connect(nonEnabled).getCurrentFeeConfig()
-    expect(res.gasLimit).to.be.equal(C_FEES.gasLimit)
-  });
-
-  it("nonEnabled should not be able to set fee config", async function () {
```
richardpringle, line 222: Same coverage in `contract should not be able to change fee without enabled`

contract-examples/test/fee_manager.ts
```diff
@@ -9,257 +9,89 @@ import {
...
-      })
-    }
-    catch (err) {
-      return
+      return;
     }
-    expect.fail("should have errored")
-  })
-
-  it("manager should be able to change fees through contract", async function () {
```
richardpringle, line 238: another redundant test

contract-examples/test/fee_manager.ts
```diff
@@ -9,257 +9,89 @@ import {
...
-        maxPriorityFeePerGas: 0
-      })
-    await enableTx.wait()
 
-    var res = await contract.connect(manager).getCurrentFeeConfig()
-    expect(res.minBaseFee).to.be.equal(WAGMI_FEES.minBaseFee)
-  })
-
-
-  it("non-enabled should not be able to change fees through contract", async function () {
```
richardpringle, line 251: another redundant test

contract-examples/test/fee_manager.ts
```diff
@@ -9,257 +9,87 @@ import {
...
-  gasLimit: 8_000_000, // gasLimit
-  targetBlockRate: 2, // targetBlockRate
-  minBaseFee: 25_000_000_000, // minBaseFee
-  targetGas: 15_000_000, // targetGas
-  baseFeeChangeDenominator: 36, // baseFeeChangeDenominator
-  minBlockGasCost: 0, // minBlockGasCost
-  maxBlockGasCost: 1_000_000, // maxBlockGasCost
-  blockGasCostStep: 100_000 // blockGasCostStep
-}
+const testFn = (fnName, overrides = {}) => async function () {
```
aaronbuchwald, line 20: Can we add a comment here to explain how this is used?

richardpringle: Yes, I will do that at the end as I may need to add more functionality with the other tests. I'm going to pull it out and add it in a `utils.ts` file somewhere after.

contract-examples/test/tx_allow_list.ts
```diff
@@ -19,13 +20,55 @@ const ROLES = {
   ADMIN: 2
 }
 
+const testFn = (fnName, overrides = {}, debug = false) => async function () {
```
richardpringle, line 23: Ignore the duplication for now, will find the common abstraction before this is ready to merge

contract-examples/test/fee_manager.ts
```diff
@@ -9,257 +9,87 @@ import {
...
+  const tx = await this.testContract["test_" + fnName](overrides)
+  const txReceipt = await tx.wait()
+  const failed = await this.testContract.callStatic.failed()
+
+  if (failed) {
+    console.log('')
+
+    txReceipt
+      .events
```
richardpringle, line 29: Add `console.warn` here if `events` is `undefined`/`null`. If that happens, we need to look into what's going on.

contract-examples/contracts/ExampleDeployerList.sol
```diff
@@ -4,12 +4,148 @@ pragma solidity ^0.8.0;
...
 // ExampleDeployerList shows how ContractDeployerAllowList precompile can be used in a smart contract
 // All methods of [allowList] can be directly called. There are example calls as tasks in hardhat.config.ts file.
 contract ExampleDeployerList is AllowList {
   // Precompiled Allow List Contract Address
-  address constant DEPLOYER_LIST = 0x0200000000000000000000000000000000000000;
-
   constructor() AllowList(DEPLOYER_LIST) {}
+
+  function deployToken() public {
+    new AllowList(DEPLOYER_LIST);
```
richardpringle, line 18: As per offline discussion with @ceyonur, should try using a different contract.

richardpringle: Could be because of `IAllowList` intialization... something weird is going on there.

contract-examples/contracts/ExampleFeeManager.sol
```diff
@@ -84,3 +104,131 @@ contract ExampleFeeManager is AllowList {
...
+    assertNotEq(config.gasLimit, newFeeConfig.gasLimit);
+    // target block rate is the same for wagmi and cchain
+    // assertNotEq(config.targetBlockRate, newFeeConfig.targetBlockRate);
+    assertNotEq(config.minBaseFee, newFeeConfig.minBaseFee);
+    assertNotEq(config.targetGas, newFeeConfig.targetGas);
+    assertNotEq(config.baseFeeChangeDenominator, newFeeConfig.baseFeeChangeDenominator);
+    // min block gas cost is the same for wagmi and cchain
+    // assertNotEq(config.minBlockGasCost, newFeeConfig.minBlockGasCost);
+    assertNotEq(config.maxBlockGasCost, newFeeConfig.maxBlockGasCost);
+    assertNotEq(config.blockGasCostStep, newFeeConfig.blockGasCostStep);
```
darioush, line 181: should we instead use the `WAGMI_*`constants as expected values and assertEq instead?
(same below with `CCHAIN_*` constants for the expected values after updating fees?)

contract-examples/contracts/ExampleFeeManager.sol
```diff
@@ -84,3 +104,131 @@ contract ExampleFeeManager is AllowList {
...
+    assertEq(changedFeeConfig.targetGas, newFeeConfig.targetGas);
+    assertEq(changedFeeConfig.baseFeeChangeDenominator, newFeeConfig.baseFeeChangeDenominator);
+    // min block gas cost is the same for wagmi and cchain
+    // assertEq(changedFeeConfig.minBlockGasCost, newFeeConfig.minBlockGasCost);
+    assertEq(changedFeeConfig.maxBlockGasCost, newFeeConfig.maxBlockGasCost);
+    assertEq(changedFeeConfig.blockGasCostStep, newFeeConfig.blockGasCostStep);
+
+    assertEq(example.getFeeConfigLastChangedAt(), block.number);
+
+    example.enableCustomFees(config);
```
darioush, line 200: is this done as a cleanup (should we add a comment or use a tearDown fn, not sure if that exists?)

darioush (commented):
partial review, could we:
- remove the redundant tests in a separate PR
- fix the addDeployer issue that exists in the current tests?

this way this PR can be reviewed side by side and we can start removing the `it.skip` directives.

contract-examples/contracts/ExampleFeeManager.sol
```diff
@@ -3,52 +3,72 @@ pragma solidity ^0.8.0;
...
+  uint256 baseFeeChangeDenominator;
+  uint256 minBlockGasCost;
+  uint256 maxBlockGasCost;
+  uint256 blockGasCostStep;
+}
+
 // ExampleFeeManager shows how FeeManager precompile can be used in a smart contract
 // All methods of [allowList] can be directly called. There are example calls as tasks in hardhat.config.ts file.
 contract ExampleFeeManager is AllowList {
   // Precompiled Fee Manager Contract Address
```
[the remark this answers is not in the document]
richardpringle: (I don't think this is a `nit`, this comment is confusing since it's meant for the other line)

contract-examples/contracts/ExampleFeeManager.sol
richardpringle: There's no tear-down function, but we could add one. This is literally the only place where I had to do a cleanup, so I didn't want to add a teardown but I think it's a good idea.

contract-examples/test/tx_allow_list.ts
```diff
@@ -1,199 +1,128 @@
...
+  test("should add contract deployer as admin", "contractOwnerIsAdmin")
 
-  it("precompile should see test address has no role", async function () {
-    // test precompile first
-    const allowList = await ethers.getContractAt("IAllowList", TX_ALLOW_LIST_ADDRESS, admin)
-    let role = await allowList.readAllowList(noRole.address)
-    expect(role).to.be.equal(ROLES.NONE)
-  })
+  // this is testing something that is setup in the before hook 
+  // I don't really understand why...
```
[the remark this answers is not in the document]
richardpringle: I don't think adding assertions to the `setUp` functions would make for very clear errors. If for whatever reason the deployer was not setup as an admin properly, most of the tests would fail and it should be clear-ish that something is seriously wrong.

I don't have a super strong opinion here though, if you think I should add error-logging (for assertions) to the `setUp` block, I can do that.

contract-examples/test/tx_allow_list.ts
```diff
@@ -1,199 +1,128 @@
...
+  it.skip("should add contract deployer as admin", async function () {})
 
-  it("contract should report test address has no admin role", async function () {
-    const result = await contract.isAdmin(noRole.address)
-    expect(result).to.be.false
-  })
+  test("precompile should see admin address has admin role", "precompileHasDeployerAsAdmin")
 
+  // this isn't really testing the functionality of the precompile but instead testing if the genesis config 
+  // set things up as expected.
```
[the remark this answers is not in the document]
richardpringle: Agreed. Is there no better place to test this functionality, though? Are we not testing the rest of the genesis stuff somewhere else?

contract-examples/contracts/ExampleFeeManager.sol
richardpringle: Ah, the `assertNotEq `assertions here are actually just making sure that _current config_ doesn't have any of the same values as `newFeeConfig`. Changing the actual fees happens below this, then we verify with `assertEq` that all the fees were changed to exactly what was expected.

contract-examples/test/contract_deployer_allow_list.ts
```diff
@@ -1,191 +1,115 @@
...
-const ALLOWLIST_ADDRESS = "0x0200000000000000000000000000000000000000";
+const ADMIN_ADDRESS: string = "0x8db97C7cEcE249c2b98bDC0226Cc4C2A57BF52FC"
+const DEPLOYER_ALLOWLIST_ADDRESS = "0x0200000000000000000000000000000000000000";
 
-const ROLES = {
-  NONE: 0,
-  DEPLOYER: 1,
-  ADMIN: 2
-};
+const testFn = (fnNameOrNames: string | string[], overrides = {}, debug = false) => async function () {
```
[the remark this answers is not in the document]
darioush: I'd slightly prefer `test_blahBlahBlah` since it avoids building function names with strings. Could go either way so long as the `test` fn has a comment explaining what it does such as the one I mentioned.

contract-examples/test/fee_manager.ts
```diff
@@ -1,265 +1,53 @@
...
-    }
-    catch (err) {
-      return
-    }
-    expect.fail("should have errored")
-  })
+  test("should reject a transaction below the minimum", [
+    "test_raiseMinFeeByOne",
+    {
+      method: "test_minFeeTransaction",
```
richardpringle, line 48: I should be specifying the fee here

contract-examples/test/utils.ts
```diff
@@ -0,0 +1,65 @@
+import { ethers } from "hardhat"
+const assert = require("assert")
+
+type MethodObject = { method: string, debug?: boolean, overrides?: any, shouldFail?: boolean }
+type FnNameOrObject = string | string[] | MethodObject | MethodObject[]
+type MethodWithDebugAndOverrides = MethodObject & { debug: boolean, overrides: any, shouldFail: boolean }
```
richardpringle, line 6: TODO: Leave comments so someone else can also maintain

contract-examples/contracts/ExampleRewardManager.sol
```diff
@@ -31,3 +34,107 @@ contract ExampleRewardManager is Ownable {
...
+
+    rewardManager.setEnabled(exampleAddress);
+    example.setRewardAddress(exampleAddress);
+
+    exampleReceiveFees = example;
+    exampleBalance = exampleAddress.balance;
+  }
+
+  function test_receiveFees() public {
+    // noop
```
ceyonur, line 102: why we have this function?

contract-examples/contracts/ExampleFeeManager.sol
```diff
@@ -84,3 +103,131 @@ contract ExampleFeeManager is AllowList {
...
+    assertEq(changedFeeConfig.maxBlockGasCost, newFeeConfig.maxBlockGasCost);
+    assertEq(changedFeeConfig.blockGasCostStep, newFeeConfig.blockGasCostStep);
+
+    assertEq(example.getFeeConfigLastChangedAt(), block.number);
+
+    example.enableCustomFees(config);
+  }
+
+  function test_minFeeTransaction() public {
+    // noop
```
ceyonur, line 203: what is this function doing?

contract-examples/test/contract_deployer_allow_list.ts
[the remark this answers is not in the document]
richardpringle: I actually prefer `test_blahBlah` as well, it's a little (not a lot) more obvious how it maps to the smart contract call. I changed this already

contract-examples/contracts/ExampleFeeManager.sol
```diff
@@ -84,3 +103,131 @@ contract ExampleFeeManager is AllowList {
...
+    address exampleAddress = address(example);
+
+    IFeeManager manager = IFeeManager(FEE_MANAGER_ADDRESS);
+
+    manager.setEnabled(exampleAddress);
+
+    FeeConfig memory config = example.getCurrentFeeConfig();
+    config.minBaseFee = config.minBaseFee + 1;
+
+    example.enableCustomFees(config);
```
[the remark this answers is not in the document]
ceyonur: if these functions are used for before/after tests can we specify them in the name like test_beforeXXX/test_afterXXX where XXX is the test name

contract-examples/contracts/ExampleRewardManager.sol
```diff
@@ -31,3 +34,107 @@ contract ExampleRewardManager is Ownable {
...
+  }
+
+  function test_checkReceiveFees() public {
+    assertGt(address(exampleReceiveFees).balance, exampleBalance);
+  }
+
+  function test_areFeeRecipientsAllowed() public {
+    ExampleRewardManager example = new ExampleRewardManager();
+    assertTrue(!example.areFeeRecipientsAllowed());
```
ceyonur, line 111: assertFalse?

contract-examples/contracts/ExampleRewardManager.sol
```diff
@@ -31,3 +34,107 @@ contract ExampleRewardManager is Ownable {
...
+    exampleReceiveFees = example;
+    exampleBalance = exampleAddress.balance;
+  }
+
+  function test_receiveFees() public {
+    // noop
+  }
+
+  function test_checkReceiveFees() public {
+    assertGt(address(exampleReceiveFees).balance, exampleBalance);
```
ceyonur, line 106: what is doing?

contract-examples/contracts/ExampleTxAllowList.sol
```diff
@@ -1,14 +1,195 @@
...
+
+  function test_enableThroughContract() public {
+    ExampleTxAllowList example = new ExampleTxAllowList();
+    ExampleTxAllowList other = new ExampleTxAllowList();
+    address exampleAddress = address(example);
+    address otherAddress = address(other);
+
+    IAllowList allowList = IAllowList(TX_ALLOW_LIST);
+
+    assertTrue(!example.isEnabled(exampleAddress));
```
ceyonur, line 101: assertFalse?

richardpringle: https://github.com/dapphub/ds-test/blob/master/src/test.sol

contract-examples/contracts/ExampleRewardManager.sol
```diff
@@ -31,3 +34,107 @@ contract ExampleRewardManager is Ownable {
...
   }
 }
+
+contract ExampleRewardManagerTest is DSTest {
+  uint blackholeBalance;
+  ExampleRewardManager exampleReceiveFees;
+  uint exampleBalance;
+
+  function setUp() public {
+    blackholeBalance = BLACKHOLE_ADDRESS.balance;
```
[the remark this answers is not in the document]
richardpringle: Please be more specific

contract-examples/contracts/ExampleRewardManager.sol
```diff
@@ -31,3 +34,107 @@ contract ExampleRewardManager is Ownable {
...
+  }
+
+  function test_sendFeesToBlackhole() public {
+    blackholeBalance = BLACKHOLE_ADDRESS.balance;
```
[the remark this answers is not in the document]
richardpringle: It sets the `blackholeBalance` variable

contract-examples/contracts/ExampleRewardManager.sol
richardpringle: Some tests require multiple transactions. Adding a noop with a specific function name hints at the purpose of the transaction. 

In order to understand the tests, you need to look at how the functions are being called:
https://github.com/ava-labs/subnet-evm/blob/3eec654c97d73dfc7a498e9ad1e19f5ad0bc1a1b/contract-examples/test/reward_manager.ts#L40

`test_setupReceiveFees` -> `test_receiveFees` -> `test_checkReceiveFees` are all called in conjunction.

contract-examples/contracts/ExampleRewardManager.sol
richardpringle: https://github.com/ava-labs/subnet-evm/pull/601/files#r1178078510

contract-examples/contracts/ExampleFeeManager.sol
richardpringle: https://github.com/ava-labs/subnet-evm/blob/3eec654c97d73dfc7a498e9ad1e19f5ad0bc1a1b/contract-examples/test/fee_manager.ts#L40

Testing that the transactions succeeds

contract-examples/contracts/ExampleFeeManager.sol
```diff
@@ -84,3 +103,131 @@ contract ExampleFeeManager is AllowList {
...
+    address exampleAddress = address(example);
+
+    IFeeManager manager = IFeeManager(FEE_MANAGER_ADDRESS);
+
+    manager.setEnabled(exampleAddress);
+
+    FeeConfig memory config = example.getCurrentFeeConfig();
+    config.minBaseFee = config.minBaseFee - 1;
+
+    example.enableCustomFees(config);
```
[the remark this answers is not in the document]
richardpringle: see here
https://github.com/ava-labs/subnet-evm/pull/601/files#r1178095800

contract-examples/contracts/ExampleFeeManager.sol
```diff
@@ -84,3 +103,131 @@ contract ExampleFeeManager is AllowList {
...
+
+  function test_enableWAGMIFeesFailure() public {
+    ExampleFeeManager example = new ExampleFeeManager();
+
+    IFeeManager manager = IFeeManager(FEE_MANAGER_ADDRESS);
+
+    // TODO: make roles const (role: None = 0)
+    assertEq(manager.readAllowList(address(example)), 0);
+
+    try example.enableWAGMIFees() {
```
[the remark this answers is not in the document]
richardpringle: What is `expectRevert`?

contract-examples/contracts/ExampleDeployerList.sol
```diff
@@ -4,12 +4,145 @@ pragma solidity ^0.8.0;
 import "@openzeppelin/contracts/access/Ownable.sol";
 import "./IAllowList.sol";
 import "./AllowList.sol";
+import "ds-test/src/test.sol";
+
+address constant DEPLOYER_LIST = 0x0200000000000000000000000000000000000000;
+address constant OTHER_ADDRESS = 0x0Fa8EA536Be85F32724D57A37758761B86416123;
```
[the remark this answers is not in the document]
richardpringle: Why not?

contract-examples/contracts/ERC20NativeMinter.sol
```diff
@@ -56,3 +58,147 @@ contract ERC20NativeMinter is ERC20, AllowList {
...
+
+    uint initialTokenBalance = token.balanceOf(testAddress);
+    uint initialNativeBalance = testAddress.balance;
+
+    uint amount = 100;
+
+    token.mintdraw(amount);
+
+    assertEq(token.balanceOf(testAddress), initialTokenBalance - amount);
+    assertEq(testAddress.balance, initialNativeBalance + amount);
```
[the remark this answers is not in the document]
richardpringle: Please give me a specific example of the level of detail you think is required to understand a test like this. What specifically don't you understand?

contract-examples/contracts/ERC20NativeMinter.sol
```diff
@@ -56,3 +58,147 @@ contract ERC20NativeMinter is ERC20, AllowList {
...
+    address tokenAddress = address(token);
+
+    INativeMinter nativeMinter = INativeMinter(MINTER_ADDRESS);
+
+    assertEq(nativeMinter.readAllowList(tokenAddress), 0);
+
+    try token.mintdraw(100) {
+      assertTrue(false, "mintdraw should fail");
+    } catch {} // TODO should match on an error to make sure that this is failing in the way that's expected
+  }
```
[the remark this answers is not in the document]
richardpringle: If you have a specific suggestion on the best way to do so, I'm open to suggestions

contract-examples/contracts/ERC20NativeMinter.sol
```diff
@@ -56,3 +58,147 @@ contract ERC20NativeMinter is ERC20, AllowList {
...
+
+  function deposit(uint value) external {
+    token.deposit{value: value}();
+  }
+}
+
+contract ERC20NativeMinterTest is DSTest {
+  function setUp() public {
+    // noop
+  }
```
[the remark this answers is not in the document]
richardpringle: We actually don't need any of the example-contracts for testing anymore. The whole structure of `contract-examples` should change. I would argue that `contract-examples` shouldn't necessarily be in this repository anymore, but there are definitely valid arguments against that. 

For now, the changes are a lot simpler if we keep the tests in here before separating out code that's purpose is documentation vs code that's purpose is testing.

contract-examples/contracts/ExampleDeployerList.sol
richardpringle: created an issue here:
https://github.com/ava-labs/subnet-evm/issues/630

richardpringle: Opened #630

contract-examples/test/fee_manager.ts
richardpringle: Added below

contract-examples/test/utils.ts
richardpringle: Done in #628

contract-examples/contracts/ExampleFeeManager.sol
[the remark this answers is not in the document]
ceyonur: https://book.getfoundry.sh/cheatcodes/expect-revert
But I don't think we can use it due to `hevm` requirement.

contract-examples/contracts/ExampleFeeManager.sol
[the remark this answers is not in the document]
ceyonur: We don't need to enforce these names but just use them as the convention. IMO this would make it much easier to read and not mix them with actual tests.

contract-examples/contracts/ExampleDeployerList.sol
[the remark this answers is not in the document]
ceyonur: These contracts were meant to be deployable/example contracts for a simple understanding how precompiles can be called from contracts. I think they become a bit hard to follow in this form with tests. We put these examples to provide a simple example for those developing with precompiles. and also add another layer of testing to precompiles. Adding any test logic to actual contracts might be confusing and could lead others to include those in their own contracts.

contract-examples/contracts/ERC20NativeMinter.sol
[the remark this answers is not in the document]
ceyonur: Answered: https://github.com/ava-labs/subnet-evm/pull/601#discussion_r1180405929

contract-examples/contracts/ERC20NativeMinter.sol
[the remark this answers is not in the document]
ceyonur: oh I was actually referring to `TODO`. I thought we already had a way to test them.

contract-examples/contracts/ERC20NativeMinter.sol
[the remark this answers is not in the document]
ceyonur: We can have some inlined comments like here `ERC20 token balance should be decreased` etc. Also comments above setup/teardown functions can be helpful to understand which functions are testing and which are doing auxiliary work.

contract-examples/contracts/ExampleFeeManager.sol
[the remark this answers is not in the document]
ceyonur: shouldn't we assert that (and on above) that `getCurrentFeeConfig().minBaseFee ==  config.minBaseFee - 1`?

Also IMO `test_` prefix sounds a bit weird when we don't actually test anything :/

contract-examples/contracts/ExampleRewardManager.sol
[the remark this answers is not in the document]
ceyonur: can we rename this to something like `captureBlackholeBalance` etc? in this form it sounds like misses the actual sending fees part. 

And without a before/setup reference to the actual test, this should assert that the current rewarding address is blackhole address.

contract-examples/contracts/ExampleRewardManager.sol
```diff
@@ -31,3 +34,107 @@ contract ExampleRewardManager is Ownable {
...
+  }
+
+  function test_checkSendFeesToBlackhole() public {
+    assertGt(BLACKHOLE_ADDRESS.balance, blackholeBalance);
```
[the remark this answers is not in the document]
ceyonur: Can we put some comments above (and similarly to others) this depends on another function?

contract-examples/contracts/ExampleRewardManager.sol
```diff
@@ -31,3 +34,107 @@ contract ExampleRewardManager is Ownable {
...
+    ExampleRewardManager example = new ExampleRewardManager();
+    address exampleAddress = address(example);
+
+    IRewardManager rewardManager = IRewardManager(REWARD_MANAGER_ADDRESS);
+
+    rewardManager.setEnabled(exampleAddress);
+    example.setRewardAddress(exampleAddress);
+
+    exampleReceiveFees = example;
+    exampleBalance = exampleAddress.balance;
```
ceyonur, line 98: we should probably assert getRewardAddress == exampleRewardAddress

contract-examples/contracts/ExampleRewardManager.sol
ceyonur: I see. Can we add some comments why this function is empty?

contract-examples/contracts/ExampleFeeManager.sol
ceyonur: Can we add comments here that explains why it is empty?

richardpringle: Then I have to add the same comment to every empty function and I suspect there will be many of these. I really don't want to optimize for being able to view this file without also looking at the accompanying `.ts` file.

contract-examples/contracts/ExampleDeployerList.sol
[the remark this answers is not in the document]
richardpringle: Makes sense. It seems like it's also a little confusing that we have both example contracts and tests in the same place (all under `example-contracts`). Does it make sense to separate the two? It seems like based on original intent, `contract-examples` should be separated to its own repo, especially with potential vulnerabilities in npm modules (one of the original motivations for moving away from HardHat). 

@ceyonur, I think you already had some context here, but could use @aaronbuchwald's opinion as well (regarding splitting out the examples into its own repo). 

I believe we can trigger a set of tests in a new repo on every merge to `master` on this repo too. 

We would want to be sure that examples work (nothing as frustrating as examples that don't), but that should never block code from getting merged into this repo. We would then have a nice clean separation between what is an example and what is an e2e test.

contract-examples/contracts/ExampleFeeManager.sol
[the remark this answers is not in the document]
richardpringle: But you can't "not mix them with the actual tests,"... though that would be nice. 

A `test_*` function itself does not constitute a test, but it should be run in a testing context. The `test_` prefix is enforced at the call site though there's nothing preventing a `test_` function from being called outside of a test. 

The issue with adding `test_beforeMyTest` and `test_afterMyTest` is that it implies `test_myTest` is a test, which actually isn't the case. You *must* read the TypeScript code in order to fully understand what is being tested, and when you do, the order in which the `test_*` methods are invoked is clear. *Not fully understanding* what's being tested in `*.sol` files is essentially a feature. 

While this is far from ideal (if only we could plug `subnet-evm` directly into `Forge`), it's actually no worse than what was previously here. While understanding a test *still* requires context in both Solidity and TypeScript, the vast majority of that context is now in Solidity, which IMO is an improvement.

contract-examples/contracts/ExampleFeeManager.sol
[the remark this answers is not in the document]
richardpringle: See [my comment here](https://github.com/ava-labs/subnet-evm/pull/601#discussion_r1180778950). `test_` implies a test context, not that we are necessarily testing something. 

If there's a prefix that would be more suitable, I'm open to suggestions. 

And I don't think we need to check to see if the fee is lowered. We already have tests that check to ensure we can change the fee properly.  If that's not working for whatever reason, the other test should be failing.

contract-examples/contracts/ERC20NativeMinter.sol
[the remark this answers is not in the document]
richardpringle: but it should be abundantly clear that the token balance should be decreased.
```solidity
assertEq(token.balanceOf(testAddress), initialTokenBalance - amount);
```
If you think this would be a little clearer
```solidity
unit currentTokenBalance = token.balanceOf(testAddress);
assertEq(currentTokenBalance, intialTokenBalance - amount); 
```
I'm happy to make that change. 

Excessive commenting is usually a symptom of unclear code that can likely be improved by actual code changes (better naming or otherwise) rather than adding comments.

contract-examples/contracts/ExampleFeeManager.sol
richardpringle: Maybe... I thought about that. I like having some kind of name when reading the test in TS though.

contract-examples/contracts/ExampleRewardManager.sol
richardpringle: It's redundant. The test will fail if that assertion does not pass. Testing `getRewardAddress` can happen in its own test (can't remember if it does or not). 

When you run the test suite and see both `should be able to change the reward address` AND `should be able to receive fees` failing, it should be more obvious where to look for the issue. If you only see that `should be able to receive fees` is failing, then you have to look at the exact assertion that's failing (which, unfortunately, isn't very clear, we don't have solidity stack-traces).  

If that test is missing, we can add it in another PR. These tests need some cleanup, regardless.

contract-examples/contracts/ExampleDeployerList.sol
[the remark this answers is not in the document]
ceyonur: I agree. If we cannot get rid of NPM (or any insecure dep management) currently we should move this out of the repo. IMO we can even keep the old hardhat tests in this folder along with these new solidity tests. This can still helpful for developers to use precompiles with JS/TS functions.

We can move necessary files for e2e (under tests/e2e/precompile) to this new repo and run those tests with the latest subnet-evm (master?). I think we can even put contract-examples repo as a submodule under subnet-evm repo. I think if we use this as a submodule we can run those e2e solidity tests in PR branches as well(with a script replaces subnet-evm via `go replace github.com/ava-labs/subnet-evm => ../`). This can be run as part of subnet-evm e2e tests and can report pass/failure before merging the PR to master.

contract-examples/contracts/ExampleFeeManager.sol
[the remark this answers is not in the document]
ceyonur: I get these cannot be separated from the test framework, however this is not clearly understandable from this file. As I said on my previous comments, these contracts also meant to be a documentation/tutorial for precompiles; so these should be clear/clean as possible. I'm fine with either:
1- adding comments that refers to the actual test in TS file and for full context reader should be looking at there as well.  
2- Or adding comments to functions/renaming function names in .sol file, explaining them in the .sol file.

I still strongly think we should move those test parts to another file under `contracts/tests/ExampleFeeManagerTest.sol` etc.

contract-examples/contracts/ExampleFeeManager.sol
darioush: > Then I have to add the same comment to every empty function and I suspect there will be many of these. I really don't want to optimize for being able to view this file without also looking at the accompanying .ts file.

I don't think this is a problem. Sounds very useful to anyone reading the code without context. eg,
```
// this function is empty as the test only requires the transaction to succeed. 
```

contract-examples/contracts/ExampleFeeManager.sol
```diff
@@ -84,3 +103,132 @@ contract ExampleFeeManager is AllowList {
...
+
+    IFeeManager manager = IFeeManager(FEE_MANAGER_ADDRESS);
+
+    // TODO: make this a const (role: ADMIN = 2)
+    assertEq(manager.readAllowList(thisAddress), 2);
+    assertEq(manager.readAllowList(exampleAddress), 0);
+
+    manager.setEnabled(exampleAddress);
+
+    // TODO: make this a const (role: ENABLED = 1)
```
darioush, line 146: Could we do this as part of this PR?

contract-examples/contracts/ExampleFeeManager.sol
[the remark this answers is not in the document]
darioush: > I still strongly think we should move those test parts to another file under contracts/tests/ExampleFeeManagerTest.sol etc.

This seem preferable to me as well.

Should we refer to these functions as `testStep_` or `testTx_` or `step_` instead of `test_`? 

`test_` implies that this function is a full test in most testing frameworks.

contract-examples/contracts/ExampleFeeManager.sol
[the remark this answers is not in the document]
darioush: Should we refer to these functions as `testStep_` or `testTx_` or `step_` instead of `test_`? 

`test_` implies that this function is a full test in most testing frameworks.

contract-examples/contracts/ERC20NativeMinter.sol
```diff
@@ -56,3 +58,147 @@ contract ERC20NativeMinter is ERC20, AllowList {
...
+  constructor(address tokenAddress) {
+    token = ERC20NativeMinter(tokenAddress);
+  }
+
+  function mintdraw(uint amount) external {
+    token.mintdraw(amount);
+  }
+
+  function deposit(uint value) external {
+    token.deposit{value: value}();
```
darioush, line 78: is this line the reason the wrapper is necessary? (could we make that clear)?

contract-examples/contracts/ExampleDeployerList.sol
```diff
@@ -4,12 +4,145 @@ pragma solidity ^0.8.0;
...
-  address constant DEPLOYER_LIST = 0x0200000000000000000000000000000000000000;
-
   constructor() AllowList(DEPLOYER_LIST) {}
+
+  function deployContract() public {
+    new Example();
+  }
+}
+
+contract Example {}
```
darioush, line 23: What is this for? could we add some comments?

richardpringle:
Changes above, rebase below

contract-examples/contracts/ExampleDeployerList.sol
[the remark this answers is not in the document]
richardpringle: What I'm suggesting a copy-paste of `contract-examples` from `master` (before any of my changes, so, excluding `utils.ts`). It should still use `HardHat` as that seems to be a community best practice for integration tests (`Forge` is more of a unit-testing framework since you're stubbing out the EVM implementation for `hevm`) 

Then this repo, we can actually use `Foundry` to compile all the solidity and rely on `ethers` and `mocha` as the only dependencies. `npm` itself isn't inherently flawed from a security standpoint, it's just the dependencies. `ethers` doesn't have sub-dependencies and `mocha` has very few, but they are *very* well vetted. 

@ceyonur, what do you think? I can actually spin up that repo and make sure the tests work before we merge anything here. I think it makes sense to have working examples somewhere before these changes are merged. A little scope creep, but I think it's for the better.

cc @aaronbuchwald

contract-examples/contracts/ExampleFeeManager.sol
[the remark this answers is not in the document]
richardpringle: @ceyonur and @darioush, please see my comment here:
https://github.com/ava-labs/subnet-evm/pull/601#discussion_r1181830561

and feel free to continue the discussion in that thread. 

If you guys _**don't agree**_ with a separate repo with the examples, then I will move the tests to `contract/tests/*.sol`

contract-examples/contracts/ExampleFeeManager.sol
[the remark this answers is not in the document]
richardpringle: https://github.com/ava-labs/subnet-evm/pull/636

contract-examples/contracts/ExampleRewardManager.sol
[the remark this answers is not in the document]
richardpringle: Done!

contract-examples/contracts/ExampleRewardManager.sol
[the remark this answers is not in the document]
richardpringle: What do you mean by 
>depends on another function

? 

Would renaming `test_` to `step_` clear things up? 
https://github.com/ava-labs/subnet-evm/pull/636

contract-examples/contracts/ExampleRewardManager.sol
richardpringle: done

contract-examples/contracts/ExampleFeeManager.sol
richardpringle: done

contract-examples/contracts/ExampleFeeManager.sol
richardpringle: done

contract-examples/contracts/ERC20NativeMinter.sol
richardpringle: You'll have to be a little more specific about _what_ you want me to clarify. I agree that using this `Minter` contract really muddies the waters regarding clarity. 

To me, it's clear that calling `Minter.deposit(<some_amount>)` is telling an instance of the `Minter` contract to send `<some_amount>` of its ETH to the token it was instantiated with. What's less clear is why we need the indirection. The answer to that is related to the allow-list and testing to make sure that a new address (in this case, the Minter contract) that isn't on the allow-list can receive `<some_amount>` of token. 

Are you more confused yet? haha 

In large part, the reason that this is so confusing is because we are testing the `ERC20NativeMinter` example-contract functionality instead of simply testing the `NativeMinter` precompile directly. This isn't the only place where there exists this indirection between the example-contract and the precompile. 

If we actually isolate contracts for the purpose of testing and contracts that act as examples, the precompile tests will be much easier to understand. The reason I didn't do that right away is because I wanted one PR to migrate as much logic to DS-Test as possible with the exact same coverage. Then another PR (one per precompile would probably be better), to remove the examples entirely where they aren't necessary. The examples should live somewhere else where the only serve as examples, not e2e tests.

With all that said, I propose we leave this as is, and I can start with this one as far as removing examples goes. I left this in another comment somewhere, but the examples should also be split out before this PR lands.

contract-examples/contracts/ExampleDeployerList.sol
richardpringle: I feel like there's a bit of an impedance mismatch between what *I think* comments should be used for and how we treat them in `subnet-evm` in general. 

The only way you can tell what any `contract`, `struct`, or variables, etc is used for is to look at the call-site. 
Here in this case:
https://github.com/ava-labs/subnet-evm/blob/6bb15f7a6f60bbddf95546574b53a11d22d315da/contract-examples/contracts/ExampleDeployerList.sol#L18-L20

Then there are a couple of different places that `deployContract` is called:
https://github.com/ava-labs/subnet-evm/blob/6bb15f7a6f60bbddf95546574b53a11d22d315da/contract-examples/contracts/ExampleDeployerList.sol#L65-L73
and
https://github.com/ava-labs/subnet-evm/blob/6bb15f7a6f60bbddf95546574b53a11d22d315da/contract-examples/contracts/ExampleDeployerList.sol#L107-L125

The first call to `deployContract` tests whether or not `tx.origin` (the signer) can deploy a contract. This seems relatively clear to me, but I wrote it, so maybe it's not... but then I would think that you would ask me to clarify `test_noRoleCannotDeploy` instead of what `contarct Example {}` is. 

The second is making sure that we can deploy a contract through a different contract call.


So, just to be clear, I didn't add a big long comment here to try and make the case that the tests are crystal clear and couldn't use comments or better code to make them clearer, I just don't see how added a comment at this line helps clarify anything. If today I add a comment here explaining what `Example` is used for, then that changes, I likely wouldn't remember to go update the comment, and it would be misleading; it's a "nonlocal" comment.

contract-examples/contracts/ExampleDeployerList.sol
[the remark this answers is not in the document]
anusha-ctrl: Also agree the contracts functions should be simple and just call the precompile. People fork subnet-evm all the time and stateful precompiles need to be as simple as possible.  Also SGTM on moving the example contracts to another repo just for us to hardcore test.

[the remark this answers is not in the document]
darioush: As long as we can keep running e2e tests through github actions on PRs to this repository, I don't really care where they are placed (maybe git submodules).

contract-examples/contracts/ExampleDeployerList.sol
darioush: In general, non-obvious code and top level definitions should be accompanied by comments.

This is so the reviewers and maintainers can understand the code without seeking its callsites. It also helps confirm their understanding of the code after they read it, or allows them to skip over it and come back later if what they are looking for is not there.

You are correct in that this leads to some duplication and possibility of being outdated, however these comments are part of the code and we try to maintain their accuracy at a similar level to code.

For example, below we could have
```
// ExampleDeployerListTest defines transactions that are used to test the DeployerAllowList precompile by instantiating and calling the ExampleDeployerList and making assertions. The transactions are put together as steps of a complete test in contract_deployer_allow_list.ts.
```

contract-examples/contracts/ExampleDeployerList.sol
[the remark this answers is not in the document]
ceyonur: SGTM ✅ 

I think we can also copy e2e tests to that new example repository so we can run those hardhat tests in that repository without manually running the subnet-evm network elsewhere. 

Optionally we can add that repository as a submodule to subnet-evm repo so it would be easier to maintain the contract-examples repo when we add a new precompile.

contract-examples/test/contract_deployer_allow_list.ts
```diff
@@ -1,117 +1,50 @@
...
+    const signer = await ethers.getSigner(ADMIN_ADDRESS)
+    const allowListPromise = ethers.getContractAt("IAllowList", DEPLOYER_ALLOWLIST_ADDRESS, signer)
 
-  it("precompile should see owner address has admin role", async function () {
-    // test precompile first
-    const allowList = await ethers.getContractAt("IAllowList", ALLOWLIST_ADDRESS, owner);
-    let adminRole = await allowList.readAllowList(owner.address);
-    expect(adminRole).to.be.equal(ROLES.ADMIN)
-  });
+    return ethers.getContractFactory("ExampleDeployerListTest", { signer })
```
darioush, line 16: what happens to the return value of `beforeEach`? (is the result of  tx.wait() checked or used somewhere?)

contract-examples/contracts/ExampleTxAllowList.sol
```diff
@@ -1,14 +1,195 @@
 //SPDX-License-Identifier: MIT
-pragma solidity ^0.8.0;
+pragma solidity ^0.8.5;
```
darioush, line 2: why is this changed? could we either stay on 0.8.0 or upgrade to 0.8.5 version for all tests?

contract-examples/contracts/ExampleDeployerList.sol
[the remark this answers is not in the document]
ceyonur: From: https://github.com/ava-labs/subnet-evm/pull/601#discussion_r1181834836

I think these are 2 different issues. I still strongly think we should separate out test logic from actual contracts (regardless of moving hardhat tests into new repo). I don't think this is related to Foundry/Hardhat choice directly but to keeping things minimal and readable in general. 

Do we have any concern about using a separate test file for contracts?

contract-examples/contracts/ExampleTxAllowList.sol
richardpringle: Good catch, will go back to 0.8.0, 0.8.5 produces warnings and I have since removed the 0.8.5 feature I was using. We might need to upgrade to better deal with errors in the future (when our precompiles return better errors), but we don't need to do that now.

contract-examples/test/contract_deployer_allow_list.ts
richardpringle: Nothing happens to the return value, you need to either return or await the promise. `return` is more robust as you can convert the function to a regular function (instead of `async`) and it will behave the same way.

contract-examples/contracts/ExampleDeployerList.sol
richardpringle: I don't want to keep going back and forth so I'll do it, but this violates "clean-code" principles.
https://medium.com/codex/clean-code-comments-833e11a706dc#e45c for reference. 

Using comments to understand code could lead to a false understanding if anything at the call site changes.

contract-examples/test/contract_deployer_allow_list.ts
darioush: Should we await so we can assert the tx's receipt status?

contract-examples/contracts/ExampleDeployerList.sol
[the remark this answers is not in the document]
ceyonur: Hmm if I did not understand this wrong we're about to remove those examples from this repository and turn them into test contracts that will contain no extra logic (like erc20/c-chain&wagmi fees etc)? Previously I was under impression that we will keep those examples in both repositories.

The plan sounds good, we can simplify those contracts and turn them into raw test contracts that will test only precompile logic. However I don't still think it's an utter necessity to keep them together until that point; especially if we want to merge this one to master without waiting the other.

For some clarification: In a perfect world we should not need those solidity tests for precompile testing. We already have unit tests in their related packages, and integration tests under `plugin/evm/vm_test.go`. However our stateful precompiles are a bit different than EVM precompiles. We wanted to make sure we're not breaking any invariant there and put up some wrapper contracts which also can be used as examples. Instead of testing the precompile logic directly (which should be tested by go tests), we tested the example contracts and related precompiles through them. In the process we might have added some extra assertions for precompiles directly to make sure our test paths are correctly executed.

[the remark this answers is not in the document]
richardpringle: >However I don't still think it's an utter necessity to keep them together until that point

It's more work to move them to separate files, import all the examples, then when we go to delete the examples, we would want to move all of the test contracts up a directory because `contracts` wouldn't contain anything besides `tests`. I guess it would also contain the interfaces, but I think it's fine for those to live together the same way we didn't have a separate directory for examples.

>For some clarification:...

Yep, makes sense. but now that we're writing our tests in Solidity, we are already testing calling the precompiles from a contract.

contract-examples/test/contract_deployer_allow_list.ts
richardpringle: I don't understand, the next beforeEach block or test function won't be triggered until that final `tx.wait()` resolves below.

contract-examples/contracts/ERC20NativeMinter.sol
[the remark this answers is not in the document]
richardpringle: Unfortunately, we don't have a better way to do this (in sol or TS).

contract-examples/contracts/ExampleFeeManager.sol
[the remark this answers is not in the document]
richardpringle: Update: moved everything to `contract/test/*.sol`

contract-examples/contracts/test/AllowList.sol
ceyonur, line 1: can we add `Test` suffix to test files? I.e `AllowListTest.sol`

ceyonur (commented):
looks good 👍 .
just a file rename comment.

contract-examples/contracts/test/ExampleDeployerListTest.sol
```diff
@@ -0,0 +1,136 @@
...
+  ExampleDeployerList private example;
+
+  function setUp() public {
+    example = new ExampleDeployerList();
+    IAllowList allowList = IAllowList(DEPLOYER_LIST);
+    allowList.setNone(OTHER_ADDRESS);
+  }
+
+  function step_verifySenderIsAdmin() public {
+    IAllowList allowList = IAllowList(DEPLOYER_LIST);
```
ceyonur, line 24: similar pattern here with: https://github.com/ava-labs/subnet-evm/pull/601#discussion_r1186145582

contract-examples/contracts/test/ExampleFeeManagerTest.sol
```diff
@@ -0,0 +1,135 @@
...
+
+  function step_addContractDeployerAsOwner() public {
+    ExampleFeeManager manager = new ExampleFeeManager();
+    assertEq(address(this), manager.owner());
+  }
+
+  function step_enableWAGMIFeesFailure() public {
+    ExampleFeeManager example = new ExampleFeeManager();
+
+    IFeeManager manager = IFeeManager(FEE_MANAGER_ADDRESS);
```
ceyonur, line 26: similar pattern here with: https://github.com/ava-labs/subnet-evm/pull/601#discussion_r1186145582

contract-examples/contracts/test/ExampleRewardManagerTest.sol
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  function step_checkSendFeesToBlackhole() public {
+    assertGt(BLACKHOLE_ADDRESS.balance, blackholeBalance);
+  }
+
+  function step_doesNotSetRewardAddressBeforeEnabled() public {
+    ExampleRewardManager example = new ExampleRewardManager();
+    address exampleAddress = address(example);
+    IRewardManager rewardManager = IRewardManager(REWARD_MANAGER_ADDRESS);
```
ceyonur, line 28: similar pattern here with: https://github.com/ava-labs/subnet-evm/pull/601#discussion_r1186145582

contract-examples/contracts/test/ExampleTxAllowListTest.sol
```diff
@@ -0,0 +1,180 @@
...
+    allowList.setNone(OTHER_ADDRESS);
+  }
+
+  function step_contractOwnerIsAdmin() public {
+    ExampleTxAllowList example = new ExampleTxAllowList();
+    assertTrue(example.isAdmin(address(this)));
+  }
+
+  function step_precompileHasDeployerAsAdmin() public {
+    IAllowList allowList = IAllowList(TX_ALLOW_LIST);
```
ceyonur, line 21: similar pattern here with: https://github.com/ava-labs/subnet-evm/pull/601#discussion_r1186145582

ceyonur (commented):
some similar patterns in other tests
https://github.com/ava-labs/subnet-evm/pull/601#discussion_r1186145582

richardpringle:
@ceyonur 

> some similar patterns in other tests [#601 (comment)](https://github.com/ava-labs/subnet-evm/pull/601#discussion_r1186145582)

Applying the pattern right now, but consider this scenario:

You have a failing test, and the `step_` function is at the bottom of the contract. I have to scroll up to the top to see what `rewardManager` is (for example). I find that in tests, it's actually preferable and more flexible not to follow DRY principles and instead of _overly_ explicit where possible. Tests are not part of a greater system, and each test should be able to pass on its own; therefore, repetition is okay. Just wanted to share my opinion.

ceyonur:
> @ceyonur
> 
> > some similar patterns in other tests [#601 (comment)](https://github.com/ava-labs/subnet-evm/pull/601#discussion_r1186145582)
> 
> Applying the pattern right now, but consider this scenario:
> 
> You have a failing test, and the `step_` function is at the bottom of the contract. I have to scroll up to the top to see what `rewardManager` is (for example). I find that in tests, it's actually preferable and more flexible not to follow DRY principles and instead of _overly_ explicit where possible. Tests are not part of a greater system, and each test should be able to pass on its own; therefore, repetition is okay. Just wanted to share my opinion.

I don't have a strong opinion on this one. I agree that tests should be explicit as possible. I just feel like calling the precompile should be done once in each contract/test. In both cases we should be consistent amongst test files, so just wanted to make sure we follow same approach in every test file. Thanks for working on this one.

contract-examples/contracts/ExampleTxAllowList.sol
```diff
@@ -1,14 +1,21 @@
 //SPDX-License-Identifier: MIT
 pragma solidity ^0.8.0;
 
-import "@openzeppelin/contracts/access/Ownable.sol";
 import "./AllowList.sol";
+import "./IAllowList.sol";
```
aaronbuchwald, line 5:
```suggestion
```

looks like a double import

contract-examples/contracts/test/ERC20NativeMinterTest.sol
```diff
@@ -0,0 +1,145 @@
...
+
+    Minter minter = new Minter(tokenAddress);
+    address minterAddress = address(minter);
+
+    nativeMinter.setEnabled(tokenAddress);
+
+    uint initialTokenBalance = token.balanceOf(minterAddress);
+    uint initialNativeBalance = minterAddress.balance;
+
+    assertRole(initialTokenBalance, AllowList.Role.None);
```
darioush, line 89: should this assertion be checking the Role against expected role and initial balance with 0?

contract-examples/contracts/test/ERC20NativeMinterTest.sol
```diff
@@ -0,0 +1,145 @@
...
+    assertRole(initialNativeBalance, AllowList.Role.None);
+
+    token.mint(minterAddress, amount);
+
+    uint initialTokenBalance = token.balanceOf(minterAddress);
+    assertEq(initialTokenBalance, amount);
+
+    minter.mintdraw(amount);
+
+    assertRole(token.balanceOf(minterAddress), AllowList.Role.None);
```
darioush, line 142: same as above

contract-examples/contracts/test/ERC20NativeMinterTest.sol
```diff
@@ -0,0 +1,145 @@
...
+
+    Minter minter = new Minter(tokenAddress);
+    address minterAddress = address(minter);
+
+    nativeMinter.setEnabled(tokenAddress);
+
+    uint amount = 100;
+
+    uint initialNativeBalance = minterAddress.balance;
+    assertRole(initialNativeBalance, AllowList.Role.None);
```
darioush, line 133: same as above

contract-examples/test/contract_native_minter.ts
```diff
@@ -1,141 +1,38 @@
...
-    expect(balance).to.be.equal(initBalance.sub(mintValue))
+  test("admin should mintdraw", "step_adminMintdraw")
 
-    let nativeBalance = await ethers.provider.getBalance(owner.address)
-    let gasUsed: BigNumber = txRec.cumulativeGasUsed
-    let gasPrice: BigNumber = txRec.effectiveGasPrice
-    let txFee = gasUsed.mul(gasPrice)
-    expect(nativeBalance).to.be.equal(initNativeBalance.add(mintValue).sub(txFee))
-  })
+  test("minter should not mintdraw ", "step_minterMintdrawFailure")
```
darioush, line 33:
```suggestion
  test("minter should not mintdraw", "step_minterMintdrawFailure")
```

contract-examples/contracts/test/ERC20NativeMinterTest.sol
richardpringle: wow... good catch

contract-examples/contracts/test/ExampleTxAllowListTest.sol
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  function step_noRoleCannotEnableItself() public {
+    ExampleTxAllowList example = new ExampleTxAllowList();
+
+    assertRole(allowList.readAllowList(address(example)), AllowList.Role.None);
+
+    try example.setEnabled(address(example)) {
+      assertTrue(false, "setEnabled should fail");
+    } catch {} // TODO should match on an error to make sure that this is failing in the way that's expected
```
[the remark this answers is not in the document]
richardpringle: >low-level call operation

Code example? Maybe put that code-example in a new issue and assign it to me, then I can follow up in another PR?

contract-examples/test/contract_deployer_allow_list.ts
```diff
@@ -1,117 +1,50 @@
...
-    deployer = signers.slice(-1)[0]
-
-    // Fund deployer address
-    await owner.sendTransaction({
-      to: deployer.address,
-      value: ethers.utils.parseEther("1")
-    })
-
-  });
+  beforeEach('Setup DS-Test contract', async function () {
```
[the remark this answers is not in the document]
richardpringle: I tried to make it so the tests do not need to be run in order. Before this PR, they all needed to be run in a specific order, and most could not be run in isolation, which is a bad testing practice. I believe I have run every single one of these tests in isolation, the only thing I did not test is if you could run them out of order, but I think that's more effort to verify than it's worth.

Also, DS-Test will set its own state to `failed` once it fails a single test. You should still be able to see what other tests pass/fail when a single test fails. Tests in general should be independent.

contract-examples/contracts/test/ERC20NativeMinterTest.sol
richardpringle: Actually, I meant to be checking the balance. Changing this to `assertRole` was the result of a search and replace whose search params were a little too wide, apparently.

contract-examples/contracts/test/ERC20NativeMinterTest.sol
```diff
@@ -0,0 +1,144 @@
+//SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "../ERC20NativeMinter.sol";
+import "../INativeMinter.sol";
+import "./AllowListTest.sol";
+
+// TODO:
```
aaronbuchwald, line 8: +1 for stripping this down to only the testing logic

aaronbuchwald:
It looks like this PR increases the time that the e2e tests take from about 5min to 11min in CI and about 200s to 600s running locally on my computer.

This seems to come from the fact that the tests are completing more total transactions now that it's re-deploying the contract and calling setup for each test.

contract-examples/contracts/test/ExampleTxAllowListTest.sol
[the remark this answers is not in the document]
aaronbuchwald: Created another issue here: https://github.com/ava-labs/subnet-evm/issues/655

aaronbuchwald:
Modified tests to confirm they fail as expected and got this great ouput:

```
Combined output:

Compiling 1 file with 0.8.0
Compilation finished successfully


  ExampleDeployerList
    ✓ precompile should see owner address has admin role (4056ms)

Error: a == b not satisfied [uint]
      Left
BigNumber { value: "1" }
     Right
BigNumber { value: "0" }

    1) precompile should see test address has no role
    ✓ contract should report test address has no admin role (4051ms)
    ✓ contract should report owner address has admin role (4047ms)
    ✓ should not let test address deploy (4084ms)

Error: Assertion Failed

    2) should allow admin to add contract as admin
    ✓ should allow admin to add deployer address as deployer through contract (4049ms)
    ✓ should let deployer address to deploy (4053ms)
    ✓ should let admin revoke deployer (4067ms)


  7 passing (2m)
  2 failing

  1) ExampleDeployerList
       precompile should see test address has no role:

      AssertionError [ERR_ASSERTION]: step_newAddressHasNoRole failed
      + expected - actual

      -false
      +true
      
      at Context.<anonymous> (test/utils.ts:123:9)
      at step (test/utils.ts:33:23)
      at Object.next (test/utils.ts:14:53)
      at fulfilled (test/utils.ts:5:58)
      at processTicksAndRejections (internal/process/task_queues.js:95:5)

  2) ExampleDeployerList
       should allow admin to add contract as admin:

      AssertionError [ERR_ASSERTION]: step_adminAddContractAsAdmin failed
      + expected - actual

      -false
      +true
      
      at Context.<anonymous> (test/utils.ts:123:9)
      at step (test/utils.ts:33:23)
      at Object.next (test/utils.ts:14:53)
      at fulfilled (test/utils.ts:5:58)
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (internal/process/task_queues.js:95:5)





Err: exit status 2
  [FAILED] Expected
      <*exec.ExitError | 0xc001182000>: {
          ProcessState: {
              pid: 54074,
              status: 512,
              rusage: {
                  Utime: {
                      Sec: 7,
                      Usec: 863281,
                      Pad_cgo_0: [0, 0, 0, 0],
                  },
                  Stime: {
                      Sec: 0,
                      Usec: 806771,
                      Pad_cgo_0: [0, 0, 0, 0],
                  },
                  Maxrss: 322871296,
                  Ixrss: 0,
                  Idrss: 0,
                  Isrss: 0,
                  Minflt: 161476,
                  Majflt: 15,
                  Nswap: 0,
                  Inblock: 0,
                  Oublock: 0,
                  Msgsnd: 676,
                  Msgrcv: 675,
                  Nsignals: 2,
                  Nvcsw: 1554,
                  Nivcsw: 13148,
              },
          },
          Stderr: nil,
      }
  to be nil
  In [It] at: /Users/aaronbuchwald/go/src/github.com/ava-labs/subnet-evm/tests/utils/subnet.go:46 @ 05/23/23 16:37:04.524
  < Exit [It] contract deployer allow list - /Users/aaronbuchwald/go/src/github.com/ava-labs/subnet-evm/tests/precompile/solidity/suites.go:43 @ 05/23/23 16:37:04.524 (1m57.465s)
• [FAILED] [117.465 seconds]

```

aaronbuchwald (approved):
LGTM once @darioush approves

darioush (approved):

PR #604, opened 2023-04-04
Bump avalanchego to v1.10.0

aaronbuchwald:
This PR bumps avalanchego dependency to v1.10.0-fuji-rc.2 for the Fuji Cortina Release.

This also updates the Subnet-EVM version to v0.5.0-fuji-rc.0

### DIFF
--- a/compatibility.json
+++ b/compatibility.json
@@ -1,5 +1,6 @@
 {
   "rpcChainVMProtocolVersion": {
+    "v0.5.0-fuji-rc.0": 25,
     "v0.4.12": 24,
     "v0.4.11": 24,
     "v0.4.10": 23,
--- a/compatibility.json
+++ b/compatibility.json
@@ -1,6 +1,6 @@
 {
   "rpcChainVMProtocolVersion": {
-    "v0.5.0-fuji-rc.0": 25,
+    "v0.5.0": 25,
     "v0.4.12": 24,
     "v0.4.11": 24,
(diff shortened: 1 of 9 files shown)


aaronbuchwald:
This PR will be merged with the full AvalancheGo v1.10.0 release instead of the Fuji Cortina release, so marking this as blocked for now.

go.mod
```diff
@@ -4,7 +4,7 @@ go 1.19
 
 require (
 	github.com/VictoriaMetrics/fastcache v1.10.0
-	github.com/ava-labs/avalanchego v1.9.11
+	github.com/ava-labs/avalanchego v1.10.0-fuji-rc.2
```
StephenButtolph, line 7: Why rc.2 rather than the actual fuji release?

aaronbuchwald: I wasn't sure which of the rcs and -fuji were the final one, so checked a slack channel to see which one was used. Since these are the same, going to leave this as is since it only needs to update the `RPCChainVMProtocolVersion`.

darioush (approved):

compatibility.json
```diff
@@ -1,5 +1,6 @@
 {
   "rpcChainVMProtocolVersion": {
+    "v0.5.0": 25,
```
ceyonur, line 3: can we also update README?

ceyonur (approved):

patrick-ogrady (approved):

PR #610, opened 2023-04-07
Bench precompile

aaronbuchwald:
## Why this should be merged

This PR adds a helper to run already existing precompile tests as benchmarks for each individual case.
This also adds a new GH Action to CI to benchmark the precompile contracts (only)

## How this works

This PR adds a `Bench` function to the existing `PrecompileTest` struct, so that we can run the existing tests and add new ones that can be used as both unit tests and benchmarks.

This also switches existing tests that use `*testing.T` to use `testing.TB` to support the new benchmarks.

## How this was tested

Run new benchmarks locally and in CI

## How is this documented

No new documentation necessary

### DIFF
--- a/precompile/allowlist/test_allowlist.go
+++ b/precompile/allowlist/test_allowlist.go
@@ -26,7 +26,7 @@ func AllowListTests(module modules.Module) map[string]testutils.PrecompileTest {
 		"set admin": {
 			Caller:     TestAdminAddr,
 			BeforeHook: SetDefaultRoles(contractAddress),
-			InputFn: func(t *testing.T) []byte {
+			InputFn: func(t testing.TB) []byte {
 				input, err := PackModifyAllowList(TestNoRoleAddr, AdminRole)
 				require.NoError(t, err)
 
@@ -35,7 +35,7 @@ func AllowListTests(module modules.Module) map[string]testutils.PrecompileTest {
 			SuppliedGas: ModifyAllowListGasCost,
 			ReadOnly:    false,
 			ExpectedRes: []byte{},
-			AfterHook: func(t *testing.T, state contract.StateDB) {
+			AfterHook: func(t testing.TB, state contract.StateDB) {
 				res := GetAllowListStatus(state, contractAddress, TestNoRoleAddr)
 				require.Equal(t, AdminRole, res)
 			},
@@ -43,7 +43,7 @@ func AllowListTests(module modules.Module) map[string]testutils.PrecompileTest {
 		"set enabled": {
 			Caller:     TestAdminAddr,
 			BeforeHook: SetDefaultRoles(contractAddress),
-			InputFn: func(t *testing.T) []byte {
+			InputFn: func(t testing.TB) []byte {
 				input, err := PackModifyAllowList(TestNoRoleAddr, EnabledRole)
 				require.NoError(t, err)
 
@@ -52,7 +52,7 @@ func AllowListTests(module modules.Module) map[string]testutils.PrecompileTest {
 			SuppliedGas: ModifyAllowListGasCost,
 			ReadOnly:    false,
 			ExpectedRes: []byte{},
-			AfterHook: func(t *testing.T, state contract.StateDB) {
+			AfterHook: func(t testing.TB, state contract.StateDB) {
 				res := GetAllowListStatus(state, contractAddress, TestNoRoleAddr)
 				require.Equal(t, EnabledRole, res)
 			},
@@ -60,7 +60,7 @@ func AllowListTests(module modules.Module) map[string]testutils.PrecompileTest {
 		"set no role": {
 			Caller:     TestAdminAddr,
 			BeforeHook: SetDefaultRoles(contractAddress),
-			InputFn: func(t *testing.T) []byte {
+			InputFn: func(t testing.TB) []byte {
 				input, err := PackModifyAllowList(TestEnabledAddr, NoRole)
 				require.NoError(t, err)
 
@@ -69,7 +69,7 @@ func AllowListTests(module modules.Module) map[string]testutils.PrecompileTest {
 			SuppliedGas: ModifyAllowListGasCost,
 			ReadOnly:    false,
 			ExpectedRes: []byte{},
-			AfterHook: func(t *testing.T, state contract.StateDB) {
+			AfterHook: func(t testing.TB, state contract.StateDB) {
 				res := GetAllowListStatus(state, contractAddress, TestEnabledAddr)
 				require.Equal(t, NoRole, res)
 			},
@@ -77,7 +77,7 @@ func AllowListTests(module modules.Module) map[string]testutils.PrecompileTest {
 		"set no role from no role": {
 			Caller:     TestNoRoleAddr,
 			BeforeHook: SetDefaultRoles(contractAddress),
-			InputFn: func(t *testing.T) []byte {
+			InputFn: func(t testing.TB) []byte {
 				input, err := PackModifyAllowList(TestEnabledAddr, NoRole)
 				require.NoError(t, err)
 
@@ -90,7 +90,7 @@ func AllowListTests(module modules.Module) map[string]testutils.PrecompileTest {
 		"set enabled from no role": {
 			Caller:     TestNoRoleAddr,
 			BeforeHook: SetDefaultRoles(contractAddress),
-			InputFn: func(t *testing.T) []byte {
+			InputFn: func(t testing.TB) []byte {
 				input, err := PackModifyAllowList(TestNoRoleAddr, EnabledRole)
 				require.NoError(t, err)
 
@@ -103,7 +103,7 @@ func AllowListTests(module modules.Module) map[string]testutils.PrecompileTest {
 		"set admin from no role": {
 			Caller:     TestNoRoleAddr,
 			BeforeHook: SetDefaultRoles(contractAddress),
-			InputFn: func(t *testing.T) []byte {
+			InputFn: func(t testing.TB) []byte {
 				input, err := PackModifyAllowList(TestEnabledAddr, AdminRole)
 				require.NoError(t, err)
 
@@ -116,7 +116,7 @@ func AllowListTests(module modules.Module) map[string]testutils.PrecompileTest {
 		"set no role from enabled": {
 			Caller:     TestEnabledAddr,
 			BeforeHook: SetDefaultRoles(contractAddress),
-			InputFn: func(t *testing.T) []byte {
+			InputFn: func(t testing.TB) []byte {
 				input, err := PackModifyAllowList(TestAdminAddr, NoRole)
 				require.NoError(t, err)
 
@@ -129,7 +129,7 @@ func AllowListTests(module modules.Module) map[string]testutils.PrecompileTest {
 		"set enabled from enabled": {
 			Caller:     TestEnabledAddr,
 			BeforeHook: SetDefaultRoles(contractAddress),
-			InputFn: func(t *testing.T) []byte {
+			InputFn: func(t testing.TB) []byte {
 				input, err := PackModifyAllowList(TestNoRoleAddr, EnabledRole)
 				require.NoError(t, err)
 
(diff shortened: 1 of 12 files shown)


scripts/build_bench_precompiles.sh
```diff
@@ -0,0 +1,19 @@
...
+    cd .. && pwd
+)
+
+# Load the versions
+source "$SUBNET_EVM_PATH"/scripts/versions.sh
+
+# Load the constants
+source "$SUBNET_EVM_PATH"/scripts/constants.sh
+
+go test ./precompile/contracts/... -bench=./... -covermode=atomic -timeout="10m" $@
```
darioush, line 19: does enabling coverage interfere with benchmarking?

darioush (approved):

precompile/contracts/rewardmanager/contract_test.go
```diff
@@ -128,17 +127,19 @@ func TestRewardManagerRun(t *testing.T) {
 			ReadOnly:    false,
 			ExpectedRes: func() []byte {
 				res, err := PackCurrentRewardAddressOutput(testAddr)
-				require.NoError(t, err)
+				if err != nil {
+					panic(err)
+				}
```
anusha-ctrl, line 132: curious why we use `panic` here instead of `t.Fatal`

scripts/build_bench_precompiles.sh
anusha-ctrl, line 19: This is great. Can we get the average numbers for all the precompile benchmarks and fail a test if we are significantly over? Currently, we just report, but maybe in the future we can report and compare?

precompile/testutils/test_precompile.go
```diff
@@ -69,18 +98,76 @@ func (test PrecompileTest) Run(t *testing.T, module modules.Module, state contra
...
+		ReadOnly:        test.ReadOnly,
+	}
+}
+
+func (test PrecompileTest) Bench(b *testing.B, module modules.Module, state contract.StateDB) {
+	runParams := test.setup(b, module, state)
+
+	if runParams.Input == nil {
+		b.Skip("Skipping precompile benchmark due to nil input (used for configuration tests)")
 	}
```
anusha-ctrl, line 117: It seems like we do `errorcheck`, `benchmark`, and sanity check `errorcheck`? Can we combine the errorcheck inside benchmark? and just do one total combo `errorcheck +benchmark`?

anusha-ctrl: https://scene-si.org/2017/06/06/benchmarking-go-programs/

precompile/contracts/rewardmanager/contract_test.go
aaronbuchwald: No `t` is available here anymore and it should never error here for a correctly written test - a cleaner way to do this would be to construct the needed byte array before these tests

precompile/testutils/test_precompile.go
aaronbuchwald: I'm not sure I'm following your suggestion, could you be more specific?

scripts/build_bench_precompiles.sh
aaronbuchwald: ya this would be a good future change to track performance over time

aaronbuchwald: I'll remove it since it will add overhead and doesn't seem necessary

precompile/allowlist/test_allowlist.go
```diff
@@ -270,3 +270,22 @@ func RunPrecompileWithAllowListTests(t *testing.T, module modules.Module, newSta
...
+	tests := AllowListTests(module)
+	// Add the contract specific tests to the map of tests to run.
+	for name, test := range contractTests {
+		if _, exists := tests[name]; exists {
+			b.Fatalf("duplicate bench name: %s", name)
+		}
+		tests[name] = test
+	}
+
+	for name, test := range tests {
```
ceyonur, line 286: seems we're running tests as our benchmark cases. Do you think it would be better if we define benchmark cases in each precompile and let it run them instead of unit tests?

precompile/testutils/test_precompile.go
```diff
@@ -69,18 +98,76 @@ func (test PrecompileTest) Run(t *testing.T, module modules.Module, state contra
...
+	snapshot := stateDB.Snapshot()
+
+	ret, remainingGas, err := module.Contract.Run(runParams.AccessibleState, runParams.Caller, runParams.ContractAddress, runParams.Input, runParams.SuppliedGas, runParams.ReadOnly)
+	if len(test.ExpectedErr) != 0 {
+		require.ErrorContains(b, err, test.ExpectedErr)
+	} else {
+		require.NoError(b, err)
+	}
+	require.Equal(b, uint64(0), remainingGas)
+	require.Equal(b, test.ExpectedRes, ret)
```
ceyonur, line 128: why are we initially running this here?

precompile/testutils/test_precompile.go
```diff
@@ -69,18 +98,76 @@ func (test PrecompileTest) Run(t *testing.T, module modules.Module, state contra
...
+
+	for i := 0; i < b.N; i++ {
+		// Revert to the previous snapshot and take a new snapshot, so we can reset the state after execution
+		stateDB.RevertToSnapshot(snapshot)
+		snapshot = stateDB.Snapshot()
+
+		// Ignore return values for benchmark
+		_, _, _ = module.Contract.Run(runParams.AccessibleState, runParams.Caller, runParams.ContractAddress, runParams.Input, runParams.SuppliedGas, runParams.ReadOnly)
+	}
+	b.StopTimer()
```
ceyonur, line 146: at here, aren't we also including the time for reverting to the snapshot?

precompile/testutils/test_precompile.go
```diff
@@ -69,18 +98,76 @@ func (test PrecompileTest) Run(t *testing.T, module modules.Module, state contra
...
+
+	elapsed := uint64(time.Since(start))
+	if elapsed < 1 {
+		elapsed = 1
+	}
+	gasUsed := runParams.SuppliedGas * uint64(b.N)
+	b.ReportMetric(float64(runParams.SuppliedGas), "gas/op")
```
ceyonur, line 153: shouldn't this be `remainingGas` from each benhmark run? I don't think SuppliedGas param shows a reliable metric.

precompile/testutils/test_precompile.go
```diff
@@ -69,18 +98,76 @@ func (test PrecompileTest) Run(t *testing.T, module modules.Module, state contra
...
+		require.ErrorContains(b, err, test.ExpectedErr)
+	} else {
+		require.NoError(b, err)
+	}
+	require.Equal(b, uint64(0), remainingGas)
+	require.Equal(b, test.ExpectedRes, ret)
+
+	if test.AfterHook != nil {
+		test.AfterHook(b, state)
 	}
```
ceyonur, line 172: hm this sounds like another unit test to me rather than a benchmark

ceyonur (commented):
great work! left some comments/questions.

this probably requires a change in precompile template as well.

precompile/allowlist/test_allowlist.go
aaronbuchwald: The intention is to make each case portable to be run as both a unit test and benchmark

ceyonur: hmm but I don't think some unit tests (those are testing failing cases like sufficient gas, invalid input etc) will be helpful for benchmarks.

precompile/testutils/test_precompile.go
aaronbuchwald: The original iteration of this used `StartTimer` and `StopTimer`, but this resulted in an interesting bug.

The golang benchmarks use the execution time tracked during an iteration of the for loop in order to figure out how many times it should re-run the tests to get an accurate benchmark.

This means that if the time spent in the for loop is extremely small, it will end up running the loop a much larger number of times.

A strange result of this pattern is that benchmarks need to be sure that a majority of the time spent in each iteration of the loop is tracked. Otherwise, you can run into an issue where a single iteration takes 1ns, so the benchmark decides to perform it 1 million times. However, if the setup/teardown in the for loop actually takes 1 second, then it will still execute the benchmark 1 million times and it will execute the setup/teardown code a large number of times as well.

The original iteration of this called `StartTimer` and `StopTimer`, but it turned out that capturing the memory statistics took a much longer amount of time than the execution that was being benchmarked, so the benchmarks actually took minutes to complete.

Here's a more intuitive example snippet:
```go
func BenchmarkSleepBenchmark(b *testing.B) {
	for i := 0; i < b.N; i++ {
		time.Sleep(time.Nanosecond)
		// time.Sleep(time.Millisecond)
		b.StopTimer()
		b.StartTimer()
	}
}
```

precompile/testutils/test_precompile.go
```diff
@@ -69,18 +98,76 @@ func (test PrecompileTest) Run(t *testing.T, module modules.Module, state contra
...
+		// Ignore return values for benchmark
+		_, _, _ = module.Contract.Run(runParams.AccessibleState, runParams.Caller, runParams.ContractAddress, runParams.Input, runParams.SuppliedGas, runParams.ReadOnly)
+	}
+	b.StopTimer()
+
+	elapsed := uint64(time.Since(start))
+	if elapsed < 1 {
+		elapsed = 1
+	}
+	gasUsed := runParams.SuppliedGas * uint64(b.N)
```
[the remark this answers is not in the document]
aaronbuchwald: Yup that's right, this uses the `SuppliedGas` param from the test itself.

These tests assume that they will consume exactly the amount of gas that was supplied or else they will error - including running into an `ErrOutOfGas`. You can see above, it requires that the amount of remaining gas after the execution is 0.

If the test runs into `ErrOutOfGas`, then it will be considered to have consumed all of the gas. Therefore, for these benchmarks we would expect that they have a higher `gas/sec` benchmark because they should consume all of the gas without actually doing the full operation.

precompile/testutils/test_precompile.go
aaronbuchwald: Same as above, these test cases are assumed to supply the exact amount of gas that they need or else they will fail. This means that the total gas executed for these tests over a period of `b.N` runs is `suppliedGas * b.N` assuming that they pass.

precompile/testutils/test_precompile.go
aaronbuchwald: As per the comment, this is intended to ensure that the benchmark is passing and not benchmarking an execution that is silently failing, which we would want to catch.

precompile/allowlist/test_allowlist.go
aaronbuchwald: They should still be valid benchmarks and I actually really like the fact that we can easily define a case that is portable for both a unit test and benchmark.

Any valid test case will still be a valid benchmark, they may just produce a much higher mgas/sec if they are running into an insufficient gas or invalid input since they will consume gas without performing the full operation.

precompile/testutils/test_precompile.go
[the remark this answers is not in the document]
ceyonur: wouldn't this cause problems if gas requirement changes dynamically and the param is just set to a safe value? I'm not sure if this assumption to consume the exact same gas is very reliable (at least compared to the remaningGas returned from `Run`).

precompile/testutils/test_precompile.go
ceyonur: hmm yea weird behaviour, it's probably expected to use them before/after running each iteration (not inside the loop).

precompile/allowlist/test_allowlist.go
ceyonur: I was more thinking of simple success cases for each function (readRole, setRole etc). I.e `setRole` case for assigning roles to a predetermined size of addresses in each iteration (without reverting to original state). E.g `BenchmarkSetRole1000`:
```
        size := 1000
	for i := 0; i < b.N; i++ {
		state := newStateDB()
                for k := 0; k < size; k++{
                    address := randAddress(k)
                    input := packSetRole(address)
		    ..., err := contract.Run(state, caller, contractAddress, input, gas, readonly)
		    req.NoErr(err)
		}
	}
```

precompile/testutils/test_precompile.go
aaronbuchwald: These are actually used fairly regularly inside of the for loop in benchmarks, they just run into issues if the computation being benchmarked is extremely small.

PR #611, opened 2023-04-10
Use shorter ctx for snap reads in state sync

darioush:
## Why this should be merged
Under degraded conditions, it is possible that database reads become slow.
In that scenario, if the snapshot data is outdated (especially the 1st segment of it),
and the entire response time is spent on reading the snapshot, state sync responses will return as empty.

## How this works
Improves network resilience under degraded conditions by reserving a portion of the response
time to correct data from the trie as needed.

## How this was tested
UT

## How is this documented
No updates needed

### DIFF
--- a/sync/handlers/leafs_request.go
+++ b/sync/handlers/leafs_request.go
@@ -31,6 +31,10 @@ const (
 	// in message.LeafsRequest if it is greater than this value
 	maxLeavesLimit = uint16(1024)
 
+	// Maximum percent of the time left to deadline to spend on optimistically
+	// reading the snapshot to find the response
+	maxSnapshotReadTimePercent = 75
+
 	segmentLen = 64                // divide data from snapshot to segments of this size
 	keyLength  = common.HashLength // length of the keys of the trie to sync
 )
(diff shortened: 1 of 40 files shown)


sync/handlers/leafs_request_test.go
```diff
@@ -649,6 +651,51 @@ func TestLeafsRequestHandler_OnLeafsRequest(t *testing.T) {
...
+				}
+			},
+			assertResponseFn: func(t *testing.T, request message.LeafsRequest, response []byte, err error) {
+				require.NoError(t, err)
+				var leafsResponse message.LeafsResponse
+				_, err = message.Codec.Unmarshal(response, &leafsResponse)
+				require.NoError(t, err)
+				require.NotZero(t, mockHandlerStats.SnapshotReadAttemptCount)
+				require.Zero(t, mockHandlerStats.SnapshotReadSuccessCount)
+				require.NotEmpty(t, leafsResponse.Keys)
```
aaronbuchwald, line 696: Can we require that it returns the full 500 keys/values here instead of not empty?

I assume this was done to handle the case that serving each section from the trie times out, but this should pass consistently for the local test with an in-memory database.

sync/handlers/leafs_request_test.go
```diff
@@ -649,6 +651,53 @@ func TestLeafsRequestHandler_OnLeafsRequest(t *testing.T) {
 				assertRangeProofIsValid(t, &request, &leafsResponse, false)
 			},
 		},
+		"reading from snapshot with stale data times out": {
+			prepareTestFn: func() (context.Context, message.LeafsRequest) {
+				// delayedReader allows us to cause the snapshot read context
+				// to timeout, so we can test the trie is read from disk as a fallback.
```
aaronbuchwald, line 657: Can we make this more obviously correct by just blocking the snapshot on `<-ctx.Done()` ?

This will make it a bit more clear that we wait for the `snapCtx` to timeout instead of needing to think about how many delays occur and what the exact time outs are.

darioush: Do you find the new version better?
It is a bit tricky since we can't wait for the entire context to expire and need to know how long we should wait, but this makes the math a bit more explicit

sync/handlers/leafs_request.go
```diff
@@ -220,7 +224,19 @@ func (rb *responseBuilder) fillFromSnapshot(ctx context.Context) (bool, error) {
 	// modified since the requested root. If this assumption can be verified with
 	// range proofs and data from the trie, we can skip iterating the trie as
 	// an optimization.
-	snapKeys, snapVals, err := rb.readLeafsFromSnapshot(ctx)
+	// Since we are performing this read optimistically, we use a separate context
+	// with reduced timeout so there is enough time to read the trie if the snapshot
+	// read does not contain up-to-date data.
+	snapCtx := ctx
```
richardpringle, line 230: Is `snapCtx` used anywhere? It looks like it's only assigned here and once inside the `if` block, but we don't do anything with it.

aaronbuchwald: It's used on line 239

aaronbuchwald (approved):

anusha-ctrl (approved):

sync/handlers/leafs_request.go
richardpringle: 😞 whoops

PR #612, opened 2023-04-10
Enables trie clean cache rejournaling by default

darioush:
## Why this should be merged
Writing the clean cache periodically may reduce the amount of time it takes to recover from an unclean shutdown.
Currently we have a default of 4096 blocks per commit interval which amounts to 1 root each ~2.3hrs
This means on an average unclean shutdown, nodes would have to reprocess (without snapshot) ~1.1hrs worth of blocks.

## How this works
Specifies default values for rejournal and the data directory for caching.

## How this was tested
existing UTs + manual testing w/ DFK subnet.

## How is this documented
TBD

### DIFF
--- a/plugin/evm/config.go
+++ b/plugin/evm/config.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/ava-labs/avalanchego/snow"
 	"github.com/ava-labs/subnet-evm/core"
 	"github.com/ava-labs/subnet-evm/eth"
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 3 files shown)


plugin/evm/config.go
```diff
@@ -219,7 +221,12 @@ func (c Config) EthBackendSettings() eth.Settings {
 	return eth.Settings{MaxBlocksPerRequest: c.MaxBlocksPerRequest}
 }
 
-func (c *Config) SetDefaults() {
+func (c *Config) SetDefaults(ctx *snow.Context) {
+	var trieCleanJournal string
+	if ctx != nil {
+		trieCleanJournal = ctx.ChainDataDir
+	}
```
aaronbuchwald, line 228: Can we add a suffix to this file path, so that we can re-used the `ChainDataDir` for other use cases in the future as well?

We should also change this from taking in `*snow.Context` to just using `ChainDataDir` so that it takes in the minimal amount of information that is needed here.

darioush: updated

aaronbuchwald:
I think we should migrate off of the JSON config and `SetDefaults()` stuff we do now to use the `pflags` package before making this change, so that it is simpler to handle our configs.

PR #614, opened 2023-04-12
Make state sync request size configurable

darioush:
## Why this should be merged
Makes the number of state sync leafs requested configurable. Allows clients to try a smaller number if servers are overloaded.

## How this works
Adds a configuration option

## How this was tested
UT

## How is this documented
TBD.


aaronbuchwald (approved):
LGTM

hexfusion (approved):

dhrubabasu (approved):

PR #619, opened 2023-04-19
Simulator improvements

aaronbuchwald:
## Why this should be merged

This PR is intended to improve the load simulator. The original code included a lot of unnecessary error handling that complicated the logic. This code should simply:

- fund keys
- generate sequences of transactions
- issue sequence of transactions
- await tx acceptance to mark time to finality from perspective of a client
- confirm all of the transactions

## How this works

This PR breaks the load simulator into:

- insecure key management (intended only for testing)
- tx sequence generation
- worker to handle issuance, await acceptance, confirmation via `eth_getTransactionByHash`

## How this was tested

CI and local testing. To verify, create a local network with the avalanche-network-runner and try running the simulator with the following commands to test different scenarios (note: if you run a large amount of volume through the network, then it may increase the base fee, such that previously issued transactions may be underpriced)

## How is this documented

This PR updates `./cmd/simulator/README.md` with up to date documentation including flag options in the command output.

To see the simulator options run (from the root of the repo):

```bash
go run cmd/simulator/main/*.go --help
```

Output:
```
# command-line-arguments
ld: warning: could not create compact unwind for _blst_sha256_block_data_order: does not use RBP or RSP based frame
Usage of simulator:
      --config-file string    Specify the config path to use to load a YAML config for the simulator
      --endpoints strings     Specify a comma separated list of RPC Websocket Endpoints (default [ws://127.0.0.1:9650/ext/bc/C/ws])
      --key-dir string        Specify the directory to save private keys in (INSECURE: only use for testing) (default ".simulator/keys")
      --log-level string      Specify the log level to use in the simulator. (default "info")
      --max-fee-cap int       Specify the maximum fee cap to use for transactions denominated in GWei (default 50)
      --max-tip-cap int       Specify the max tip cap for transactions denominated in GWei (default 1)
      --timeout duration      Specify the timeout for the simulator to complete. (default 5m0s)
      --txs-per-worker uint   Specify the number of transactions to create per worker. (default 100)
      --version               Print the version and exit.
      --workers int           Specify the number of workers to create for the simulator. (default 1)
```


cmd/simulator/load/loader.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/ava-labs/subnet-evm/cmd/simulator/key"
+	"github.com/ava-labs/subnet-evm/ethclient"
+	"github.com/ava-labs/subnet-evm/params"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/log"
+)
+
+func CreateLoader(ctx context.Context, config config.Config) (*WorkerGroup, error) {
+	// Construct the arguments for the load simulator
+	switch {
```
richardpringle, line 23: I don't know if this matters, but wouldn't it make more sense to have:
```go
switch endpointCount := len(config.Endpoints) {
case 0:
  // ...
case endpointCount < config.Workers:
  // ...
}
```
IMO, function calls within the case arms or whatever they're called make things harder to read. I'm reaching here though, it's not that bad.

cmd/simulator/load/loader.go
```diff
@@ -0,0 +1,107 @@
...
+	case len(config.Endpoints) == 0:
+		fmt.Printf("Must specify at least one clientURI\n")
```
richardpringle, line 25: Can we not validate the config before we get to this point?

cmd/simulator/load/loader.go
```diff
@@ -0,0 +1,107 @@
...
+		os.Exit(1)
+	case len(config.Endpoints) < config.Workers:
+		// Ensure there are at least [config.Workers] config.Endpoints by creating
+		// duplicates as needed.
+		for i := 0; len(config.Endpoints) < config.Workers; i++ {
```
richardpringle, line 30: Yeah, having an `endpointCount` would be nicer than making the same call over and over. You've hit the magic rule of three here (and there's at least one more below).

cmd/simulator/load/tx_generator.go
```diff
@@ -0,0 +1,117 @@
...
+type CreateTxData func() *TxData
+
+type TxData struct {
+	To    *common.Address
+	Data  []byte
+	Value *big.Int
+	Gas   uint64
+}
+
+func GenerateFundDistributionTxSequence(key *ecdsa.PrivateKey, chainID *big.Int, signer types.Signer, startingNonce uint64, gasFeeCap *big.Int, gasTipCap *big.Int, fundAddrs []common.Address, value *big.Int) ([]*types.Transaction, error) {
```
richardpringle, line 28: Ouff... a struct for the arguements would be nice here...

cmd/simulator/load/tx_generator.go
```diff
@@ -0,0 +1,117 @@
...
+		})
+		if err != nil {
+			return nil, fmt.Errorf("failed to sign tx at index %d: %w", i, err)
+		}
+		txs = append(txs, tx)
+	}
+	return txs, nil
+}
+
+func GenerateNoopTxSequence(key *ecdsa.PrivateKey, chainID *big.Int, signer types.Signer, startingNonce uint64, gasFeeCap *big.Int, gasTipCap *big.Int, address common.Address, numTxs uint64) ([]*types.Transaction, error) {
```
richardpringle, line 76: Would be cleaner to have a constructor, `NilTxData()` and use that in the closure, calling `GenerateTxSequence` instead of having this wrapper function.

richardpringle (commented):
My review is a little overly critical; this is how I learn. Feel free to enlighten me on some of my comments, or you can even say, "I agree with your comment, but it's not worth the effort to make the change". 

I also may have made comments on files that you moved and didn't change any of the logic... if that's the case, whoops...

cmd/simulator/load/funder.go
```diff
@@ -0,0 +1,106 @@
...
+
+// DistributeFunds ensures that each address in keys has at least [minFundsPerAddr] by sending funds
+// from the key with the highest starting balance.
+// This function should never return a set of keys with length less than [numKeys]
+func DistributeFunds(ctx context.Context, client ethclient.Client, keys []*key.Key, numKeys int, minFundsPerAddr *big.Int) ([]*key.Key, error) {
+	if len(keys) < numKeys {
+		return nil, fmt.Errorf("insufficient number of keys %d < %d", len(keys), numKeys)
+	}
+	fundedKeys := make([]*key.Key, 0, numKeys)
+	// TODO: clean up fund distribution.
```
anusha-ctrl, line 26: should we validate that the client is not nil ?

cmd/simulator/load/loader.go
```diff
@@ -0,0 +1,107 @@
...
+			if err := newKey.Save(config.KeyDir); err != nil {
+				return nil, fmt.Errorf("failed to save %d new key: %w", i, err)
+			}
+			keys = append(keys, newKey)
+		}
+	}
+
+	maxFeeCap := new(big.Int).Mul(big.NewInt(params.GWei), big.NewInt(config.MaxFeeCap))
+	minFundsPerAddr := new(big.Int).Mul(maxFeeCap, big.NewInt(int64(config.TxsPerWorker*params.TxGas)))
+	log.Info("Distributing funds", "numTxsPerWorker", config.TxsPerWorker, "minFunds", minFundsPerAddr)
```
anusha-ctrl, line 63: can we put more comments around the `maxFeeCap` calc and the `minFundsPerAddr`

cmd/simulator/load/tx_generator.go
```diff
@@ -0,0 +1,117 @@
...
+
+type CreateTxData func() *TxData
+
+type TxData struct {
+	To    *common.Address
+	Data  []byte
+	Value *big.Int
+	Gas   uint64
+}
+
```
anusha-ctrl, line 27: some comments on the function would be great for all of these

cmd/simulator/load/workers.go
```diff
@@ -0,0 +1,96 @@
...
+	for _, worker := range wg.Workers {
+		worker := worker
+		eg.Go(func() error {
+			return f(ctx, worker)
+		})
+	}
+
+	return eg.Wait()
+}
+
```
anusha-ctrl, line 43: do these functions need to be public?

cmd/simulator/load/workers.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+	endTime := time.Now()
+	log.Info("Transaction confirmation completed", "confirmationElapsed", endTime.Sub(confirmationStart))
+	totalTxs := 0
+	for _, worker := range wg.Workers {
+		totalTxs += len(worker.txs)
+	}
+
+	// We exclude the time for ConfirmAllTransactions because we assume once the nonce is reported, the transaction has been finalized.
+	// For Subnet-EVM, this assumes that unfinalized data is not exposed via API.
```
anusha-ctrl, line 92: Sounds good

cmd/simulator/load/tx_generator.go
anusha-ctrl: Agree, not sure we need another wrapper function for generating the  Tx Sequence

cmd/simulator/load/loader.go
```diff
@@ -0,0 +1,107 @@
...
+	if len(clients) < config.Workers {
+		return nil, fmt.Errorf("less clients %d than requested workers %d", len(clients), config.Workers)
+	}
+	if len(senders) < config.Workers {
+		return nil, fmt.Errorf("less senders %d than requested workers %d", len(senders), config.Workers)
+	}
+	if len(txSequences) < config.Workers {
+		return nil, fmt.Errorf("less txSequences %d than requested workers %d", len(txSequences), config.Workers)
+	}
+
```
anusha-ctrl, line 93: All these assumptions abut number of workers, clients, etc can we put them in the `--help` so that people can enter the flags correctly

cmd/simulator/load/loader.go
aaronbuchwald: I think `len(...)` is fine here. If it were another function call, I'd agree.

cmd/simulator/load/loader.go
aaronbuchwald: Good call

cmd/simulator/load/loader.go
aaronbuchwald: Generally, I'd agree with normal function calls, but I'd say I actually prefer `len(...)` in this case.

cmd/simulator/load/tx_generator.go
```diff
@@ -0,0 +1,117 @@
...
+		key,
+		chainID,
+		signer,
+		startingNonce,
+		gasFeeCap,
+		gasTipCap,
+		func() *TxData {
+			return &TxData{
+				To:    &address,
+				Data:  nil,
```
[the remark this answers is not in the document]
aaronbuchwald: ya these are no-op transactions

cmd/simulator/load/loader.go
aaronbuchwald: Good call

cmd/simulator/load/funder.go
aaronbuchwald: no

cmd/simulator/load/workers.go
aaronbuchwald: Yes, these are intended to be public so that an external package can execute them step by step.

cmd/simulator/load/tx_generator.go
aaronbuchwald: Removed and re-factored this

cmd/simulator/load/tx_generator.go
aaronbuchwald: Good callout, refactored this to simplify the generators a bit and put more of the work on the caller. This half-way point was not a good middleground.

cmd/simulator/config/flags.go
```diff
@@ -80,12 +106,12 @@ func BuildFlagSet() *pflag.FlagSet {
...
-	fs.Uint64(TxsPerWorkerKey, 100, "Specify the number of transactions to create per worker.")
-	fs.Int(WorkersKey, 1, "Specify the number of workers to create for the simulator.")
+	fs.StringSlice(EndpointsKey, []string{"ws://127.0.0.1:9650/ext/bc/C/ws"}, "Specify a comma separated list of RPC Websocket Endpoints (minimum of 1 endpoint)")
+	fs.Int64(MaxFeeCapKey, 50, "Specify the maximum fee cap to use for transactions denominated in GWei (must be > 0)")
+	fs.Int64(MaxTipCapKey, 1, "Specify the max tip cap for transactions denominated in GWei (must be >= 0)")
+	fs.Uint64(TxsPerWorkerKey, 100, "Specify the number of transactions to create per worker (must be > 0)")
+	fs.Int(WorkersKey, 1, "Specify the number of workers to create for the simulator (must be > 0).")
 	fs.String(KeyDirKey, ".simulator/keys", "Specify the directory to save private keys in (INSECURE: only use for testing)")
-	fs.Duration(TimeoutKey, 5*time.Minute, "Specify the timeout for the simulator to complete.")
+	fs.Duration(TimeoutKey, 5*time.Minute, "Specify the timeout for the simulator to complete (0 indicates no timeout).")
```
darioush, line 115: nit: should we be consistent with `.` at the end?

cmd/simulator/config/flags.go
```diff
@@ -0,0 +1,117 @@
...
+	if c.TxsPerWorker == 0 {
+		return c, ErrNoTxs
+	}
+	// Note: it's technically valid for the fee/tip cap to be 0, but cannot
+	// be less than 0.
+	if c.MaxFeeCap < 0 {
+		return c, fmt.Errorf("invalid max fee cap %d < 0", c.MaxFeeCap)
+	}
+	if c.MaxTipCap < 0 {
+		return c, fmt.Errorf("invalid max tip cap %d <= 0", c.MaxTipCap)
```
darioush, line 72:
```suggestion
		return c, fmt.Errorf("invalid max tip cap %d < 0", c.MaxTipCap)
```

scripts/run_simulator.sh
```diff
@@ -22,18 +22,18 @@ run_simulator() {
     #################################
     echo "building simulator"
     pushd ./cmd/simulator
-    go install -v .
+    go build -o ./simulator main/*.go
```
darioush, line 25: can we add a comment to the top of this file explaining what running it does?

cmd/simulator/load/funder.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/ava-labs/subnet-evm/core/types"
+	"github.com/ava-labs/subnet-evm/ethclient"
+	"github.com/ava-labs/subnet-evm/params"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/log"
+)
+
+// DistributeFunds ensures that each address in keys has at least [minFundsPerAddr] by sending funds
+// from the key with the highest starting balance.
+// This function will never return a set of keys with length less than [numKeys]
```
darioush, line 22:
```suggestion
// This function returns a set of at least [numKeys] keys, each having a minimum balance [minFundsPerAddr].
```

cmd/simulator/load/workers.go
```diff
@@ -0,0 +1,104 @@
...
+	defer func() {
+		log.Info("Completed execution", "totalTime", time.Since(start))
+	}()
+
+	executionStart := time.Now()
+	log.Info("Executing transactions", "startTime", executionStart)
+	if err := wg.IssueTxs(ctx); err != nil {
+		return err
+	}
+	awaitStart := time.Now()
```
darioush, line 81: nit: could we add wspace before this or remove the one on line 86 for consistency?

cmd/simulator/load/workers.go
```diff
@@ -0,0 +1,104 @@
...
+
+	"github.com/ava-labs/subnet-evm/core/types"
+	"github.com/ava-labs/subnet-evm/ethclient"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/log"
+	"golang.org/x/sync/errgroup"
+)
+
+type WorkerGroup struct {
+	Workers []*Worker
```
darioush, line 18: nit: should we use `workers` instead?

cmd/simulator/load/funder.go
```diff
@@ -0,0 +1,124 @@
...
+	gasTipCap, err := client.SuggestGasTipCap(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch suggested gas tip: %w", err)
+	}
+	signer := types.LatestSignerForChainID(chainID)
+
+	// Generate a sequence of transactions to distribute the required funds.
+	log.Info("Generating distribution transactions")
+	addrs := make([]common.Address, len(needFundsAddrs))
+	copy(addrs, needFundsAddrs)
```
darioush, line 85: why is `addrs` and this copy necessary?

cmd/simulator/load/loader.go
```diff
@@ -0,0 +1,125 @@
...
+)
+
+// CreateLoader creates a WorkerGroup from [config] to perform the specified simulation.
+func CreateLoader(ctx context.Context, config config.Config) (*WorkerGroup, error) {
+	// Construct the arguments for the load simulator
+	if len(config.Endpoints) < config.Workers {
+		// Ensure there are at least [config.Workers] config.Endpoints by creating
+		// duplicates as needed.
+		for i := 0; len(config.Endpoints) < config.Workers; i++ {
+			config.Endpoints = append(config.Endpoints, config.Endpoints[i])
```
darioush, line 29: could we avoid modifying the `config` object so it always corresponds to the command line arguments as provided?

darioush (commented):
LGTM, some nits. Can approve after they are addressed.

cmd/simulator/config/flags.go
richardpringle: No `.`! Technically, you should not have a `.` if you aren't forming a complete sentence and often, error messages or logs are only sentence fragments. To `.` or not to `.` is the bane of my existence, and if you ever review any of my code, I'm sure you will see inconsistent use (especially in comments).

cmd/simulator/load/workers.go
richardpringle: Agree with @darioush here. I'm a big fan of the empty line before and after `if`/`for`/`switch` blocks. Buuuut, not the end of the world. I'm also a big fan of following up a PR with logical changes with another PR address formatting nits...

richardpringle (approved):

cmd/simulator/config/flags.go
darioush: yes I mean either consistent with having or not having them :)

cmd/simulator/config/flags.go
aaronbuchwald: the tip cap can be 0

cmd/simulator/config/flags.go
aaronbuchwald: Gonna update this to no `.`

cmd/simulator/load/workers.go
aaronbuchwald: Updating this

cmd/simulator/load/funder.go
aaronbuchwald: Not necessary, removing

cmd/simulator/load/loader.go
aaronbuchwald: Updated

darioush (approved):
lgtm

anusha-ctrl (approved):

PR #622, opened 2023-04-24
metrics: use atomic type

darioush:
## Why this should be merged
Uses the atomic types instead of `atomic.`

## How this works
Takes [this change](https://github.com/ethereum/go-ethereum/pull/27121) from upstream

## How this was tested
UTs

## How is this documented
No changes necessary


aaronbuchwald (approved):

richardpringle (approved):
🚀

PR #627, opened 2023-04-25
save e2e anr artifacts

felipemadero:
## Why this should be merged

It saves anr temporaries as CI artifact ready to download. Keep a record of full anr execution to better study/address problems/bug observed on CI, either during PR development or as a result of dependency changes. As CI is a particular env with limited resources, some times certain bugs are most easily observed here.


darioush:
@felipemadero could you please update the PR description with why this is done and what it will be used for? Thanks

aaronbuchwald (approved):

PR #631, opened 2023-04-27
bump linting timeout to 10m

aaronbuchwald:
## Why this should be merged

This PR bumps the linting timeout to 10min in CI since it has flaked a few times. For reference, AvalancheGo does not specify a linting timeout and takes around 4min.

## How this works

increases timeout

## How this was tested

ci

## How is this documented

n/a


darioush (approved):
hopefully this is the fix

patrick-ogrady (approved):

PR #636, opened 2023-05-01
Rename test_ to step_ for clarity

richardpringle:
## Why this should be merged
In the DS-Test style tests, there is some ambiguity with what should/shouldn't constitue a test. This clears up that ambiguity by using `step_` as a clear differentiator of a prefixed method that should only be used in a testing context (as opposed to a helper). The issue with `test_` is that one might think they can fully understand a test from the `Solidity` method definition, without looking at how the method is being called from `TypeScript`. 

## How this works
Same as before, just a different name

### DIFF
--- a/contract-examples/test/README.md
+++ b/contract-examples/test/README.md
@@ -1,6 +1,6 @@
 # Testing Precompiles
 
-If you can, put all of your test logic into DS-test tests. Prefix test functions with `test_`. There's also a `setUp` function that gets called before the test contract is deployed. The current best-practice is to re-deploy one test contract per `test` function called in `*.ts` test definitions. The `setUp` method should be called once, then `test_` functions passed in as the 2nd argument to `test("<description>", <test_function_name_OR_array_of_test_function_names>)` will be called in order. `test.only` and `test.skip` behave the same way as `it.only` and `it.skip`. There's also a `test.debug` that combines `test.only` with some extra event logging (you can use `emit log_string` to help debug Solidity test code).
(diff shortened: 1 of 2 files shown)


contract-examples/test/README.md
```diff
@@ -37,7 +37,7 @@ test(
 Many contract functions can be called as a part of the same test:
 
 ```ts
-test("<test_name>", ["<test_fn1>", "<test_fn2>", "<test_fn3>"])
+test("<step_name>", ["<step_fn1>", "<step_fn2>", "<step_fn3>"])
```
darioush, line 40: should we do:
```suggestion
test("<test_name>", ["<step_fn1>", "<step_fn2>", "<step_fn3>"])
```
?
Also if each step is a transaction maybe that can be reflected in the name somehow?

richardpringle: Global search and replace failed me 😞.

>Also if each step is a transaction maybe that can be reflected in the name somehow?

I thought the same thing... couldn't think of anything that didn't look weird though. I think it's fine as is. One really needs to look at the TS to understand what's going on and I think it's clear that `contract.some_method()` is a transaction from any context outside of solidity. And I guess there's nothing preventing someone from defining the following in solidity:
```solidity
function step_1a() public {
    // ...
}
function step_1b() public {
    // ...
}

function step_1() public {
    this.step_1a();
    this.step_1b();
}
```

Though I wouldn't encourage it. 

If you have any ideas that aren't `tx_step_`, I'm open...

darioush (approved):
Looks like an improvement.

As a side note, could we avoid force pushing branches when not needed?

ceyonur (approved):

aaronbuchwald (approved):

Issue #638, opened 2023-05-03
Prepare Subnet-EVM for next mandatory network upgrade 


ceyonur:
We're planning to put some upgrades/changes to precompiles, which will require mandatory upgrades.(#345, #510) Currently we have `NetworkUpgrades` but these must be specified from genesis/upgrade bytes. We should be able to define a timestamp in the code that would go effective without requiring a corresponding entry in the genesis/JSON upgrade.

PR #640, opened 2023-05-04
mandatory network upgrades

ceyonur:
## Why this should be merged

We should be able to specify network upgrades in the code, without expecting them to be in upgrade bytes.

Closes: #638 

## How this works

Removes network uprades from UpgradeConfig and move them to specifiable variables for networks (fuji + mainnet + local)

## How this was tested

Locally tested

## How is this documented

No need for documentation


richardpringle (commented):
LGTM
(not leaving a full review as I am not super familiar with the code)

params/network_upgrades.go
```diff
@@ -5,18 +5,52 @@ package params
...
 		return newCompatError("SubnetEVM fork block timestamp", n.SubnetEVMTimestamp, newcfg.SubnetEVMTimestamp)
 	}
+	if isForkIncompatible(n.DUpgradeTimestamp, newcfg.DUpgradeTimestamp, headTimestamp) {
+		return newCompatError("DUpgrade fork block timestamp", n.DUpgradeTimestamp, newcfg.DUpgradeTimestamp)
+	}
 
 	return nil
 }
+
+func GetNetworkUpgrades(networkID uint32) NetworkUpgrades {
```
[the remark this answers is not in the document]
ceyonur: There are 2 different options:
1- Allow these timestamps to be overriden by genesis/external upgrades
2- Allow separate optional network upgrades.

There could be some problems allowing overrides for avalanche network-wide upgrades. Like AvalancheGo versions, a critical bug etc etc. 

I can revert changes that removes specifiable network upgrades, and convert them to "Optional" network upgrades. In any case we should keep this mandatory network upgrade notion in the code, without expecting to be activated by genesis/upgrade files.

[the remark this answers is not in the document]
ceyonur: Added optional network upgrades

plugin/evm/vm.go
```diff
@@ -310,6 +313,10 @@ func (vm *VM) Initialize(
 
 	vm.ethConfig = ethconfig.NewDefaultConfig()
 	vm.ethConfig.Genesis = g
+	// NetworkID of EVM is equal to the ChainID
+	// But this is different than Avalanche's NetworkID
+	// which represents the Avalanche network the EVM is running on
+	// like Fuji, Mainnet, Local, etc.
```
aaronbuchwald, line 319: what's the reason for this comment?

the networkID is not necessarily the same as the chainID in the EVM, but some tooling incorrectly assumes it is and adds incorrect checks to verify this which caused us to follow this convention so that the tooling would work correctly 😅

Could we either remove this comment or clarify this point?

plugin/evm/vm_test.go
```diff
@@ -3016,6 +2964,21 @@ func TestSkipChainConfigCheckCompatible(t *testing.T) {
 	require.NoError(t, reinitVM.Shutdown(context.Background()))
 }
 
+func TestMandatoryUpgrades(t *testing.T) {
+	// use modified genesis to set SubnetEVM timestamp to 1
+	genesisJsonModifiedSubnetEVM := strings.Replace(genesisJSONSubnetEVM, "\"subnetEVMTimestamp\":0", "\"subnetEVMTimestamp\":1", 1)
```
aaronbuchwald, line 2969: This seems like something we could easily break if we changed the string formatting slightly.

Could we alter this format to unmarshal the JSON, update the value, and re-marshal it to pass it in as we do in other places?

aaronbuchwald: I see, so the expected behavior is that mandatory upgrades are overridden completely based on the Avalanche network enabled network upgrades.

params/network_upgrades.go
```diff
@@ -5,18 +5,74 @@ package params
...
+	if isForkIncompatible(m.SubnetEVMTimestamp, newcfg.SubnetEVMTimestamp, headTimestamp) {
+		return newCompatError("SubnetEVM fork block timestamp", m.SubnetEVMTimestamp, newcfg.SubnetEVMTimestamp)
+	}
+	if isForkIncompatible(m.DUpgradeTimestamp, newcfg.DUpgradeTimestamp, headTimestamp) {
+		return newCompatError("DUpgrade fork block timestamp", m.DUpgradeTimestamp, newcfg.DUpgradeTimestamp)
 	}
+	return nil
+}
 
+func (m *MandatoryNetworkUpgrades) MandatoryForkOrder() []fork {
```
aaronbuchwald, line 48: Can we un-export this, so that we do not have an exported function returning an internal type?

params/network_upgrades.go
```diff
@@ -5,18 +5,74 @@ package params
...
+type OptionalNetworkUpgrades struct {
+}
+
+func (n *OptionalNetworkUpgrades) CheckOptionalCompatible(newcfg *OptionalNetworkUpgrades, headTimestamp *big.Int) *ConfigCompatError {
 	return nil
 }
+
+func (n *OptionalNetworkUpgrades) OptionalForkOrder() []fork {
+	return []fork{}
+}
```
aaronbuchwald, line 78: Same comment here on returning an internal type from an exported function

aaronbuchwald (approved):

plugin/evm/vm_test.go
darioush: nit: can we adopt the convention of using backtick (``) to quote strings containing `"`?

plugin/evm/vm_test.go
```diff
@@ -3016,6 +2964,21 @@ func TestSkipChainConfigCheckCompatible(t *testing.T) {
...
+	_, vm, _, _ := GenesisVM(t, true, genesisJsonModifiedSubnetEVM, "", "")
+
+	defer func() {
+		if err := vm.Shutdown(context.Background()); err != nil {
+			t.Fatal(err)
```
darioush, line 2974: nit: can use require here.

plugin/evm/vm_upgrade_bytes_test.go
```diff
@@ -157,76 +157,10 @@ func TestVMUpgradeBytesPrecompile(t *testing.T) {
 	assert.Equal(t, signedTx1.Hash(), txs[0].Hash())
 }
 
-func TestVMUpgradeBytesNetworkUpgrades(t *testing.T) {
```
darioush, line 160: Can we keep these tests and modify them to be with optional upgrades instead?
Since the optional upgrades should still function with upgrade bytes?

plugin/evm/vm.go
ceyonur: changed it

params/network_upgrades.go
ceyonur: changed it

plugin/evm/vm_test.go
ceyonur: I used this to keep the convention with other tests, but yea I think we can change it. 

We should definitely modify these tests to use require. Created an issue https://github.com/ava-labs/subnet-evm/issues/701

plugin/evm/vm_upgrade_bytes_test.go
ceyonur: We don't currently have any actual upgrade to test out. I commented out a working case and added a TODO.

darioush (approved):

PR #644, opened 2023-05-12
core/tx_pool: document why write lock is needed for flatten

gyuho:
## Why this should be merged

Document write locks in tx pool

## How this works

Just add doc comment.

## How this was tested

N/A

## How is this documented

This is documentation :)


ceyonur (approved):

aaronbuchwald:
Could you make this a PR to geth first?

aaronbuchwald:
Closing in favor of making this change upstream

PR #648, opened 2023-05-12
plugin/evm: remove unncessary tx.Hash op in map iter

gyuho:
## Why this should be merged

Simplify, eliminate unnecessary Hash op.

## How this works

Just use the key in map iteration.

## How this was tested

N/A

## How is this documented

N/A


ceyonur (approved):
LGTM. If we have concerns with `tx.Hash()` (cycle/mem) we can also remove them below by converting `txs` to a txHash > tx map.

aaronbuchwald (approved):

PR #649, opened 2023-05-12
use Kurtosis instead of the run script

h4ck3rk3y:
## Why this should be merged

This replaces the current run script with Kurtosis which is dockerized & cleans up after itself.

## How this works

Installs Kurtosis in the CI and then runs test using Kurtosis. This forces the node thats spinning up to expose ports on 9650, so that we don't have to edit the existing testsuite too much

## How this was tested

This is a test; CI runs fine

## How is this documented

Not sure where we document such things


h4ck3rk3y:
@aaronbuchwald Hey! I took over Tedi's project to replace the script with Kurtosis as he stopped working part time a few days ago.

.github/workflows/ci.yml
```diff
@@ -65,6 +65,12 @@ jobs:
         uses: actions/setup-go@v3
         with:
           go-version: "1.19"
+      - name: Install & Restart Kurtosis
+        run: |
+          echo "deb [trusted=yes] https://apt.fury.io/kurtosis-tech/ /" | sudo tee /etc/apt/sources.list.d/kurtosis.list
+          sudo apt update
+          sudo apt install kurtosis-cli
+          kurtosis engine restart
```
aaronbuchwald, line 73: Why does this require a restart after install?

h4ck3rk3y: I have changed it to `start`. I am just super used to using `kurtosis engine restart` locally!

tests/utils/node_launcher.go
```diff
@@ -0,0 +1,75 @@
...
+	isPartitioningEnabled    = false
+	enclaveIdPrefix          = "test"
+	avalancheStarlarkPackage = "github.com/kurtosis-tech/avalanche-package"
+	// forces the node to launch on 9650 instead of ephemeral ports
+	forceExposeOn9650  = `{"test_mode": true}`
+	defaultParallelism = 4
+	firstNodeId        = "node-0"
+)
+
+func SpinupAvalancheNode() (error, string, func()) {
```
aaronbuchwald, line 22: The error should be the last return value instead of the first

h4ck3rk3y: Yes, my bad. Fixed the ordering.

tests/utils/node_launcher.go
```diff
@@ -0,0 +1,75 @@
...
+	isPartitioningEnabled    = false
+	enclaveIdPrefix          = "test"
+	avalancheStarlarkPackage = "github.com/kurtosis-tech/avalanche-package"
+	// forces the node to launch on 9650 instead of ephemeral ports
+	forceExposeOn9650  = `{"test_mode": true}`
+	defaultParallelism = 4
+	firstNodeId        = "node-0"
+)
+
+func SpinupAvalancheNode() (string, func(), error) {
```
aaronbuchwald, line 22: What would it take to change this from `SpinupAvalancheNode` to spin up a network of N nodes and how would you specify configs of each individual node?

tests/utils/node_launcher.go
```diff
@@ -0,0 +1,75 @@
...
+
+	if runResult.InterpretationError != nil {
+		return "", nil, errors.New("error interpreting Starlark code")
+	}
+	if len(runResult.ValidationErrors) != 0 {
+		return "", nil, errors.New("error validating Starlark code")
+	}
+	if runResult.ExecutionError != nil {
+		return "", nil, errors.New("error executing Starlark code")
+	}
```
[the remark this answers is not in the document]
h4ck3rk3y: I have propagated the errors. We check three different types of errors as Kurtosis executes in [Multiple Phases](https://docs.kurtosis.com/explanations/why-multi-phase-runs)

Perhaps our SDK could have an API that abstracts away the errors and returns `scriptOutput, err` where err is true regardless of it being interpretation, validation, execution or any kind of errors. So far we have been following a pattern where we assert all of them are nils individually like [here](https://github.com/kurtosis-tech/awesome-kurtosis/blob/main/quickstart/go-test/main_test.go#L64-L67)

tests/utils/node_launcher.go
h4ck3rk3y: To run multiple nodes; we'll have to change the args to include `node_count` like

```
forceExposeOn9650        = `{"test_mode": true, "node_count": 5}`
```

This would run 5 different nodes. Note Kurtosis by default tries to expose ephemeral ports but we're using the `"test_mode": true` to force public ports to be a certain number(first node rpc gets 9650, second gets 9652 and so forth). Ideally we remove the `test_mode` and pass around the ephemeral port generated inside these tests.

Re config - At the moment all nodes start with the same config except for bootstrap information. The second bootstraps from the first one; the third from the first two. What other config would you like to tweak? It shouldn't be too hard to set it; but I wanted to get something minimal out

tests/utils/node_launcher.go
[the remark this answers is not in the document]
h4ck3rk3y: We spent some cycles this morning and have changed the API so that only one error needs to be checked and it propagates the right error to you. Making it simpler for people to use the Go SDK.

tests/utils/node_launcher.go
aaronbuchwald: Cool that makes sense. AvalancheGo requires uses two ports the HTTP port (https://docs.avax.network/nodes/maintain/avalanchego-config-flags#--http-port-int) and the Staking port (https://docs.avax.network/nodes/maintain/avalanchego-config-flags#--staking-port-int).

If it's easier for Kurtosis to use ephemeral ports, you may just want to assign those two ports via these flags.

As in the documentation, the staking port is where the node listens for incoming p2p connections and the http port is used to run an API server.

tests/utils/node_launcher.go
```diff
@@ -0,0 +1,68 @@
...
+	rpcPortNumber := rpcPortSpec.GetNumber()
+
+	tearDownFunction := func() {
+		fmt.Println(fmt.Printf("Destroying enclave with id '%v'", enclaveId))
+		if err = kurtosisCtx.StopEnclave(ctx, enclaveId); err != nil {
+			fmt.Printf("An error occurred while stopping the enclave with id '%v'\n", enclaveId)
+		}
+		if err = kurtosisCtx.DestroyEnclave(ctx, enclaveId); err != nil {
+			fmt.Printf("An error occurred while cleaning up the enclave with id '%v'\n", enclaveId)
+		}
```
aaronbuchwald, line 64: should these errors be propagated somewhere?

I think failure to teardown should be treated as unexpected behavior and cause the test to fail. If this causes tests to flake intermittently, then I'd consider it a bug.

tests/utils/node_launcher.go
```diff
@@ -0,0 +1,68 @@
...
+	"fmt"
+	"time"
+
+	"github.com/kurtosis-tech/kurtosis/api/golang/engine/lib/kurtosis_context"
+)
+
+const (
+	isPartitioningEnabled    = false
+	enclaveIdPrefix          = "test"
+	avalancheStarlarkPackage = "github.com/kurtosis-tech/avalanche-package"
```
aaronbuchwald, line 17: Is this running a static version of AvalancheGo/Subnet-EVM? For CI, we would want to build a Docker image of AvalancheGo that has the VM binary present in the plugin directory and launch a node running that otherwise it seems that this will not actually be testing the latest code.

scripts/run_ginkgo.sh
```diff
@@ -3,7 +3,6 @@ set -e
 
 # This script assumes that an AvalancheGo and Subnet-EVM binaries are available in the standard location
 # within the $GOPATH
-# The AvalancheGo and PluginDir paths can be specified via the environment variables used in ./scripts/run.sh.
```
aaronbuchwald, line 6: Since this comes with new requirements (install kurtosis cli), would it be possible to add or link instructions that are necessary to run the e2e tests with this change?

tests/utils/node_launcher.go
h4ck3rk3y: So we have one container per Avalanche Node inside the package. Within the Docker network the ports are set to the defaults 9650 and 9651; Kurtosis then produces (by default) an ephemeral port on localhost. That is the port that the tests interact with; with `test_mode` true I am forcing Kurtosis to produce local port mappings that are expected

So 9650,9651 on the first node will map to 9650 and 9651 on localhost
For the second avalanche node they map to 9652 and 9653
.. so on and so forth!

I am using `test_mode` as the tests seem to hardcode 

```
var (
	DefaultLocalNodeURI = "http://127.0.0.1:9650"
	NodeURIs            = []string{DefaultLocalNodeURI, "http://127.0.0.1:9652", "http://127.0.0.1:9654", "http://127.0.0.1:9656", "http://127.0.0.1:9658"}
)
```

and I didn't want to change your code base too much! There's not much extra effort for Kurtosis here!

Perhaps I can make the load test a multi node load test based on what I see here

```
tests/load/load_test.go
56:		rpcEndpoints := make([]string, 0, len(utils.NodeURIs))
57:		for _, uri := range []string{utils.DefaultLocalNodeURI} { // TODO: use NodeURIs instead, hack until fixing multi node in a network behavior
```

h4ck3rk3y: I have changed the load test slightly and now I pass utils.NodeURIs

tests/utils/node_launcher.go
h4ck3rk3y: I've added image building as a step in the CI test and I pass the built image in the avalanche package now using the `:test` tag

tests/utils/node_launcher.go
h4ck3rk3y: Makes sense! I am now asserting that the teardown has no errors.

scripts/run_ginkgo.sh
h4ck3rk3y: I have added instructions about installing Kurtosis & restarting the engine. Let me know what you think :-)

tests/utils/node_launcher.go
h4ck3rk3y: I have reverted the multi node load test as it flaked on one run - https://github.com/h4ck3rk3y/subnet-evm/actions/runs/5093958012/jobs/9157181604

aaronbuchwald:
Closing in favor of migrating to https://github.com/ava-labs/avalanchego/tree/master/tests/e2e#avalanche-e2e-test-suites

PR #650, opened 2023-05-16
deployExampleRewardManager.ts hardhat script fixed

izzetemredemir:
## Why this should be merged
Fixed incorrect contract call in deployExampleRewardManager.ts hardhat script; getContractFactory now correctly calls "ExampleRewardManager"


ceyonur (approved):
Good catch 👍

PR #659, opened 2023-05-29
export gingkgo suites

ceyonur:
## Why this should be merged

It exports necessary test suite functions (before/after) to be used in other repos

## How this works

Moves common after/before suites to utils package and exports them.

## How this was tested

E2E tests 

## How is this documented

It will be documented through external precompile registration tutorial.

### DIFF
--- a/tests/utils/command.go
+++ b/tests/utils/command.go
@@ -50,13 +50,11 @@ func RunCommand(bin string, args ...string) (*cmd.Cmd, error) {
 }
 
 func RegisterPingTest() {
-	ginkgo.Describe("[Precompiles]", ginkgo.Ordered, func() {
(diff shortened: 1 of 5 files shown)


tests/utils/command.go
```diff
@@ -43,3 +48,46 @@ func RunCommand(bin string, args ...string) (*cmd.Cmd, error) {
 
 	return curCmd, nil
 }
+
+func RegisterPingTest() {
+	ginkgo.Describe("[Precompiles]", ginkgo.Ordered, func() {
```
darioush, line 53: should the reference to `[Precompiles]` still be here?

darioush (approved):

tests/utils/command.go
ceyonur: ah good catch, no need for that.

PR #663, opened 2023-06-08
Publish npm module

ceyonur:
Prepares files for NPM module publishing.


contract-examples/README.md
```diff
@@ -89,8 +83,8 @@ Alternatively, you can copy and paste the `local` network configuration to creat
         "0x86f78c5416151fe3546dece84fda4b4b1e36089f2dbc48496faf3a950f16157c",
         "0x750839e9dbbd2a0910efe40f50b2f3b2f2f59f5580bb4b83bd8c1201cf9a010a"
       ],
-      pollingInterval: "1s"
-    },
+      "pollingInterval": "1s"
```
richardpringle, line 86: 🤔  The hardhat config is usually a `js` file isn't it?

contract-examples/hardhat.config.ts
richardpringle, line 1: 🥇

contract-examples/test/utils.ts
```diff
@@ -74,7 +74,7 @@ const buildTestFn = (fnNameOrObject: FnNameOrObject, overrides = {}, debug = fal
 
   // return the test function that will be used by `it`
   // this function must be defined with the `function` keyword so that `this` is bound to the Mocha context
-  return async function() {
+  return async function () {
```
richardpringle, line 77: 👎 

(haha I don't like this... but you don't actually need to change it)

.github/workflows/ci.yml
```diff
@@ -68,9 +68,9 @@ jobs:
       - name: Use Node.js
         uses: actions/setup-node@v3
         with:
-          node-version: "14.x"
-      - name: Yarn install
-        run: yarn
+          node-version: "18.x"
+      - name: NPM install
+        run: npm install
```
richardpringle, line 73: Should we use `npm ci` here instead? It will install from `package-lock.json` instead of updating any dependency patches, etc.

richardpringle (approved):
Looks good @ceyonur, 🥇 ! 

Please see my comments, nothing overly concerning though.

contract-examples/README.md
ceyonur: TS is officially supported as well

.github/workflows/ci.yml
ceyonur: good idea 👍

contract-examples/package.json
```diff
@@ -27,23 +28,19 @@
   },
   "license": "BSD-3-Clause",
   "scripts": {
+    "build": "rm -rf dist/ && npx tsc -b",
```
iknowhtml, line 31: Why are we using `npx` if we have `typescript` installed locally?

contract-examples/tsconfig.json
```diff
@@ -1,5 +1,19 @@
 {
   "compilerOptions": {
-    "resolveJsonModule": true
-  }
+    "outDir": "./dist/",
+    "baseUrl": ".",
+    "module": "commonjs",
+    "target": "es2020",
+    "lib": ["dom", "es2016"],
```
iknowhtml, line 7: I don't believe you need the `dom` library

iknowhtml (commented):
General comment that isn't related to the PR, but it may make sense to have both an ESM & CJS build for better DX for package users as opposed to just CJS which is how this package is currently configured

contract-examples/package.json
ceyonur: well I always find `npx` as the "safest" option. I believe it will run local script if found. But you're correct we probably don't need it here as we declared typescript as a dependency.

contract-examples/tsconfig.json
ceyonur: right 👍 deleted.

richardpringle (approved):

PR #672, opened 2023-06-15
Fix contract examples package json

ceyonur:
Fixes NPM package issue for imports
allows customized CMD to be run for e2e tests.
renames contract-examples to contracts


iknowhtml (approved):

richardpringle (approved):

tests/utils/subnet.go
```diff
@@ -23,7 +23,7 @@ import (
 	"github.com/onsi/gomega"
 )
 
-func RunHardhatTests(test string, rpcURI string) {
+func RunTestCMD(testCMD *exec.Cmd, rpcURI string) {
```
aaronbuchwald, line 26: Does this need to be external or do we only use `RunDefaultHardHatTests` ? If exported, could we add a comment?

aaronbuchwald (approved):
There's one comment I think we should address before merging, but other than that this LGTM

tests/utils/subnet.go
ceyonur: `RunTestCMD` could be used if paths and test commands are different. Developers can customize their commands to run different tests (not only hardhat) in different networks (not only in local). I added more comments.

aaronbuchwald (approved):

richardpringle:
🎉

PR #694, opened 2023-06-22
Sync coreth v0.12.4 rc0

ceyonur:
## Why this should be merged

Syncs Coreth v0.12.14-rc0

## How this works

https://github.com/ava-labs/subnet-evm/discussions/693

## How this was tested

- [x] UTs should work
- [x] Local network testing
- [x] let it bootstrapped in an actual subnet
- [ ] Canary an API Node 

## How is this documented

- [ ] Precompile templates have been changed, will reflect to docs and hello world PRs


accounts/abi/abi_test.go
```diff
@@ -738,12 +738,12 @@ func TestBareEvents(t *testing.T) {
...
-//	    received(msg.sender, msg.value, memo);
-//	    receivedAddr(msg.sender);
-//	  }
+//		event received(address sender, uint amount, bytes memo);
+//		event receivedAddr(address sender);
+//		function receive(bytes memo) external payable {
+//			received(msg.sender, msg.value, memo);
+//			receivedAddr(msg.sender);
+//		}
 //	}
```
anusha-ctrl, line 747: nit: I'm seeing some more spacing fixes on this file

core/bench_test.go
```diff
@@ -157,22 +158,17 @@ func benchInsertChain(b *testing.B, disk bool, gen func(int, *BlockGen)) {
 
 	// Generate a chain of b.N blocks using the supplied block
 	// generator function.
-	gspec := Genesis{
+	gspec := &Genesis{
 		Config:   params.TestChainConfig,
 		Alloc:    GenesisAlloc{benchRootAddr: {Balance: benchRootFunds}},
-		Coinbase: constants.BlackholeAddr,
+		GasLimit: params.TestChainConfig.FeeConfig.GasLimit.Uint64(),
```
anusha-ctrl, line 164: why did we add gaslimit?

ceyonur: TestChainConfig enables Subnet-EVM fork and Subnet-EVM Fork checks if header.GasLimit == config.GasLimit.

accounts/abi/abi_test.go
ceyonur: I tried to minimize the diff between coreth:
https://github.com/ava-labs/coreth/blob/master/accounts/abi/abi_test.go

accounts/abi/unpack.go
```diff
@@ -43,43 +44,72 @@ var (
 )
 
 // ReadInteger reads the integer based on its kind and returns the appropriate value.
-func ReadInteger(typ Type, b []byte) interface{} {
+func ReadInteger(typ Type, b []byte) (interface{}, error) {
```
aaronbuchwald, line 47: 👀

core/genesis.go
```diff
@@ -176,67 +175,66 @@ func (e *GenesisMismatchError) Error() string {
...
-		return nil, fmt.Errorf("gas limit in fee config (%d) does not match gas limit in header (%d)", gasLimitConfig, genesis.GasLimit)
+	if genesis.Config.IsSubnetEVM(genesis.Timestamp) {
+		gasLimitConfig := genesis.Config.FeeConfig.GasLimit.Uint64()
+		if gasLimitConfig != genesis.GasLimit {
+			return nil, common.Hash{}, fmt.Errorf("gas limit in fee config (%d) does not match gas limit in header (%d)", gasLimitConfig, genesis.GasLimit)
+		}
+		// Verify config
+		if err := genesis.Config.Verify(); err != nil {
+			return nil, common.Hash{}, err
+		}
```
aaronbuchwald, line 195: Why are we moving this to inside of `IsSubnetEVM` ?

accounts/abi/unpack.go
aaronbuchwald: need to double check that this is backwards compatible where we use the ABI package

core/genesis.go
ceyonur: Config verification is added with Subnet-EVM.

accounts/abi/unpack.go
ceyonur: looking into this

core/vm/jump_table.go
```diff
@@ -77,6 +78,13 @@ func newSubnetEVMInstructionSet() JumpTable {
 	return validate(instructionSet)
 }
 
+func newDUpgradeInstructionSet() JumpTable {
+	instructionSet := newSubnetEVMInstructionSet()
+	enable3855(&instructionSet) // PUSH0 instruction
+	enable3860(&instructionSet) // Limit and meter initcode
+	return validate(instructionSet)
+}
```
aaronbuchwald, line 86: Can we move the DUpgrade and SubnetEVM instruction set functions below `validate(...)` ?

core/genesis.go
aaronbuchwald: We should always enforce this regardless of when SubnetEVM is enabled. If we don't, then we are no longer enforcing the invariant that chains are started with a verified set of precompile/state upgrades. Can we move this back out of the `IsSubnetEVM` branch?

Since it was enforced before, I don't see any reason to move it into this branch.

darioush: Yes I agree with above, we should keep the fee check in IsSubnetEVM but not `Config.Verify`

ceyonur: `SetupGenesisBlock` call path has changed. It was only used in eth/backend before which called only by VM. With these changes now it's being called by `NewBlockchain` which is heavily used by tests and `simulated`.  If we want to move them out of the subnet-evm then we need to make sure that all tests use a FeeConfig and proper gas limit.

core/vm/jump_table.go
ceyonur: done

core/genesis.go
darioush: I suggest we do something like: https://github.com/ava-labs/subnet-evm/pull/749/commits/03560a4ad8d2bd751a3e82b44c3f913f91e82fe8

aaronbuchwald: Ah I see,  I didn't follow the reason for the change originally.

Since we add the need for calling `Verify()` from our own code. Would it make sense to call this from `vm.go` when initializing the VM? I think we can also move the FeeConfig gas limit matches the genesis config check into the same spot.

If we move it into `IsSubnetEVM`, then we will rely on not enforcing that, which will become an issue in future mandatory hard forks where we need to enable those and still don't want to enforce this config check.

ceyonur: I think it makes sense

aaronbuchwald (approved):

core/genesis.go
```diff
@@ -245,30 +234,35 @@ func SetupGenesisBlock(
...
-		if compatErr != nil && height != 0 && compatErr.RewindTo != 0 {
-			return newcfg, compatErr
+		if compatErr != nil && ((height != 0 && compatErr.RewindToBlock != 0) || (timestamp != 0 && compatErr.RewindToTime != 0)) {
+			return newcfg, stored, compatErr
 		}
 	}
-	rawdb.WriteChainConfig(db, stored, newcfg)
-	return newcfg, nil
+	// Don't overwrite if the old is identical to the new
+	if newData, _ := json.Marshal(newcfg); !bytes.Equal(storedData, newData) {
```
aaronbuchwald, line 250: We need to remove this check so that it writes the config to disk (including the upgrade config).

Issue #662, opened 2023-06-06
Add metrics for gas price / usage


darioush:
We should add a gauge for the base fee of the last accepted block, and perhaps total fees as well.
These metrics can be useful to debug network activity.

PR #700, opened 2023-06-26
Add metrics for gas price / usage

nytzuga:
## Why this should be merged

Adds more metrics requested in #662

## How this works

New metrics has been added to extend the visibility of nodes.

## How this was tested

To be updated

## How is this documented

It is a minor request, maybe the documents for the metrics should be updated.


core/blockchain.go
```diff
@@ -1382,6 +1384,18 @@ func (bc *BlockChain) insertBlock(block *types.Block, writes bool) error {
...
+	}
+
+	totalFees := big.NewInt(0)
+	for _, tx := range block.Transactions() {
+		gasPrice := tx.GasPrice()
+		gasUsed := new(big.Int).SetUint64(tx.Gas())
+		txFees := new(big.Int).Mul(gasPrice, gasUsed)
+		totalFees = new(big.Int).Add(totalFees, txFees)
+	}
+	blockTotalFeesGauge.Update(totalFees.Int64())
```
aaronbuchwald, line 1398: Can we move these metrics into Accept here: https://github.com/ava-labs/subnet-evm/blob/master/core/blockchain.go#L1055 ? 

We should also update the `totalFees` calculation to account for the base fee and effective gas price. Unfortunately, `GasPrice` got a bit more complicated with EIP-1559: https://timroughgarden.org/papers/eip1559.pdf.

The transaction's `GasPrice()` function returns the gas price for legacy transactions (https://github.com/ava-labs/subnet-evm/blob/master/core/types/legacy_tx.go#L109) and the `GasFeeCap` for transactions added post-EIP1559 (https://github.com/ava-labs/subnet-evm/blob/master/core/types/dynamic_fee_tx.go#L102), which does not account for the actual fee paid anymore.

We actually do this incorrectly in the logs here too: https://github.com/ava-labs/subnet-evm/blob/master/miner/worker.go#L391, so I think we can leave this in this PR for now and make a ticket to correct this and factor out the code so that it can be shared across these two places.

nytzuga: Thanks for the feedback @aaronbuchwald, I moved the changes to the place you suggested I also update the total Fees calculation to include `GasFeeCap()` and `GasTipCap()`.

core/blockchain.go
```diff
@@ -1053,6 +1055,21 @@ func (bc *BlockChain) Accept(block *types.Block) error {
...
+	if baseFee := block.BaseFee(); baseFee != nil {
+		blockGasPriceGauge.Update(baseFee.Int64())
+	}
+
+	totalFees := big.NewInt(0)
+	for _, tx := range block.Transactions() {
+		gasFeeCap := tx.GasFeeCap()
+		gasTipCap := tx.GasTipCap()
+		gasUsed := new(big.Int).SetUint64(tx.Gas())
+		if gasUsed != nil && gasTipCap != nil && gasFeeCap != nil {
```
darioush, line 1067:
```suggestion
		if gasTipCap != nil && gasFeeCap != nil {
```

core/blockchain.go
```diff
@@ -1053,6 +1055,21 @@ func (bc *BlockChain) Accept(block *types.Block) error {
...
+			txFees := new(big.Int).Mul(new(big.Int).Add(gasFeeCap, gasTipCap), gasUsed)
+			totalFees = new(big.Int).Add(totalFees, txFees)
```
darioush, line 1069: nit: wonder if `totalFees.Add(totalFees, txFees)` saves an allocation.

core/blockchain.go
```diff
@@ -1053,6 +1055,21 @@ func (bc *BlockChain) Accept(block *types.Block) error {
...
 	acceptedTxsCounter.Inc(int64(len(block.Transactions())))
+	if baseFee := block.BaseFee(); baseFee != nil {
+		blockGasPriceGauge.Update(baseFee.Int64())
+	}
+
+	totalFees := big.NewInt(0)
+	for _, tx := range block.Transactions() {
+		gasFeeCap := tx.GasFeeCap()
+		gasTipCap := tx.GasTipCap()
+		gasUsed := new(big.Int).SetUint64(tx.Gas())
```
darioush, line 1066: I could be wrong.. but I believe that `tx.Gas()` returns the gas supplied and the gas used is available in receipts. eg:
https://github.com/ava-labs/coreth/blob/master/miner/worker.go#L370
I think we should follow the same computation with `tx.EffectiveGasTipValue(baseFee)` as well.

aaronbuchwald: That's correct

core/blockchain.go
```diff
@@ -1053,9 +1055,47 @@ func (bc *BlockChain) Accept(block *types.Block) error {
...
+			minerFee = tx.GasPrice()
+		} else {
+			minerFee = new(big.Int).Add(baseFee, tx.EffectiveGasTipValue(baseFee))
+		}
+		feesWei.Add(feesWei, new(big.Int).Mul(new(big.Int).SetUint64(receipts[i].GasUsed), minerFee))
+	}
+	return feesWei, nil
+}
+
+func TotalFeesFloat(block *types.Block, receipts []*types.Receipt) *big.Float {
```
aaronbuchwald, line 1090: Could we change the return type here to return the error instead of a debug log and zero?

core/blockchain.go
```diff
@@ -81,6 +81,8 @@ var (
 	acceptorQueueGauge           = metrics.NewRegisteredGauge("chain/acceptor/queue/size", nil)
 	acceptorWorkTimer            = metrics.NewRegisteredCounter("chain/acceptor/work", nil)
 	acceptorWorkCount            = metrics.NewRegisteredCounter("chain/acceptor/work/count", nil)
+	blockGasPriceGauge           = metrics.NewRegisteredGauge("chain/block/gas/price", nil)
```
aaronbuchwald, line 84:
```suggestion
	lastAcceptedBlockBaseFeeGauge           = metrics.NewRegisteredGauge("chain/block/gas/basefee", nil)
```

core/blockchain.go
```diff
@@ -81,6 +81,8 @@ var (
...
+	blockTotalFeesGauge          = metrics.NewRegisteredGauge("chain/block/fees", nil)
```
aaronbuchwald, line 85:
```suggestion
	blockTotalFeesGauge          = metrics.NewRegisteredGauge("chain/block/gas/fees", nil)
```

core/blockchain.go
```diff
@@ -1053,9 +1055,47 @@ func (bc *BlockChain) Accept(block *types.Block) error {
 	bc.addAcceptorQueue(block)
 	acceptedBlockGasUsedCounter.Inc(int64(block.GasUsed()))
 	acceptedTxsCounter.Inc(int64(len(block.Transactions())))
+	if baseFee := block.BaseFee(); baseFee != nil {
+		blockGasPriceGauge.Update(baseFee.Int64())
+	}
+	total, err := TotalFees(block, bc.GetReceiptsByHash(block.Hash()))
+	if err != nil {
+		log.Debug(fmt.Sprintf("TotalFees %s", err))
```
aaronbuchwald, line 1063: Can we change this to `Error` level? This error should never happen since the block has been accepted and we should have the same number of transactions and receipts for the indexed block.

core/blockchain.go
```diff
@@ -1053,9 +1055,47 @@ func (bc *BlockChain) Accept(block *types.Block) error {
...
+		}
+		feesWei.Add(feesWei, new(big.Int).Mul(new(big.Int).SetUint64(receipts[i].GasUsed), minerFee))
+	}
+	return feesWei, nil
+}
+
+func TotalFeesFloat(block *types.Block, receipts []*types.Receipt) *big.Float {
+	total, err := TotalFees(block, receipts)
+	if err != nil {
+		log.Debug(fmt.Sprintf("TotalFees %s", err))
```
aaronbuchwald, line 1093: this should be an `Error` log as well, since it should never happen

core/blockchain.go
```diff
@@ -1053,9 +1055,46 @@ func (bc *BlockChain) Accept(block *types.Block) error {
...
+	}
+	for i, tx := range block.Transactions() {
+		var minerFee *big.Int
+		if baseFee == nil {
+			// legacy block, no baseFee
+			minerFee = tx.GasPrice()
+		} else {
+			minerFee = new(big.Int).Add(baseFee, tx.EffectiveGasTipValue(baseFee))
+		}
+		feesWei.Add(feesWei, new(big.Int).Mul(new(big.Int).SetUint64(receipts[i].GasUsed), minerFee))
```
ceyonur, line 1085: This is a bit hard to read, can we split this into two different variables?

core/blockchain.go
```diff
@@ -78,12 +78,14 @@ var (
...
-	acceptorWorkTimer            = metrics.NewRegisteredCounter("chain/acceptor/work", nil)
-	acceptorWorkCount            = metrics.NewRegisteredCounter("chain/acceptor/work/count", nil)
-	processedBlockGasUsedCounter = metrics.NewRegisteredCounter("chain/block/gas/used/processed", nil)
-	acceptedBlockGasUsedCounter  = metrics.NewRegisteredCounter("chain/block/gas/used/accepted", nil)
-	badBlockCounter              = metrics.NewRegisteredCounter("chain/block/bad/count", nil)
+	acceptorQueueGauge            = metrics.NewRegisteredGauge("chain/acceptor/queue/size", nil)
+	acceptorWorkTimer             = metrics.NewRegisteredCounter("chain/acceptor/work", nil)
+	acceptorWorkCount             = metrics.NewRegisteredCounter("chain/acceptor/work/count", nil)
+	lastAcceptedBlockBaseFeeGauge = metrics.NewRegisteredGauge("chain/block/gas/basefee", nil)
+	blockTotalFeesGauge           = metrics.NewRegisteredGauge("chain/block/gas/fees", nil)
```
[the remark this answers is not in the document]
aaronbuchwald:
```suggestion
	lastAcceptedBlockBaseFeeGauge = metrics.NewRegisteredGauge("chain/block/fee/basefee", nil)
	blockTotalFeesGauge           = metrics.NewRegisteredGauge("chain/block/fee/total", nil)
```

good idea

core/blockchain.go
```diff
@@ -1053,9 +1055,47 @@ func (bc *BlockChain) Accept(block *types.Block) error {
 	bc.addAcceptorQueue(block)
 	acceptedBlockGasUsedCounter.Inc(int64(block.GasUsed()))
 	acceptedTxsCounter.Inc(int64(len(block.Transactions())))
+	if baseFee := block.BaseFee(); baseFee != nil {
+		lastAcceptedBlockBaseFeeGauge.Update(baseFee.Int64())
+	}
+	total, err := TotalFees(block, bc.GetReceiptsByHash(block.Hash()))
+	if err != nil {
+		log.Error(fmt.Sprintf("TotalFees error: %s", err))
```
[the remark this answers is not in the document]
aaronbuchwald: I think error is the right level here since this should never happen

core/blockchain.go
aaronbuchwald: I think this is fine for this PR since it's just copied from `miner/worker.go`

core/blockchain.go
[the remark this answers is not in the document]
ceyonur: Why dont we return the error?

aaronbuchwald (approved):

darioush (approved):

PR #702, opened 2023-06-27
added config setting to prune warpdb

morrisettjohn:
## Why this should be merged
Adds a config that lets validators prune their warpdb messages on initialization
## How this works
Clears via clearprefix

## How this was tested
Via a unit test included in test


go.mod
```diff
@@ -38,10 +38,10 @@ require (
...
 	github.com/tyler-smith/go-bip39 v1.1.0
 	github.com/urfave/cli/v2 v2.17.2-0.20221006022127-8f469abc00aa
-	golang.org/x/crypto v0.1.0
-	golang.org/x/sync v0.1.0
-	golang.org/x/sys v0.8.0
-	golang.org/x/text v0.7.0
+	golang.org/x/crypto v0.10.0
+	golang.org/x/sync v0.3.0
+	golang.org/x/sys v0.9.0
+	golang.org/x/text v0.10.0
```
aaronbuchwald, line 44: can we reset the changes to `go.mod` and `go.sum` ? 

We usually cautiously update dependencies and try to avoid coupling it to unrelated changes

plugin/evm/vm.go
```diff
@@ -272,6 +272,13 @@ func (vm *VM) Initialize(
 	vm.metadataDB = prefixdb.New(metadataPrefix, vm.db)
 	vm.warpDB = prefixdb.New(warpPrefix, vm.db)
 
+	if vm.config.PruneWarpDB {
+		err := database.ClearPrefix(vm.db, vm.db, warpPrefix)
+		if err != nil {
+			return fmt.Errorf("failed to prune warpDb")
+		}
+	}
```
aaronbuchwald, line 280: Will this clear the correct data passing in `warpPrefix` to `database.ClearPrefix` to clear the database created by `prefixdb.New(warpPrefix, ...)` ?

plugin/evm/vm_test.go
```diff
@@ -3204,3 +3206,83 @@ func TestSignatureRequestsToVM(t *testing.T) {
 	err = vm.Network.AppRequest(context.Background(), ids.GenerateTestNodeID(), 1, time.Now().Add(60*time.Second), requestBytes)
 	require.NoError(t, err)
 }
+
+// Test that the PruneWarpDB config flag applies as expected
+func TestPruneWarpDB(t *testing.T) {
+	// Because the chain is being reinitialized, metrics need to be disabled and upgrades skipped to allow initialization
+	t.Skip("no skipable upgrades")
```
aaronbuchwald, line 3213: Should we remove `Skip` ? It seems that this test fails for me if I remove it atm

go.mod
morrisettjohn: Sure thing, I wasn't entirely sure how those were updated anyways

plugin/evm/vm.go
morrisettjohn: Sounds good to me!

plugin/evm/vm_test.go
morrisettjohn: Probably better to remove it, also since the unit test is being moved to the backend this entire unit test can probably be removed anyways

aaronbuchwald (approved):

plugin/evm/config.go
```diff
@@ -122,6 +123,7 @@ type Config struct {
 	AllowMissingTries               bool    `json:"allow-missing-tries"`                // If enabled, warnings preventing an incomplete trie index are suppressed
 	PopulateMissingTries            *uint64 `json:"populate-missing-tries,omitempty"`   // Sets the starting point for re-populating missing tries. Disables re-generation if nil.
 	PopulateMissingTriesParallelism int     `json:"populate-missing-tries-parallelism"` // Number of concurrent readers to use when re-populating missing tries on startup.
+	PruneWarpDB                     bool    `json:"prune-warp-db-enabled"`              // Determines if the warpDb should be cleared when re-initialzing subnet-evm
```
anusha-ctrl, line 127: nit: re-initialzing -> reinitializing

warp/backend_test.go
```diff
@@ -21,6 +21,42 @@ var (
...
+
+	// ensure all messages have been deleted
+	for _, messageID := range messageIDs {
+		_, err := backend.GetSignature(messageID)
+		require.ErrorContains(t, err, "failed to get warp message")
+	}
+}
+
 func TestAddAndGetValidMessage(t *testing.T) {
 	db := memdb.New()
```
anusha-ctrl, line 62: do we need to add a e2e test in in `vm_test.go`?

darioush (approved):

plugin/evm/config.go
```diff
@@ -122,6 +123,7 @@ type Config struct {
...
```
ceyonur, line 126: Should we add add a validation for the case `PruneWarpDB` was enabled but `Pruning` was not? If these are not related, can we rename it to something like `warp-db-cleanup-enabled` etc? lIMHO one can confuse this with `pruning` and try to enable both of them.

ceyonur (commented):
Overall LGTM. just a concern with the config name

plugin/evm/config.go
aaronbuchwald: These are two different types of pruning.

I think pruning is the right term here tbh

aaronbuchwald, line 126:
```suggestion
	PruneWarpDB                     bool    `json:"prune-warp-db-enabled"`              // Determines if the warpDB should be cleared on startup
```

plugin/evm/config.go
aaronbuchwald: good catch updated to cleared on startup

aaronbuchwald (approved):

PR #705, opened 2023-06-28
add benchmark to template

ceyonur:
## Why this should be merged

Adds benchmarks to test template

## How this works

Changes the precompile_contract_test_template

## How this was tested

https://github.com/ava-labs/subnet-evm/pull/704

## How is this documented

Not yet


accounts/abi/bind/precompilebind/precompile_contract_test_template.go
```diff
@@ -139,4 +142,23 @@ func TestRun(t *testing.T) {
 	}
 	{{- end}}
 }
+
+func Benchmark{{.Contract.Type}}(b *testing.B) {
+	{{- if .Contract.AllowList}}
+	// Benchmark tests with allowlist tests.
+	// This adds allowlist run tests to your custom tests
```
darioush, line 149: is `allowlist run` correct? also referenced above

ceyonur: it is correct but not very obvious, so removed `run`

PR #706, opened 2023-06-28
Add metrics to simulator

anusha-ctrl:
## Why this should be merged
Adds Metrics to Simulator Tests
- TPS
- totalTime
- issuanceTime
- confirmationTime
- issuanceBatchTime
- confirmationBatchTime

- Expose individual tx time quantiles through a prometheus server. 

## How this was tested
locally
## How is this documented
N/A


cmd/simulator/txs/agent.go
```diff
@@ -53,6 +56,9 @@ func (a issueNAgent[T]) Execute(ctx context.Context) error {
 	}
 
 	txChan := a.sequence.Chan()
+	totalTxs := len(txChan)
```
darioush, line 59: could we count the txs instead of taking the chan len?

cmd/simulator/txs/agent.go
```diff
@@ -80,6 +86,11 @@ func (a issueNAgent[T]) Execute(ctx context.Context) error {
 			if err := a.worker.ConfirmTx(ctx, tx); err != nil {
 				return fmt.Errorf("failed to await transaction %d: %w", i, err)
 			}
+			confirmedCount++
+			if confirmedCount == totalTxs {
+				totalTime := time.Since(start).Seconds()
+				log.Info("Execution complete", "totalTxs", totalTxs, "totalTime", totalTime, "TPS", float64(totalTxs)/totalTime)
+			}
```
darioush, line 93: should we add per-batch metrics, and also for issueTx?

anusha-ctrl: yah we probably should

cmd/simulator/txs/agent.go
anusha-ctrl: yup

ethclient/ethclient.go
```diff
@@ -384,7 +384,7 @@ func (ec *client) SubscribeNewAcceptedTransactions(ctx context.Context, ch chan<
 	return ec.c.EthSubscribe(ctx, ch, "newAcceptedTransactions")
 }
 
-// SubscribeNewAcceptedTransactions subscribes to notifications about the accepted transaction hashes on the given channel.
+// SubscribeNewPendingTransactions subscribes to notifications about the pending transaction hashes on the given channel.
 func (ec *client) SubscribeNewPendingTransactions(ctx context.Context, ch chan<- *common.Hash) (interfaces.Subscription, error) {
```
anusha-ctrl, line 388: fixed incorrect comment

cmd/simulator/txs/agent.go
anusha-ctrl: I need a totalCount to know when to fully stop

cmd/simulator/txs/agent.go
```diff
@@ -75,11 +92,71 @@ func (a issueNAgent[T]) Execute(ctx context.Context) error {
...
+				confirmedDuration := confirmedEnd.Sub(confirmedStart)
+				log.Info("Confirmed Batch Done", "batch", batchI, "time", confirmedDuration.Seconds())
+				confirmedTime = confirmedTime.Add(confirmedDuration)
+
+				// Get the final duration by comparing it to the zero time
+				issuedFinalDuration := issuedTime.Sub(time.Time{})
+				confirmedFinalDuration := confirmedTime.Sub(time.Time{})
+
+				log.Info("Execution complete", "totalTxs", totalTxs, "totalTime", totalTime, "TPS", float64(totalTxs)/totalTime,
+					"issuanceTime", issuedFinalDuration.Seconds(), "confirmedTime", confirmedFinalDuration.Seconds())
```
darioush, line 122: nit: I don't think we need to call `.Seconds()`, we can log the duration directly

cmd/simulator/txs/agent.go
```diff
@@ -75,11 +92,71 @@ func (a issueNAgent[T]) Execute(ctx context.Context) error {
...
+		issuedTime = issuedTime.Add(issuedDuration)
 
+		// Start confirmation batch
+		confirmedStart := time.Now()
 		for i, tx := range txs {
 			if err := a.worker.ConfirmTx(ctx, tx); err != nil {
 				return fmt.Errorf("failed to await transaction %d: %w", i, err)
 			}
+			confirmedCount++
+			if confirmedCount == totalTxs {
```
darioush, line 108: I think we should switch the chan end from `return a.worker.Close(ctx)` to breaking the loop and do this after, or less preferable switching this to a defer statement

cmd/simulator/txs/agent.go
```diff
@@ -75,11 +92,71 @@ func (a issueNAgent[T]) Execute(ctx context.Context) error {
...
+		confirmedEnd := time.Now()
+		confirmedDuration := confirmedEnd.Sub(confirmedStart)
+		log.Info("Confirmed Batch Done", "batch", batchI, "time", confirmedDuration.Seconds())
+		// Add the confirmed batch time to the total confirmedTime
+		confirmedTime = confirmedTime.Add(confirmedDuration)
+		batchI++
+	}
+}
+
+func logOtherMetrics() error {
```
darioush, line 137: we should account for the time it calls for this http call

cmd/simulator/txs/agent.go
```diff
@@ -75,11 +92,71 @@ func (a issueNAgent[T]) Execute(ctx context.Context) error {
...
+		return fmt.Errorf("failed getting metrics: %w", err)
+	}
+
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return fmt.Errorf("failed reading response body of metrics: %w", err)
 	}
+
+	bodyString := string(body)
+	re := regexp.MustCompile(".*avalanche_C_vm_metervm_build_block_sum.*")
```
darioush, line 149: doesn't this look at C chain metrics?

cmd/simulator/txs/agent.go
darioush: see the comment below

cmd/simulator/txs/agent.go
anusha-ctrl: .If I remove the Seconds() we will get different units. For example, issuanceTime (ms) and confirmationTime(s). I want to keep them standard.

cmd/simulator/txs/agent.go
anusha-ctrl: done

cmd/simulator/txs/agent.go
anusha-ctrl: Yah in this simulation, we are issuing transactions to the C chain.

cmd/simulator/txs/agent.go
anusha-ctrl: Discussed offline

cmd/simulator/load/funder.go
```diff
@@ -21,7 +21,7 @@ import (
 // DistributeFunds ensures that each address in keys has at least [minFundsPerAddr] by sending funds
 // from the key with the highest starting balance.
 // This function returns a set of at least [numKeys] keys, each having a minimum balance [minFundsPerAddr].
-func DistributeFunds(ctx context.Context, client ethclient.Client, keys []*key.Key, numKeys int, minFundsPerAddr *big.Int) ([]*key.Key, error) {
+func DistributeFunds(ctx context.Context, client ethclient.Client, keys []*key.Key, numKeys int, minFundsPerAddr *big.Int, blockchainID string) ([]*key.Key, error) {
```
darioush, line 24: could we revert this change and the wspace change for this file?

darioush (approved):
LGTM after fixing comments

cmd/simulator/load/funder.go
anusha-ctrl: yes nice catch

cmd/simulator/load/loader.go
```diff
@@ -129,5 +141,39 @@ func ExecuteLoader(ctx context.Context, config config.Config) error {
...
+
+	logOtherMetrics(blockchainIDStr)
+	return nil
+}
+
+func logOtherMetrics(blockchainIDStr string) error {
+	getCallStart := time.Now()
+	resp, err := http.Get("http://127.0.0.1:9650/ext/metrics")
+	getCallEnd := time.Now()
+	getCallDuration := getCallEnd.Sub(getCallStart)
```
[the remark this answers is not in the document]
anusha-ctrl: thanks

cmd/simulator/txs/agent.go
```diff
@@ -53,33 +56,70 @@ func (a issueNAgent[T]) Execute(ctx context.Context) error {
...
+
+	// Tracks the total amount of time waiting for issuing and confirming txs
+	var (
+		totalIssuedTime    time.Duration
+		totalConfirmedTime time.Duration
+	)
+
+	defer func() error {
+		return a.worker.Close(ctx)
+	}()
```
aaronbuchwald, line 70: why are we returning an error from a deferred function that we're not checking?

ethclient/ethclient.go
aaronbuchwald: good catch

cmd/simulator/txs/agent.go
anusha-ctrl: fixed

cmd/simulator/load/loader.go
```diff
@@ -31,6 +35,25 @@ func ExecuteLoader(ctx context.Context, config config.Config) error {
...
+	re := regexp.MustCompile(`bc\/(.*)\/`)
+	matches := re.FindStringSubmatch(config.Endpoints[0])
+	if len(matches) < 1 {
+		return fmt.Errorf("failed to get blockchainIDStr from the clientURI %s", config.Endpoints[0])
+	}
+	// Get the last element in matches
+	blockchainIDStr := matches[len(matches)-1]
+	log.Info("Extracted blockchainIDStr from the clientURI", "blockchainIDStr", blockchainIDStr)
+
+	re = regexp.MustCompile(`127.0.0.1:(.*)/ext/bc`)
```
aaronbuchwald, line 48: can we add the blockchainID as a parameter instead of extracting it from the endpoint here? We should not assume that the blockchainID string will be included in the URI

cmd/simulator/load/loader.go
```diff
@@ -129,5 +152,51 @@ func ExecuteLoader(ctx context.Context, config config.Config) error {
...
+	}
+
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return fmt.Errorf("failed reading response body of metrics: %w", err)
+	}
+
+	bodyString := string(body)
+	re := regexp.MustCompile(fmt.Sprintf(".*avalanche_%s_vm_metervm_build_block_sum.*", blockchainIDStr))
+	matches := re.FindAllStringSubmatch(bodyString, -1)
```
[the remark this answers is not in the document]
anusha-ctrl: yup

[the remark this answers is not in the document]
aaronbuchwald: Is there a better way to do this than creating a regex over the body?

cmd/simulator/load/loader.go
```diff
@@ -129,5 +152,51 @@ func ExecuteLoader(ctx context.Context, config config.Config) error {
...
 	}
 	log.Info("Tx agents completed successfully.")
+
+	logOtherMetrics(blockchainIDStr, endpoint)
+	return nil
+}
+
+func logOtherMetrics(blockchainIDStr string, endpoint string) error {
+	getCallStart := time.Now()
+	metricsAPI := fmt.Sprintf("http://127.0.0.1:%s/ext/metrics", endpoint)
```
aaronbuchwald, line 162: we should not call the port and endpoint.

We should not add any assumptions about the format of the URIs passed in for the load test.

cmd/simulator/load/loader.go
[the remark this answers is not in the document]
anusha-ctrl: thanks

cmd/simulator/load/loader.go
anusha-ctrl: sure

anusha-ctrl: done thanks

cmd/simulator/load/loader.go
anusha-ctrl: What can I assume about the endpoint? I want to be able to use the `/ext/metrics` endpoint. I should be able to use that right?

cmd/simulator/load/loader.go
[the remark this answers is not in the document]
anusha-ctrl: the metrics API just returns a huge block of text so I have to search for exactly what I want

scripts/run_simulator.sh
```diff
@@ -36,7 +36,9 @@ run_simulator() {
         --timeout=30s \
         --workers=1 \
         --max-fee-cap=300 \
-        --max-tip-cap=100
+        --max-tip-cap=100 \
+        --blockchain-id-str=$BLOCKCHAIN_ID
+        --metrics = $METRICS
```
aaronbuchwald, line 41:
```suggestion
        --metrics=$METRICS
```

tests/load/load_test.go
```diff
@@ -33,14 +33,23 @@ var _ = ginkgo.Describe("[Load Simulator]", ginkgo.Ordered, func() {
...
 			rpcEndpoints = append(rpcEndpoints, fmt.Sprintf("%s/ext/bc/%s/rpc", uri, blockchainID))
 		}
 		commaSeparatedRPCEndpoints := strings.Join(rpcEndpoints, ",")
 		err := os.Setenv("RPC_ENDPOINTS", commaSeparatedRPCEndpoints)
 		gomega.Expect(err).Should(gomega.BeNil())
 
+		err = os.Setenv("BLOCKCHAIN_ID", blockchainID.String())
+		gomega.Expect(err).Should(gomega.BeNil())
+
+		err = os.Setenv("METRICS", fmt.Sprintf("%s/ext/metrics", nodeURIs[0]))
```
aaronbuchwald, line 50: nice 🤝

cmd/simulator/config/flags.go
```diff
@@ -36,26 +38,30 @@ var (
...
+	Endpoints       []string      `json:"endpoints"`
+	MaxFeeCap       int64         `json:"max-fee-cap"`
+	MaxTipCap       int64         `json:"max-tip-cap"`
+	Workers         int           `json:"workers"`
+	TxsPerWorker    uint64        `json:"txs-per-worker"`
+	KeyDir          string        `json:"key-dir"`
+	Timeout         time.Duration `json:"timeout"`
+	BatchSize       uint64        `json:"batch-size"`
+	BlockchainIDStr string        `json:"blockchainIDStr"`
+	Metrics         string        `json:"metrics"`
```
aaronbuchwald, line 50: can we change this to `[]string` so that we can gather metrics from multiple nodes when relevant same as hitting multiple rpc endpoints?

Could we also rename to `metrics-endpoints`

cmd/simulator/config/flags.go
```diff
@@ -16,17 +16,19 @@ import (
...
+	MaxFeeCapKey       = "max-fee-cap"
+	MaxTipCapKey       = "max-tip-cap"
+	WorkersKey         = "workers"
+	TxsPerWorkerKey    = "txs-per-worker"
+	KeyDirKey          = "key-dir"
+	VersionKey         = "version"
+	TimeoutKey         = "timeout"
+	BatchSizeKey       = "batch-size"
+	BlockchainIDStrKey = "blockchain-id-str"
+	MetricsKey         = "metrics"
```
aaronbuchwald, line 31: Can we rename to `metrics-endpoints` ?

cmd/simulator/load/loader.go
```diff
@@ -129,5 +135,67 @@ func ExecuteLoader(ctx context.Context, config config.Config) error {
 		return err
 	}
 	log.Info("Tx agents completed successfully.")
+
+	logOtherMetrics(config.Metrics, config.BlockchainIDStr)
 	return nil
 }
+
+func logOtherMetrics(metrics string, blockchainIDStr string) error {
```
aaronbuchwald, line 143: nit: can we rename to `logExtraMetrics`

cmd/simulator/config/flags.go
anusha-ctrl: sure

cmd/simulator/load/loader.go
anusha-ctrl: sure

scripts/run_simulator.sh
anusha-ctrl: thanks

cmd/simulator/load/loader.go
```diff
@@ -129,5 +135,67 @@ func ExecuteLoader(ctx context.Context, config config.Config) error {
...
+	verifyTime, err := findMatchFromString(bodyString, fmt.Sprintf(".*avalanche_%s_vm_metervm_verify_sum.*", blockchainIDStr))
+	if err != nil {
+		log.Info("No verify metrics found from metrics API for blockchainIDStr", "blockchainIDStr", blockchainIDStr, "smetricsAPI", metrics)
+		return nil
+	}
+	log.Info("Sum of time (in ns) of a verify", "time", verifyTime)
+	return nil
+}
+
+func findMatchFromString(fullString string, regex string) (string, error) {
```
aaronbuchwald, line 193: thanks for breaking this out into a function, is there a prometheus library that we could use instead so we don't have to handle regexs here?

cmd/simulator/load/loader.go
```diff
@@ -129,5 +135,67 @@ func ExecuteLoader(ctx context.Context, config config.Config) error {
...
+	}
+
+	bodyString := string(body)
+
+	buildBlockTime, err := findMatchFromString(bodyString, fmt.Sprintf(".*avalanche_%s_vm_metervm_build_block_sum.*", blockchainIDStr))
+	if err != nil {
+		log.Info("No buildBlock metrics found from metrics API for blockchainIDStr", "blockchainIDStr", blockchainIDStr, "metric", metrics)
+		return nil
+	}
+	log.Info("Sum of time (in ns) of a build_block", "time", buildBlockTime)
```
aaronbuchwald, line 175: why do we return early if we're missing this metric instead of moving on to the rest?

Do we expect to ever fail to find this metric? Should we still log everything we can or error if we do encounter an unexpected error?

cmd/simulator/txs/agent.go
```diff
@@ -53,33 +56,70 @@ func (a issueNAgent[T]) Execute(ctx context.Context) error {
...
+	batchI := 1
+
+	// Tracks the total amount of time waiting for issuing and confirming txs
+	var (
+		totalIssuedTime    time.Duration
+		totalConfirmedTime time.Duration
+	)
+
+	defer func() {
+		a.worker.Close(ctx)
```
aaronbuchwald, line 69:
```suggestion
		_ = a.worker.Close(ctx)
```

if we are ok with ignoring the error, we should make it explicit that we're ignoring it, so that it's obvious that it was not mistakenly ignored

cmd/simulator/txs/agent.go
```diff
@@ -53,33 +56,70 @@ func (a issueNAgent[T]) Execute(ctx context.Context) error {
 	}
 
 	txChan := a.sequence.Chan()
+	confirmedCount := 0
+	batchI := 1
```
aaronbuchwald, line 60:
```suggestion
	batchI := 0
```

cmd/simulator/txs/agent.go
anusha-ctrl: thanks

cmd/simulator/txs/agent.go
anusha-ctrl: thanks

anusha-ctrl: thought it would be nicer from a user reading the logs to see it 1 -indexed

cmd/simulator/load/loader.go
anusha-ctrl: None that I could see https://pkg.go.dev/github.com/prometheus/client_golang/prometheus. Seems to be focused on collecting and exposing metrics.

cmd/simulator/config/flags.go
anusha-ctrl: sure

cmd/simulator/load/loader.go
anusha-ctrl: Yah we don't need to return. In fact, I don't think we should err anywhere in this function and we should just log it. Doesn't make sense to fail the simulation if we mess up metrics

cmd/simulator/txs/agent.go
aaronbuchwald: Fair point haha

cmd/simulator/config/flags.go
```diff
@@ -118,4 +124,6 @@ func addSimulatorFlags(fs *pflag.FlagSet) {
 	fs.Duration(TimeoutKey, 5*time.Minute, "Specify the timeout for the simulator to complete (0 indicates no timeout)")
 	fs.String(LogLevelKey, "info", "Specify the log level to use in the simulator")
 	fs.Uint64(BatchSizeKey, 100, "Specify the batchsize for the worker to issue and confirm txs")
+	fs.String(BlockchainIDStrKey, "C", "Specify the blockchainIDStr for the chain you are executing the simulation on. Ex: C, X, P, pUSPHTrvs8ASVfPjrf9xdQ6SwDFhSmCZ3EKuCUMASHky2Wvr")
```
aaronbuchwald, line 127:
```suggestion
	fs.String(BlockchainIDStrKey, "C", "Specify the blockchainIDStr for the chain you are executing the simulation on. Ex: C, X, P, pUSPHTrvs8ASVfPjrf9xdQ6SwDFhSmCZ3EKuCUMASHky2Wvr (full 32 byte hash blockchainID)")
```

tests/load/load_test.go
```diff
@@ -33,15 +33,27 @@ var _ = ginkgo.Describe("[Load Simulator]", ginkgo.Ordered, func() {
 		blockchainID := subnetDetails.BlockchainID
 
 		nodeURIs := subnetDetails.ValidatorURIs
-		rpcEndpoints := make([]string, 0, len(nodeURIs))
+		nodeURIsLen := len(nodeURIs)
+		gomega.Expect(nodeURIsLen).Should(gomega.BeNumerically(">", 0))
```
aaronbuchwald, line 37: Was this an issue testing locally that `nodeURIs` ended up being 0?

tests/load/load_test.go
```diff
@@ -33,15 +33,27 @@ var _ = ginkgo.Describe("[Load Simulator]", ginkgo.Ordered, func() {
...
 
-		log.Info("Sleeping with network running", "rpcEndpoints", commaSeparatedRPCEndpoints)
+		commaSeparatedMetricsEndpoints := strings.Join(metricsEndpoints, ",")
+		err = os.Setenv("METRICS_ENDPOINTS", commaSeparatedMetricsEndpoints)
+		gomega.Expect(err).Should(gomega.BeNil())
+
+		err = os.Setenv("BLOCKCHAIN_ID", blockchainID.String())
+		gomega.Expect(err).Should(gomega.BeNil())
+
+		log.Info("Sleeping with network running", "rpcEndpoints", commaSeparatedRPCEndpoints, "metricsEndpoints", commaSeparatedMetricsEndpoints, "blockchainIDStr", blockchainID.String())
```
aaronbuchwald, line 56:
```suggestion
		log.Info("Running load simulator...", "rpcEndpoints", commaSeparatedRPCEndpoints, "metricsEndpoints", commaSeparatedMetricsEndpoints, "blockchainIDStr", blockchainID.String())
```

cmd/simulator/config/flags.go
anusha-ctrl: thanks

tests/load/load_test.go
anusha-ctrl: i can remove this now. i was accessing an element in a previous revision so i wanted to check if it had more than 0 values.

anusha-ctrl:
image
output on a run

anusha-ctrl:
`./simulator --timeout=90s --workers=1 --max-fee-cap=1000 --max-tip-cap=10 --txs-per-worker=10 --batch-size=3`

tests/load/load_test.go
anusha-ctrl: thanks

anusha-ctrl:
image

cmd/simulator/config/flags.go
```diff
@@ -118,4 +124,6 @@ func addSimulatorFlags(fs *pflag.FlagSet) {
 	fs.Duration(TimeoutKey, 5*time.Minute, "Specify the timeout for the simulator to complete (0 indicates no timeout)")
 	fs.String(LogLevelKey, "info", "Specify the log level to use in the simulator")
 	fs.Uint64(BatchSizeKey, 100, "Specify the batchsize for the worker to issue and confirm txs")
+	fs.String(BlockchainIDStrKey, "C", "Specify the blockchainIDStr for the chain you are executing the simulation on. Ex: C, X, P, pUSPHTrvs8ASVfPjrf9xdQ6SwDFhSmCZ3EKuCUMASHky2Wvr (full 32 byte hash blockchainID)")
```
darioush, line 127:
```suggestion
	fs.String(BlockchainIDStrKey, "C", "Specify the blockchain ID for the chain you are executing the simulation on. Ex: C, X, P, pUSPHTrvs8ASVfPjrf9xdQ6SwDFhSmCZ3EKuCUMASHky2Wvr (full 32 byte hash blockchainID)")
```

cmd/simulator/load/worker.go
```diff
@@ -44,16 +51,34 @@ func NewSingleAddressTxWorker(ctx context.Context, client ethclient.Client, addr
 	return tw
 }
 
-func (tw *singleAddressTxWorker) IssueTx(ctx context.Context, tx *types.Transaction) error {
-	return tw.client.SendTransaction(ctx, tx)
+func (tw *singleAddressTxWorker) IssueTx(ctx context.Context, timeTx txs.TimeTx) error {
+	start := time.Now()
+	err := tw.client.SendTransaction(ctx, timeTx.Tx)
+	issuanceDuration := time.Since(start)
```
darioush, line 57: nit: could we move this after checking err?

cmd/simulator/txs/tx_generator.go
```diff
@@ -7,36 +7,46 @@ import (
...
+var _ TxSequence[TimeTx] = (*txSequence)(nil)
 
 type CreateTx func(key *ecdsa.PrivateKey, nonce uint64) (*types.Transaction, error)
 
+type TimeTx struct {
+	Tx                             *types.Transaction
+	IssuanceDuration               time.Duration
+	ConfirmationDuration           time.Duration
+	IssuanceStart                  time.Time
+	IssuanceToConfirmationDuration time.Duration
```
darioush, line 26: if this is equal to `IssuanceDuration + ConfirmationDuration` doesn't seem like we need to track it explicitly

cmd/simulator/txs/tx_generator.go
```diff
@@ -7,36 +7,46 @@ import (
...
 	"github.com/ava-labs/subnet-evm/ethclient"
 	ethcrypto "github.com/ethereum/go-ethereum/crypto"
 )
 
-var _ TxSequence[*types.Transaction] = (*txSequence)(nil)
+var _ TxSequence[TimeTx] = (*txSequence)(nil)
 
 type CreateTx func(key *ecdsa.PrivateKey, nonce uint64) (*types.Transaction, error)
 
+type TimeTx struct {
```
darioush, line 21: nit: perhaps `TimedTx`?

cmd/simulator/load/worker.go
```diff
@@ -44,16 +51,34 @@ func NewSingleAddressTxWorker(ctx context.Context, client ethclient.Client, addr
...
 
+	start := time.Now()
 	for {
 		// If the is less than what has already been accepted, the transaction is confirmed
 		if txNonce < tw.acceptedNonce {
+			confirmationEnd := time.Now()
+			timeTx.ConfirmationDuration = confirmationEnd.Sub(start)
+			timeTx.IssuanceToConfirmationDuration = confirmationEnd.Sub(timeTx.IssuanceStart)
+			tw.issuanceToConfirmationHistogram = append(tw.issuanceToConfirmationHistogram, timeTx.IssuanceToConfirmationDuration.Seconds())
+			tw.confirmationHistogram = append(tw.confirmationHistogram, timeTx.ConfirmationDuration.Seconds())
```
darioush, line 80: since we confirm txs in batches sequentially, if one tx is taking long to confirm, this will skew the metrics.
not sure if this is okay for our intended use case here.
We could alternatively spawn threads to wait on txs individually or subscribe to a feed.
If we switch to non-sequential confirm time measurement, this part will need synchronization.

cmd/simulator/config/flags.go
```diff
@@ -36,26 +38,30 @@ var (
...
-	BatchSize    uint64        `json:"batch-size"`
+	Endpoints        []string      `json:"endpoints"`
+	MaxFeeCap        int64         `json:"max-fee-cap"`
+	MaxTipCap        int64         `json:"max-tip-cap"`
+	Workers          int           `json:"workers"`
+	TxsPerWorker     uint64        `json:"txs-per-worker"`
+	KeyDir           string        `json:"key-dir"`
+	Timeout          time.Duration `json:"timeout"`
+	BatchSize        uint64        `json:"batch-size"`
+	BlockchainIDStr  string        `json:"blockchainIDStr"`
```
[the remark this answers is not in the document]
anusha-ctrl: yes thanks

go.mod
```diff
@@ -53,6 +53,7 @@ require (
 	github.com/Microsoft/go-winio v0.5.2 // indirect
 	github.com/NYTimes/gziphandler v1.1.1 // indirect
 	github.com/ava-labs/coreth v0.12.2-rc.0 // indirect
+	github.com/aybabtme/uniplot v0.0.0-20151203143629-039c559e5e7e // indirect
```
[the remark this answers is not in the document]
anusha-ctrl: thanks

cmd/simulator/txs/tx_generator.go
anusha-ctrl: thanks

cmd/simulator/txs/tx_generator.go
anusha-ctrl: yeah i can probably remove this and simplify.

cmd/simulator/load/worker.go
anusha-ctrl: I think we want to show all individual tx times. We want to see the ones that took a while too. Maybe we should show this in terms of percentile instead of a histogram.  cc: @aaronbuchwald

aaronbuchwald (commented):
Couple proposed changes:

- We should switch from tracking metrics via `TimedTx` in the worker, to tracking them in the agent. This way we can use the same tracking code based on `txID` regardless of the transaction type as opposed to coupling the confirmation metrics to the worker
- can we remove the `logExtraMetrics` in favor of ingesting these metrics into prometheus
- can we switch from adding a new dependency to create a histogram to creating a prometheus registry and endpoint to get client side load simulator metrics (histogram of confirmation latency)

anusha-ctrl:
So `logExtraMetrics` is reading metrics being exposed by prometheus from the c-chain/ subnet chain. We have no access to those metrics other than the endpoint

cmd/simulator/load/loader.go
```diff
@@ -129,5 +136,10 @@ func ExecuteLoader(ctx context.Context, config config.Config) error {
 		return err
 	}
 	log.Info("Tx agents completed successfully.")
+
+	// Start a prometheus server to expose individual tx metrics
+	http.Handle("/metrics", promhttp.HandlerFor(reg, promhttp.HandlerOpts{Registry: reg}))
+	log.Info("localhost:8082/metrics")
+	http.ListenAndServe(":8082", nil)
```
aaronbuchwald, line 143: If we start this here, doesn't this mean that the metrics will only be exposed after the agents have already completed?

tests/load/load_test.go
```diff
@@ -33,15 +33,17 @@ var _ = ginkgo.Describe("[Load Simulator]", ginkgo.Ordered, func() {
 		blockchainID := subnetDetails.BlockchainID
 
 		nodeURIs := subnetDetails.ValidatorURIs
-		rpcEndpoints := make([]string, 0, len(nodeURIs))
+		nodeURIsLen := len(nodeURIs)
+
+		rpcEndpoints := make([]string, 0, nodeURIsLen)
```
aaronbuchwald, line 38: Can we revert this change? Seems unnecessary

cmd/simulator/metrics/metrics.go
```diff
@@ -0,0 +1,31 @@
...
+// NewMetrics creates and returns a Metrics and registers it with a Collector
+func NewMetrics(reg prometheus.Registerer) *Metrics {
+	m := &Metrics{
+		ConfirmationTxTimes: prometheus.NewSummary(prometheus.SummaryOpts{
+			Name:       "confirmation_tx_times",
+			Help:       "Individual Tx Confirmation Times for a Load Test",
+			Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
+		}),
+		IssuanceTxTimes: prometheus.NewSummary(prometheus.SummaryOpts{
+			Name:       "issuance_tx_times",
```
aaronbuchwald, line 23:
```suggestion
			Name:       "tx_confirmation",
			Help:       "Individual Tx Confirmation Times for a Load Test",
			Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
		}),
		IssuanceTxTimes: prometheus.NewSummary(prometheus.SummaryOpts{
			Name:       "tx_issuance",
```

cmd/simulator/metrics/metrics.go
```diff
@@ -0,0 +1,31 @@
+package metrics
```
aaronbuchwald, line 1: nit: add license

cmd/simulator/metrics/metrics.go
anusha-ctrl: sounds good, but i believe it should have time in there to denote unit

cmd/simulator/load/loader.go
```diff
@@ -115,12 +119,15 @@ func ExecuteLoader(ctx context.Context, config config.Config) error {
...
+	reg := prometheus.NewRegistry()
+	m := metrics.NewMetrics(reg)
+
 	log.Info("Starting tx agents...")
 	eg := errgroup.Group{}
 	for _, agent := range agents {
 		agent := agent
 		eg.Go(func() error {
-			return agent.Execute(ctx)
+			return agent.Execute(ctx, m)
```
aaronbuchwald, line 130: can we add metrics to the constructor of a new agent instead of a parameter to `Execute`

cmd/simulator/metrics/metrics.go
anusha-ctrl: thanks

cmd/simulator/load/loader.go
anusha-ctrl: yup

cmd/simulator/load/funder.go
```diff
@@ -109,7 +111,10 @@ func DistributeFunds(ctx context.Context, client ethclient.Client, keys []*key.K
 	worker := NewSingleAddressTxWorker(ctx, client, maxFundsKey.Address)
 	txFunderAgent := txs.NewIssueNAgent[*types.Transaction](txSequence, worker, numTxs)
 
-	if err := txFunderAgent.Execute(ctx); err != nil {
+	reg := prometheus.NewRegistry()
+	m := metrics.NewMetrics(reg)
+
+	if err := txFunderAgent.Execute(ctx, m); err != nil {
```
aaronbuchwald, line 117: Can we switch to creating a single instance of metrics to be shared throughout the test instead of creating this once for funding and a separate time for the duration of the test? It seems that these metrics get thrown away and never used

anusha-ctrl: sure

tests/load/load_test.go
anusha-ctrl: sure

cmd/simulator/load/loader.go
anusha-ctrl: Yes, that's what we want so we get all the data?

anusha-ctrl: Also listening is a blocking operation so it made sense to put it last

cmd/simulator/load/loader.go
```diff
@@ -129,5 +138,12 @@ func ExecuteLoader(ctx context.Context, config config.Config) error {
 		return err
 	}
 	log.Info("Tx agents completed successfully.")
+
+	if config.ExposeMetrics {
+		// Start a prometheus server to expose individual tx metrics
+		http.Handle("/metrics", promhttp.HandlerFor(reg, promhttp.HandlerOpts{Registry: reg}))
+		log.Info("localhost:8082/metrics")
+		http.ListenAndServe(":8082", nil)
```
aaronbuchwald, line 146: We should handle the error returned here and make sure this handles ctrl c correctly.

Can we update this so that the server runs at the same time as the test so we can ingest metrics while it's running?

Blocking forever to run a metrics server is not great behavior here. Would it make sense to optionally hit this HTTPServer ourselves to get the result rather than running the server and blocking?

cmd/simulator/load/loader.go
```diff
@@ -129,5 +138,12 @@ func ExecuteLoader(ctx context.Context, config config.Config) error {
 		return err
 	}
 	log.Info("Tx agents completed successfully.")
+
+	if config.ExposeMetrics {
+		// Start a prometheus server to expose individual tx metrics
+		http.Handle("/metrics", promhttp.HandlerFor(reg, promhttp.HandlerOpts{Registry: reg}))
+		log.Info("localhost:8082/metrics")
```
aaronbuchwald, line 145: can we provide a more informative message here and make this port a constant

cmd/simulator/txs/agent.go
```diff
@@ -53,33 +59,80 @@ func (a issueNAgent[T]) Execute(ctx context.Context) error {
...
 
+		// Wait for txs in this batch to confirm
+		confirmedStart := time.Now()
 		for i, tx := range txs {
+			confirmedIndividualStart := time.Now()
 			if err := a.worker.ConfirmTx(ctx, tx); err != nil {
 				return fmt.Errorf("failed to await transaction %d: %w", i, err)
 			}
+			confirmationIndividualTime := time.Since(confirmedIndividualStart)
+			m.ConfirmationTxTimes.Observe(confirmationIndividualTime.Seconds())
```
aaronbuchwald, line 119: The confirmation time should be the time from issuing the transaction to when it is confirmed.

The current metric captures the amount of time it takes from starting to try and confirm the transaction (which is after the transaction has already been issued) to when we've finished confirming it.

cmd/simulator/load/loader.go
```diff
@@ -133,17 +141,37 @@ func ExecuteLoader(ctx context.Context, config config.Config) error {
...
+				}
+				log.Info("Received a SIGINT signal: Gracefully shutting down metrics server")
+			}()
+			sigChan := make(chan os.Signal, 1)
+			signal.Notify(sigChan, syscall.SIGINT)
+			<-sigChan
+		}()
+
+		http.Handle("/metrics", promhttp.HandlerFor(reg, promhttp.HandlerOpts{Registry: reg}))
+		log.Info("Metrics Server: localhost:8082/metrics")
```
darioush, line 163: seems this should refer to the new constant

cmd/simulator/load/loader.go
anusha-ctrl: done

cmd/simulator/load/loader.go
anusha-ctrl: fixed

cmd/simulator/load/loader.go
anusha-ctrl: thanks

cmd/simulator/txs/agent.go
anusha-ctrl: fixed

morrisettjohn (approved):
LGTM, I'm not sure if the changes I suggested are critical

cmd/simulator/load/loader.go
```diff
@@ -124,6 +141,31 @@ func ExecuteLoader(ctx context.Context, config config.Config) error {
...
+			signal.Notify(sigChan, syscall.SIGINT)
+			<-sigChan
+		}()
+
+		http.Handle("/metrics", promhttp.HandlerFor(reg, promhttp.HandlerOpts{Registry: reg}))
+		log.Info(fmt.Sprintf("Metrics Server: localhost%s/metrics", MetricsPort))
+		if err := server.ListenAndServe(); !errors.Is(err, http.ErrServerClosed) {
+			log.Error("Metrics server error: %v", err)
+		}
+	}(ctx)
```
aaronbuchwald, line 167: How can we make this simpler?

cmd/simulator/txs/agent.go
```diff
@@ -7,42 +7,53 @@ import (
...
-type TxSequence[T any] interface {
+type TxSequence[T THash] interface {
 	Chan() <-chan T
 }
 
 // Worker defines the interface for issuance and confirmation of transactions.
 // The caller is responsible for calling Close to cleanup resources used by the
 // worker at the end of the simulation.
-type Worker[T any] interface {
+type Worker[T THash] interface {
```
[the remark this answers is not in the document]
aaronbuchwald: This is intentional. This package defines its own dependency that a worker needs to implement in order to be compatible.

See https://en.wikipedia.org/wiki/Dependency_injection

cmd/simulator/load/loader.go
anusha-ctrl: Simplified this by moving this to a new function, improving readability by adding comments,  and moving all var declarations to the beginning of the function. I tried to mimic the avalanchego code [here. ](https://github.com/ava-labs/avalanchego/blob/master/vms/rpcchainvm/vm.go#L35-L70)

aaronbuchwald: Nice, this is 10x clearer

aaronbuchwald (approved):
LGTM

cmd/simulator/load/loader.go
```diff
@@ -141,36 +157,61 @@ func ExecuteLoader(ctx context.Context, config config.Config) error {
...
+		Addr: MetricsPort,
+	}
+
+	// Start up go routine to listen for SIGINT notifications to gracefully shut down server
+	go func() {
+		defer func() {
+			if err := server.Shutdown(ctx); err != nil {
+				log.Error("Metrics server error: %v", err)
+			}
+			log.Info("Received a SIGINT signal: Gracefully shutting down metrics server")
```
darioush, line 184: seems slightly cleaner if this is moved after <-ctx.Done() instead of defer

cmd/simulator/load/loader.go
```diff
@@ -41,6 +43,20 @@ func ExecuteLoader(ctx context.Context, config config.Config) error {
 		defer cancel()
 	}
 
+	// Create buffered sigChan to send SIGINT notifications
```
darioush, line 46:
```suggestion
	// Create buffered sigChan to receive SIGINT notifications
```

cmd/simulator/load/loader.go
```diff
@@ -41,6 +43,20 @@ func ExecuteLoader(ctx context.Context, config config.Config) error {
...
+	sigChan := make(chan os.Signal, 1)
+	signal.Notify(sigChan, syscall.SIGINT)
+
+	// Create context with cancel
+	ctx, cancel := context.WithCancel(ctx)
+
+	go func() {
+		// Blocks until we receive a SIGNINT notification
```
darioush, line 54:
```suggestion
		// Blocks until we receive a SIGINT notification
```

darioush (approved):

cmd/simulator/load/loader.go
```diff
@@ -6,21 +6,35 @@ package load
...
 	"github.com/ethereum/go-ethereum/common"
 	ethcrypto "github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/log"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"golang.org/x/sync/errgroup"
 )
 
+const (
+	MetricsPort = ":8082" // Port for the Prometheus Metrics Server
```
ceyonur, line 35: can we define this as a flag and default to this?

cmd/simulator/load/loader.go
```diff
@@ -124,10 +157,61 @@ func ExecuteLoader(ctx context.Context, config config.Config) error {
...
+			}
+			log.Info("Received a SIGINT signal: Gracefully shutting down metrics server")
+		}()
+
+		// Blocks until signal is received
+		<-ctx.Done()
+	}()
+
+	// Start metrics server
+	http.Handle("/metrics", promhttp.HandlerFor(reg, promhttp.HandlerOpts{Registry: reg}))
```
ceyonur, line 192: can we constantize this `/metrics` endpoint?

cmd/simulator/load/loader.go
```diff
@@ -124,10 +157,61 @@ func ExecuteLoader(ctx context.Context, config config.Config) error {
...
+	log.Info(fmt.Sprintf("Metrics Server: localhost%s/metrics", MetricsPort))
+	if err := server.ListenAndServe(); !errors.Is(err, http.ErrServerClosed) {
+		log.Error("Metrics server error: %v", err)
+	}
+}
+
+func getOutputFromMetricsServer() {
```
ceyonur, line 199: can we return the result instead of printing it here?

cmd/simulator/load/loader.go
```diff
@@ -124,10 +157,61 @@ func ExecuteLoader(ctx context.Context, config config.Config) error {
...
 
+	go startMetricsServer(ctx, reg)
+
 	log.Info("Waiting for tx agents...")
 	if err := eg.Wait(); err != nil {
 		return err
 	}
 	log.Info("Tx agents completed successfully.")
+
+	getOutputFromMetricsServer()
```
ceyonur, line 168: Do we need to call this here?

cmd/simulator/load/loader.go
```diff
@@ -29,6 +43,20 @@ func ExecuteLoader(ctx context.Context, config config.Config) error {
...
+
+	// Create context with cancel
+	ctx, cancel := context.WithCancel(ctx)
+
+	go func() {
+		// Blocks until we receive a SIGNINT notification
+		<-sigChan
+		// Cancel the context and end all shared processes
+		cancel()
+	}()
```
ceyonur, line 58: Seems that this could leak if context was done/cancelled before sigInt. So we should probably return if ctx was done. See: https://gist.github.com/matejb/87064825093c42c1e76e7175665d9a9b

ceyonur (changes requested):

cmd/simulator/load/loader.go
anusha-ctrl: thanks

cmd/simulator/load/loader.go
anusha-ctrl: thanks

cmd/simulator/load/loader.go
anusha-ctrl: sure

cmd/simulator/load/loader.go
ceyonur: I see. in that case can we change the function name to `printOutput...`.  ? 
Thanks!

anusha-ctrl: thanks

anusha-ctrl:
Merging master is making the unit tests and e2e CI tests fail.

aaronbuchwald (approved):
LGTM

PR #708, opened 2023-06-29
Add wrapped validator state for primary to subnet warp messages

aaronbuchwald:
## Why this should be merged

This PR adds a wrapped validator state type to pass in to warp message verification. This reduces the number of signatures required to verify a message from the primary network using the assumptions:

1. A subnet already relies on its own validator set
2. A subnet's validator set is also validating the primary network

## How this works

Wraps `validators.State` with a special case that will return the local snow context's validator set if the primary network validator set is requested.

## How this was tested

na

## How is this documented

comment in the code explaining the rationale

### DIFF
--- a/warp/validators/state.go
+++ b/warp/validators/state.go
@@ -1,0 +1,50 @@
+// (c) 2023, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package validators
+
+import (
+	"context"
+
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/snow"
+	"github.com/ava-labs/avalanchego/snow/validators"
+)
+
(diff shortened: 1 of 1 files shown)


warp/validators/state.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+func (s *State) GetValidatorSet(
+	ctx context.Context,
+	height uint64,
+	subnetID ids.ID,
+) (map[ids.NodeID]*validators.GetValidatorOutput, error) {
+	// If the subnetID is anything other than the Primary Network, this is a direct
+	// passthrough
+	if subnetID != ids.Empty {
```
darioush, line 43: nit: maybe should refer to `constants.PrimaryNetworkID`

darioush (approved):

anusha-ctrl:
lgtm

PR #709, opened 2023-06-29
upload network runner artifact after e2e test run

aaronbuchwald:
## Why this should be merged

This PR uploads the anr data dir after the e2e tests run.

## How this works

Uses upload-artifact GH action to upload relevant data for debugging e2e tests

## How this was tested

CI

## How is this documented

na

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,3 +81,10 @@ jobs:
       - name: Run E2E Tests
         shell: bash
         run: AVALANCHEGO_BUILD_PATH=/tmp/e2e-test/avalanchego DATA_DIR=/tmp/e2e-test/data ./scripts/run_ginkgo.sh
+      - name: Upload Artifact
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -81,3 +81,10 @@ jobs:
       - name: Run E2E Tests
         shell: bash
         run: AVALANCHEGO_BUILD_PATH=/tmp/e2e-test/avalanchego DATA_DIR=/tmp/e2e-test/data ./scripts/run_ginkgo.sh
+      - name: Upload Artifact
+        if: always()
+        uses: actions/upload-artifact@v3
```
anusha-ctrl, line 86: do we always want this step to run even if another step fails?

aaronbuchwald: This is intended to run in the case of failure so that we can view the logs of the ANR from an e2e test failure. It would be great to run this iff `Run E2E Tests` runs, but not sure of a cleaner way to accomplish exactly that.

anusha-ctrl (approved):

PR #710, opened 2023-06-30
Add predicate packing helper

aaronbuchwald:
## Why this should be merged

This PR adds helpers for packing predicate bytes into the access list of a transaction.

## How this works

Packs a predicate of N bytes as:

- Slice of N bytes
- Delimiter byte 0xff
- Append 0s to the nearest multiple of 32 bytes

## How this was tested

unit tests + minimal fuzz test

## How is this documented

Added README

### DIFF
--- a/utils/predicate/predicate_bytes.go
+++ b/utils/predicate/predicate_bytes.go
@@ -1,7 +1,7 @@
 // (c) 2023, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE for licensing terms.
 
-package utils
+package predicateutils
 
 import (
 	"fmt"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,3 +81,10 @@ jobs:
       - name: Run E2E Tests
         shell: bash
         run: AVALANCHEGO_BUILD_PATH=/tmp/e2e-test/avalanchego DATA_DIR=/tmp/e2e-test/data ./scripts/run_ginkgo.sh
+      - name: Upload Artifact
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
+          name: subnet-evm-e2e-logs
+          path: /tmp/network-runner-root-data*/
+          retention-days: 5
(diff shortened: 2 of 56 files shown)


darioush (approved):

utils/predicate/predicate_bytes.go
```diff
@@ -0,0 +1,68 @@
...
+// For messages with a length that does not comply to that, this delimiter is used to
+// append/remove padding.
+var PredicateEndByte = byte(0xff)
+
+// PackPredicate packs [predicate] by delimiting the actual message with [PredicateEndByte]
+// and zero padding to reach a length that is a multiple of 32.
+func PackPredicate(predicate []byte) []byte {
+	predicate = append(predicate, PredicateEndByte)
+	return common.RightPadBytes(predicate, (len(predicate)+31)/32*32)
+}
```
anusha-ctrl, line 24: Won't the final predicate byte length be  (len(predicate) + some multiple of 32), instead of what we want which is some multiple of 32?

anusha-ctrl (approved):
lgtm you got the zero case too

utils/predicate/predicate_bytes.go
```diff
@@ -0,0 +1,68 @@
...
+// which means that its length must be a multiple of 32 (common.HashLength).
+// For messages with a length that does not comply to that, this delimiter is used to
+// append/remove padding.
+var PredicateEndByte = byte(0xff)
+
+// PackPredicate packs [predicate] by delimiting the actual message with [PredicateEndByte]
+// and zero padding to reach a length that is a multiple of 32.
+func PackPredicate(predicate []byte) []byte {
+	predicate = append(predicate, PredicateEndByte)
+	return common.RightPadBytes(predicate, (len(predicate)+31)/32*32)
```
ceyonur, line 23: can we use `common.HashLength` instead of using hardcoded `32`?

ceyonur: also in the rest of the file

ceyonur (approved):
just a nit

utils/predicate/predicate_bytes.go
aaronbuchwald: This is copying the style used here: https://github.com/ethereum/go-ethereum/blob/master/accounts/abi/pack.go#L33. I think it's clearer to use 32 in this case since we're using 31 as well, so we can either have the current code or:

always use constant and relative constant
```
return common.RightPadBytes(predicate, (len(predicate)+(common.HashLength-1))/common.HashLength*common.HashLength)
```

use 31 but the constant for 32
```
return common.RightPadBytes(predicate, (len(predicate)+31)/common.HashLength*common.HashLength)
```

ceyonur: if these are supposed to be same with hash length then I'd say we should (and geth should) use constant. but no pressure.

PR #712, opened 2023-06-30
Ginkgo tests parallel

nytzuga:
## Why this should be merged

This PR runs all the Ginkgo tests in parallel, making the whole testing time twice as fast.

## How this works

Ginkgo support running tests in parallel. This PR introduces a synchronized way to start an Avalanche Go node once, and setup one subnet for each tests to be run, and share those subnets IDs with the other processes.

## How this was tested

We should have a green pipeline.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,7 +68,7 @@ jobs:
       - name: Use Node.js
         uses: actions/setup-node@v3
         with:
-          node-version: "18.x"
+          node-version: "18.15"
       - name: NPM Clean Install
         run: npm ci
         working-directory: ./contracts
--- a/contracts/package.json
+++ b/contracts/package.json
@@ -28,7 +28,7 @@
   },
   "license": "BSD-3-Clause",
   "scripts": {
-    "build": "rm -rf dist/ && tsc -b",
+    "build": "rm -rf dist/ && tsc -b && npx hardhat compile --network local",
     "compile": "npx hardhat compile",
     "console": "npx hardhat console",
     "test": "npx hardhat test",
@@ -41,6 +41,6 @@
   },
   "engines": {
     "npm": ">6.0.0",
-    "node": ">=18.16.0"
+    "node": ">=18.15.0"
   }
 }
--- a/scripts/run_ginkgo.sh
+++ b/scripts/run_ginkgo.sh
@@ -20,12 +20,14 @@ echo "building precompile.test"
 # to install the ginkgo binary (required for test build and run)
 go install -v github.com/onsi/ginkgo/v2/ginkgo@${GINKGO_VERSION}
 
-ACK_GINKGO_RC=true ginkgo build ./tests/precompile ./tests/load
+TEST_SOURCE_ROOT=$(pwd)
+
+ACK_GINKGO_RC=true ginkgo build ./tests/load
 
 # By default, it runs all e2e test cases!
 # Use "--ginkgo.skip" to skip tests.
 # Use "--ginkgo.focus" to select tests.
-./tests/precompile/precompile.test \
+TEST_SOURCE_ROOT="$TEST_SOURCE_ROOT" ginkgo run -procs=5 tests/precompile \
   --ginkgo.vv \
   --ginkgo.label-filter=${GINKGO_LABEL_FILTER:-""}
 
--- a/tests/precompile/solidity/suites.go
+++ b/tests/precompile/solidity/suites.go
@@ -12,7 +12,7 @@ import (
 	ginkgo "github.com/onsi/ginkgo/v2"
 )
 
-var _ = ginkgo.Describe("[Precompiles]", ginkgo.Ordered, func() {
+var _ = ginkgo.Describe("[Precompiles]", func() {
 	// Register the ping test first
 	utils.RegisterPingTest()
 
@@ -22,35 +22,35 @@ var _ = ginkgo.Describe("[Precompiles]", ginkgo.Ordered, func() {
 		ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
 		defer cancel()
 
-		utils.RunDefaultHardhatTests(ctx, "contract_native_minter")
+		utils.RunDefaultHardhatTests(ctx, utils.BlockchainIDs["contract_native_minter"], "contract_native_minter")
 	})
 
 	ginkgo.It("tx allow list", ginkgo.Label("Precompile"), ginkgo.Label("TxAllowList"), func() {
 		ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
 		defer cancel()
 
-		utils.RunDefaultHardhatTests(ctx, "tx_allow_list")
+		utils.RunDefaultHardhatTests(ctx, utils.BlockchainIDs["tx_allow_list"], "tx_allow_list")
 	})
 
 	ginkgo.It("contract deployer allow list", ginkgo.Label("Precompile"), ginkgo.Label("ContractDeployerAllowList"), func() {
 		ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
 		defer cancel()
 
-		utils.RunDefaultHardhatTests(ctx, "contract_deployer_allow_list")
+		utils.RunDefaultHardhatTests(ctx, utils.BlockchainIDs["contract_deployer_allow_list"], "contract_deployer_allow_list")
 	})
 
 	ginkgo.It("fee manager", ginkgo.Label("Precompile"), ginkgo.Label("FeeManager"), func() {
 		ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
 		defer cancel()
 
-		utils.RunDefaultHardhatTests(ctx, "fee_manager")
+		utils.RunDefaultHardhatTests(ctx, utils.BlockchainIDs["fee_manager"], "fee_manager")
 	})
 
 	ginkgo.It("reward manager", ginkgo.Label("Precompile"), ginkgo.Label("RewardManager"), func() {
 		ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
 		defer cancel()
 
-		utils.RunDefaultHardhatTests(ctx, "reward_manager")
+		utils.RunDefaultHardhatTests(ctx, utils.BlockchainIDs["reward_manager"], "reward_manager")
 	})
 
 	// and then runs the hardhat tests for each one without forcing precompile developers to modify this file.
--- a/tests/utils/command.go
+++ b/tests/utils/command.go
@@ -5,8 +5,10 @@ package utils
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 	"time"
 
@@ -58,13 +60,36 @@ func RegisterPingTest() {
 	})
 }
 
-// RegisterNodeRun registers a before suite that starts an AvalancheGo process to use for the e2e tests
-// and an after suite that stops the AvalancheGo process
+// At boot time subnets are created, one for each test suite. This global
+// variable has all the subnets IDs that can be used.
+//
+// One process creates the AvalancheGo node and all the subnets, and these
+// subnets IDs are passed to all other processes and stored in this global
+// variable
+var BlockchainIDs map[string]string
+
+// Timeout to boot the AvalancheGo node
+var bootAvalancheNodeTimeout = 5 * time.Minute
+
+// Timeout for the health API to check the AvalancheGo is ready
+var healthCheckTimeout = 5 * time.Second
+
 func RegisterNodeRun() {
-	// BeforeSuite starts an AvalancheGo process to use for the e2e tests
+	// Keep track of the AvalancheGo external bash script, it is null for most
+	// processes except the first process that starts AvalancheGo
 	var startCmd *cmd.Cmd
-	_ = ginkgo.BeforeSuite(func() {
-		ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+
+	// Our test suite runs in a separated processes, ginkgo hasI
+	// SynchronizedBeforeSuite() which is promised to run once, and its output is
+	// passed over to each worker.
+	//
+	// In here an AvalancheGo node instance is started, and subnets are created for
+	// each test case. Each test case has its own subnet, therefore all tests can
+	// run in parallel without any issue.
+	//
+	// This function also compiles all the solidity contracts
+	var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
+		ctx, cancel := context.WithTimeout(context.Background(), bootAvalancheNodeTimeout)
 		defer cancel()
 
 		wd, err := os.Getwd()
@@ -76,16 +101,34 @@ func RegisterNodeRun() {
 
 		// Assumes that startCmd will launch a node with HTTP Port at [utils.DefaultLocalNodeURI]
 		healthClient := health.NewClient(DefaultLocalNodeURI)
-		healthy, err := health.AwaitReady(ctx, healthClient, 5*time.Second, nil)
+		healthy, err := health.AwaitReady(ctx, healthClient, healthCheckTimeout, nil)
 		gomega.Expect(err).Should(gomega.BeNil())
 		gomega.Expect(healthy).Should(gomega.BeTrue())
 		log.Info("AvalancheGo node is healthy")
+
+		blockchainIds := make(map[string]string)
+		files, err := filepath.Glob("./tests/precompile/genesis/*.json")
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+		for _, file := range files {
+			basename := filepath.Base(file)
+			index := basename[:len(basename)-5]
+			blockchainIds[index] = CreateNewSubnet(ctx, file)
+		}
+
+		blockchainIDsBytes, err := json.Marshal(blockchainIds)
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+		return blockchainIDsBytes
+	}, func(data []byte) {
+		err := json.Unmarshal(data, &BlockchainIDs)
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
 	})
 
-	ginkgo.AfterSuite(func() {
+	// SynchronizedAfterSuite() takes two functions, the first runs after each test suite is done and the second
+	// function is executed once when all the tests are done. This function is used
+	// to gracefully shutdown the AvalancheGo node.
+	var _ = ginkgo.SynchronizedAfterSuite(func() {}, func() {
 		gomega.Expect(startCmd).ShouldNot(gomega.BeNil())
 		gomega.Expect(startCmd.Stop()).Should(gomega.BeNil())
-		// TODO add a new node to bootstrap off of the existing node and ensure it can bootstrap all subnets
-		// created during the test
 	})
 }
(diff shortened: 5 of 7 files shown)


tests/precompile/solidity/suites.go
```diff
@@ -6,13 +6,70 @@ package solidity
...
-var _ = ginkgo.Describe("[Precompiles]", ginkgo.Ordered, func() {
+var subnets map[string]string
+var startCmd *cmd.Cmd
+
+var _ = ginkgo.SynchronizedAfterSuite(func() {}, func() {
+	gomega.Expect(startCmd).ShouldNot(gomega.BeNil())
+	gomega.Expect(startCmd.Stop()).Should(gomega.BeNil())
+})
+
+var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
```
darioush, line 31: could we add a comment explaining why we are using `SynchronizedBeforeSuite` here

tests/precompile/solidity/suites.go
```diff
@@ -6,13 +6,70 @@ package solidity
...
+	wd, err := os.Getwd()
+	gomega.Expect(err).Should(gomega.BeNil())
+	log.Info("Starting AvalancheGo node", "wd", wd)
+	cmd, err := utils.RunCommand("./scripts/run.sh")
+	startCmd = cmd
+	gomega.Expect(err).Should(gomega.BeNil())
+
+	// Assumes that startCmd will launch a node with HTTP Port at [utils.DefaultLocalNodeURI]
+	healthClient := health.NewClient(utils.DefaultLocalNodeURI)
+	healthy, err := health.AwaitReady(ctx, healthClient, 5*time.Second, nil)
```
darioush, line 44: nit: could we add a named constant or local var for the 5*time.Second?

tests/precompile/solidity/suites.go
```diff
@@ -6,13 +6,70 @@ package solidity
...
+	"github.com/ava-labs/avalanchego/api/health"
 	"github.com/ava-labs/subnet-evm/tests/utils"
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/go-cmd/cmd"
 	ginkgo "github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
 )
 
-var _ = ginkgo.Describe("[Precompiles]", ginkgo.Ordered, func() {
+var subnets map[string]string
```
darioush, line 23: could we define both of these in a var block and add a comment describing `subnets`?

darioush (commented):
Could we either update the PR title to not include wip if it's ready or mark as draft if it is wip?
Also would you please update the PR description?

tests/precompile/solidity/suites.go
```diff
@@ -6,13 +6,87 @@ package solidity
...
 	"github.com/ava-labs/subnet-evm/tests/utils"
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/go-cmd/cmd"
 	ginkgo "github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
 )
 
-var _ = ginkgo.Describe("[Precompiles]", ginkgo.Ordered, func() {
+var healthCheckTimeout = 5 * time.Second
+var bootAvalancheNodeTimeout = 5 * time.Minute
```
darioush, line 23: nit: could we make these const and in a block?

tests/precompile/solidity/suites.go
```diff
@@ -6,13 +6,87 @@ package solidity
...
+
+// At boot time subnets are created, one for each test suite. This global
+// variable has all the subnets Ids, that can be used.
```
darioush, line 26:
```suggestion
// variable has all the subnets IDs that can be used.
```

tests/precompile/solidity/suites.go
```diff
@@ -6,13 +6,87 @@ package solidity
...
+//
+// One process creates the AvalanceGo node and all the subnets, and these
```
darioush, line 28:
```suggestion
// One process creates the AvalancheGo node and all the subnets, and these
```

tests/precompile/solidity/suites.go
```diff
@@ -6,13 +6,87 @@ package solidity
...
+// subnets IDs are passed to all other process and stored in this global
```
darioush, line 29:
```suggestion
// subnets IDs are passed to all other processes and stored in this global
```

tests/precompile/solidity/suites.go
```diff
@@ -6,13 +6,87 @@ package solidity
...
+// variable
+var subnetsIds map[string]string
+
+// Keep track of the AvalancheGo external bash script, it is null for most
+// process but the first process that creates the AvalanceGo
```
darioush, line 34:
```suggestion
// Keep track of the AvalancheGo external bash script, it is null for most
// processes except the first process that starts AvalancheGo
```

tests/precompile/solidity/suites.go
```diff
@@ -6,13 +6,87 @@ package solidity
...
+var startCmd *cmd.Cmd
+
+// Out test suite runs in a separated processes, ginkgo has
+// SynchronizedBeforeSuite() which is promised to run once, and its output is
+// passed over to each worker.
+//
+// In here an AvalancheGo node instance is started, and subnets are created for
+// each test case. Each test case has their own subnet, therefore all tests can
+// be run in parallel without any issue.
```
darioush, line 43:
```suggestion
// each test case. Each test case has its own subnet, therefore all tests can
// run in parallel without any issue.
```

tests/precompile/solidity/suites.go
```diff
@@ -6,13 +6,87 @@ package solidity
...
+	subnetsBytes, err := json.Marshal(subnets)
+	gomega.Expect(err).NotTo(gomega.HaveOccurred())
+	utils.CompileContracts()
+	return subnetsBytes
+}, func(data []byte) {
+	err := json.Unmarshal(data, &subnetsIds)
+	gomega.Expect(err).NotTo(gomega.HaveOccurred())
+})
+
+// SynchronizedAfterSuite() runs once each test suite is done and second
```
darioush, line 81:
```suggestion
// SynchronizedAfterSuite() takes two functions, the first runs after each test suite is done and the second
```

tests/precompile/solidity/suites.go
```diff
@@ -6,13 +6,87 @@ package solidity
...
+// function is executed once when all the tests are done. This function is used
+// to shutdown gracefully the AvalancheGo node.
```
darioush, line 83:
```suggestion
// to gracefully shutdown the AvalancheGo node.
```

tests/utils/subnet.go
```diff
@@ -127,3 +124,14 @@ func RunDefaultHardhatTests(ctx context.Context, test string) {
 
 	RunTestCMD(cmd, chainURI)
 }
+
+func CompileContracts() {
```
darioush, line 128: could we move this fn to the package it's used from and unexport it?

darioush (approved):

aaronbuchwald:
> ## How this was tested
> We should have a green pipeline.

Nice

tests/precompile/solidity/suites.go
```diff
@@ -6,13 +6,106 @@ package solidity
...
+	}
+
+	subnetsBytes, err := json.Marshal(subnets)
+	gomega.Expect(err).NotTo(gomega.HaveOccurred())
+	compileContracts()
+	return subnetsBytes
+}, func(data []byte) {
+	err := json.Unmarshal(data, &subnetsIds)
+	gomega.Expect(err).NotTo(gomega.HaveOccurred())
+})
```
aaronbuchwald, line 86: This is really nice. Putting this right here, this code is really clear that the `SynchronizedBeforeSuite` function generates this result and then the next function running in every process will unmarshal that result to ensure that this variable is set correctly in every process ginkgo is running.

tests/utils/command.go
```diff
@@ -57,35 +56,3 @@ func RegisterPingTest() {
 		gomega.Expect(healthy.Healthy).Should(gomega.BeTrue())
 	})
 }
-
-// RegisterNodeRun registers a before suite that starts an AvalancheGo process to use for the e2e tests
```
aaronbuchwald, line 61: We started using this `RegisterNodeRun` function here: https://github.com/ava-labs/precompile-evm/blob/main/tests/precompile/precompile_test.go#L18 to support builders that want to develop their own precompiles without maintaining a full fork of Subnet-EVM.

Could we add back this functionality, so that this PR doesn't break Precompile-EVM and then we can make a separate GitHub issue to leverage the same change to parallelize precompile e2e tests inside of Precompile-EVM as well?

I don't think we need to block this PR to figure out the best way to do that, so we should just make sure this doesn't break Precompile-EVM in the meantime.

aaronbuchwald: cc @ceyonur

tests/precompile/solidity/suites.go
```diff
@@ -6,13 +6,106 @@ package solidity
...
+// variable has all the subnets IDs that can be used.
+//
+// One process creates the AvalancheGo node and all the subnets, and these
+// subnets IDs are passed to all other processes and stored in this global
+// variable
+var subnetsIds map[string]string
+
+// Keep track of the AvalancheGo external bash script, it is null for most
+// processes except the first process that starts AvalancheGo
+var startCmd *cmd.Cmd
```
aaronbuchwald, line 40: Could we put this inside one var block and update `subnetsIds` to `subnetIDs` ?

Unfortunately, we don't have a style guide for this stuff 😅

happy to make a follow on PR if you prefer

tests/precompile/solidity/suites.go
```diff
@@ -6,13 +6,106 @@ package solidity
...
+
+// Out test suite runs in a separated processes, ginkgo has
+// SynchronizedBeforeSuite() which is promised to run once, and its output is
```
aaronbuchwald, line 43: Is this intended to be Our test suite?

aaronbuchwald (changes requested):
Left a couple of optional nits and one requested change to add back the exported function `RegisterNodeRun`

Great work!

tests/utils/command.go
nytzuga: Great feedback and context. I redo the PR without touching this function and the final PR is much smaller.

tests/utils/command.go
```diff
@@ -76,16 +102,47 @@ func RegisterNodeRun() {
...
 		log.Info("AvalancheGo node is healthy")
+
+		subnets := make(map[string]string)
+		files, err := filepath.Glob("./tests/precompile/genesis/*.json")
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+		for _, file := range files {
+			basename := filepath.Base(file)
+			index := basename[:len(basename)-5]
+			subnets[index] = CreateNewSubnet(ctx, file)
```
ceyonur, line 117: can we create those subnets in parallel? i.e in their old places?

tests/utils/command.go
```diff
@@ -76,16 +102,47 @@ func RegisterNodeRun() {
...
+		}
+
+		subnetsBytes, err := json.Marshal(subnets)
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+		compileContracts()
```
ceyonur, line 122: this is too specific for this function. We should not assume subtests will use a contract test.

scripts/run_ginkgo.sh
```diff
@@ -20,12 +20,14 @@ echo "building precompile.test"
...
-ACK_GINKGO_RC=true ginkgo build ./tests/precompile ./tests/load
+TEST_SOURCE_ROOT=$(pwd)
+
+ACK_GINKGO_RC=true ginkgo build ./tests/load
 
 # By default, it runs all e2e test cases!
 # Use "--ginkgo.skip" to skip tests.
 # Use "--ginkgo.focus" to select tests.
-./tests/precompile/precompile.test \
+TEST_SOURCE_ROOT="$TEST_SOURCE_ROOT" ginkgo run -procs=5 tests/precompile \
```
ceyonur, line 30: Why do we need `TEST_SOURCE_ROOT`?

tests/utils/command.go
ceyonur: yea I agree. Optimistically, changes here should not affect current non-parallel tests and they should be able to run fine with changes here. I.e they should be running fine without `ginkgo run -procs=5`. If it's not possible, then we should have different functions for parallel testing, still compatible with sub-functions like running hardhat tests etc.

ceyonur (commented):
I'm really hyped up for those parallel tests 🚀 We should just make sure that we're not breaking our existing tests as Subnet-EVM can be used as a library for e2e tests as well.

tests/utils/command.go
nytzuga: No, because they share the same wallet and if we do that we will have race conditions.

tests/utils/command.go
nytzuga: I'll move this compiling functions elsewhere.

scripts/run_ginkgo.sh
nytzuga: Because the cwd is different if we compile and run the tests from a binary vs invoking them from the cli.

.github/workflows/ci.yml
```diff
@@ -68,7 +68,7 @@ jobs:
       - name: Use Node.js
         uses: actions/setup-node@v3
         with:
-          node-version: "18.x"
+          node-version: "20.x"
```
nytzuga, line 71: I had to upgrade the CI's node version due to this obscure bug that is hard to debug https://github.com/NomicFoundation/hardhat/issues/3877

.github/workflows/ci.yml
```diff
@@ -68,7 +68,7 @@ jobs:
       - name: Use Node.js
         uses: actions/setup-node@v3
         with:
-          node-version: "18.x"
+          node-version: "18.15"
```
nytzuga, line 71: I had to downgrade due to https://github.com/NomicFoundation/hardhat/issues/3877

contracts/package.json
```diff
@@ -41,6 +41,6 @@
   },
   "engines": {
     "npm": ">6.0.0",
-    "node": ">=18.16.0"
+    "node": ">=18.15.0"
```
nytzuga, line 44: I had to downgrade due to https://github.com/NomicFoundation/hardhat/issues/3877

.github/workflows/ci.yml
aaronbuchwald: Nice find

tests/precompile/solidity/suites.go
```diff
@@ -22,35 +22,35 @@ var _ = ginkgo.Describe("[Precompiles]", ginkgo.Ordered, func() {
 		ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
 		defer cancel()
 
-		utils.RunDefaultHardhatTests(ctx, "contract_native_minter")
+		utils.RunDefaultHardhatTests(ctx, utils.SubnetsIDs["contract_native_minter"], "contract_native_minter")
```
aaronbuchwald, line 25: The variable that needs to be passed in here is the `blockchainID` as opposed to the `subnetID`, could we change the name of this variable?

tests/utils/subnet.go
```diff
@@ -110,14 +110,10 @@ func GetDefaultChainURI(blockchainID string) string {
 // 2. Hardhat contract environment is located at ./contracts
 // 3. Hardhat test file is located at ./contracts/test/<test>.ts
 // 4. npx is available in the ./contracts directory
-func RunDefaultHardhatTests(ctx context.Context, test string) {
-	log.Info("Executing HardHat tests on a new blockchain", "test", test)
-
-	genesisFilePath := fmt.Sprintf("./tests/precompile/genesis/%s.json", test)
-
-	blockchainID := CreateNewSubnet(ctx, genesisFilePath)
+func RunDefaultHardhatTests(ctx context.Context, blockchainID string, test string) {
```
aaronbuchwald, line 113: could we update the comment here to remove the assumptions that no longer apply?

aaronbuchwald (approved):
LGTM

contracts/package.json
```diff
@@ -28,7 +28,7 @@
   },
   "license": "BSD-3-Clause",
   "scripts": {
-    "build": "rm -rf dist/ && tsc -b",
+    "build": "rm -rf dist/ && tsc -b && npx hardhat compile --network local",
```
darioush, line 31: curious why this is needed?

tests/utils/command.go
```diff
@@ -58,13 +60,36 @@ func RegisterPingTest() {
...
+// One process creates the AvalancheGo node and all the subnets, and these
+// subnets IDs are passed to all other processes and stored in this global
+// variable
+var BlockchainIDs map[string]string
+
+// Timeout to boot the AvalancheGo node
+var bootAvalancheNodeTimeout = 5 * time.Minute
+
+// Timeout for the health API to check the AvalancheGo is ready
+var healthCheckTimeout = 5 * time.Second
```
darioush, line 75: style nit: we should define top-level vars in a block where possible, also seems the latter 2 could be constants

tests/utils/command.go
```diff
@@ -58,13 +60,36 @@ func RegisterPingTest() {
...
+
 func RegisterNodeRun() {
-	// BeforeSuite starts an AvalancheGo process to use for the e2e tests
+	// Keep track of the AvalancheGo external bash script, it is null for most
+	// processes except the first process that starts AvalancheGo
 	var startCmd *cmd.Cmd
-	_ = ginkgo.BeforeSuite(func() {
-		ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+
+	// Our test suite runs in a separated processes, ginkgo hasI
```
darioush, line 82:
```suggestion
	// Our test suite runs in separate processes, ginkgo has
```

tests/utils/command.go
```diff
@@ -58,13 +60,36 @@ func RegisterPingTest() {
...
+	// SynchronizedBeforeSuite() which is promised to run once, and its output is
```
darioush, line 83:
```suggestion
	// SynchronizedBeforeSuite() which is run once, and its return value is
```

tests/utils/command.go
```diff
@@ -58,13 +60,36 @@ func RegisterPingTest() {
...
+	// passed over to each worker.
+	//
+	// In here an AvalancheGo node instance is started, and subnets are created for
```
darioush, line 86:
```suggestion
	// Here an AvalancheGo node is started, and subnets are created for
```

tests/utils/subnet.go
```diff
@@ -104,20 +104,15 @@ func GetDefaultChainURI(blockchainID string) string {
...
-
-	blockchainID := CreateNewSubnet(ctx, genesisFilePath)
+// 1. Hardhat contract environment is located at ./contracts
+// 2. Hardhat test file is located at ./contracts/test/<test>.ts
+// 3. npx is available in the ./contracts directory
+func RunDefaultHardhatTests(ctx context.Context, blockchainID string, test string) {
 	chainURI := GetDefaultChainURI(blockchainID)
-	log.Info("Created subnet successfully", "ChainURI", chainURI)
+	log.Info("Executing HardHat tests on a new blockchain", "blockchainID", blockchainID, "test", test)
+	log.Info("Using subnet", "ChainURI", chainURI)
```
darioush, line 115: could we merge these 2 logs?

contracts/package.json
nytzuga: I was debugging an issue where the hardhat compile would just give up with no error, so I moved the compilation earlier. I thought it was because we would using the compiler in parallel.

This was the issue though https://github.com/NomicFoundation/hardhat/issues/3877

PR #713, opened 2023-06-30
Add support for predicate gas

aaronbuchwald:
## Why this should be merged

This PR adds support for predicate gas.

## How this works

Enables precompiles to create a predicate that charges gas within `IntrinsicGas(...)`.

## How this was tested

CI

## How is this documented

na

### DIFF
--- a/core/predicate_check.go
+++ b/core/predicate_check.go
@@ -15,6 +15,13 @@ import (
 
 // CheckPredicates checks that all precompile predicates are satisfied within the current [predicateContext] for [tx]
 func CheckPredicates(rules params.Rules, predicateContext *precompileconfig.ProposerPredicateContext, tx *types.Transaction) error {
+	intrinsicGas, err := IntrinsicGas(tx.Data(), tx.AccessList(), tx.To() == nil, rules)
+	if err != nil {
+		return err
+	}
+	if tx.Gas() < intrinsicGas {
+		return fmt.Errorf("insufficient gas for predicate verification (%d) < intrinsic gas (%d)", tx.Gas(), intrinsicGas)
+	}
 	if err := checkPrecompilePredicates(rules, &predicateContext.PrecompilePredicateContext, tx); err != nil {
 		return err
 	}
--- a/core/predicate_check.go
+++ b/core/predicate_check.go
@@ -15,6 +15,8 @@ import (
 
 // CheckPredicates checks that all precompile predicates are satisfied within the current [predicateContext] for [tx]
 func CheckPredicates(rules params.Rules, predicateContext *precompileconfig.ProposerPredicateContext, tx *types.Transaction) error {
+	// Check that the transaction can cover its IntrinsicGas (including the gas required by the predicate) before
+	// verifying the predicate.
 	intrinsicGas, err := IntrinsicGas(tx.Data(), tx.AccessList(), tx.To() == nil, rules)
 	if err != nil {
 		return err
--- a/core/predicate_check_test.go
+++ b/core/predicate_check_test.go
@@ -22,19 +22,35 @@ var (
 )
 
 type mockPredicater struct {
-	predicateFunc func(*precompileconfig.PrecompilePredicateContext, []byte) error
+	predicateFunc    func(*precompileconfig.PrecompilePredicateContext, []byte) error
+	predicateGasFunc func([]byte) (uint64, error)
 }
 
 func (m *mockPredicater) VerifyPredicate(predicateContext *precompileconfig.PrecompilePredicateContext, b []byte) error {
 	return m.predicateFunc(predicateContext, b)
 }
 
+func (m *mockPredicater) PredicateGas(b []byte) (uint64, error) {
+	if m.predicateGasFunc == nil {
+		return 0, nil
+	}
+	return m.predicateGasFunc(b)
+}
+
 type mockProposerPredicater struct {
-	predicateFunc func(*precompileconfig.ProposerPredicateContext, []byte) error
+	predicateFunc    func(*precompileconfig.ProposerPredicateContext, []byte) error
+	predicateGasFunc func([]byte) (uint64, error)
 }
 
 func (m *mockProposerPredicater) VerifyPredicate(predicateContext *precompileconfig.ProposerPredicateContext, b []byte) error {
 	return m.predicateFunc(predicateContext, b)
+}
+
+func (m *mockProposerPredicater) PredicateGas(b []byte) (uint64, error) {
+	if m.predicateGasFunc == nil {
+		return 0, nil
+	}
+	return m.predicateGasFunc(b)
 }
 
 type predicateCheckTest struct {
--- a/core/predicate_check_test.go
+++ b/core/predicate_check_test.go
@@ -58,6 +58,7 @@ type predicateCheckTest struct {
 	predicater            precompileconfig.PrecompilePredicater
 	proposerPredicater    precompileconfig.ProposerPredicater
 	accessList            types.AccessList
+	gas                   uint64
 	emptyProposerBlockCtx bool
 	expectedErr           error
 }
@@ -65,9 +66,11 @@ type predicateCheckTest struct {
 func TestCheckPredicate(t *testing.T) {
 	for name, test := range map[string]predicateCheckTest{
 		"no predicates, no access list passes": {
+			gas:         53000,
 			expectedErr: nil,
 		},
 		"no predicates, with access list passes": {
+			gas: 57300,
 			accessList: types.AccessList([]types.AccessTuple{
 				{
 					Address: common.HexToAddress("0x8db97C7cEcE249c2b98bDC0226Cc4C2A57BF52FC"),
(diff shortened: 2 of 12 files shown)


precompile/testutils/test_predicate.go
```diff
@@ -0,0 +1,87 @@
...
+	var (
+		gas                  uint64
+		gasErr, predicateErr error
+		predicate            = test.Config.(precompileconfig.ProposerPredicater)
+	)
+
+	gas, gasErr = predicate.PredicateGas(test.StorageSlots)
+	if test.GasErr != nil {
+		require.ErrorIs(t, gasErr, test.GasErr)
+		return // If PredicateGas returns an error, the predicate is invalidated and no further checks are needed
```
darioush, line 35: nit: what does it mean for a predicate to be invalidated?

precompile/testutils/test_predicate.go
```diff
@@ -0,0 +1,87 @@
...
+
+import (
+	"testing"
+	"time"
+
+	"github.com/ava-labs/subnet-evm/precompile/precompileconfig"
+	"github.com/stretchr/testify/require"
+)
+
+type PredicateTest struct {
```
darioush, line 14: could we add some comment about how this struct is intended for use?

core/state_transition.go
```diff
@@ -163,12 +165,64 @@ func IntrinsicGas(data []byte, accessList types.AccessList, isContractCreation b
...
+		return gas, nil
 	}
+
+	for i, accessTuple := range accessList {
+		address := accessTuple.Address
+		if !rules.PredicateExists(address) {
+			totalGas, overflow := commonMath.SafeAdd(gas, params.TxAccessListAddressGas+uint64(len(accessTuple.StorageKeys))*params.TxAccessListStorageKeyGas)
+			if overflow {
+				return 0, fmt.Errorf("overflow adding storage slot gas at index %d", i)
+			}
```
darioush, line 196: technically this does represent a behavior change. Since it's enabled via a network upgrade it should be fine, but still perhaps we should retain old behavior? (is it even possible to overflow here?)

core/state_transition.go
```diff
@@ -163,12 +165,64 @@ func IntrinsicGas(data []byte, accessList types.AccessList, isContractCreation b
...
 	return gas, nil
 }
 
+func ApplyPredicateGas(rules params.Rules, accessTuple types.AccessTuple) (uint64, error) {
+	predicate, ok := rules.PredicatePrecompiles[accessTuple.Address]
+	if ok {
+		return predicate.PredicateGas(utils.HashSliceToBytes(accessTuple.StorageKeys))
+	}
+	proposerPredicates, ok := rules.ProposerPredicates[accessTuple.Address]
+	if !ok {
```
darioush, line 220: nit: could we re-write as above for symmetry?

core/state_transition.go
```diff
@@ -163,12 +165,64 @@ func IntrinsicGas(data []byte, accessList types.AccessList, isContractCreation b
...
+			return 0, fmt.Errorf("overflow adding access list gas")
+		}
+		gas = totalGas
+	}
+
+	return gas, nil
+}
+
+func AccessListGas(rules params.Rules, accessList types.AccessList) (uint64, error) {
+	var gas uint64
```
anusha-ctrl, line 183: does this method need to be public

core/state_transition.go
```diff
@@ -163,12 +165,64 @@ func IntrinsicGas(data []byte, accessList types.AccessList, isContractCreation b
...
 		gas += uint64(len(accessList)) * params.TxAccessListAddressGas
 		gas += uint64(accessList.StorageKeys()) * params.TxAccessListStorageKeyGas
+		return gas, nil
 	}
+
+	for i, accessTuple := range accessList {
+		address := accessTuple.Address
+		if !rules.PredicateExists(address) {
+			totalGas, overflow := commonMath.SafeAdd(gas, params.TxAccessListAddressGas+uint64(len(accessTuple.StorageKeys))*params.TxAccessListStorageKeyGas)
+			if overflow {
```
anusha-ctrl, line 194: nit: can this be on its own line? its a bit hard to read `params.TxAccessListAddressGas+uint64(len(accessTuple.StorageKeys))*params.TxAccessListStorageKeyGas)`

precompile/testutils/test_predicate.go
```diff
@@ -0,0 +1,87 @@
...
+	ProposerPredicateContext *precompileconfig.ProposerPredicateContext
+
+	StorageSlots         []byte
+	Gas                  uint64
+	GasErr, PredicateErr error
+}
+
+func (test PredicateTest) Run(t testing.TB) {
+	t.Helper()
+	var (
```
anusha-ctrl, line 26: can we add `require := require.New(t)` and reduce the number of arguments for all require

core/state_transition.go
anusha-ctrl: also can this addition ever overflow ^

precompile/testutils/test_predicate.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+	b.StopTimer()
+	elapsed := uint64(time.Since(start))
+	if elapsed < 1 {
+		elapsed = 1
+	}
+
+	gasUsed := test.Gas * uint64(b.N)
+	b.ReportMetric(float64(test.Gas), "gas/op")
+	// Keep it as uint64, multiply 100 to get two digit float later
```
anusha-ctrl, line 74: this seems like the totalGas not "gas/op".

precompile/testutils/test_predicate.go
anusha-ctrl, line 35: confused as to where the `test.GasErr` is being set

anusha-ctrl: same for `test.Gas`

precompile/testutils/test_predicate.go
```diff
@@ -0,0 +1,87 @@
...
+
+	predicateErr = predicate.VerifyPredicate(test.ProposerPredicateContext, test.StorageSlots)
+	if test.PredicateErr == nil {
+		require.NoError(t, predicateErr)
+	} else {
+		require.ErrorIs(t, predicateErr, test.PredicateErr)
+	}
+}
+
+func RunPredicateTests(t *testing.T, predicateTests map[string]PredicateTest) {
```
anusha-ctrl, line 49: will call this function with actual `predicateTests`? or in a later pr

core/state_transition.go
aaronbuchwald: nope good callout

precompile/testutils/test_predicate.go
aaronbuchwald: Yup

precompile/testutils/test_predicate.go
aaronbuchwald: yup

precompile/testutils/test_predicate.go
aaronbuchwald: just that it fails. I think I may have wrote invalidated because it fails for a reason other than the verify predicate function returning an error.

precompile/testutils/test_predicate.go
aaronbuchwald: later PR

core/state_transition.go
aaronbuchwald: Removing the `s`

core/state_transition.go
aaronbuchwald: With the current code and gas charges for specifying state storage slot tuples, it would take multiple TBs of data in order for an access list to overflow uint64.

This only changes the behavior when a predicate could return an arbitrary amount of gas, such that an overflow becomes realistic here.

I'm going to change this to multiple lines and keep the safemath when adding to the total since a precompile predicate gas implementation can return a `uint64` that could result in an overflow, so we should check that case here.

core/state_transition.go
aaronbuchwald: Good call, moving to a separate line and adding a comment

core/state_transition.go
```diff
@@ -163,12 +165,69 @@ func IntrinsicGas(data []byte, accessList types.AccessList, isContractCreation b
...
+			}
+			gas = totalGas
+		}
+	}
+
 	return gas, nil
 }
 
+func ApplyPredicateGas(rules params.Rules, accessTuple types.AccessTuple) (uint64, error) {
+	predicate, ok := rules.PredicatePrecompiles[accessTuple.Address]
```
anusha-ctrl, line 220: does this function also need to be public?

core/predicate_check.go
```diff
@@ -15,6 +15,13 @@ import (
 
 // CheckPredicates checks that all precompile predicates are satisfied within the current [predicateContext] for [tx]
 func CheckPredicates(rules params.Rules, predicateContext *precompileconfig.ProposerPredicateContext, tx *types.Transaction) error {
+	intrinsicGas, err := IntrinsicGas(tx.Data(), tx.AccessList(), tx.To() == nil, rules)
+	if err != nil {
+		return err
+	}
+	if tx.Gas() < intrinsicGas {
+		return fmt.Errorf("insufficient gas for predicate verification (%d) < intrinsic gas (%d)", tx.Gas(), intrinsicGas)
+	}
```
anusha-ctrl, line 24: do we test the insufficient gas case below?

core/predicate_check_test.go
```diff
@@ -80,23 +83,53 @@ func TestCheckPredicate(t *testing.T) {
...
-				} else {
-					return fmt.Errorf("unexpected bytes: 0x%x", b)
-				}
-			}},
+			gas:     53000,
+			predicater: &mockPredicater{
+				predicateFunc: func(_ *precompileconfig.PrecompilePredicateContext, b []byte) error {
+					if bytes.Equal(b, common.Hash{1}.Bytes()) {
+						return nil
+					} else {
```
darioush, line 103: style nit: we can drop one branch here in favor of early return (+ other occurrences in this file)

core/state_transition.go
```diff
@@ -190,7 +190,12 @@ func AccessListGas(rules params.Rules, accessList types.AccessList) (uint64, err
 	for i, accessTuple := range accessList {
 		address := accessTuple.Address
 		if !rules.PredicateExists(address) {
-			totalGas, overflow := commonMath.SafeAdd(gas, params.TxAccessListAddressGas+uint64(len(accessTuple.StorageKeys))*params.TxAccessListStorageKeyGas)
+			// Previous access list gas calculation does not use safemath because an overflow would not be possible with
+			// the size of access lists that could be included in a block and standard access list gas costs.
+			// Therefore, we only check for overflow when adding to [totalGas], which could include the sum of values
+			// returned by a predicate.
```
darioush, line 196: Could we clarify here or elsewhere what happens if we overflow?

darioush (approved):

anusha-ctrl (approved):

core/predicate_check.go
```diff
@@ -15,6 +15,13 @@ import (
 
 // CheckPredicates checks that all precompile predicates are satisfied within the current [predicateContext] for [tx]
 func CheckPredicates(rules params.Rules, predicateContext *precompileconfig.ProposerPredicateContext, tx *types.Transaction) error {
+	intrinsicGas, err := IntrinsicGas(tx.Data(), tx.AccessList(), tx.To() == nil, rules)
```
ceyonur, line 18: Why do we check `IntrinsicGas` here in addition to tx pool and state transition? Do we want to return an early err?

core/state_transition.go
aaronbuchwald: Nope good callout

core/predicate_check_test.go
aaronbuchwald: Yup good catch

core/state_transition.go
aaronbuchwald: I don't think this is necessary. If you look at `IntrinsicGas` it can already return an overflow error, which invalidates the transaction.

I'm going to change the returned error to be the same as the overflow error used in `IntrinsicGas`

core/predicate_check.go
aaronbuchwald: Yes, I'll add a comment to that effect.

Verifying the predicate costs some amount of computation, so we want to make sure that the transaction is capable of paying for that computation before performing it.

PR #715, opened 2023-07-04
fix goroutine leaks in core tests

darioush:
## Why this should be merged
Tests should not leak goroutines, this slows down the environment for other tests.

## How this works
calls Stop methods or uses test stubs which do not launch goroutines instead of implementation.

## How this was tested
UTs + goleak

## How is this documented
N/A


core/tx_pool_test.go
```diff
@@ -136,7 +136,7 @@ func (bc *testBlockChain) GetFeeConfigAt(parent *types.Header) (commontype.FeeCo
 }
 
 func (bc *testBlockChain) SenderCacher() *TxSenderCacher {
-	return newTxSenderCacher(1)
+	return &TxSenderCacher{}
```
darioush, line 139: This avoids starting goroutines.
Not a very nice way to shut down this without lots of textual modifications.. So I prefer this.

core/blockchain_snapshot_test.go
```diff
@@ -360,16 +361,18 @@ func (snaptest *wipeCrashSnapshotTest) test(t *testing.T) {
...
 		t.Fatalf("Failed to recreate chain: %v", err)
 	}
+	defer newchain.Stop()
 	// Simulate the blockchain crash.
 
 	newchain, err = NewBlockChain(snaptest.db, DefaultCacheConfig, params.TestChainConfig, snaptest.engine, vm.Config{}, snaptest.lastAcceptedHash)
 	if err != nil {
 		t.Fatalf("Failed to recreate chain: %v", err)
 	}
+	defer newchain.Stop()
```
darioush, line 375: looks like this is already in upstream.

core/blockchain_snapshot_test.go
```diff
@@ -360,16 +361,18 @@ func (snaptest *wipeCrashSnapshotTest) test(t *testing.T) {
 		Pruning:        true,
 		CommitInterval: 4096,
 	}
-	_, err = NewBlockChain(snaptest.db, config, params.TestChainConfig, snaptest.engine, vm.Config{}, snaptest.lastAcceptedHash)
+	newchain, err = NewBlockChain(snaptest.db, config, params.TestChainConfig, snaptest.engine, vm.Config{}, snaptest.lastAcceptedHash)
 	if err != nil {
 		t.Fatalf("Failed to recreate chain: %v", err)
 	}
+	defer newchain.Stop()
```
darioush, line 368: looks like this is already in upstream.

core/blockchain_test.go
```diff
@@ -471,6 +473,7 @@ func testRepopulateMissingTriesParallel(t *testing.T, parallelism int) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	blockchain.Stop()
```
darioush, line 476: is missing from subnet-evm

core/blockchain_snapshot_test.go
```diff
@@ -245,6 +245,7 @@ func (snaptest *crashSnapshotTest) test(t *testing.T) {
 	// Pull the plug on the database, simulating a hard crash
 	db := chain.db
 	db.Close()
+	chain.Stop()
```
darioush, line 248: done with .stopWithoutSaving in upstream

core/blockchain_test.go
```diff
@@ -698,6 +701,8 @@ func TestCanonicalHashMarker(t *testing.T) {
 		if err != nil {
 			t.Fatalf("failed to create tester chain: %v", err)
 		}
+		defer chain.Stop()
```
darioush, line 704: coreth has a `chain.Stop()` at the end of this test..

core/test_blockchain.go
```diff
@@ -130,6 +130,7 @@ func checkBlockChainState(
 	if err != nil {
 		t.Fatalf("Failed to create new blockchain instance: %s", err)
 	}
+	defer newBlockChain.Stop()
```
darioush, line 133: already in coreth

darioush:
waiting on this till https://github.com/ava-labs/coreth/pull/273/files is merged

PR #586, opened 2023-03-21
Warp e2e

aaronbuchwald:
## Why this should be merged

This PR implements the warp precompile. Addresses https://github.com/ava-labs/subnet-evm/issues/440

## How this works

This PR adds a warp precompile to integrate warp messaging into Subnet-EVM. This includes a Solidity interface for contracts that want to send a message via warp here.

## How this was tested

This code has been tested through precompile unit tests, an e2e VM level test for the VM's handling of verification and acceptance of warp interactions, and an e2e ginkgo test that sends a warp message from subnet A to subnet B where the validator sets of subnet A and B are disjoint.

## How is this documented

https://github.com/ava-labs/subnet-evm/blob/warp-e2e/precompile/contracts/warp/README.md

## TODO

- [x] Split out network runner changes + minor changes to load test into https://github.com/ava-labs/subnet-evm/pull/579
- [x] Move cleanup into a separate PR: (Receipt handling fix: https://github.com/ava-labs/subnet-evm/pull/586#discussion_r1144821965, BLS key setup for unit tests: https://github.com/ava-labs/subnet-evm/pull/586#discussion_r1144822689)
- [x] Clean up parsing of AddressedPayload within an UnsignedMessage
- [x] Improve helpers for unpacking predicate bytes
- [x] Clean up comments and add README to warp precompile package
- [x] Add pre-verification gas charges to `IntrinsicGas` to account for work done in predicate verification
- [x] Add support for receiving messages from Primary Network
- [x] Add synchronous signature aggregator API call
- [x] Benchmark precompile including parsing, P-Chain validator set lookup, and signature verification to calculate gas costs
- [x] Add unit test for VerifyWithContext on an already validated block with a new context that is valid/invalid
- [x] Add warp precompile contract execution unit tests
- [x] Add warp precompile predicate unit tests
- [ ] Stress testing
- [ ] Cleanup and review

Issue #440, opened 2023-01-18
Avalanche Warp Messaging


aaronbuchwald:
# Documentation

https://github.com/ava-labs/subnet-evm/blob/warp-e2e/precompile/contracts/warp/README.md

# Tasks

- #444
- #445
- #446
- #442
- #443
- #447
- #471
- #448
- #449
- #698

michaelkaplan13:
Calling out a dev dependency: We'll need the precompile solidity interface in order to filter for messages to be inserted into the messages database. 

May want to consider moving the precompile solidity interface up to the top.

aaronbuchwald:
> Calling out a dev dependency: We'll need the precompile solidity interface in order to filter for messages to be inserted into the messages database.
> 
> May want to consider moving the precompile solidity interface up to the top.

The messages database in the first ticket should be the database of messages the validator is willing to sign, so there won't be a need to filter that as any message emitted by the precompile should go in.

Any message emitted by the Warp precompile, should be added to the database when it gets accepted, so it should just need the following interface:

```
type WarpBackend interface {
	AddMessage(ctx context.Context, unsignedMessage *teleporter.UnsignedMessage) error
	GetSignature(ctx context.Context, messageHash ids.ID) ([]byte, error)
}
```

aaronbuchwald:
@minghinmatthewlam to add a separate ticket to clean up the warp messages database and add to this ticket.

minghinmatthewlam:
Added #471 into list of issues

PR #718, opened 2023-07-10
Add warp contract implementation

aaronbuchwald:
## Why this should be merged

This PR implements the warp precompile with experimental support. Addresses https://github.com/ava-labs/subnet-evm/issues/440

This PR replaces https://github.com/ava-labs/subnet-evm/pull/586

Future TODOs before production readiness are:

1. Compare predicate approach with stapling warp verification results to the block (approach taken by HyperSDK and mentioned in the README) https://github.com/ava-labs/subnet-evm/issues/735
2. Add support for signing block hashes directly https://github.com/ava-labs/subnet-evm/issues/734
3. Finalize benchmarks and gas costs after P-Chain DB improvements

## How this works

This PR adds a warp precompile to integrate warp messaging into Subnet-EVM. This includes a Solidity interface for contracts that want to send a message via warp here.

## How this was tested

This code has been tested through precompile unit tests, an e2e VM level test for the VM's handling of verification and acceptance of warp interactions, and an e2e ginkgo test that sends a warp message from subnet A to subnet B where the validator sets of subnet A and B are disjoint.

## How is this documented

https://github.com/ava-labs/subnet-evm/blob/warp-e2e/precompile/contracts/warp/README.md


x/warp/contract.go
```diff
@@ -0,0 +1,277 @@
...
+	packedOutput, err := PackGetBlockchainIDOutput(accessibleState.GetSnowContext().ChainID)
+	if err != nil {
+		return nil, remainingGas, err
+	}
+
+	// Return the packed output and the remaining gas
+	return packedOutput, remainingGas, nil
+}
+
+// PackGetVerifiedWarpMessage packs the include selector (first 4 func signature bytes).
```
darioush, line 101: could we fix this comment?

x/warp/contract.go
```diff
@@ -0,0 +1,277 @@
...
+// getVerifiedWarpMessage retrieves the pre-verified warp message from the predicate storage slots and returns
+// the expected ABI encoding of the message to the caller.
+func getVerifiedWarpMessage(accessibleState contract.AccessibleState, caller common.Address, addr common.Address, _ []byte, suppliedGas uint64, readOnly bool) (ret []byte, remainingGas uint64, err error) {
+	remainingGas = suppliedGas
+	// XXX Note: there is no base cost for retrieving a verified warp message. Instead, we charge for each piece of gas,
+	// prior to each execution step.
+	// Ignore input since there are no arguments
+	predicateBytes, exists := accessibleState.GetStateDB().GetPredicateStorageSlots(ContractAddress)
+	// If there is no such value, return false to the caller.
+	// Note: decoding errors will return an error instead.
```
darioush, line 125:
```suggestion
	// Note: encoding errors will return an error instead.
```

x/warp/contract.go
```diff
@@ -0,0 +1,277 @@
...
+	if !exists {
+		packedOutput, err := PackGetVerifiedWarpMessageOutput(GetVerifiedWarpMessageOutput{
+			Exists: false,
+		})
+		if err != nil {
+			return nil, remainingGas, err
```
darioush, line 131: it seems we charge no gas on this path? is this ok?

plugin/evm/vm_warp_test.go
```diff
@@ -0,0 +1,338 @@
...
+	}
+
+	// Verify that the constructed block contains the expected log with an unsigned warp message in the log data
+	ethBlock1 := blk.(*chain.BlockWrapper).Block.(*Block).ethBlock
+	require.Len(t, ethBlock1.Transactions(), 1)
+	receipts := rawdb.ReadReceipts(vm.chaindb, ethBlock1.Hash(), ethBlock1.NumberU64(), vm.chainConfig)
+	require.NotNil(t, receipts)
+	require.Len(t, receipts, 1)
+
+	logData := receipts[0].Logs[0].Data
```
darioush, line 104: can we also verify the topics are as expected?

tests/warp/warp_test.go
```diff
@@ -0,0 +1,384 @@
...
+	ginkgo.RunSpecs(t, "subnet-evm warp e2e test")
+}
+
+func toWebsocketURI(uri string, blockchainID string) string {
+	return fmt.Sprintf("ws://%s/ext/bc/%s/ws", strings.TrimPrefix(uri, "http://"), blockchainID)
+}
+
+// BeforeSuite starts the default network and adds 10 new nodes as validators with BLS keys
+// registered on the P-Chain.
+// Adds two disjoin sets of 5 of the new validator nodes to validate two new subnets with a
```
darioush, line 55:
```suggestion
// Adds two disjoint sets of 5 of the new validator nodes to validate two new subnets with a
```

plugin/evm/vm_warp_test.go
```diff
@@ -0,0 +1,338 @@
...
+	"github.com/ava-labs/subnet-evm/x/warp"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/stretchr/testify/require"
+)
+
+func TestSendWarpMessage(t *testing.T) {
+	genesis := &core.Genesis{}
+	if err := genesis.UnmarshalJSON([]byte(genesisJSONSubnetEVM)); err != nil {
+		t.Fatal(err)
```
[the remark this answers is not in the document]
aaronbuchwald: Good call, updating

plugin/evm/vm_warp_test.go
```diff
@@ -0,0 +1,338 @@
...
+
+	if status := blk.Status(); status != choices.Processing {
+		t.Fatalf("Expected status of built block to be %s, but found %s", choices.Processing, status)
+	}
+
+	// Verify that the constructed block contains the expected log with an unsigned warp message in the log data
+	ethBlock1 := blk.(*chain.BlockWrapper).Block.(*Block).ethBlock
+	require.Len(t, ethBlock1.Transactions(), 1)
+	receipts := rawdb.ReadReceipts(vm.chaindb, ethBlock1.Hash(), ethBlock1.NumberU64(), vm.chainConfig)
+	require.NotNil(t, receipts)
```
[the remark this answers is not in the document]
aaronbuchwald: Removing

plugin/evm/vm_warp_test.go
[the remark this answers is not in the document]
aaronbuchwald: Updating

contracts/contracts/IWarpMessenger.sol
```diff
@@ -0,0 +1,53 @@
...
+    event SendWarpMessage(
+        bytes32 indexed destinationChainID,
+        bytes32 indexed destinationAddress,
+        bytes32 indexed sender,
+        bytes message
+    );
+
+    // sendWarpMessage emits a request for the subnet to send a warp message from [msg.sender]
+    // with the specified parameters.
+    // This emits a SendWarpMessage log, which will be picked up by validators to queue the signing of
```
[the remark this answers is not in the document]
aaronbuchwald: This is intentionally `validators` not relayers. The validators handle the produced log on accept and add it to the warp backend, so that they are willing to sign it.

I'll re-phrase this to make it more clear that validators are willing to sign and aggregating signatures is left to relayers.

warp/aggregator/aggregator.go
```diff
@@ -42,7 +42,7 @@ func (a *Aggregator) AggregateSignatures(ctx context.Context, unsignedMessage *a
 		a.subnetID,
 		quorumNum,
 		quorumNum,
-		params.WarpQuorumDenominator,
+		warpPrecompile.QuorumDenominator,
```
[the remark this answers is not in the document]
aaronbuchwald: Moved to params package

contracts/contracts/ExampleWarp.sol
```diff
@@ -0,0 +1,42 @@
...
+        require(message.destinationChainID == destinationChainID);
+        require(message.destinationAddress == destinationAddress);
+        require(keccak256(message.payload) == keccak256(payload));
+    }
+
+    // validateGetBlockchainID checks that the blockchainID returned by warp matches the argument
+    function validateGetBlockchainID(bytes32 blockchainID) external view {
+        require(blockchainID == warp.getBlockchainID());
+    }
+}
```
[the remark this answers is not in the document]
aaronbuchwald: I think we could probably remove `contracts/tasks.ts` completely tbh

plugin/evm/vm_warp_test.go
```diff
@@ -0,0 +1,338 @@
...
+	vm.ctx.ValidatorState.(*validators.TestState).GetValidatorSetF = func(ctx context.Context, height uint64, subnetID ids.ID) (map[ids.NodeID]*validators.GetValidatorOutput, error) {
+		return nil, testErr
+	}
+	require.ErrorIs(t, block2VerifyWithCtx.VerifyWithContext(context.Background(), &block.Context{
+		PChainHeight: 9,
+	}), testErr)
+	require.Equal(t, choices.Processing, block2.Status())
+
+	// Accept the block after performing multiple VerifyWithContext operations
+	require.NoError(t, block2.Accept(context.Background()))
```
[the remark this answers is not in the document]
aaronbuchwald: Precompile accept is handled when the block is accepted, not in the acceptor queue, so this is not currently necessary. I'll add it to make sure this test does not become flaky if we decide to move it to the acceptor queue.

x/warp/README.md
```diff
@@ -0,0 +1,189 @@
...
+
+For example, let's say that there are two Subnets Subnet A and Subnet B each with a single Blockchain we'll call Blockchain A and Blockchain B.
+
+First, Blockchain A produces a BLS Multisignature of a message to send to Blockchain B:
+
+1. A transaction is issued on Blockchain A to send a message to Subnet B
+2. The transaction is accepted on Blockchain A (must wait for acceptance)
+3. The VM powering Blockchain A (ex: Subnet-EVM) should either
+    a) be willing to sign the message to allow a signature aggregator to aggregate signaures from the Subnet's validator set
+    b) aggregate signatures from the validator set of Subnet A to produce its own aggregate signature
```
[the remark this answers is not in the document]
aaronbuchwald: Added `(currently implemented solution)`

x/warp/README.md
```diff
@@ -0,0 +1,189 @@
...
+
+The signature of Subnet A's validator set attests to the message being sent by Blockchain A.
+
+To receive the message, Blockchain B must be running a Snowman VM that is wrapped in the Snowman++ ProposerVM. The ProposerVM provides the P-Chain Context which a block on Blockchain B was issued in. See (here)[https://github.com/ava-labs/avalanchego/tree/master/vms/proposervm#snowman-block-extension] for more details on the ProposerVM block wrapper.
```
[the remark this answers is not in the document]
aaronbuchwald: Adding another sentence explaining the use of the P-Chain height here. Good callout

x/warp/README.md
```diff
@@ -0,0 +1,189 @@
...
+
+To validate and deliver the message, an off-chain component delivers the signed message from Blockchain A to Blockchain B (this is out of scope for Avalanche Warp Messaging itself).
```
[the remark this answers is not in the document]
aaronbuchwald: This is intended as a description of the high level avalanche warp messaging protocol, not specific to Subnet-EVM or how it's integrated here, so I don't think a description of the tx where it's delivered would make sense to include here.

x/warp/README.md
```diff
@@ -0,0 +1,189 @@
...
+
+When Blockchain B receives the message, it validates and delivers/executes the message:
+
+1. Read the SourceChainID of the signed message (Blockchain A)
+2. Look up the SubnetID that validates Blockchain A: Subnet A
+3. Look up the validator set of Subnet A and the registered BLS Public Keys of Subnet A
```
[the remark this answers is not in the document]
aaronbuchwald: Good call

x/warp/README.md
```diff
@@ -0,0 +1,189 @@
...
+- `sender`
+
+The actual `message` is the entire [Avalanche Warp Unsigned Message](https://github.com/ava-labs/avalanchego/blob/master/vms/platformvm/warp/unsigned_message.go#L14) including the Subnet-EVM [Addressed Payload](../../../warp/payload/payload.go).
+
+
+### getVerifiedMessage
+
+`getVerifiedMessage` is used to read the contents of the delivered Avalanche Warp Message into the expected format.
+
+It returns the message if present and a boolean indicating whether or not there is an Avalanche Warp Message included in the transaction and pre-verified in the block.
```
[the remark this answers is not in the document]
aaronbuchwald: Re-phrasing this

x/warp/contract.go
aaronbuchwald: This is correct as is. If there is an error decoding the message (as opposed to the message is not present), then the predicate will fail verification.

x/warp/contract.go
aaronbuchwald: We've already charged for the `CALL` opcode to get here. The cost of a map lookup to get the predicate bytes is negligible so if this function exits along this path, this should not create a DoS vector imo. We could charge a negligible gas cost as well, but I didn't think this was necessary.

x/warp/README.md
```diff
@@ -0,0 +1,189 @@
...
+
+`getBlockchainID` returns the blockchainID of the blockchain that Subnet-EVM is running on.
+
+This is different from the conventional Eth ChainID registered to https://chainlist.org/.
+
+The `blockchainID` in Avalanche refers to the txID that created the blockchain on the Avalanche P-Chain ([docs](https://docs.avax.network/specs/platform-transaction-serialization#unsigned-create-chain-tx)).
+
+### Predicate Encoding
+
+Avalanche Warp Messages are encoded as a signed Avalanche [Warp Message](https://github.com/ava-labs/avalanchego/blob/master/vms/platformvm/warp/message.go#L7) where the [UnsignedMessage](https://github.com/ava-labs/avalanchego/blob/master/vms/platformvm/warp/unsigned_message.go#L14)'s payload includes an [AddressedPayload](../../../warp/payload/payload.go).
```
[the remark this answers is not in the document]
aaronbuchwald: Updated to link v1.10.4

tests/warp/warp_test.go
```diff
@@ -0,0 +1,384 @@
...
+		newHeads := make(chan *types.Header, 10)
+		sub, err := chainBWSClient.SubscribeNewHead(ctx, newHeads)
+		gomega.Expect(err).Should(gomega.BeNil())
+		defer sub.Unsubscribe()
+
+		// Trigger building of a new block at the current timestamp.
+		// This timestamp should be after the ProposerVM activation time or ApricotPhase4 block timestamp.
+		// This should generate a PostForkBlock because its parent block (genesis) has a timestamp (0) that is greater than or equal
+		// to the fork activation time of 0.
+		// Therefore, when we build a subsequent block it should be built with BuildBlockWithContext
```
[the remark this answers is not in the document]
aaronbuchwald: `BuildBlockWithContext` will be called if the VM implements it and the ProposerVM has activated, this determines at what point the proposervm activates.

tests/warp/warp_test.go
```diff
@@ -0,0 +1,384 @@
...
+
+		nonce++
+		// Try building another block to see if that one ends up as a PostForkBlock
+		triggerTx2, err := types.SignTx(types.NewTransaction(nonce, fundedAddress, common.Big1, 21_000, big.NewInt(225*params.GWei), nil), txSigner, fundedKey)
+		gomega.Expect(err).Should(gomega.BeNil())
+
+		err = chainBWSClient.SendTransaction(ctx, triggerTx2)
+		gomega.Expect(err).Should(gomega.BeNil())
+		newHead = <-newHeads
+		log.Info("Transaction2 triggered new block", "blockHash", newHead.Hash())
```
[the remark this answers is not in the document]
aaronbuchwald: Removing

plugin/evm/vm_warp_test.go
aaronbuchwald: Updated

contracts/contracts/IWarpMessenger.sol
```diff
@@ -23,8 +23,11 @@ interface WarpMessenger {
 
     // sendWarpMessage emits a request for the subnet to send a warp message from [msg.sender]
     // with the specified parameters.
-    // This emits a SendWarpMessage log, which will be picked up by validators to queue the signing of
-    // the corresponding Warp Message after the block containing this log has been accepted.
+    // This emits a SendWarpMessage log from the precompile. When the corresponding block is accepted
+    // the Accept hook of the Warp precompile is invoked with any logs coming from the Warp
+    // precompile address (assuming they were not reverted).
```
darioush, line 28:
```suggestion
    // precompile address.
```

Since this refers to accepted logs I don't think we need to exclude reverted.

warp/payload/payload.go
```diff
@@ -44,7 +44,7 @@ func ParseAddressedPayload(b []byte) (*AddressedPayload, error) {
 }
 
 // Initialize recalculates the result of Bytes().
```
darioush, line 46: nit: should update the casing of the comment

contracts/contracts/ExampleWarp.sol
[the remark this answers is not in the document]
darioush: Sounds good to me.

x/warp/contract.go
darioush: nit: could consider moving this Note closer to the error being returned.

contracts/contracts/IWarpMessenger.sol
aaronbuchwald: Sounds good, I'll take another look and remove.

warp/payload/payload.go
aaronbuchwald: Good call

x/warp/contract.go
aaronbuchwald: Moved to right above `UnpackPredicate`

x/warp/config.go
```diff
@@ -0,0 +1,213 @@
...
+	if err := acceptCtx.Warp.AddMessage(unsignedMessage); err != nil {
+		return fmt.Errorf("failed to add warp message during accept (TxHash: %s, LogIndex: %d): %w", txHash, logIndex, err)
+	}
+	return nil
+}
+
+// verifyWarpMessage checks that [warpMsg] can be parsed as an addressed payload and verifies the Warp Message Signature
+// within [predicateContext].
+func (c *Config) verifyWarpMessage(predicateContext *precompileconfig.ProposerPredicateContext, warpMsg *warp.Message) error {
+	// Use default quourum numerator unless config specifies a non-default option
```
anusha-ctrl, line 113: nit: quourum -> quorum

x/warp/config_test.go
```diff
@@ -0,0 +1,110 @@
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			require := require.New(t)
+
+			require.Equal(tt.expected, tt.config.Equal(tt.other))
+		})
+	}
+}
```
anusha-ctrl, line 110: Why do we only test 2 of the config functions? Can we test the `VerifyPredicate` and `PredicateGas`?

x/warp/README.md
```diff
@@ -0,0 +1,190 @@
...
+
+For example, let's say that there are two Subnets: Subnet A and Subnet B each with a single Blockchain we'll call Blockchain A and Blockchain B.
+
+First, Blockchain A produces a BLS Multisignature of a message to send to Blockchain B:
+
+1. A transaction is issued on Blockchain A to send a message to Subnet B
+2. The transaction is accepted on Blockchain A (must wait for acceptance)
+3. The VM powering Blockchain A (ex: Subnet-EVM) should either
+    a) be willing to sign the message to allow a signature aggregator to aggregate signaures from the Subnet's validator set (existing implementation)
+    b) aggregate signatures from the validator set of Subnet A to produce its own aggregate signature
```
anusha-ctrl, line 36: can we clarify definition of signature aggregator on line 35

x/warp/README.md
```diff
@@ -0,0 +1,190 @@
...
+
+The Avalanche P-Chain tracks only its current state and reverse diff layers (reversing the changes from past blocks) in order to re-calculate the validator set at a historical height. This means calculating a very old validator set that is used to verify a Warp Message in an old block may become prohibitively expensive.
+
+Therefore, we need a heuristic to ensure that the network can correctly re-process old blocks (note: re-processing old blocks is a requirement to perform bootstrapping and is used in some VMs including Subnet-EVM to serve or verify historical data).
+
+As a result, we require that the block itself provides a deterministic hint which determines which Avalanche Warp Messages were considered valid/invalid during the block's execution. This ensures that we can always re-process blocks and use the hint to decide whether an Avalanche Warp Message should be treated as valid/invalid even after the P-Chain state that was used at the original execution time may no longer support fast lookups.
+
+To provide that thint, we've explored two designs:
+
```
anusha-ctrl, line 162: nit: thint -> hint

x/warp/README.md
```diff
@@ -0,0 +1,190 @@
...
+The original reason for this was that the notion of predicates for precompiles was designed with Shared Memory in mind. In the case of shared memory, there is no canonical "P-Chain height" in the block which determines whether or not Avalanche Warp Messages are valid.
+
+Instead, the VM interprets a shared memory import operation as valid as soon as the UTXO is available in shared memory. This means that if it were up to the block producer to staple the valid/invalid results of whether or not an attempted atomic operation should be treated as valid, a byzantine block producer could arbitrarily report that such atomic operations were invalid and cause a griefing attack to burn the gas of users that attempted to perform an import.
+
+Therefore, a transaction specified predicate is required to implement the shared memory precompile to prevent such a griefing attack.
+
+In contrast, Avalanche Warp Messages are validated within the context of an exact P-Chain height. Therefore, if a block producer attempted to lie about the validity of such a message, the network would interpret that block as invalid.
+
+### Guarantees Offered by Warp Precompile vs. Built on Top
+
```
anusha-ctrl, line 177: Is this section missing the built on top?

plugin/evm/vm_test.go
```diff
@@ -3153,24 +3153,61 @@ func TestCrossChainMessagestoVM(t *testing.T) {
...
 
-	// Generate a SignatureRequest for an unknown message
-	var signatureRequest message.Request = message.SignatureRequest{
-		MessageID: ids.GenerateTestID(),
-	}
-
-	requestBytes, err := message.Codec.Marshal(message.Version, &signatureRequest)
+	// Generate a new warp unsigned message and add to warp backend
+	warpMessage, err := avalancheWarp.NewUnsignedMessage(vm.ctx.ChainID, ids.GenerateTestID(), []byte{1, 2, 3})
 	require.NoError(t, err)
```
anusha-ctrl, line 3166: nit: can slightly simplify with `require := require.New(t)`

plugin/evm/vm_warp_test.go
```diff
@@ -0,0 +1,316 @@
...
+	issuer, vm, _, _ := GenesisVM(t, true, string(genesisJSON), "", "")
+
+	defer func() {
+		require.NoError(vm.Shutdown(context.Background()))
+	}()
+
+	acceptedLogsChan := make(chan []*types.Log, 10)
+	logsSub := vm.eth.APIBackend.SubscribeAcceptedLogsEvent(acceptedLogsChan)
+	defer logsSub.Unsubscribe()
+
```
anusha-ctrl, line 54: You defer unsubscribe on line 53 and on line 118 you call `logsSub.Unsubscribe()`. Can we just keep the defer?

utils/predicate/predicate_bytes_test.go
```diff
@@ -53,21 +53,21 @@ func FuzzPackPredicate(f *testing.F) {
...
-	f.Fuzz(func(t *testing.T, b []byte) {
-		// Ensure that adding the invalid padding to any length correctly packed predicate
-		// results in failing to unpack it.
-		for _, l := range []int{0, 1, 31, 32, 33, 63, 64, 65} {
-			validPredicate := PackPredicate(utils.RandomBytes(l))
-			invalidPredicate := append(validPredicate, b...)
+	for _, l := range []int{0, 1, 31, 32, 33, 63, 64, 65} {
+		validPredicate := PackPredicate(utils.RandomBytes(l))
+
+		for _, padding := range paddingCases {
```
anusha-ctrl, line 67: would be nice to have a comment here of what a valid predicate is and how the padding breaks it

warp/payload/payload.go
```diff
@@ -0,0 +1,62 @@
...
+)
+
+// AddressedPayload defines an optional format for the bytes payload of a Warp message.
+type AddressedPayload struct {
+	SourceAddress      ids.ID `serialize:"true"`
+	DestinationAddress ids.ID `serialize:"true"`
+	Payload            []byte `serialize:"true"`
+
+	bytes []byte
+}
```
anusha-ctrl, line 19: what is this `bytes` field for

x/warp/config.go
```diff
@@ -0,0 +1,213 @@
...
+		quorumNumerator = c.QuorumNumerator
+	}
+
+	// Verify the warp payload can be decoded to the expected type
+	_, err := warpPayload.ParseAddressedPayload(warpMsg.UnsignedMessage.Payload)
+	if err != nil {
+		return fmt.Errorf("%w: %s", errInvalidAddressedPayload, err)
+	}
+
+	log.Info("verifyingWarpMessage", "warpMsg", warpMsg, "quorumNum", quorumNumerator, "quorumDenom", params.WarpQuorumDenominator)
```
darioush, line 125:
```suggestion
	log.Debug("verifying warp message", "warpMsg", warpMsg, "quorumNum", quorumNumerator, "quorumDenom", params.WarpQuorumDenominator)
```
I think should be debug since invalid txs could spam this log?

x/warp/config.go
```diff
@@ -0,0 +1,213 @@
...
+	if err := warpMsg.Signature.Verify(
```
darioush, line 126: Should we fail verification when `warpMsg.Signature.NumSigners() == 0` before calling `.Verify(...)`?

x/warp/contract.go
```diff
@@ -0,0 +1,278 @@
...
+	// GasCostPerSourceSubnetValidator uint64 = 1 // TODO: charge O(n) cost for subnet validator set lookup
+)
+
+var (
+	errInvalidSendInput = errors.New("invalid sendWarpMessage input")
+)
+
+// Singleton StatefulPrecompiledContract and signatures.
+var (
+
```
darioush, line 45: nit: extra wspace

plugin/evm/vm_warp_test.go
```diff
@@ -0,0 +1,316 @@
...
+	)
+	require.NoError(err)
+	require.NoError(executionRes.Err)
+	require.Equal(expectedOutput, executionRes.ReturnData)
+
+	// Build, verify, and accept block with valid proposer context.
+	validProposerCtx := &block.Context{
+		PChainHeight: 10,
+	}
+	vm.clock.Set(vm.clock.Time().Add(2 * time.Second))
```
darioush, line 276: what does this do?

x/warp/contract.go
```diff
@@ -0,0 +1,278 @@
...
+		if err != nil {
+			return nil, remainingGas, err
+		}
+		return packedOutput, remainingGas, nil
+	}
+
+	msgBytesGas, overflow := math.SafeMul(GasCostPerWarpMessageBytes, uint64(len(predicateBytes)))
+	if overflow {
+		return nil, remainingGas, vmerrs.ErrOutOfGas
+	}
```
darioush, line 138: It seems we charged for this gas once in PredicateGas? (i.e, actual cost per byte is 2* GasCostPerWarpMessageBytes)

warp/payload/payload_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+	addressedPayloadBytes := addressedPayload.Bytes()
+	addressedPayload2, err := ParseAddressedPayload(addressedPayloadBytes)
+	require.NoError(err)
+	require.Equal(addressedPayload, addressedPayload2)
+}
+
+func TestParseAddressedPayloadJunk(t *testing.T) {
+	_, err := ParseAddressedPayload(utils.RandomBytes(1024))
+	require.Error(t, err)
```
anusha-ctrl, line 33: can we keep the tests consistent with `TestAddressedPayload` by adding this line `require := require.New(t)`

x/warp/contract.go
```diff
@@ -0,0 +1,278 @@
...
+	"github.com/ethereum/go-ethereum/common/math"
+)
+
+const (
+	GetBlockchainIDGasCost uint64 = 2 // Based on GasQuickStep used in existing EVM instructions
+	// Sum of base log gas cost, cost of producing 4 topics, and producing + serving a BLS Signature (sign + trie write)
+	// Note: using trie write for the gas cost results in a conservative overestimate since the message is stored in a
+	// flat database that can be cleaned up after a period of time instead of the EVM trie.
+	SendWarpMessageGasCost uint64 = params.LogGas + 4*params.LogTopicGas + 20_000 + contract.WriteGasCostPerSlot
+	// SendWarpMessageGasCostPerByte cost accounts for producing a signed message of a given size
```
anusha-ctrl, line 30: what is the 20_000 and should it be a constant?

x/warp/contract.go
```diff
@@ -0,0 +1,278 @@
...
+	if overflow {
+		return nil, 0, vmerrs.ErrOutOfGas
+	}
+	if remainingGas, err = contract.DeductGas(remainingGas, payloadGas); err != nil {
+		return nil, 0, err
+	}
+	if readOnly {
+		return nil, remainingGas, vmerrs.ErrWriteProtection
+	}
+	// unpack the arguments
```
anusha-ctrl, line 207: should `readOnly` be moved up the function so we don't have to deduct gas from remainingGas

x/warp/config_test.go
aaronbuchwald: These are covered by `predicate_test.go`

x/warp/README.md
aaronbuchwald: Changing to off-chain relayer. I think it's ok to assume an understanding of what an off-chain relayer would do in this context

plugin/evm/vm_test.go
aaronbuchwald: If we do that here, then when we use `t.Run(...)` require would have already taken in the outer `*testing.T` context.

I'd rather make the switch to use `require.New(t)` throughout this file in a separate PR

x/warp/README.md
```diff
@@ -0,0 +1,190 @@
...
+The actual `message` is the entire [Avalanche Warp Unsigned Message](https://github.com/ava-labs/avalanchego/blob/master/vms/platformvm/warp/unsigned_message.go#L14) including the Subnet-EVM [Addressed Payload](../../../warp/payload/payload.go).
+
+
+### getVerifiedMessage
+
+`getVerifiedMessage` is used to read the contents of the delivered Avalanche Warp Message into the expected format.
+
+It returns the message if present and a boolean indicating if a message is present.
+
+To use this function, the transaction must include the signed Avalanche Warp Message encoded in the [predicate](#predicate-encoding) of the transaction. Prior to executing a block, the VM iterates through transactions and pre-verifies all predicates. If a transaction's predicate is invalid, then it is considered invalid to include in the block and dropped.
```
ceyonur, line 118: Does this mean we allow only 1 warp message per block?

x/warp/README.md
```diff
@@ -0,0 +1,190 @@
...
+- `sender`
+
+The actual `message` is the entire [Avalanche Warp Unsigned Message](https://github.com/ava-labs/avalanchego/blob/master/vms/platformvm/warp/unsigned_message.go#L14) including the Subnet-EVM [Addressed Payload](../../../warp/payload/payload.go).
+
+
+### getVerifiedMessage
+
+`getVerifiedMessage` is used to read the contents of the delivered Avalanche Warp Message into the expected format.
+
+It returns the message if present and a boolean indicating if a message is present.
```
ceyonur, line 116: Does this mean getVerifiedMessage returns only if there is an accepted warp message in the same block? How do other contracts can use this to see if a particulat message was accepted?

x/warp/config.go
```diff
@@ -0,0 +1,213 @@
...
+
+// Equal returns true if [s] is a [*Config] and it has been configured identical to [c].
+func (c *Config) Equal(s precompileconfig.Config) bool {
+	// typecast before comparison
+	other, ok := (s).(*Config)
+	if !ok {
+		return false
+	}
+	equals := c.Upgrade.Equal(&other.Upgrade)
+	return equals && c.QuorumNumerator == other.QuorumNumerator
```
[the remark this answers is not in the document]
aaronbuchwald: Agreed

contracts/contracts/ExampleWarp.sol
```diff
@@ -0,0 +1,42 @@
...
+    address constant WARP_ADDRESS = 0x0200000000000000000000000000000000000005;
+    WarpMessenger warp = WarpMessenger(WARP_ADDRESS);
+
+    // sendWarpMessage sends a warp message to the specified destination chain and address pair containing the payload
+    function sendWarpMessage(
+        bytes32 destinationChainID,
+        bytes32 destinationAddress,
+        bytes calldata payload
+    ) external {
+        warp.sendWarpMessage(destinationChainID, destinationAddress, payload);
```
ceyonur, line 17: maybe I completely got this wrong but how do we plan other contracts to use these functions in case they want to issue a message? We currently don't return any ID we can store here and we also don't let historical search in `validateWarpMessage` with message IDs(or hashes). Does `validateWarpMessage` returns true if it's called for a later block? I.e warp accepted in x but contract called that in x+1.

x/warp/README.md
aaronbuchwald: Good callout. This information is included in the section below, but didn't get its own header. Adding that now

tests/warp/warp_test.go
```diff
@@ -0,0 +1,374 @@
+// Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+// Implements solidity tests.
+package solidity
```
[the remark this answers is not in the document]
aaronbuchwald: Renamed

tests/warp/warp_test.go
```diff
@@ -0,0 +1,374 @@
...
+		chainAURIs, chainBURIs         []string
+		chainAWSClient, chainBWSClient ethclient.Client
+		chainID                        = big.NewInt(99999)
+		fundedKey                      *ecdsa.PrivateKey
+		fundedAddress                  common.Address
+		payload                        = []byte{1, 2, 3}
+		txSigner                       = types.LatestSignerForChainID(chainID)
+	)
+
+	var err error
```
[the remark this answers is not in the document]
aaronbuchwald: Moved

tests/warp/warp_test.go
```diff
@@ -0,0 +1,374 @@
...
+		chainAURIs = append(chainAURIs, subnetADetails.ValidatorURIs...)
+
+		subnetB := subnetIDs[1]
+		subnetBDetails, ok := manager.GetSubnet(subnetB)
+		gomega.Expect(ok).Should(gomega.BeTrue())
+		blockchainIDB := subnetBDetails.BlockchainID
+		gomega.Expect(len(subnetBDetails.ValidatorURIs)).Should(gomega.Equal(5))
+		chainBURIs = append(chainBURIs, subnetBDetails.ValidatorURIs...)
+
+		log.Info("Created URIs for both subnets", "ChainAURIs", chainAURIs, "ChainBURIs", chainBURIs, "blockchainIDA", blockchainIDA.String(), "blockchainIDB", blockchainIDB)
```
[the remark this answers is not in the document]
aaronbuchwald: Updated

tests/warp/warp_test.go
```diff
@@ -0,0 +1,374 @@
...
+	// Aggregate a Warp Signature using the node's Signature Aggregation API call and verifying that its output matches the
+	// the manual construction
+	ginkgo.It("Aggregate Warp Signature via Aggregator", ginkgo.Label("Warp", "ReceiveWarp", "AggregatorWarp"), func() {
+		ctx := context.Background()
+
+		// Verify that the signature aggregation matches the results of manually constructing the warp message
+		warpClient, err := warpBackend.NewWarpClient(chainAURIs[0], blockchainIDA.String())
+		gomega.Expect(err).Should(gomega.BeNil())
+
+		signedWarpMessageBytes, err := warpClient.GetAggregateSignature(ctx, unsignedWarpMessageID, 100)
```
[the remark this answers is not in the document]
aaronbuchwald: Updated to use the denominator constant and adding this comment

tests/warp/warp_test.go
```diff
@@ -0,0 +1,374 @@
...
+		log.Info("Waiting for new block confirmation")
+		newHead = <-newHeads
+		blockHash := newHead.Hash()
+		log.Info("Fetching relevant warp logs and receipts from new block")
+		logs, err := chainBWSClient.FilterLogs(ctx, interfaces.FilterQuery{
+			BlockHash: &blockHash,
+			Addresses: []common.Address{warp.Module.Address},
+		})
+		gomega.Expect(err).Should(gomega.BeNil())
+		gomega.Expect(len(logs)).Should(gomega.Equal(0))
```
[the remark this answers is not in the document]
aaronbuchwald: This confirms the behavior that `getVerifiedWarpMessage` does not emit a log whereas sending a warp message does.

warp/payload/payload.go
```diff
@@ -0,0 +1,62 @@
...
+
+package payload
+
+import (
+	"fmt"
+
+	"github.com/ava-labs/avalanchego/ids"
+)
+
+// AddressedPayload defines an optional format for the bytes payload of a Warp message.
```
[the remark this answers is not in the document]
aaronbuchwald: Re-phrasing to the format for a specific type of message. I meant that this was one possible message type, but optional is misleading.

x/warp/config.go
aaronbuchwald: This is in predicate gas, which should only fail imo if it is unable to calculate the amount of gas, not if the message should be invalid. The call to `Verify(...)` should perform operations in the correct order, such that if there is a sign the message will fail verification, we will minimize wasted work, but I don't think we should try to fail early within `PredicateGas(...)`

plugin/evm/vm_warp_test.go
aaronbuchwald: This sets the time 2 seconds into the future so that block production occurs at the target rate and does not increase the base fee. This is used throughout the rest of the VM tests as well.

x/warp/config.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+func (c *Config) Accept(acceptCtx *precompileconfig.AcceptContext, txHash common.Hash, logIndex int, topics []common.Hash, logData []byte) error {
+	unsignedMessage, err := warp.ParseUnsignedMessage(logData)
+	if err != nil {
+		return fmt.Errorf("failed to parse warp log data into unsigned message (TxHash: %s, LogIndex: %d): %w", txHash, logIndex, err)
+	}
+	log.Info("Accepted warp unsigned message", "txHash", txHash, "logIndex", logIndex, "logData", common.Bytes2Hex(logData))
```
ceyonur, line 103: Isn't this signed and verified by `VerifyPredicate` at this point?

warp/payload/payload_test.go
aaronbuchwald: I'd prefer a two line to a three line test here since it's not going to be re-used at all.

x/warp/config.go
aaronbuchwald: The codec will fail to unmarshal if it's going to return a nil value.

x/warp/contract.go
aaronbuchwald: Moved to `AddWarpMessageGasCost` to reflect the cost of signing and storing the message.

utils/predicate/predicate_tx.go
```diff
@@ -0,0 +1,43 @@
+// (c) 2023, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package predicateutils
```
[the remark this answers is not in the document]
aaronbuchwald: Updated to `predicate`

x/warp/README.md
```diff
@@ -0,0 +1,190 @@
...
+
+1. Read the SourceChainID of the signed message (C-Chain)
+2. Look up the SubnetID that validates C-Chain: Primary Network
+3. Look up the validator set of Subnet B (instead of the Primary Network) and the registered BLS Public Keys of Subnet B at the P-Chain height specified by the ProposerVM header
+4. Filter the validators of Subnet B to include only the validators that the signed message claims as signers
+5. Verify the claimed included validators represent a sufficient quorum of stake to verify the message
+6. Aggregate the BLS Public Keys of the claimed signers into an aggregated BLS Public Key
+7. Validate the aggregate signature matches the claimed aggregate BLS Public Key
+
+This means that if Subnet B has 10 validators, then C-Chain to Subnet communication only requires a threshold of those 10 validators rather than the entire Primary Network.
```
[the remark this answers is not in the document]
aaronbuchwald: It's a threshold of stake, not a specific number of validators since everything is stake weighted.

x/warp/contract.go
aaronbuchwald: We could save gas in the case that someone calls this incorrectly in a `readOnly` state. I don't think this is a necessary change.

x/warp/README.md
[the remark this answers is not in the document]
aaronbuchwald: Updating the comment to make it clear this is stake weighted.

x/warp/README.md
```diff
@@ -0,0 +1,190 @@
...
+
+- `SourceChainID` - blockchainID of the sourceChain on the Avalanche P-Chain
+- `SourceAddress` - `msg.sender` encoded as a 32 byte value that calls `sendWarpMessage`
+- `DestinationChainID` - `bytes32` argument specifies the blockchainID on the Avalanche P-Chain that should receive the message
+- `DestinationAddress` - 32 byte value that represents the destination address that should receive the message (on the EVM this is the 20 byte address left zero extended)
+- `Payload` - `payload` argument specified in the call to `sendWarpMessage`
+
+Calling this function will issue a `SendWarpMessage` event from the Warp Precompile. Since the EVM limits the number of topics to 4 including the EventID, this message includes only the topics that would be expected to help filter messages emitted from the Warp Precompile the most.
+
+Specifically, the `payload` is not emitted because this would need to be a hash and it would not be expected to be helpful to determine whether or not the log is of any interest to anyone parsing the chain.
```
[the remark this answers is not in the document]
aaronbuchwald: Yup, good point

x/warp/contract.go
```diff
@@ -0,0 +1,278 @@
...
+
+	WarpPrecompile = createWarpPrecompile()
+)
+
+// WarpMessage is an auto generated low-level Go binding around an user-defined struct.
+type WarpMessage struct {
+	OriginChainID       [32]byte
+	OriginSenderAddress [32]byte
+	DestinationChainID  [32]byte
+	DestinationAddress  [32]byte
```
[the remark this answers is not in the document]
aaronbuchwald: It's just an alias, but this is what was produced by the precompile template. If we want to use `common.Hash` instead of `[32]byte` we should update the precompile template to handle the special case. Would you prefer I update these to use `common.Hash` in the meantime?

x/warp/contract.go
```diff
@@ -0,0 +1,278 @@
...
+}
+
+type GetVerifiedWarpMessageOutput struct {
+	Message WarpMessage
+	Exists  bool
+}
+
+type SendWarpMessageInput struct {
+	DestinationChainID [32]byte
+	DestinationAddress [32]byte
```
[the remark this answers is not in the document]
aaronbuchwald: Same comment as above

x/warp/contract.go
```diff
@@ -0,0 +1,278 @@
...
+
+// getBlockchainID returns the snow Chain Context ChainID of this blockchain.
+func getBlockchainID(accessibleState contract.AccessibleState, caller common.Address, addr common.Address, input []byte, suppliedGas uint64, readOnly bool) (ret []byte, remainingGas uint64, err error) {
+	if remainingGas, err = contract.DeductGas(suppliedGas, GetBlockchainIDGasCost); err != nil {
+		return nil, 0, err
+	}
+	packedOutput, err := PackGetBlockchainIDOutput(accessibleState.GetSnowContext().ChainID)
+	if err != nil {
+		return nil, remainingGas, err
+	}
```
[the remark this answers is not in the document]
aaronbuchwald: This is stored in memory on `*snow.Context`, so this should already be extremely cheap to access and using it through this interface makes it cleaner and more testable than if we tried to set it up on configuration.

x/warp/config.go
```diff
@@ -0,0 +1,213 @@
+// (c) 2023, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package warp
```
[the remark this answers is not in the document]
aaronbuchwald: This is meant to be moved directly into the precompile package as the contract implementation.

I think we should either have one package with everything warp related as a sub-package or separate it into the warp contract and everything else as it is now. I prefer this approach because it follows the convention of having each precompile contract implementation within `precompile/contracts/XXX`

x/warp/README.md
aaronbuchwald: No this is encoded on a per transaction basis, so there can only be one warp message per transaction currently.

contracts/contracts/ExampleWarp.sol
aaronbuchwald: Currently, the messageID is the hash of the unsigned warp message which is emitted as the un-indexed log data from the relevant warp log.

The current implementation relies on off-chain relayers to ingest these logs, parse the message, and request the signatures from VMs, so the contract emitting the message does not need to know the messageID that it's producing. It would certainly be reasonable to return the messageID to the caller though.

`validateWarpMessage` performs signature verification based off of the P-Chain. It's not related to where the message was produced on this chain since it's verifying a message that originated on a different chain.

x/warp/config.go
aaronbuchwald: No. When you send a message it produces a log. When you verify a message, it verifies a predicate and makes it read-able throughout the transaction's execution.

x/warp/contract.go
```diff
@@ -0,0 +1,278 @@
...
+func getVerifiedWarpMessage(accessibleState contract.AccessibleState, caller common.Address, addr common.Address, _ []byte, suppliedGas uint64, readOnly bool) (ret []byte, remainingGas uint64, err error) {
+	remainingGas = suppliedGas
+	// XXX Note: there is no base cost for retrieving a verified warp message. Instead, we charge for each piece of gas,
+	// prior to each execution step.
+	// Ignore input since there are no arguments
+	predicateBytes, exists := accessibleState.GetStateDB().GetPredicateStorageSlots(ContractAddress)
+	// If there is no such value, return false to the caller.
+	if !exists {
+		packedOutput, err := PackGetVerifiedWarpMessageOutput(GetVerifiedWarpMessageOutput{
+			Exists: false,
```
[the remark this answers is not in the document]
aaronbuchwald: I think this would be a fairly minor optimization. Happy to merge a follow on PR to do so.

x/warp/config.go
```diff
@@ -0,0 +1,213 @@
...
+		params.WarpQuorumDenominator,
+	); err != nil {
+		return fmt.Errorf("warp signature verification failed: %w", err)
+	}
+
+	return nil
+}
+
+// PredicateGas returns the amount of gas necessary to verify the predicate
+func (c *Config) PredicateGas(predicateBytes []byte) (uint64, error) {
```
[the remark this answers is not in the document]
aaronbuchwald: I don't think additional comments would make this code much clearer. Could you be more specific about which steps could use further explanation?

x/warp/contract.go
```diff
@@ -0,0 +1,278 @@
...
+		outputStruct.Message,
+		outputStruct.Exists,
+	)
+}
+
+// getVerifiedWarpMessage retrieves the pre-verified warp message from the predicate storage slots and returns
+// the expected ABI encoding of the message to the caller.
+func getVerifiedWarpMessage(accessibleState contract.AccessibleState, caller common.Address, addr common.Address, _ []byte, suppliedGas uint64, readOnly bool) (ret []byte, remainingGas uint64, err error) {
+	remainingGas = suppliedGas
+	// XXX Note: there is no base cost for retrieving a verified warp message. Instead, we charge for each piece of gas,
```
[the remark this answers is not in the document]
aaronbuchwald: I'm happy to add a base cost for this, but this is accessing a map from statedb and does not actually touch disk.

x/warp/contract.go
darioush: Could we add a comment about this?

aaronbuchwald:
Creating this GitHub issue to ensure that the Warp README provides sufficient documentation for the target audiences: https://github.com/ava-labs/subnet-evm/issues/754.

darioush (approved):
I think this looks good and we can continue the README changes in a follow-up PR

x/warp/README.md
```diff
@@ -0,0 +1,190 @@
...
+Since the Primary Network has the largest possible number of validators for any Subnet on Avalanche, it would also be the most expensive Subnet to verify Avalanche Warp Messages from (most signatures required to verify it). Luckily, we can do something much smarter.
+
+When a Subnet receives a message from a blockchain on the Primary Network, we use the validator set of the receiving Subnet instead of the entire network when validating the message. This means that the C-Chain sending a message can be the exact same as Subnet to Subnet communciation.
+
+However, when Subnet B receives a message from the C-Chain, it changes the semantics to the following:
+
+1. Read the SourceChainID of the signed message (C-Chain)
+2. Look up the SubnetID that validates C-Chain: Primary Network
+3. Look up the validator set of Subnet B (instead of the Primary Network) and the registered BLS Public Keys of Subnet B at the P-Chain height specified by the ProposerVM header
+4. Filter the validators of Subnet B to include only the validators that the signed message claims as signers
```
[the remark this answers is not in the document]
ceyonur: I see. I was trying to say if it would be practical to assume "everyone" has signed it or do we guarantee that B validators will receive those messages (as priority). But I don't think this is in the scope at the moment?

x/warp/config.go
```diff
@@ -0,0 +1,213 @@
...
+		return fmt.Errorf("failed to add warp message during accept (TxHash: %s, LogIndex: %d): %w", txHash, logIndex, err)
+	}
+	return nil
+}
+
+// verifyWarpMessage checks that [warpMsg] can be parsed as an addressed payload and verifies the Warp Message Signature
+// within [predicateContext].
+func (c *Config) verifyWarpMessage(predicateContext *precompileconfig.ProposerPredicateContext, warpMsg *warp.Message) error {
+	// Use default quorum numerator unless config specifies a non-default option
+	quorumNumerator := params.WarpDefaultQuorumNumerator
```
ceyonur, line 114: instead of this, can we configure the default in module.go/#Configure and set to `c.QuorumNumerator`?

ceyonur: related to https://github.com/ava-labs/subnet-evm/pull/718#discussion_r1271963249

x/warp/config.go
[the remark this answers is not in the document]
ceyonur: I was thinking of something like this:https://github.com/ava-labs/subnet-evm/blob/warp-contract/core/state_transition.go#L364-L370

It's a nit but could be handy once we implement that TODO.

x/warp/contract.go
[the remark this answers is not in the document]
ceyonur: If it also makes sense to you, we can both convert here and in template. I can change the template in another PR.

x/warp/contract.go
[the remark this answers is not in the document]
ceyonur: I'd sleep better if we have a static cost :D

x/warp/config.go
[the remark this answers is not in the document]
ceyonur: I agree it's better to have this under precompile package.

x/warp/contract.go
aaronbuchwald: Adding

x/warp/README.md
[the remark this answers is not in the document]
aaronbuchwald: We do not require everyone has signed it only a threshold of stake.

x/warp/config.go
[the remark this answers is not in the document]
aaronbuchwald: Ahhh, you're right that is really nice. Adding a comment for that.

x/warp/contract.go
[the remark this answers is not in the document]
aaronbuchwald: Updated to `common.Hash`

x/warp/contract.go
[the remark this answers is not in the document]
aaronbuchwald: Adding a base cost

x/warp/config.go
aaronbuchwald: I think it's simpler to keep this entirely in-memory on the config for warp itself and provide that from the config during verification, so that we don't need to write it into the state. I think writing it in the state makes sense when we want it to be mutable via on-chain events.

ceyonur (approved):

x/warp/contract_test.go
```diff
@@ -141,10 +141,10 @@ func TestSendWarpMessage(t *testing.T) {
 				addressedPayload, err := warpPayload.ParseAddressedPayload(unsignedWarpMsg.Payload)
 				require.NoError(t, err)
 
-				require.Equal(t, unsignedWarpMsg.DestinationChainID, destinationChainID)
+				// require.Equal(t, unsignedWarpMsg.DestinationChainID, destinationChainID)
```
darioush, line 144: nit: could we delete this?

darioush (approved):

warp/payload/payload.go
```diff
@@ -6,22 +6,22 @@ package payload
...
 
 // AddressedPayload defines the format for delivering a point to point message across VMs
 // ie. (ChainA, AddressA) -> (ChainB, AddressB)
 type AddressedPayload struct {
-	SourceAddress      ids.ID `serialize:"true"`
-	DestinationChainID ids.ID `serialize:"true"`
-	DestinationAddress ids.ID `serialize:"true"`
-	Payload            []byte `serialize:"true"`
+	SourceAddress      common.Address `serialize:"true"`
+	DestinationChainID common.Hash    `serialize:"true"`
```
darioush, line 16: why do we change this from `ids.ID` to common.Hash?

aaronbuchwald: It's a no-op change from one alias to the other, but changed it to `common.Hash` to stick with using the `common` package instead of one from each.

Daniel-K-Ivanov (commented):
Hey, Daniel from [LimeChain](https://limechain.tech/) here. Thanks for providing a thorough explanation of the design of the Warp protocol. Here are some of my considerations that popped up while I was reading the README. Sorry if some of those questions seem off / not relevant. Disclaimer - I do not have a deep understanding of the finality guarantees and configuration options for Subnets:

- **Minimum Enforced Latency** - How would the `receiving` subnet guard against reorgs on the `destination` subnet? Is there a `minimum_delay` for messages that would cover the cases where the `source` subnet changes its state (due to not being finalised). The protocol must guard `receiving` subnets from applying messages from a `sender` subnet before being “finalised”, otherwise they risk their own finalisation. Applying a message from a network, which gets reorged after the message is delivered must trigger the receiving network to reorgs as-well and re-apply all state transitions at the point where the message was “applied”.  However, if the receiving network indeed re-applies the state, that can be an attack vector for malicious Subnets to disturb the finalisation of the receiving networks. A solution to this problem might be a _configuration for a minimal enforced latency/delay on messages for each subnet that the given subnet accepts messages from_.
- **DDoS Protection**
    - How is the C-Chain protected against malicious subnets DDoS-ing it by sending lots of Warp messages? Is there an economic security that would guard the receiving networks against DDoS attacks from malicious subnets that would create lots of messages which would need to be verified by receiving subnets? This becomes even harder if the message is multicast and gets to multiple subnets. Ideally, there should be *something* that the receiving network gets in exchange for “verifying” the message. Gas costs in the straightforward implementation do not count. They are security guarantees for the validators in the `sending` subnet to guard against malicious users that want to DDoS the sending subnet, however they do not protect `receiving` subnet validators from malicious `sending` validators.
    - I am not sure whether a maximum number of messages that can get applied (queue) in the receiving network per block has been considered. Ideally this number would be configurable similar to the block’s `gas limit`.
- **Syncing the network** - It seems that the P chain cannot be used to provide the data (validator sets) so that new nodes can sync the network trustlessly, meaning verify the authenticity of messages on their own. I am not quite sure how the issue has been resolved using transaction predicates. My assumption on how transaction predicates work might not be correct, however as far as I understand, the “message bytes” are added as data in the access list of the EOA transaction. If that is the case, there would be DevEx trade-offs such as:
    - How would the EOA know what the message required for the TX is?
    - What if the EOA interacts with a dApp that interacts with a second dApp that uses the Warp protocol. How would the UI of the first dApp know that the warp message should be placed within the access list?
    
    In general I think that it would be great if additional information on how the Warp protocol will ensure that _new nodes will be able to follow / sync the network **ideally** (although it might not be technically possible) without trusting the consensus at the time of message inclusion_ but rather being able to compute / deduce the same facts themselves.
    
    If the syncing nodes are relying on the “facts” that a message _was considered valid at block `X` that would mean that the full chain history after `block X` becomes trusted and not computed by the syncing node.
    
- **Misc**
    - **Latency** - Developers would be interested in message latency, meaning how long it would take to deliver a message on a destination subnet.
    - **Costs** - Developers would be interested in gas costs for `sendWarpMessage` and `getVerifiedMessage`

aaronbuchwald:
Thanks for taking a look!

> * **Minimum Enforced Latency** - How would the `receiving` subnet guard against reorgs on the `destination` subnet? Is there a `minimum_delay` for messages that would cover the cases where the `source` subnet changes its state (due to not being finalised). The protocol must guard `receiving` subnets from applying messages from a `sender` subnet before being “finalised”, otherwise they risk their own finalisation. Applying a message from a network, which gets reorged after the message is delivered must trigger the receiving network to reorgs as-well and re-apply all state transitions at the point where the message was “applied”.  However, if the receiving network indeed re-applies the state, that can be an attack vector for malicious Subnets to disturb the finalisation of the receiving networks. A solution to this problem might be a _configuration for a minimal enforced latency/delay on messages for each subnet that the given subnet accepts messages from_.

Avalanche Subnets offer fast finality after a few seconds, such that there are no reorgs after a block has been accepted. Subnets only sign warp messages after a block has been accepted, so receiving subnets do not need to worry about handling reorgs.

> * **DDoS Protection**
>   
>   * How is the C-Chain protected against malicious subnets DDoS-ing it by sending lots of Warp messages? Is there an economic security that would guard the receiving networks against DDoS attacks from malicious subnets that would create lots of messages which would need to be verified by receiving subnets? This becomes even harder if the message is multicast and gets to multiple subnets. Ideally, there should be _something_ that the receiving network gets in exchange for “verifying” the message. Gas costs in the straightforward implementation do not count. They are security guarantees for the validators in the `sending` subnet to guard against malicious users that want to DDoS the sending subnet, however they do not protect `receiving` subnet validators from malicious `sending` validators.
>   * I am not sure whether a maximum number of messages that can get applied (queue) in the receiving network per block has been considered. Ideally this number would be configurable similar to the block’s `gas limit`.

The cost for sending a message is paid fully on the source subnet and the cost of verifying/receiving a message is paid fully on the destination subnet. In other words, you can send as many messages as you want from a Subnet to the C-Chain or between any two Subnets, which may create a surplus of outstanding messages to be delivered to the C-Chain, but delivering each one requires paying the full cost of verifying/receiving that message on the C-Chain.

> * **Syncing the network**

Yes, we assume that the message was verified at the time it was verified by the Subnet. We make the same assumption in order to make bootstrapping faster in a few other places.

To perform full verification for all Avalanche Warp Messages, we could either rely on fast archival lookups on the P-Chain to verify historical messages or switching to including proofs of the validator sets used to verify messages at a specific point in time.

For now, we use the assumption that the network correctly verified warp messages when it was accepted by the network. It would be a reasonable feature request to add an option to fully verify every operation during a full sync.

>   * How would the EOA know what the message required for the TX is?

Warp is planned to be used as a primitive to build cross-chain communication protocols on top of. A relayer would listen for Avalanche Warp Messages and deliver them to the recipient chain.

>   * What if the EOA interacts with a dApp that interacts with a second dApp that uses the Warp protocol. How would the UI of the first dApp know that the warp message should be placed within the access list?

Great question, a communication layer built on top of Warp could handle this by delivering messages across multiple transactions and storing those messages in the state to be read all at once later. Alternatively, you could build a communication on top by verifying signed block hashes and proving the contents of the block, all of its ancestors, and any transactions/messages sent based off of that information.

>   * **Latency** - Developers would be interested in message latency, meaning how long it would take to deliver a message on a destination subnet.

The latency depends on how fast the relayer is able to aggregate signatures from the validator set of the sending Subnet and how quickly it can deliver the transaction on the receiving Subnet. In general, the latency should be dominated by the time for a block to be accepted on the source subnet (cannot aggregate signatures until this point) + the latency of confirming the transaction on the destination, which will be approximately two block confirmations one on each of the source and receiver.

>   * **Costs** - Developers would be interested in gas costs for `sendWarpMessage` and `getVerifiedMessage`

The gas costs are present in the Warp contract code. We can add documentation to break down the gas costs for `sendWarpMessage` and `getVerifiedMessage` as well. We are performing some optimizations to P-Chain lookups, and once that's completed we'll derive the final gas costs based off of those benchmarks.

PR #724, opened 2023-07-10
Warp contract feature

morrisettjohn:
## Why this should be merged
Extended warp e2e to use an example contract deployed to subnet A and B.  Also fixed minor bug where both subnets used the same URI, causing them to have messages sent on the same network.


tests/warp/warp_test.go
```diff
@@ -163,22 +220,28 @@ var _ = ginkgo.Describe("[Warp]", ginkgo.Ordered, func() {
 
 		gomega.Expect(err).Should(gomega.BeNil())
 
+		// TODO: switch to using SubscribeFilterLogs to retrieve the warp log
```
morrisettjohn, line 223:
```suggestion
```

tests/warp/warp_test.go
```diff
@@ -154,7 +165,53 @@ var _ = ginkgo.Describe("[Warp]", ginkgo.Ordered, func() {
...
+	// Deploy the test warp contracts to subnets A & B
+	ginkgo.It("Deploy Test Warp Contracts", ginkgo.Label("Warp", "DeployWarp"), func() {
+		ctx := context.Background()
+
+		auth, err := bind.NewKeyedTransactorWithChainID(fundedKey, chainID)
+		gomega.Expect(err).Should(gomega.BeNil())
 
+		//deploy on subnetA
+		newHeads := make(chan *types.Header, 10)
+		sub, err := chainAWSClient.SubscribeNewHead(ctx, newHeads)
```
anusha-ctrl, line 179: should we have a buffer of 10 for `newHeads`

tests/warp/warp_test.go
```diff
@@ -169,16 +226,21 @@ var _ = ginkgo.Describe("[Warp]", ginkgo.Ordered, func() {
...
 			Payload:            payload,
 		})
 		gomega.Expect(err).Should(gomega.BeNil())
+
 		tx := types.NewTx(&types.DynamicFeeTx{
 			ChainID:   chainID,
-			Nonce:     0,
-			To:        &warp.Module.Address,
+			Nonce:     1,
+			To:        &chainAExAddr,
```
anusha-ctrl, line 243: why is this nonce hardcoded to 1?

contracts/contracts/ExampleWarp.sol
```diff
@@ -36,7 +35,7 @@ contract ExampleWarp {
     }
 
     // validateGetBlockchainID checks that the blockchainID returned by warp matches the argument
-    function validateGetBlockchainID(blockchainID bytes32) external {
+    function validateGetBlockchainID(bytes32 blockchainID) external view {
```
richardpringle, line 38: @aaronbuchwald, what would you do without @morrisettjohn?

tests/warp/warp_test.go
```diff
@@ -39,6 +41,12 @@ var (
...
...5b600060a0828403121561069a5761069961051e565b5b6106a460a0610594565b905060006106b4848285016104dc565b60008301525060206106c8848285016104dc565b60208301525060406106dc848285016104dc565b60408301525060606106f0848285016104dc565b606083015250608082015167ffffffffffffffff811115610714576107136105af565b5b61072084828501610656565b60808301525092915050565b60008115159050919050565b6107418161072c565b811461074c57600080fd5b50565b60008151905061075e81610738565b92915050565b6000806040838503121561077b5761077a6102fc565b5b600083015167ffffffffffffffff81111561079957610798610301565b5b6107a585828601610684565b92505060206107b68582860161074f565b9150509250929050565b600081905092915050565b82818337600083830152505050565b60006107e683856107c0565b93506107f38385846107cb565b82840190509392505050565b600061080c8284866107da565b91508190509392505050565b61082181610306565b82525050565b600082825260208201905092915050565b60006108448385610827565b93506108518385846107cb565b61085a83610523565b840190509392505050565b600060608201905061087a6000830187610818565b6108876020830186610818565b818103604083015261089a818486610838565b90509594505050505056fea2646970667358221220b2cefd832fa6ca66aff8dba2e446584042974aa24a3f54e0f99e83fc9cf34edf64736f6c63430008140033")
```
richardpringle, line 47: Is there a better way to do this rather than having a massive line like this?

contracts/contracts/ExampleWarp.sol
aaronbuchwald: I just hope he doesn't go back to school 😅. We can give him a degree right?

tests/warp/warp_test.go
morrisettjohn: Previously, I used a goembed function, keeping the abi and bin in a separate folder.  Aaron suggested it might be better to keep the data directly in this file.  In other files contract abi/bins have been stored directly like this, so I think we're mostly trying to adhere to the same style (though those have mostly been much shorter)

contracts/contracts/ExampleWarp.sol
morrisettjohn: Haha that would be cool!

tests/warp/warp_test.go
```diff
@@ -111,10 +119,13 @@ var _ = ginkgo.Describe("[Warp]", ginkgo.Ordered, func() {
 		blockchainIDA, blockchainIDB   ids.ID
 		chainAURIs, chainBURIs         []string
 		chainAWSClient, chainBWSClient ethclient.Client
+		chainANonce, chainBNonce       = uint64(0), uint64(0)
+		chainAExAddr, chainBExAddr     common.Address
```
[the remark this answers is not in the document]
morrisettjohn: Mostly, those are used across different ginkgo tests.  But, I think it might be better to fetch the nonce via the ethclient

tests/warp/warp_test.go
morrisettjohn: It's probably a bad idea to have it hardcoded, I think it would probably be better to fetch the nonce via the ethclient, like above

tests/warp/warp_test.go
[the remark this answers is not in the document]
aaronbuchwald: Can we switch to using `ethclient` to fetch the correct nonce where they are needed?

If a nonce for the same address is used more than once in an It node, we should store it in a local variable and increment it after the first transaction.

tests/warp/warp_test.go
```diff
@@ -39,6 +41,12 @@ var (
 	manager = runner.NewNetworkManager(config)
 )
 
+var (
```
aaronbuchwald, line 44: Could we move these local variables into the `ginkgo.Describe` var block so that we put all of the variables needed throughout the test there rather than using a global variable here?

imo we could either put all of the vars there or here, but we should put them in one place.

tests/warp/warp_test.go
```diff
@@ -154,7 +165,53 @@ var _ = ginkgo.Describe("[Warp]", ginkgo.Ordered, func() {
...
+	})
+
+	// Deploy the test warp contracts to subnets A & B
+	ginkgo.It("Deploy Test Warp Contracts", ginkgo.Label("Warp", "DeployWarp"), func() {
+		ctx := context.Background()
+
+		auth, err := bind.NewKeyedTransactorWithChainID(fundedKey, chainID)
+		gomega.Expect(err).Should(gomega.BeNil())
 
+		//deploy on subnetA
```
aaronbuchwald, line 177:
```suggestion
		// deploy warp contracts on subnetA
```

we should add one character of whitespace between `//` and the actual comment

tests/warp/warp_test.go
```diff
@@ -154,7 +165,53 @@ var _ = ginkgo.Describe("[Warp]", ginkgo.Ordered, func() {
...
+		gomega.Expect(err).Should(gomega.BeNil())
+
+		log.Info("Waiting for contract creation on chain A to be accepted")
+		<-newHeads
+		recp, err := chainAWSClient.TransactionReceipt(ctx, signed.Hash())
+		gomega.Expect(err).Should(gomega.BeNil())
+		gomega.Expect(recp.Status).Should(gomega.Equal(types.ReceiptStatusSuccessful)) //make sure status code is 1, contract deployed successfully
+		sub.Unsubscribe()
+
+		//deploy on subnetB
```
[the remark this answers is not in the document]
aaronbuchwald: Since this is the same code to deploy to both chains, could we change this to a for loop over the URIs or respective client values, and deploy the contract once to each so that we reduce the code duplication

tests/warp/warp_test.go
```diff
@@ -169,16 +226,21 @@ var _ = ginkgo.Describe("[Warp]", ginkgo.Ordered, func() {
 		gomega.Expect(err).Should(gomega.BeNil())
 		defer sub.Unsubscribe()
 
+		toKey, err := crypto.GenerateKey()
+		gomega.Expect(err).Should(gomega.BeNil())
+		sendToAddress = crypto.PubkeyToAddress(toKey.PublicKey)
+
```
aaronbuchwald, line 232: We never actually use `toKey` because we only need an address value to include in the contract arguments here and not the key itself.

Could we use a hardcoded address here instead ie. `common.HexToAddress("095e7baea6a6c7c4c2dfeb977efac326af552d87")` so that we don't need this extra code

tests/warp/warp_test.go
```diff
@@ -336,12 +399,20 @@ var _ = ginkgo.Describe("[Warp]", ginkgo.Ordered, func() {
 		log.Info("Transaction2 triggered new block", "blockHash", newHead.Hash())
 		nonce++
 
-		packedInput, err := warp.PackGetVerifiedWarpMessage()
+		//attempt to send a warp message with no sigs, which should fail
```
aaronbuchwald, line 402: same whitespace comment here

tests/warp/warp_test.go
```diff
@@ -336,12 +399,20 @@ var _ = ginkgo.Describe("[Warp]", ginkgo.Ordered, func() {
...
+		packedInput, err := WarpExampleABI.Pack(
+			"validateWarpMessage",
+			blockchainIDA,
+			chainAExAddr.Hash(), //calling address is the deployed exampleWarpTx, not the original funded address
```
aaronbuchwald, line 406: same whitespace comment here

tests/warp/warp_test.go
```diff
@@ -154,7 +165,53 @@ var _ = ginkgo.Describe("[Warp]", ginkgo.Ordered, func() {
...
+		chainANonce++
+		chainAExAddr = addr
+		signed, err := auth.Signer(auth.From, tx)
+		gomega.Expect(err).Should(gomega.BeNil())
+
+		log.Info("Waiting for contract creation on chain A to be accepted")
+		<-newHeads
+		recp, err := chainAWSClient.TransactionReceipt(ctx, signed.Hash())
+		gomega.Expect(err).Should(gomega.BeNil())
+		gomega.Expect(recp.Status).Should(gomega.Equal(types.ReceiptStatusSuccessful)) //make sure status code is 1, contract deployed successfully
```
aaronbuchwald, line 193: same whitespace comment

tests/warp/warp_test.go
```diff
@@ -154,7 +165,53 @@ var _ = ginkgo.Describe("[Warp]", ginkgo.Ordered, func() {
...
+		addr, tx, _, err := bind.DeployContract(auth, WarpExampleABI, WarpExampleBin, chainAWSClient)
+		gomega.Expect(err).Should(gomega.BeNil())
+		chainANonce++
+		chainAExAddr = addr
+		signed, err := auth.Signer(auth.From, tx)
+		gomega.Expect(err).Should(gomega.BeNil())
+
+		log.Info("Waiting for contract creation on chain A to be accepted")
+		<-newHeads
+		recp, err := chainAWSClient.TransactionReceipt(ctx, signed.Hash())
```
aaronbuchwald, line 191: nit: `txReceipt`. We should use as clear variable names as possible

aaronbuchwald (changes requested):
Can we make this PR add a test case for interacting with the warp precompile via contract rather than replacing the existing direct interaction via EOA?

tests/warp/warp_test.go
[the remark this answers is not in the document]
morrisettjohn: Sounds good to me!

tests/warp/warp_test.go
morrisettjohn: Agreed, it will definitely make it a lot cleaner

tests/warp/warp_test.go
morrisettjohn: Sounds good!

tests/warp/warp_test.go
morrisettjohn: sure!

morrisettjohn: I just left it as a ginkgo test variable

tests/warp/warp_test.go
darioush: I think we should embed this.

morrisettjohn: I think it makes everything a little cleaner, I added it back in

morrisettjohn:
closed for now in favor of https://github.com/ava-labs/subnet-evm/compare/warp-contract...warp-contract-testing?expand=1, as I had to make sure my commits were all signed

PR #728, opened 2023-07-12
precompile accepter should take logIdx

darioush:
## Why this should be merged
The 3rd argument to Accept should be the logIndex not the txIndex

## How this works
Fixes above and adds a UT

## How this was tested
UT

## How is this documented
N/A


plugin/evm/block.go
```diff
@@ -89,8 +91,7 @@ func (b *Block) Accept(context.Context) error {
 // contract.Accepter
 // This function assumes that the Accept function will ONLY operate on state maintained in the VM's versiondb.
 // This ensures that any DB operations are performed atomically with marking the block as accepted.
-func (b *Block) handlePrecompileAccept(sharedMemoryWriter *sharedMemoryWriter) error {
-	rules := b.vm.chainConfig.AvalancheRules(b.ethBlock.Number(), b.ethBlock.Timestamp())
+func (b *Block) handlePrecompileAccept(rules *params.Rules, sharedMemoryWriter *sharedMemoryWriter) error {
```
darioush, line 94: added rules to make testing easier

plugin/evm/block.go
```diff
@@ -103,20 +104,18 @@ func (b *Block) handlePrecompileAccept(sharedMemoryWriter *sharedMemoryWriter) e
...
 		return fmt.Errorf("failed to fetch receipts for accepted block with non-empty root hash (%s) (Block: %s, Height: %d)", b.ethBlock.ReceiptHash(), b.ethBlock.Hash(), b.ethBlock.NumberU64())
 	}
-
-	for txIndex, receipt := range receipts {
-		for _, log := range receipt.Logs {
+	acceptCtx := &precompileconfig.AcceptContext{
+		SnowCtx:      b.vm.ctx,
+		SharedMemory: sharedMemoryWriter,
+		Warp:         b.vm.warpBackend,
+	}
```
darioush, line 111: doesn't seem needs to be in the loop

plugin/evm/block.go
```diff
@@ -103,20 +104,18 @@ func (b *Block) handlePrecompileAccept(sharedMemoryWriter *sharedMemoryWriter) e
...
 				continue
 			}
-
-			acceptCtx := &precompileconfig.AcceptContext{
-				SnowCtx:      b.vm.ctx,
-				SharedMemory: sharedMemoryWriter,
-				Warp:         b.vm.warpBackend,
-			}
-			if err := accepter.Accept(acceptCtx, log.TxHash, txIndex, log.Topics, log.Data); err != nil {
+			if err := accepter.Accept(acceptCtx, log.TxHash, logIdx, log.Topics, log.Data); err != nil {
```
darioush, line 118: ^ actual change

ceyonur (approved):
great catch!

aaronbuchwald (approved):
LGTM, great catch!

PR #731, opened 2023-07-14
Fix BlockContext comment

martineckardt:
## Why this should be merged
The current comment suggest that the BlockContext supplied to the precompile during execution is the block context during the activation of the precompile upgrade. This is not the case. 

## How this works
Change comment to account both for execution and upgrade.

## How this was tested
No test necessary, since only a comment.


aaronbuchwald (approved):
LGTM

PR #737, opened 2023-07-18
Fix precompile test gen template

RodrigoVillar:
## Why this should be merged

This pull requests fixes the following bugs regarding the function Benchmark in `precompile_contract_test_template.go` : 

- We capitalize `module` (an undefined variable) so that we can pass in the variable `Module` defined in `module.go` for a given precompile
- We rename `newStateDB(b)` (an undefined function) to `state.NewTestStateDB(b)` (the function used to define a new test state for testing a precompile)

## How this works

We modified the arguments to `test.Bench` within `Benchmark`

## How this was tested

After running `./scripts/build_test.sh` in precompile-evm, we are able to compile` precompile_contract_test_template.go
`


aaronbuchwald (approved):
LGTM

PR #744, opened 2023-07-19
Clean up gossip

anusha-ctrl:
## Why this should be merged
- Copies changes from coreeth -> https://github.com/ava-labs/coreth/commit/61baf6dc01755513c195bd54c218b942d2d113d7
- Should speed up gossiping within subnet-evm
- Should improve TPS times 

## How this works
PR description copied from @aaronbuchwald 
https://github.com/ava-labs/coreth-internal/pull/1253

This PR removes/refactors two delays in the path of mempool gossip.

Sleep 100 ms before gossiping newly received transactions to the network
This was added prior to Snowman++ to avoid the situation that a new node received transactions and then gossiped them immediately resulting in multiple nodes building a block in the expected case, which lead to unnecessary contention in consensus.

Since we've added Snowman++, this is no longer necessary except in the case that either 1) we are the next block proposer and the effort to gossip is likely to be wasted or 2) the proposer window has expired and we have fallen back to anyone proposing.

This PR leaves case 1 to a future improvement, since this is usually a very small optimization to reduce the wasted effort.

Case 2 is ignored as well since this is no longer the normal case. If the proposer window completely elapses now, then it usually indicates a network problem where multiple nodes may be trying to build a block anyways and performing mempool gossip at the same time does not change that situation.

Use a different delay for minimum delay since last gossip and frequency of gossip
There are two cases where the gossiper will call gossipEthTxs(false) 1) it received new transactions in the mempool 2) it a ticker that gossips every 500ms.

Since the ticker gossips every 500ms, when we receive new transactions and check if we should gossip them with force=false, we will always have gossiped in the past 500ms except in the case that there is a bit of variability based on the time that we set lastGossipTime and how much time performing that gossip actually takes, which should be very small.

Therefore, there's no real reason to have both of these cases if they are going to use the exact same delay.

The goal of this PR is to reduce some unnecessary latency without doing a full overhaul of the protocol, so rather than removing one or using a different mechanism, this PR only reduces the minimum time since last gossip was performed that is used, so that we will gossip a bit more aggressively.
## How this was tested
n/a
## How is this documented
n/a

### DIFF
--- a/plugin/evm/block_builder.go
+++ b/plugin/evm/block_builder.go
@@ -17,16 +17,6 @@ import (
 )
 
 const (
-	// waitBlockTime is the amount of time to wait for BuildBlock to be
-	// called by the engine before deciding whether or not to gossip the
-	// transaction that triggered the PendingTxs message to the engine.
-	//
-	// This is done to reduce contention in the network when there is no
-	// preferred producer. If we did not wait here, we may gossip a new
-	// transaction to a peer while building a block that will conflict with
-	// whatever the peer makes.
-	waitBlockTime = 100 * time.Millisecond
-
 	// Minimum amount of time to wait after building a block before attempting to build a block
 	// a second time without changing the contents of the mempool.
 	minBlockBuildingRetryDelay = 500 * time.Millisecond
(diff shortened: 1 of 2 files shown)


aaronbuchwald (approved):
LGTM

darioush (approved):

PR #749, opened 2023-07-21
Run ginkgo precompile tests in parallel (#712)

darioush:
This PR merges master,  undoes some other whitespace changes in test_blockchain
It includes a fix to call verify on the chain config in genesis outside of subnet evm but only verifies the fee config if it is non-empty so tests pass.


anusha-ctrl (approved):
lgtm

PR #750, opened 2023-07-21
Revert "revert unpack changes"

aaronbuchwald:
This reverts commit 900eb39da6ae7690c05e64a43589568303cbd403 and migrates the changes from https://github.com/ethereum/go-ethereum/pull/26568.

This was reverted to separate it from the rest of the coreth sync changes.

### DIFF
--- a/accounts/abi/unpack.go
+++ b/accounts/abi/unpack.go
@@ -29,6 +29,7 @@ package abi
 import (
 	"encoding/binary"
 	"fmt"
+	"math"
 	"math/big"
 	"reflect"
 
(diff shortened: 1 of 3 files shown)


darioush (approved):

anusha-ctrl (approved):
lgtm seems consistent with geth.

accounts/abi/unpack.go
```diff
@@ -43,43 +44,72 @@ var (
...
+			}
+			return uint32(u64), nil
 		case 64:
-			return binary.BigEndian.Uint64(b[len(b)-8:])
+			if !isu64 {
+				return nil, errBadUint64
+			}
+			return u64, nil
 		default:
 			// the only case left for unsigned integer is uint256.
```
morrisettjohn, line 74: Could we make this a case 256 and make default return an error?  I think it's better to be overly cautious, and if something incredibly bizarre does happen where typ.size is an invalid length, it's going to be a lot easier to find the issue.

accounts/abi/unpack.go
```diff
@@ -43,43 +44,72 @@ var (
...
+		}
+		return int32(i64), nil
 	case 64:
-		return int64(binary.BigEndian.Uint64(b[len(b)-8:]))
+		if !isi64 {
+			return nil, errBadInt64
+		}
+		return i64, nil
 	default:
 		// the only case left for integer is int256
```
morrisettjohn, line 110: Same comment as above

morrisettjohn (changes requested):
LGTM, I have two nits that should never be an issue - but then again you never know!

accounts/abi/unpack.go
aaronbuchwald: I don't think there's a need to be overly cautious here. There's a finite set of valid int/uint types and I'd prefer to keep this aligned with upstream.

PR #751, opened 2023-07-23
Fix nil types tmpl

ceyonur:
## Why this should be merged

Fixes int types for uint and int8,16,32,64. also adds byte types.

## How this works

changes the `converToNil` function and adds new cases.

## How this was tested

Locally generated precompile files

## How is this documented

No need as this is part of a generator tool

### DIFF
--- a/accounts/abi/bind/bind.go
+++ b/accounts/abi/bind/bind.go
@@ -662,6 +662,8 @@ func convertToNil(input abi.Type) string {
 		return "false"
 	case abi.AddressTy:
 		return "common.Address{}"
+	case abi.HashTy:
+		return "common.Hash{}"
 	case abi.FixedBytesTy:
 		return fmt.Sprintf("[%d]byte{}", input.Size)
 	case abi.BytesTy:
--- a/accounts/abi/bind/bind.go
+++ b/accounts/abi/bind/bind.go
@@ -650,6 +650,11 @@ func decapitalise(input string) string {
 func convertToNil(input abi.Type) string {
 	switch input.T {
 	case abi.IntTy, abi.UintTy:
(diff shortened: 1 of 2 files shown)


accounts/abi/bind/bind.go
```diff
@@ -650,6 +650,11 @@ func decapitalise(input string) string {
 func convertToNil(input abi.Type) string {
 	switch input.T {
 	case abi.IntTy, abi.UintTy:
+		parts := regexp.MustCompile(`(u)?int([0-9]*)`).FindStringSubmatch(input.String())
+		switch parts[2] {
+		case "8", "16", "32", "64":
+			return "0"
+		}
```
aaronbuchwald, line 657: Can we add unit tests for `convertToNil` ?

aaronbuchwald: If we need a regex here, should we compile once with `MustCompile` and re-use it rather than compiling every time this is called?

ceyonur: I think using `*new(Type)` would be a more straightforward solution. I removed all of these and converted template to use that instead.

aaronbuchwald: Won't this generate some potentially weird code like `*new(int)` when we really just want 0?

ceyonur: Yes but `convertToNil` cannot correctly detect array types + custom structs. We can expand the function by copying the`bindTypeGo`, but seemed a bit too much copy pasta to me.

This is also being used by ABI binder: https://github.com/ava-labs/subnet-evm/blob/master/accounts/abi/bind/template.go#L344

darioush (approved):

aaronbuchwald (approved):

PR #759, opened 2023-07-27
Generalize test suits

ceyonur:
## Why this should be merged

Refactors e2e test suites to generalize functions for different genesis paths and use cases. 

## How this was tested

E2E tests should work

## How is this documented

Current documentation + comments in code covers these.


tests/utils/subnet.go
```diff
@@ -105,15 +167,42 @@ func GetDefaultChainURI(blockchainID string) string {
...
+// CreateAndRunHardhatTests creates a subnet and blockchain and then
+// runs the hardhat tests on the new blockchain with default parameters.
+//
+//	Default parameters are:
+//
+// 1. Genesis file is located at ./tests/precompile/genesis/<test>.json
+// 2. Hardhat contract environment is located at ./contracts
+// 3. Hardhat test file is located at ./contracts/test/<test>.ts
+// 4. npx is available in the ./contracts directory
+func CreateAndRunHardhatTests(ctx context.Context, test string) {
```
aaronbuchwald, line 179: Seems like we're moving towards a point where we should stop making these assumptions and pass them in from the caller.

tests/precompile/solidity/suites.go
```diff
@@ -12,59 +12,94 @@ import (
...
+				// use the utils.RunTestCMD. See utils.RunDefaultHardhatTests for an example.
+				subnetsSuite.RunHardhatTests(ctx, "your_precompile")
+			})
+		*/
 	})
-
+}
+
+// Registers the Synchronized Precompile Tests
+// Before running the tests, this function creates the Avalanche locan network without explicitly creating any subnets,
```
aaronbuchwald, line 87:
```suggestion
// Before running the tests, this function creates the Avalanche local network without explicitly creating any subnets,
```

tests/utils/command.go
```diff
@@ -19,21 +18,7 @@ import (
...
-)
-
-// At boot time subnets are created, one for each test suite. This global
-// variable has all the subnets IDs that can be used.
-//
-// One process creates the AvalancheGo node and all the subnets, and these
-// subnets IDs are passed to all other processes and stored in this global
-// variable
-var BlockchainIDs map[string]string
+const ()
```
aaronbuchwald, line 21: remove

tests/utils/command.go
```diff
@@ -103,34 +79,33 @@ func RegisterNodeRun() {
...
 		gomega.Expect(startCmd).ShouldNot(gomega.BeNil())
 		gomega.Expect(startCmd.Stop()).Should(gomega.BeNil())
+		// TODO add a new node to bootstrap off of the existing node and ensure it can bootstrap all subnets
+		// created during the test
 	})
 }
+
+// RunTestCMD runs a given test command with the given rpcURI
+// It also waits for the test ping to succeed before running the test command
+func RunTestCMD(testCMD *exec.Cmd, rpcURI string) {
```
aaronbuchwald, line 98: Rather than moving this here, could we put this code inside of `runHardhatTests`. It's only used there and this is setting a very specific environment variable coupled to the logic of that function, so it doesn't make much sense to me to export this function.

tests/utils/subnet.go
```diff
@@ -9,41 +9,103 @@ import (
...
-	log.Info("Sleeping to wait for test ping", "rpcURI", rpcURI)
-	client, err := NewEvmClient(rpcURI, 225, 2)
-	gomega.Expect(err).Should(gomega.BeNil())
-
-	bal, err := client.FetchBalance(context.Background(), common.HexToAddress(""))
-	gomega.Expect(err).Should(gomega.BeNil())
-	gomega.Expect(bal.Cmp(common.Big0)).Should(gomega.Equal(0))
+type SubnetSuite struct {
+	GetBlockchainID func(alias string) string
+}
```
aaronbuchwald, line 32: This seems pretty jank. Can we define a simple type that stores a map and provides a read/write lock to support this functionality instead?

tests/precompile/solidity/suites.go
```diff
@@ -12,59 +12,94 @@ import (
...
+// and then runs the hardhat tests for each one synchronously.
+/*
+func RegisterSyncTests() {
```
aaronbuchwald, line 90: Should we remove this block?

tests/utils/command.go
ceyonur: moved

tests/utils/subnet.go
ceyonur: done

tests/precompile/solidity/suites.go
ceyonur: My intention was to provide a way to show how synced tests can be written. But I guess we don't need that.

tests/utils/command.go
```diff
@@ -103,34 +79,33 @@ func RegisterNodeRun() {
...
+
+	err := os.Setenv("RPC_URI", rpcURI)
+	gomega.Expect(err).Should(gomega.BeNil())
+	log.Info("Running test command", "cmd", testCMD.String())
+
+	out, err := testCMD.CombinedOutput()
+	fmt.Printf("\nCombined output:\n\n%s\n", string(out))
+	if err != nil {
+		fmt.Printf("\nErr: %s\n", err.Error())
+	}
```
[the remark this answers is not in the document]
ceyonur: removed

tests/utils/subnet.go
```diff
@@ -9,41 +9,103 @@ import (
...
+	return &globalSuite
+}
 
-	out, err := testCMD.CombinedOutput()
-	fmt.Printf("\nCombined output:\n\n%s\n", string(out))
-	if err != nil {
-		fmt.Printf("\nErr: %s\n", err.Error())
-	}
-	gomega.Expect(err).Should(gomega.BeNil())
+// RunDefaultHardhatTests runs the hardhat tests on a given blockchain ID
```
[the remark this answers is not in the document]
ceyonur: changed

tests/utils/subnet.go
```diff
@@ -42,6 +51,10 @@ func CreateSubnetsSuite(genesisFiles map[string]string) *SubnetSuite {
 	// processes except the first process that starts AvalancheGo
 	var startCmd *cmd.Cmd
 
+	// TODO: find a better way rather than using a global variable
```
darioush, line 54: do we need this TODO?

darioush (approved):

tests/utils/subnet.go
ceyonur: removed

aaronbuchwald (approved):

PR #762, opened 2023-07-28
Draft: Add handshake codec

nytzuga:
The goal of the this codec is to serialize `params.UpgradeConfig` in a deterministic way, to be hashed later. This hash is going to be share by nodes at handshake, so they can determine if they have the same upgrade config.

It has to be deterministic, otherwise same configs may have different hashes, making them believe they are on different configs.

This attempt leverages JSON to serialize the `params.UpgradeConfig`, but using `sortedMarshal` which walks over the object using reflection. Each object is sorted by their key, any scalar value is just serialized with regular JSON. Because the objects are sorted, `{z: 1, b: 2}` and `{b: 2, z:1}` are going to be identical when hashing.

The ideal solution would involve writing our own `codec`, based on `reflectcodec`, but extending it so the structs are sorted by their keys, even nested structs.

Another important thing that we should support is that this new codec that is being proposed should accept properties with the `json:` or `serialize:` annotation, because we're using existing structs that are defined and maintined outside of this context. Right now JSON is being used inside `UpgradeConfigInternal` to bypass that limitation


nytzuga:
I think the main thing to be done is to write our own codec, that should extend `reflectcodec` (we can even host it in this repo) to sort all the [properties inside a Struct before serializing](https://github.com/ava-labs/subnet-evm/pull/762/files#diff-602a8dc42859ecfee7ce84f3925c5c188befe393b90d7fd3cf662c3b0624031bR31-R42). This sorting is required so, no matter how the struct is defined, we have a pretty standarized way of serializing. The ultimate goal is to have the same binary and hash for the configuration.

The new codec should also serialize anything `json:*` as well as it does with `serialize:"true"`. Otherwise, all precompiles must have the `serialize:"true"`, which is fine for now, but too error prone for the future.'

Right now this PR works, but it is suboptimal as it uses raw-JSON under the hood. I don't like that.

plugin/evm/message/handshake/upgrade_config_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"testing"
+
+	"github.com/ava-labs/subnet-evm/params"
+	"github.com/ava-labs/subnet-evm/precompile/contracts/nativeminter"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestSerialize(t *testing.T) {
```
ceyonur, line 11: can we add more test cases with: 
* Enabled/Admin addresses
* Initial configs (like mint config)
* State Upgrades

precompile/contracts/feemanager/config.go
```diff
@@ -17,7 +17,7 @@ var _ precompileconfig.Config = &Config{}
 type Config struct {
 	allowlist.AllowListConfig // Config for the fee config manager allow list
 	precompileconfig.Upgrade
```
ceyonur, line 19: why didn't we serialize those?

plugin/evm/message/handshake/upgrade_config_test.go
ceyonur: I think we need to make sure:
* all of our precompiles can be deserializable
* add serialize tags to precompile config template
* if possible add a test to `config_test.go`so that new precompiles can be tested out when generated.

plugin/evm/message/handshake/codec.go
```diff
@@ -0,0 +1,36 @@
+// (c) 2019-2023, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
```
aaronbuchwald, line 2: nit: could we add the same license header to all of the newly added files?

params/network_upgrades.go
```diff
@@ -64,16 +64,14 @@ func GetMandatoryNetworkUpgrades(networkID uint32) MandatoryNetworkUpgrades {
 	}
 }
 
-// OptionalNetworkUpgrades includes overridable and optional Subnet-EVM network upgrades.
-// These can be specified in genesis and upgrade configs.
-// Timestamps can be different for each subnet network.
-// TODO: once we add the first optional upgrade here, we should uncomment TestVMUpgradeBytesOptionalNetworkUpgrades
-type OptionalNetworkUpgrades struct{}
+type OptionalNetworkUpgrades struct {
+	Updates []Fork `json:"serialize,omitempty" serialize:"true"`
```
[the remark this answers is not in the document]
nytzuga: Working on it now

plugin/evm/message/handshake/upgrade_config_test.go
ceyonur: ^^^

precompile/contracts/feemanager/config.go
```diff
@@ -15,9 +15,9 @@ var _ precompileconfig.Config = &Config{}
 // Config implements the StatefulPrecompileConfig interface while adding in the
 // FeeManager specific precompile config.
 type Config struct {
-	allowlist.AllowListConfig // Config for the fee config manager allow list
-	precompileconfig.Upgrade
-	InitialFeeConfig *commontype.FeeConfig `json:"initialFeeConfig,omitempty"` // initial fee config to be immediately activated
+	allowlist.AllowListConfig `serialize:"true"`
+	precompileconfig.Upgrade  `serialize:"true"`
```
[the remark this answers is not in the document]
nytzuga: That's correct, I'll add it.

params/config.go
```diff
@@ -166,7 +166,7 @@ type UpgradeConfig struct {
 	// Config for optional timestamps that enable network upgrades.
 	// Note: if OptionalUpgrades is specified in the JSON all previously activated
 	// forks must be present or upgradeBytes will be rejected.
-	OptionalNetworkUpgrades *OptionalNetworkUpgrades `json:"networkUpgrades,omitempty"`
+	OptionalNetworkUpgrades *OptionalNetworkUpgrades
```
nytzuga, line 169: This is needed, as far as I did my research, so the JSON decoder (from the local config) will flatten and serialize their children instead (`OptionalNetworkUpgrades` has instructions on how to deserialize the JSON file). 

TODO: Write tests on how to parse the config

nytzuga:
This PR depends on https://github.com/ava-labs/avalanchego/pull/2171

plugin/evm/message/handshake/upgrade_config.go
```diff
@@ -0,0 +1,118 @@
...
+	// Config for modifying state as a network upgrade.
+	StateUpgrades []params.StateUpgrade `serialize:"true"`
+
+	// Config for enabling and disabling precompiles as network upgrades.
+	PrecompileUpgrades []rawPrecompileUpgrade `serialize:"true"`
+}
+
+type UpgradeConfigMessage struct {
+	Bytes []byte
+	Hash  []byte
```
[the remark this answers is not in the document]
nytzuga: I think the idea was to use bytes to make it subnet-agnostic. Do you think we should be using ids.ID?

precompile/contracts/feemanager/config.go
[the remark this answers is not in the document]
nytzuga: Done in 9c954d5c

plugin/evm/message/handshake/upgrade_config_message.go
```diff
@@ -0,0 +1,117 @@
...
+
+	bytes, err := Codec.Marshal(Version, wrappedConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	hash := crypto.Keccak256(bytes)
+	return &UpgradeConfigMessage{
+		bytes: bytes,
+		hash:  hash[:8],
```
ceyonur, line 115: can we send full hash? I don't think it worth saving 24 bytes. We can also convert `hash` to `common.Hash` and use `crypto.Keccak256Hash`.

plugin/evm/message/handshake/upgrade_config_test.go
ceyonur: This is still pending.

params/network_upgrades.go
```diff
@@ -64,16 +66,40 @@ func (m *MandatoryNetworkUpgrades) mandatoryForkOrder() []fork {
...
+
+type OptionalNetworkUpgrades struct {
+	Test *OptionalFork `json:"test,omitempty" serialize:"true,nullable"`
+}
 
 func (n *OptionalNetworkUpgrades) CheckOptionalCompatible(newcfg *OptionalNetworkUpgrades, time uint64) *ConfigCompatError {
 	return nil
 }
 
 func (n *OptionalNetworkUpgrades) optionalForkOrder() []fork {
```
ceyonur, line 99: can we use the same logic as we do it in `mandatoryForkOrder`

ceyonur (changes requested):

plugin/evm/message/handshake/upgrade_config_message.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+// Attempts to parse a networkUpgradeConfigMessage from a []byte
+//
+// This function attempts to parse a stream of bytes as a
+// networkUpgradeConfigMessage (as serialized from
+// UpgradeConfigToNetworkMessage).
+//
+// The function returns a reference of *params.UpgradeConfig
+func NewUpgradeConfigMessageFromBytes(bytes []byte) (*params.UpgradeConfig, error) {
```
ceyonur, line 45: why is this returning `*params.UpgradeConfig` instead of `UpgradeConfigMessage`?

plugin/evm/message/handshake/upgrade_config_message_test.go
```diff
@@ -0,0 +1,44 @@
...
+	require.NoError(t, err)
+
+	message2, err := NewUpgradeConfigMessage(config)
+	require.NoError(t, err)
+
+	config3, err := NewUpgradeConfigMessageFromBytes(message2.Bytes())
+	require.NoError(t, err)
+
+	message3, err := NewUpgradeConfigMessage(config3)
+	require.NoError(t, err)
```
ceyonur, line 39: These seem repetitive and unnecessary to me. If message2 and message are equal, these are guaranteed to be equal, right?.

plugin/evm/message/handshake/upgrade_config_test.go
nytzuga: I'm writing more tests now

plugin/evm/message/handshake/upgrade_config_message_test.go
nytzuga: 1. `message` is from local configs
2. `message2` is from a deserialization (simulating a network exchange)
3. `message3` is to check if deserializing from a network exchange and serializing it again is consistent (generates the same hash).

precompile/allowlist/config.go
```diff
@@ -15,9 +15,9 @@ var ErrCannotAddManagersBeforeDUpgrade = fmt.Errorf("cannot add managers before
 
 // AllowListConfig specifies the initial set of addresses with Admin or Enabled roles.
 type AllowListConfig struct {
-	AdminAddresses   []common.Address `json:"adminAddresses,omitempty"`   // initial admin addresses
-	ManagerAddresses []common.Address `json:"managerAddresses,omitempty"` // initial manager addresses
-	EnabledAddresses []common.Address `json:"enabledAddresses,omitempty"` // initial enabled addresses
+	AdminAddresses   []common.Address `json:"adminAddresses,omitempty" serialize:"true"`   // initial admin addresses
+	ManagerAddresses []common.Address `json:"managerAddresses,omitempty" serialize:"true"` // initial manager addresses
+	EnabledAddresses []common.Address `json:"enabledAddresses,omitempty" serialize:"true"` // initial enabled addresses
```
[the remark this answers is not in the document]
nytzuga: I added more tests to c261ce27. Do you think that is enough @ceyonur ? I'm checking the other configs to add more tests if needed.

params/network_upgrades.go
nytzuga: The expected output of `optionalForkOrder` is the same as `mandatoryForkOrder`. However, to make things less error-prone `OptionalFork` has a method `ToFork` that will return the expected struct.

accounts/abi/bind/precompilebind/precompile_config_template.go
```diff
@@ -26,7 +26,7 @@ var _ precompileconfig.Config = &Config{}
 // adds specific configuration for {{.Contract.Type}}.
 type Config struct {
 	{{- if .Contract.AllowList}}
-	allowlist.AllowListConfig
+	allowlist.AllowListConfig ` + "`" + `serialize:"true"` + "`" + `
 	{{- end}}
 	precompileconfig.Upgrade
```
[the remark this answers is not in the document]
nytzuga: Done

plugin/evm/message/handshake/upgrade_config_message.go
ceyonur: Can we rename the function to reflect that?

accounts/abi/bind/precompilebind/precompile_config_template.go
```diff
@@ -26,9 +28,13 @@ var _ precompileconfig.Config = &Config{}
...
 	{{- if .Contract.AllowList}}
-	allowlist.AllowListConfig
+	// Serialize allowList to let nodes exchange information about their
+	// precompile configs
+	allowlist.AllowListConfig ` + "`" + `serialize:"true"` + "`" + `
 	{{- end}}
-	precompileconfig.Upgrade
+	// Serialize the Upgrade config to let nodes exchange information about
+	// their precompile configs
+	precompileconfig.Upgrade ` + "`" + `serialize:"true"` + "`" + `
```
ceyonur, line 37: do we need these?

commontype/fee_config.go
```diff
@@ -144,6 +145,57 @@ func (f *FeeConfig) checkByteLens() error {
 	return nil
 }
 
+func (c *FeeConfig) getBigIntToSerialize() []**big.Int {
+	return []**big.Int{
+		&c.GasLimit, &c.MinBaseFee, &c.TargetGas, &c.BaseFeeChangeDenominator,
+		&c.MinBlockGasCost, &c.MaxBlockGasCost, &c.BlockGasCostStep,
+	}
+}
```
ceyonur, line 153: what is this used for? this look a bit weird.

ceyonur (changes requested):

precompile/contracts/nativeminter/config.go
```diff
@@ -19,9 +22,9 @@ var _ precompileconfig.Config = &Config{}
 // Config implements the StatefulPrecompileConfig interface while adding in the
 // ContractNativeMinter specific precompile config.
 type Config struct {
-	allowlist.AllowListConfig
-	precompileconfig.Upgrade
-	InitialMint map[common.Address]*math.HexOrDecimal256 `json:"initialMint,omitempty"` // addresses to receive the initial mint mapped to the amount to mint
+	allowlist.AllowListConfig `serialize:"true"`
```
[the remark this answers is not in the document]
nytzuga: Nope. I'll remove them now.

accounts/abi/bind/precompilebind/precompile_config_template.go
nytzuga: Fixed in 2309af50

precompile/allowlist/config.go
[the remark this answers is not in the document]
nytzuga: Coming back to this @ceyonur. Do you think that having all the handshake tests inside a single file (serializing and deserializing the configs and hashing them) is not better to this alternative?

precompile/contracts/nativeminter/config.go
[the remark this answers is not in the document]
nytzuga: Fixed in 2309af50

precompile/precompileconfig/config.go
```diff
@@ -30,6 +30,10 @@ type Config interface {
 	Equal(Config) bool
 	// Verify is called on startup and an error is treated as fatal. Configure can assume the Config has passed verification.
 	Verify(ChainConfig) error
+	// Converts the current configuration to a vector of bytes. This vector of bytes is expected to be deterministic
+	ToBytes() ([]byte, error)
+	// Instantiates an object from a vector of bytes (the output of ToBytes())
+	FromBytes([]byte) error
```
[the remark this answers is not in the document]
nytzuga: We could do that. It's up to the people.

precompile/contracts/deployerallowlist/config.go
```diff
@@ -57,3 +58,33 @@ func (c *Config) Equal(cfg precompileconfig.Config) bool {
...
+	}
+
+	return p.Bytes, nil
+}
+
+func (c *Config) FromBytes(bytes []byte) error {
+	p := wrappers.Packer{
+		Bytes: bytes,
+	}
+	if err := c.AllowListConfig.FromBytesWithPacker(&p); err != nil {
```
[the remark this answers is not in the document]
nytzuga: Allowlist is not a config, instead it is a sub-config. It can be done that way I just thought it was more efficient to give a reference to the parker, otherwise we would have to know exactly how big the slice of bytes to share.

precompile/allowlist/config.go
```diff
@@ -37,6 +40,43 @@ func (c *AllowListConfig) Configure(chainConfig precompileconfig.ChainConfig, pr
 	return nil
 }
 
+func (c *AllowListConfig) packAddresses(addresses []common.Address, p *wrappers.Packer) error {
+	sort.Slice(addresses, func(i, j int) bool {
+		return bytes.Compare(addresses[i][:], addresses[j][:]) < 0
+	})
```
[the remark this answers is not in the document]
nytzuga: To have a consistent view of the hashmap.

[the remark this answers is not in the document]
nytzuga: Just to let you know, we need to be deterministic always.

That being said

```
{"foo": 5, "bar": 10}
```

It should be identical to 

```
{"bar": 10, "foo": 5}
```

To have the same hash. That cannot be done without sorting the hashmaps.

precompile/precompileconfig/config.go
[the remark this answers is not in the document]
nytzuga: Fixed in f1d3d075

params/state_upgrade.go
```diff
@@ -16,18 +16,18 @@ import (
...
 
 // StateUpgradeAccount describes the modifications to be made to an account during
 // a state upgrade.
 type StateUpgradeAccount struct {
-	Code          hexutil.Bytes               `json:"code,omitempty"`
-	Storage       map[common.Hash]common.Hash `json:"storage,omitempty"`
-	BalanceChange *math.HexOrDecimal256       `json:"balanceChange,omitempty"`
+	Code          hexutil.Bytes               `json:"code,omitempty" serialize:"true"`
+	Storage       map[common.Hash]common.Hash `json:"storage,omitempty" serialize:"true"`
+	BalanceChange *math.HexOrDecimal256       `json:"balanceChange,omitempty" serialize:"true"`
```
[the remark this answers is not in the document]
nytzuga: We no longer rely on the codec. In the next commit, all traces of the reflect_codec will be gone.

params/network_upgrades.go
```diff
@@ -67,29 +65,25 @@ func (m *MandatoryNetworkUpgrades) mandatoryForkOrder() []fork {
 }
 
 type OptionalFork struct {
```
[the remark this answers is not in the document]
nytzuga: Good idea, I don't think we do.

plugin/evm/message/handshake/upgrade_config_message.go
nytzuga: Is the name better now @ceyonur ?

precompile/allowlist/config.go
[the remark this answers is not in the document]
nytzuga: The same thing happens with a vector of addresses, in this particular context `["addr1", "addr2"]` is equal to `["addr2", "addr1"]`

plugin/evm/message/handshake/upgrade_config_message.go
nytzuga: @ceyonur I think now it is better. I dropped all the intermediary types and I added [MarshalBinary](https://github.com/ava-labs/subnet-evm/blob/handshake-codec/params/config.go#L197-L244), [UnmarshalBinary](https://github.com/ava-labs/subnet-evm/blob/handshake-codec/params/config.go#L246-L296) and [Hash](https://github.com/ava-labs/subnet-evm/blob/handshake-codec/params/config.go#L189-L195) to the `UpgradeConfig` struct. That way it is cleaner and much simpler.

I'm working now to re-structure the tests and add more coverage.

commontype/fee_config.go
nytzuga: @aaronbuchwald Is this kinda of code not desired?  It is a canonical list of numbers and how they are marshal and unmarshal. It is way error prone and it takes advantage of the pointer to a pointer.

plugin/evm/message/handshake/codec.go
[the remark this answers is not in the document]
nytzuga: This is no longer relevant since the serialization has been moved (more details [here](https://github.com/ava-labs/subnet-evm/pull/762#discussion_r1426682952))

precompile/allowlist/config.go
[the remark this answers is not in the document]
nytzuga: Solved in 1946de42fbc6c23b9087e0b3bfd41b187a9548d7

precompile/contracts/deployerallowlist/config.go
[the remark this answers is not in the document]
nytzuga: Done in 4591ccdb

commontype/fee_config.go
aaronbuchwald: I see this is trying to make the code simpler by passing a list of pointers to pointers so we can serialize and de-serialize with a for loop instead of one by one.

It's not valid for any of these to be nil, so would it be possible to remove the double pointer and enforce that all of them are non-nil instead of allowing a boolean to indicate whether or not it's nil?

commontype/fee_config.go
```diff
@@ -144,6 +146,65 @@ func (f *FeeConfig) checkByteLens() error {
...
+	}
+	for _, bigint := range c.getBigIntToSerialize() {
+		isNil := p.UnpackBool()
+		if p.Err != nil {
+			return p.Err
+		}
+		if isNil {
+			continue
+		}
+		*bigint = big.NewInt(0).SetBytes(p.UnpackBytes())
```
aaronbuchwald, line 194: Should these be fixed bytes of size 32 since that is the max size allowed for these variables?

commontype/fee_config.go
```diff
@@ -144,6 +146,65 @@ func (f *FeeConfig) checkByteLens() error {
...
 }
 
+func (c *FeeConfig) getBigIntToSerialize() []**big.Int {
+	return []**big.Int{
+		&c.GasLimit, &c.MinBaseFee, &c.TargetGas, &c.BaseFeeChangeDenominator,
+		&c.MinBlockGasCost, &c.MaxBlockGasCost, &c.BlockGasCostStep,
+	}
+}
+
+func (c *FeeConfig) MarshalBinary() ([]byte, error) {
```
aaronbuchwald, line 156: Can we add the following test cases for the serialization of `FeeConfig`:

1. serialize and de-serialize valid config and get the same value
2. hardcoded test vector and require it results in the same value
3. unmarshal should fail if there are insufficient bytes to unpack big int values

params/config.go
```diff
@@ -67,6 +73,17 @@ var (
 		MaxBlockGasCost:  big.NewInt(1_000_000),
 		BlockGasCostStep: big.NewInt(200_000),
 	}
+
+	// For UpgradeConfig Marshal/Unmarshal
+	magicHeader              = []byte{0xff}
+	sectionHeaderSizes       = 1
+	precompileHeader         = []byte{0xaa}
+	stateUpdateHeader        = []byte{0xc8}
+	endHeader                = []byte{0}
```
aaronbuchwald, line 82: Do we need these headers? I believe these are typically used to signal what type of data they include, but we should know that when serializing/de-serializing, so I don't think these should be necessary.

params/config.go
```diff
@@ -171,6 +188,142 @@ type UpgradeConfig struct {
...
+		bytes := p.UnpackBytes()
+		if p.Err != nil {
+			return p.Err
+		}
+		if err := c.OptionalNetworkUpgrades.UnmarshalBinary(bytes); err != nil {
+			return nil
+		}
+	}
+
+	for {
```
aaronbuchwald, line 286: Let's switch to packing the number of elements of each type, so that we can proceed through field by field and unpack the number of elements, all of the elements of that type, and then proceed to the next field.

params/config.go
```diff
@@ -865,3 +1018,36 @@ func (c *ChainConfig) ToWithUpgradesJSON() *ChainConfigWithUpgradesJSON {
...
+
+// Checks if messages have the same hash
+//
+// `message` is the simulation of a configuration being parsed from the local
+// config. `message2` is parsing a message being exchanged through the network
+// (a foreign config), and `message3` is the the deserialization and
+// serialization of the foreign config. All 3 instances should have the same
+// hashing, depite maybe not being identical (some configurations may be in a
+// different order, but our hashing algorithm is resilient to those changes,
+// thanks for our serialization library, which produces always the same output.
```
aaronbuchwald, line 1030: Could you re-phrase this comment, it seems like some variable names have changed

params/config_test.go
```diff
@@ -328,3 +328,75 @@ func TestChainConfigMarshalWithUpgrades(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, config, unmarshalled)
 }
+
+func TestChainConfigMarshalWithUpgradesAndOptionalUpgrade(t *testing.T) {
+	config := ChainConfigWithUpgradesJSON{
```
aaronbuchwald, line 333: This seems to test the json marshal/unmarshal, but does not test the new marshal / unmarshal binary code.

We should add test coverage for the binary marshal / unmarshal as well.

params/config.go
```diff
@@ -865,3 +1018,36 @@ func (c *ChainConfig) ToWithUpgradesJSON() *ChainConfigWithUpgradesJSON {
...
+	bytes, err := originalConfig.MarshalBinary()
+	require.NoError(t, err)
+
+	deserializedConfig := UpgradeConfig{}
+	require.NoError(t, deserializedConfig.UnmarshalBinary(bytes))
+
+	twiceDeserialized := UpgradeConfig{}
+	newBytes, err := deserializedConfig.MarshalBinary()
+	require.NoError(t, err)
+	require.NoError(t, twiceDeserialized.UnmarshalBinary(newBytes))
```
aaronbuchwald, line 1041: When will this fail because we do it three times instead of two?

If `newBytes` is the same as `bytes` (which should be confirmed by checking that the hashes match), then I don't think that `twiceDeserialized` provides additional coverage. Afaict this will fail if `UnmarshalBinary` is itself non-deterministic on the SAME bytes.

precompile/contracts/nativeminter/config.go
```diff
@@ -95,3 +99,90 @@ func (c *Config) Verify(chainConfig precompileconfig.ChainConfig) error {
...
+		return err
+	}
+	if err := c.Upgrade.UnmarshalBinary(upgrade); err != nil {
+		return err
+	}
+	len := p.UnpackInt()
+	c.InitialMint = make(map[common.Address]*math.HexOrDecimal256, len)
+
+	for i := uint32(0); i < len; i++ {
+		key := common.BytesToAddress(p.UnpackBytes())
```
aaronbuchwald, line 176: Can we make sure to pack addresses as 20 fixed bytes and big int values as 32 fixed bytes?

precompile/contracts/feemanager/config_test.go
```diff
@@ -88,3 +89,39 @@ func TestEqual(t *testing.T) {
...
+				}, []common.Address{
+					common.BytesToAddress(common.Hex2Bytes("0000000000000000000000000000000000000000000000000000000000000050")),
+				}, &validFeeConfig), // enable at genesis
+			},
+			{
+				Config: NewDisableConfig(&t1), // disable at timestamp 1
+			},
+		},
+	}
+	params.AssertConfigHashesAndSerialization(t, &config)
```
aaronbuchwald, line 126: Could we add an expected hash to these tests, so that the tests will fail if the format changes unexpectedly? (this has the cost of needing to update the expected hash values if we change the format, so may be easiest to make this a follow up)

precompile/contracts/nativeminter/config_test.go
```diff
@@ -52,6 +53,70 @@ func TestVerify(t *testing.T) {
...
+	}
+	params.AssertConfigHashesAndSerialization(t, &config)
+}
+func TestSerializeWithAddresses(t *testing.T) {
+	var t0 uint64 = 1
+	var t1 uint64 = 11
+	config := params.UpgradeConfig{
+		PrecompileUpgrades: []params.PrecompileUpgrade{
+			{
+				Config: NewConfig(&t0, []common.Address{
```
aaronbuchwald, line 77: can we add a test with multiple addresses?

precompile/contracts/rewardmanager/config.go
```diff
@@ -20,6 +22,34 @@ type InitialRewardConfig struct {
...
+
+func (u *InitialRewardConfig) UnmarshalBinary(data []byte) error {
+	p := wrappers.Packer{
+		Bytes: data,
+	}
+	u.AllowFeeRecipients = p.UnpackBool()
+	if p.Err != nil {
+		return p.Err
+	}
+	u.RewardAddress = common.BytesToAddress(p.UnpackBytes())
```
aaronbuchwald, line 46: can we use fixed bytes of length 20?

params/state_upgrade_test.go
```diff
@@ -185,3 +185,42 @@ func TestUnmarshalStateUpgradeJSON(t *testing.T) {
...
+
+func TestSerialize(t *testing.T) {
+	var t0 uint64 = 2
+	var t1 uint64 = 1001
+	config := UpgradeConfig{
+		StateUpgrades: []StateUpgrade{
+			{
+				BlockTimestamp: &t0,
+				StateUpgradeAccounts: map[common.Address]StateUpgradeAccount{
+					common.BytesToAddress(common.Hex2Bytes("0000000000000000000000000000000000000000000000000000000000000050")): StateUpgradeAccount{
```
ceyonur, line 197: nit: can we create a variable for these addresses instead of hardcoding them into tests?

precompile/allowlist/config.go
```diff
@@ -37,6 +39,39 @@ func (c *AllowListConfig) Configure(chainConfig precompileconfig.ChainConfig, pr
 	return nil
 }
 
+func (c *AllowListConfig) packAddresses(addresses []common.Address, p *wrappers.Packer) error {
```
ceyonur, line 42: do we need these to be a receiver method (with `c`)? I think these can be good util functions in precompile/contract/utils.

accounts/abi/bind/precompilebind/precompile_config_test_template.go
```diff
@@ -61,6 +63,35 @@ func TestVerify(t *testing.T) {
...
+				{{- end}}
+				), // enable at genesis
+			},
+			{
+				Config: NewDisableConfig(&t1), // disable at timestamp 1
+			},
+		},
+	}
+	require.NotNil(t, config)
+	//params.AssertConfigHashesAndSerialization(t, &config)
```
ceyonur, line 92: why is this commented out?

params/config.go
```diff
@@ -171,6 +188,142 @@ type UpgradeConfig struct {
 	PrecompileUpgrades []PrecompileUpgrade `json:"precompileUpgrades,omitempty"`
 }
 
+func (c *UpgradeConfig) Hash() (common.Hash, error) {
+	bytes, err := c.MarshalBinary()
+	if err != nil {
+		return common.Hash{}, err
+	}
+	return crypto.Keccak256Hash(bytes), nil
+}
```
ceyonur, line 197: We should probably move all of these to a separate `upgrade_config` file in the same package.

params/config.go
```diff
@@ -865,3 +1018,36 @@ func (c *ChainConfig) ToWithUpgradesJSON() *ChainConfigWithUpgradesJSON {
...
+// Checks if messages have the same hash
+//
+// `message` is the simulation of a configuration being parsed from the local
+// config. `message2` is parsing a message being exchanged through the network
+// (a foreign config), and `message3` is the the deserialization and
+// serialization of the foreign config. All 3 instances should have the same
+// hashing, depite maybe not being identical (some configurations may be in a
+// different order, but our hashing algorithm is resilient to those changes,
+// thanks for our serialization library, which produces always the same output.
+func AssertConfigHashesAndSerialization(t *testing.T, originalConfig *UpgradeConfig) {
```
ceyonur, line 1031: Requiring this function from params package in precompiles will create a cyclic dependency most likely. I think this assertion can be reduced as suggested. So we can just readd these assertions in relevant test files without requiring to import from params package.

params/config.go
```diff
@@ -171,6 +188,142 @@ type UpgradeConfig struct {
...
+
+	isNil := p.UnpackBool()
+	if !isNil {
+		c.OptionalNetworkUpgrades = &OptionalNetworkUpgrades{}
+		bytes := p.UnpackBytes()
+		if p.Err != nil {
+			return p.Err
+		}
+		if err := c.OptionalNetworkUpgrades.UnmarshalBinary(bytes); err != nil {
+			return nil
```
ceyonur, line 282: return err?

accounts/abi/bind/precompilebind/precompile_config_template.go
```diff
@@ -26,8 +28,12 @@ var _ precompileconfig.Config = &Config{}
 // adds specific configuration for {{.Contract.Type}}.
 type Config struct {
 	{{- if .Contract.AllowList}}
+	// Serialize allowList to let nodes exchange information about their
+	// precompile configs
```
ceyonur, line 32:
```suggestion

```

accounts/abi/bind/precompilebind/precompile_config_template.go
```diff
@@ -26,8 +28,12 @@ var _ precompileconfig.Config = &Config{}
...
 	allowlist.AllowListConfig
 	{{- end}}
+	// Serialize the Upgrade config to let nodes exchange information about
+	// their precompile configs
```
ceyonur, line 36:
```suggestion
```

accounts/abi/bind/precompilebind/precompile_config_template.go
```diff
@@ -48,6 +54,14 @@ func NewConfig(blockTimestamp *uint64{{if .Contract.AllowList}}, admins []common
 	}
 }
 
+func (c * Config) MarshalBinary() ([]byte, error) {
```
ceyonur, line 57: can we add some comments here why these are important, how these can be implemented etc?

commontype/fee_config.go
nytzuga: [It is variable size](https://pkg.go.dev/math/big#Int.Bytes)

commontype/fee_config.go
nytzuga: Fixed in 2de90955940a692448c1b3fca111b0a93b3db238

params/config.go
nytzuga: Fixed in 8935259f, instead I'm having a deterministic serialization

params/config.go
nytzuga: Fixed in 8935259f

params/config.go
nytzuga: I think I made it way better in 1819327a

precompile/contracts/rewardmanager/config.go
nytzuga: Done in e079719c

accounts/abi/bind/precompilebind/precompile_config_test_template.go
```diff
@@ -61,6 +63,35 @@ func TestVerify(t *testing.T) {
...
+				[]common.Address{
+					common.BytesToAddress(common.Hex2Bytes("0000000000000000000000000000000000000000000000000000000000000020")),
+					common.BytesToAddress(common.Hex2Bytes("0000000000000000000000000000000000000000000000000000000000000030")),
+				}, []common.Address{
+					common.BytesToAddress(common.Hex2Bytes("0000000000000000000000000000000000000000000000000000000000000040")),
+				}, []common.Address{
+					common.BytesToAddress(common.Hex2Bytes("0000000000000000000000000000000000000000000000000000000000000050")),
+				},
+				{{- end}}
+				), // enable at genesis
```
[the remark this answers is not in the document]
nytzuga: Fixed in 17bf11e4

accounts/abi/bind/precompilebind/precompile_config_template.go
```diff
@@ -48,6 +54,14 @@ func NewConfig(blockTimestamp *uint64{{if .Contract.AllowList}}, admins []common
...
+	return nil, errors.New("implement MarshalBinary() method")
```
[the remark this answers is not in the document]
nytzuga: Fixed in 17bf11e4

precompile/allowlist/config.go
nytzuga: Fixed in 91bdff51

params/config.go
nytzuga: Fixed in a593a68e

commontype/fee_config.go
nytzuga: Fixed it in e8ffb133

Issue #345, opened 2022-11-04
Add "Manager" Role to AllowList


patrick-ogrady:
Add a role to the AllowList precompile that enables someone to add/remove addresses from the AllowList but not affect Admins.

This is more compatible with a hot("manager")/cold("admin") setup most partners use.

aaronbuchwald:
We should also update the existing precompiles to emit events when updates are made.

ceyonur:
I checked this one seems that it's fairly easy to change the code, but it requires a mandatory upgrade (without a precompile upgrade) if we expand the current allow list. it's probably best to wait for a avalanachego mandatory upgrade + RPC chain protocol change (both at the same time) to put this into the code if we want this to be a mandatory upgrade.

We can also add a upgradable precompile activation and check in the allow list if manager role is activated. `PrecompileAccessibleState` provides a `BlockContext` so we can check the current timestamp from `BlockContext` and `ChainConfig` to figure out if manager allow list is enabled. If not enabled, any call to manager functions will return an error.

aaronbuchwald:
That's a great idea to add this as a network upgrade to subnet-evm as opposed to one by one for each precompile by itself.

If we're going to do it that way, then we should try to include this in the next AvalancheGo network upgrade.

richardpringle:
@patrick-ogrady, could we get a little more context around the problem. I get the hot vs cold and I'm thinking adding an "Owner" role makes more sense
```
0 => no-role
1 => enabled (read perms)
2 => admin (read-write)
3 => owner (admin, but can only be removed by another owner)
```
otherwise, the `manager` role would get the 3 in order to maintain backward compatibility which would be kind of weird.

But if we can get a little more context on what exactly we want to design for when it comes to hot/cold usage, maybe we could make sure that this is going to be a little more extensible in the future.

richardpringle:
> I checked this one seems that it's fairly easy to change the code, but it requires a mandatory upgrade (without a precompile upgrade) if we expand the current allow list. it's probably best to wait for a avalanachego mandatory upgrade + RPC chain protocol change (both at the same time) to put this into the code if we want this to be a mandatory upgrade.
> 
> We can also add a upgradable precompile activation and check in the allow list if manager role is activated. `PrecompileAccessibleState` provides a `BlockContext` so we can check the current timestamp from `BlockContext` and `ChainConfig` to figure out if manager allow list is enabled. If not enabled, any call to manager functions will return an error.

[...]

Issue #692, opened 2023-06-22
Modify Precompile Contract Functions with an Activation Time


ceyonur:
We require some changes to precompile contract functions in #345 and #510. These require a way to modify/add existing precompile contract functions. 

Current idea:

* Add an optional (or mandatory) upgrade. There is another issue currently refactoring the upgrades #638 
* Expand the accessible state so contract functions can access to activation timestamps
* In precompile function use this activation time to figure out if designated upgrade has been activated.
* The strategy changes if function was added/deleted/modified
  * If a new function was added to the contract:
    * Before activation it should err out with `fmt.Errorf("invalid function selector %#x", selector)` BEFORE deducting gas i.e:
    This is equivalent of function not being found.
  * If a function was deleted from the contract:
    *  This should return the same error AFTER activation but still BEFORE deducting the gas.
  * If modified every modification should happen AFTER activation.

Some examples below. Assume we have activated `XUpgradeTimestamp`

Function added:
```
func newFunction(accessibleState contract.AccessibleState, caller common.Address, addr common.Address, input []byte, suppliedGas uint64, readOnly bool) (ret []byte, remainingGas uint64, err error) {
        if !accessibleState.Rules().IsXUpgrade() {
         return nil, suppliedGas, contract.ErrInvalidFunctionSelector 
       }

	if remainingGas, err = contract.DeductGas(suppliedGas, NewFunctionGasCost); err != nil {
		return nil, 0, err
	}
	if readOnly {
		return nil, remainingGas, vmerrs.ErrWriteProtection
	}
        ...
	// Return the packed output and the remaining gas
	return packedOutput, remainingGas, nil
}
```

Function deleted:
```
func existingFunction(accessibleState contract.AccessibleState, caller common.Address, addr common.Address, input []byte, suppliedGas uint64, readOnly bool) (ret []byte, remainingGas uint64, err error) {
        if accessibleState.Rules().IsXUpgrade() {
         return nil, suppliedGas, contract.ErrInvalidFunctionSelector 
       }

	if remainingGas, err = contract.DeductGas(suppliedGas, NewFunctionGasCost); err != nil {
		return nil, 0, err
	}
	if readOnly {
		return nil, remainingGas, vmerrs.ErrWriteProtection
	}
        ...
	// Return the packed output and the remaining gas
	return packedOutput, remainingGas, nil
}
```

Function modified:
```
[...]

PR #764, opened 2023-07-31
Add manager role

ceyonur:
## Why this should be merged

Adds manager role

Closes: #345 

## How this works

Adds a new role that can change enabled addresses, permission to only modify enabled addresses. 

* Manager Role cannot modify other managers + admins
* Manager Role is not activated until DUpgrade
* Manager role is also enabled.
* Manager roles in upgrades/genesis are not applied until DUpgrade
  * it only applies if AllowList configuration with timestamp T applied after DUpgrade. I.e if T < D, configured addresses won't have manager role *even after* Dupgrade, since it was configured before (T < D) DUpgrade and we don't reconfigure.
  * If allow list configuration is activated after DUpgrade (T >= D) and contains manager roles for addresses, those will be applied and addresses will become managers.
 * Existing precompiles will break because we have changed the `IsEnabled` signature to check if manager role is activated.
- [x] change precompile template 

## How this was tested

Added UTs

- [x] test locally with real network

## How is this documented

- [ ] Add documentation to docs repo


precompile/allowlist/role.go
```diff
@@ -5,22 +5,29 @@ package allowlist
...
+var (
+	// NoRole - this is equivalent to common.Hash{} and deletes the key from the DB when set.
+	NoRole = Role(common.BigToHash(common.Big0))
+	// EnabledRole - allowed to call the precompile.
+	EnabledRole = Role(common.BigToHash(common.Big1))
+	// Admin - allowed to both modify the allowlist and call the precompile.
+	AdminRole = Role(common.BigToHash(common.Big2))
+	// Manager - allowed to add and remove only enabled address, without being able to call the precompile (i.e not enabled). Activated only after DUpgrade.
+	// We don't specifically give enabled role to the manager, because we want to keep the backward compatibility with IsEnabled() function.
+	// TODO: we can give enabled role to the manager. For that we need to remove the IsEnabled(), so we can break other precompiles and force them to adopt.
```
ceyonur, line 21: 👁️

precompile/contract/interfaces.go
```diff
@@ -19,14 +19,17 @@ type StatefulPrecompiledContract interface {
...
+// ChainConfig defines an interface that provides information to a stateful precompile
 // about the chain configuration. The precompile can access this information to initialize
 // its state.
 type ChainConfig interface {
 	// GetFeeConfig returns the original FeeConfig that was set in the genesis.
 	GetFeeConfig() commontype.FeeConfig
 	// AllowedFeeRecipients returns true if fee recipients are allowed in the genesis.
 	AllowedFeeRecipients() bool
+	// IsDUpgrade returns true if the time is after the DUpgrade.
+	// TODO: Should we get a more general function like "AvalancheRules"?
```
ceyonur, line 31: 👁️

precompile/allowlist/role.go
ceyonur: I guess we have to give enabled to the manager because `TxAllowList` checks if the role is enabled to accept a tx. Meaning that even if manager role can alter someone else's enabled role, they won't be able to issue the tx (because their tx will be rejected by `preCheck` in [state_transition](https://github.com/ava-labs/subnet-evm/blob/master/core/state_transition.go#L350))

plugin/evm/vm.go
```diff
@@ -294,7 +295,10 @@ func (vm *VM) Initialize(
 
 	// We enforce network upgrades here, regardless of the chain config
 	// provided in the genesis file.
-	g.Config.MandatoryNetworkUpgrades = params.GetMandatoryNetworkUpgrades(chainCtx.NetworkID)
+	// Skip if we are running unit tests.
+	if chainCtx.NetworkID != avagoconstants.UnitTestID {
```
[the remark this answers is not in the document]
ceyonur: No need to special case this anymore, I gave a non-zero timestamp to DUpgrade in unit test networks. The timestamp is still in past (in 2020), but VM can alter it's time to deactivate/activate the DUpgrade.

precompile/allowlist/allowlist.go
```diff
@@ -119,11 +123,20 @@ func createAllowListRoleSetter(precompileAddr common.Address, role Role) contrac
...
 		// Verify that the caller is an admin with permission to modify the allow list
 		callerStatus := GetAllowListStatus(stateDB, precompileAddr, callerAddr)
-		if !callerStatus.IsAdmin() {
+		// At this point we know that Manager role is activated.
+		if IsManagerRoleActivated(evm) && callerStatus == ManagerRole {
+			// Get current status.
+			// Before the manager role, we never checked the status of the address we are trying to modify.
+			// So we should keep the same behaviour by special casing this.
+			modifyStatus := GetAllowListStatus(stateDB, precompileAddr, modifyAddress)
+			if !callerStatus.CanModify(IsManagerRoleActivated(evm), modifyStatus, role) {
```
darioush, line 132: nit: I think if we are only calling `CanModify` here inside the check for the manager role (to avoid reading the stateDB for the previous role, I assume), should we call it `CanManagerModify` and remove the other clauses from that function? I think otherwise it may give the impression that `CanModify` controls the behavior for other roles as well.
Also it is not needed to call `IsManagerRoleActivated` again. Possibly consider inlining the function.

precompile/contracts/nativeminter/config_test.go
```diff
@@ -17,29 +17,30 @@ import (
...
 			ExpectedError: "cannot set address",
 		},
 		"duplicate admins in config in native minter allowlist": {
-			Config:        NewConfig(utils.NewUint64(3), append(admins, admins[0]), enableds, nil),
+			Config:        NewConfig(utils.NewUint64(3), append(admins, admins[0]), enableds, managers, nil),
 			ExpectedError: "duplicate address",
 		},
 		"duplicate enableds in config in native minter allowlist": {
-			Config:        NewConfig(utils.NewUint64(3), admins, append(enableds, enableds[0]), nil),
+			Config:        NewConfig(utils.NewUint64(3), admins, append(enableds, enableds[0]), managers, nil),
```
darioush, line 31: did we add a UT to check duplicate addresses in the manager list with the other lists?

precompile/contracts/nativeminter/contract_test.go
```diff
@@ -58,6 +60,37 @@ var tests = map[string]testutils.PrecompileTest{
 			require.Equal(t, common.Big2, state.GetBalance(allowlist.TestEnabledAddr), "expected minted funds")
 		},
 	},
+	"mint funds from manager role role fails before activation": {
```
darioush, line 63:
```suggestion
	"mint funds from manager role fails before activation": {
```

precompile/contracts/rewardmanager/contract_test.go
```diff
@@ -93,6 +95,104 @@ var (
 				require.False(t, isFeeRecipients)
 			},
 		},
+		"set allow fee recipients from enabled succeeds after activation": {
```
darioush, line 98:
```suggestion
		"set allow fee recipients from manager succeeds after activation": {
```

precompile/testutils/test_precompile.go
```diff
@@ -85,8 +88,13 @@ func (test PrecompileTest) setup(t testing.TB, module modules.Module, state cont
 	}
 
 	blockContext := contract.NewMockBlockContext(big.NewInt(test.BlockNumber), 0)
-	accesibleState := contract.NewMockAccessibleState(state, blockContext, snow.DefaultContextTest())
-	chainConfig := contract.NewMockChainState(commontype.ValidTestFeeConfig, false)
+	chainConfig := test.ChainConfig
+	if chainConfig == nil {
+		// DUpgrade is activated by default
+		chainConfig = contract.NewMockChainConfig(commontype.ValidTestFeeConfig, false, utils.NewUint64(0))
```
darioush, line 94: kinda nitpicky, but it would be nice to explicitly specify this instead of modifying it here.

precompile/allowlist/test_allowlist_config.go
```diff
@@ -62,23 +111,63 @@ func AllowListConfigVerifyTests(module modules.Module) map[string]testutils.Conf
...
+			Config: mkConfigWithAllowList(module, &AllowListConfig{
+				AdminAddresses:   []common.Address{TestAdminAddr},
+				ManagerAddresses: []common.Address{TestManagerAddr},
+				EnabledAddresses: []common.Address{TestEnabledAddr},
+			}),
+			Other: mkConfigWithAllowList(module, &AllowListConfig{
+				AdminAddresses:   []common.Address{TestAdminAddr},
+				ManagerAddresses: []common.Address{TestManagerAddr},
+				EnabledAddresses: []common.Address{{3}},
+			}), Expected: false,
```
darioush, line 158: nit:
```suggestion
			}), 
			Expected: false,
```

contracts/contracts/test/ExampleDeployerListTest.sol
```diff
@@ -12,12 +12,14 @@ import "./AllowListTest.sol";
...
 contract ExampleDeployerListTest is AllowListTest {
+  address constant ENABLED_ADDRESS = 0x0Fa8EA536Be85F32724D57A37758761B86416123;
+
   IAllowList allowList = IAllowList(DEPLOYER_LIST);
   ExampleDeployerList private example;
 
   function setUp() public {
     example = new ExampleDeployerList();
-    allowList.setNone(OTHER_ADDRESS);
+    allowList.setNone(ENABLED_ADDRESS);
```
darioush, line 22: is this needed because the other test does not do a proper tear down?
(not needed to fix here, but could we open an issue or followup PR?)

darioush:
Couple high level thoughts in the context of this PR:
- How should we handle the upgrades/rules passing? Do we want to make it easier to modify precompile behavior in upgrades?
- It seems we have a lot of code duplication in tests. This is not something we need to fix in this PR, but it seems some tests should be more at the unit level so they can have less of a setup/teardown.

params/network_upgrades.go
```diff
@@ -4,14 +4,18 @@
...
 	"github.com/ava-labs/subnet-evm/utils"
 )
 
 var (
+	defaultActivationTime = time.Date(2020, time.December, 5, 5, 0, 0, 0, time.UTC)
+
 	LocalNetworkUpgrades = MandatoryNetworkUpgrades{
 		SubnetEVMTimestamp: utils.NewUint64(0),
-		DUpgradeTimestamp:  utils.NewUint64(0),
+		DUpgradeTimestamp:  utils.TimeToNewUint64(defaultActivationTime),
```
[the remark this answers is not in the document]
ceyonur: using a 0 means we activate the upgrade at genesis (if genesis.timestamp == 0), a timestamp in the past means they will be activated at the first block.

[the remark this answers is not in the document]
darioush: do we rely on this behavior in tests?

contracts/contracts/test/ExampleDeployerListTest.sol
ceyonur: I presume, yes.

ceyonur:
> Couple high level thoughts in the context of this PR:
> 
>     * How should we handle the upgrades/rules passing? Do we want to make it easier to modify precompile behavior in upgrades?
> 

We have discussed with @aaronbuchwald and agreed on using the direct upgrade times in functions: https://github.com/ava-labs/subnet-evm/issues/692. 
Another alternative would be creating precompiles (like [here](https://github.com/ava-labs/subnet-evm/issues/692)) when we activate them and provide chainConfig/accessibleState to those functions, so `createPrecompile` functions can change created functions according to upgrade.(they can map signatures to different functions, add new functions, remove existing functions etc) like:
```
func createNativeMinterPrecompile(accessibleState contract.AccessibleState) contract.StatefulPrecompiledContract {
	enabledFuncs := allowlist.CreateAllowListFunctions(ContractAddress)

	mintFunc := contract.NewStatefulPrecompileFunction(mintSignature, mintNativeCoin)

	funcs := append(enabledFuncs, mintFunc)

if accessibleState.GetChainConfig().IsXUpgrade(accessibleState.GetBlockContext().Timestamp()){
     newFunc := contract.NewStatefulPrecompileFunction(newSignature, newFunction)
     funcs = append(funcs, newFunc)
}
	// Construct the contract with no fallback function.
	contract, err := contract.NewStatefulPrecompileContract(nil, funcs)
	// TODO: Change this to be returned as an error after refactoring this precompile
	// to use the new precompile template.
	if err != nil {
		panic(err)
	}
	return contract
}
```

precompile/allowlist/role.go
```diff
@@ -24,10 +24,12 @@ var (
 type Role common.Hash
 
 // IsNoRole returns true if [s] indicates no specific role.
```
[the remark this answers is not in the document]
ceyonur: good catch, changed.

contracts/test/tx_allow_list.ts
```diff
@@ -68,4 +68,14 @@ describe("ExampleTxAllowList", function () {
...
+
+  test("should let manager to revoke allowed", "step_managerCanRevoke")
+
+  test("should let manager to add allowed", "step_managerCanAllow")
+
+  test("should not let manager to add admin", "step_managerCannotGrantAdmin")
+
+  test("should not let manager to revoke admin", "step_managerCannotRevokeAdmin")
+
+  test("should not let manager to grant admin", "step_managerCannotGrantAdmin")
```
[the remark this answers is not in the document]
ceyonur: removed

precompile/allowlist/role.go
```diff
@@ -5,56 +5,92 @@ package allowlist
...
-func (s Role) IsEnabled() bool {
-	switch s {
-	case AdminRole, EnabledRole:
+func (r Role) CanModify(isManagerActivated bool, from, target Role) bool {
+	switch r {
+	case AdminRole:
 		return true
+	case ManagerRole:
+		return isManagerActivated &&
+			(from == EnabledRole && target == NoRole) || (from == NoRole && target == EnabledRole)
```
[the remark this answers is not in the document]
ceyonur: 🦅 👁️ 
added tests

accounts/abi/bind/precompilebind/precompile_contract_test_template.go
```diff
@@ -69,6 +72,41 @@ import (
 				ExpectedErr: {{if $fail}} ErrCannot{{$func.Normalized.Name}}.Error() {{- else}} "" {{- end}},
 			},
 			{{- end}}
+			{{- if not $func.Original.IsConstant}}
```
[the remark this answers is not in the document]
ceyonur: this is not a view or constant function (i.e it can change the state)

accounts/abi/bind/precompilebind/precompile_contract_test_template.go
```diff
@@ -69,6 +72,41 @@ import (
...
+			{{- range $isBefore := mkList true false}}
+			"calling {{decapitalise $func.Normalized.Name}} from Manager should {{- if $isBefore}} fail before activation {{- else}} succeed after activation{{- end}}":  {
+				Caller:     allowlist.TestManagerAddr,
+				ChainConfig: contract.NewMockChainConfig(commontype.ValidTestFeeConfig, false, {{- if $isBefore}} nil {{- else}} utils.NewUint64(0) {{- end}}),
```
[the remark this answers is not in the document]
ceyonur: I think we can do it with formatters.

plugin/evm/vm_test.go
```diff
@@ -2197,18 +2195,47 @@ func TestBuildAllowListActivationBlock(t *testing.T) {
 // Test that the tx allow list allows whitelisted transactions and blocks non-whitelisted addresses
 func TestTxAllowListSuccessfulTx(t *testing.T) {
 	// Setup chain params
+	managerKey, err := crypto.HexToECDSA("8a1f9a8f95be41cd7ccb6168179afb4504aefe388d1e14474d32c45c72ce7b7b")
+	require.NoError(t, err)
+	managerAddress := crypto.PubkeyToAddress(managerKey.PublicKey)
```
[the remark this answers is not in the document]
ceyonur: yea we can use `testEthAddrs[1]` (we have 2 addresses in that list). 
initially I had 3 different addreses but no need that; so we can use test addr here.

plugin/evm/vm_test.go
```diff
@@ -2268,6 +2312,73 @@ func TestTxAllowListSuccessfulTx(t *testing.T) {
...
+	vm.clock.Set(reenableAllowlistTimestamp)
+
+	// Re-Submit a successful transaction
+	tx0 = types.NewTransaction(uint64(1), testEthAddrs[0], big.NewInt(1), 21000, big.NewInt(testMinGasPrice), nil)
+	signedTx0, err = types.SignTx(tx0, types.NewEIP155Signer(vm.chainConfig.ChainID), testKeys[0])
+	require.NoError(t, err)
+
+	errs = vm.txPool.AddRemotesSync([]*types.Transaction{signedTx0})
+	if err := errs[0]; err != nil {
+		t.Fatalf("Failed to add tx at index: %s", err)
```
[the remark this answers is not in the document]
ceyonur: changed them, we should do a complete refactor on those though.

plugin/evm/vm_test.go
```diff
@@ -2268,6 +2312,73 @@ func TestTxAllowListSuccessfulTx(t *testing.T) {
...
+	}
+
+	blk = issueAndAccept(t, issuer, vm)
+
+	// Verify that the constructed block only has the whitelisted tx
+	block = blk.(*chain.BlockWrapper).Block.(*Block).ethBlock
+
+	// Verify that etherBase has received fees
```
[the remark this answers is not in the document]
ceyonur: 👍

plugin/evm/vm_test.go
```diff
@@ -2268,6 +2312,73 @@ func TestTxAllowListSuccessfulTx(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Check that address 0 is whitelisted and address 1 is not
+	role = txallowlist.GetTxAllowListStatus(blkState, testEthAddrs[0])
+	if role != allowlist.AdminRole {
+		t.Fatalf("Expected allow list status to be set to admin: %s, but found: %s", allowlist.AdminRole, role)
+	}
+	// This is set as manager but should not be effectie until after the upgrade
```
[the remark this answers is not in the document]
ceyonur: 👍

precompile/allowlist/allowlist.go
ceyonur: I think it would be dangerous to provide the `CanModify` function without taking the `isManagerActivated` as we export those. if someone uses it before the activation that could potentially go wrong.

precompile/contracts/nativeminter/config_test.go
ceyonur: yes we have added them in `test_allowlist_config.go`

precompile/contracts/nativeminter/contract_test.go
ceyonur: removed

precompile/testutils/test_precompile.go
ceyonur: it would cause panics in precompile tests. added some comments for the default config.

precompile/allowlist/allowlist.go
darioush: nit: seems we could assign mgrRoleActivated to a variable to avoid calling it twice or simply pass true then?

contracts/test/tx_allow_list.ts
```diff
@@ -68,4 +68,12 @@ describe("ExampleTxAllowList", function () {
...
   test("should let admin to revoke allowed", "step_adminCanRevoke")
+
+  test("should let manager to revoke allowed", "step_managerCanRevoke")
+
+  test("should let manager to add allowed", "step_managerCanAllow")
+
+  test("should not let manager to add admin", "step_managerCannotGrantAdmin")
+
+  test("should not let manager to revoke admin", "step_managerCannotRevokeAdmin")
 })
```
anusha-ctrl, line 79: should we add " should let manager deploy"

anusha-ctrl: also manager cannot add another manager

precompile/allowlist/test_allowlist.go
```diff
@@ -152,6 +155,264 @@ func AllowListTests(module modules.Module) map[string]testutils.PrecompileTest {
...
+			ReadOnly:    false,
+			ExpectedErr: ErrManagerCannotModify.Error(),
+		},
+		"set admin role to enabled from manager after activation": {
+			Caller:      TestManagerAddr,
+			BeforeHook:  SetDefaultRoles(contractAddress),
+			ChainConfig: contract.NewMockChainConfig(commontype.ValidTestFeeConfig, false, utils.NewUint64(0) /* activated */),
+
+			InputFn: func(t testing.TB) []byte {
+				input, err := PackModifyAllowList(TestAdminAddr, EnabledRole)
```
anusha-ctrl, line 377: great coverage

precompile/testutils/test_precompile.go
```diff
@@ -85,8 +91,13 @@ func (test PrecompileTest) setup(t testing.TB, module modules.Module, state cont
...
 	blockContext := contract.NewMockBlockContext(big.NewInt(test.BlockNumber), 0)
-	accesibleState := contract.NewMockAccessibleState(state, blockContext, snow.DefaultContextTest())
-	chainConfig := contract.NewMockChainState(commontype.ValidTestFeeConfig, false)
+	chainConfig := test.ChainConfig
+	if chainConfig == nil {
+		// DUpgrade is activated by default
+		chainConfig = DefaultChainConfig
+	}
+
+	accesibleState := contract.NewMockAccessibleState(state, blockContext, snow.DefaultContextTest(), chainConfig)
```
anusha-ctrl, line 100: accesibleState -> accessibleState

precompile/allowlist/role.go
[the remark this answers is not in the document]
anusha-ctrl: nice eye @darioush

anusha-ctrl (approved):
lgtm

params/network_upgrades.go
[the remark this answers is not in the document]
ceyonur: yes, we can set the initial vm clock before this; to have a deactivated upgrade.

precompile/allowlist/allowlist.go
ceyonur: added `isManagerRoleActivated` variable on top of the function and used it here.

contracts/test/tx_allow_list.ts
ceyonur: added

contracts/contracts/test/ExampleDeployerListTest.sol
richardpringle: Making sure the state is what you want it to be in `setUp` is better than relying on teardown. I agree that we should also do teardown, but that shouldn't change the `setUp` function here.

contracts/contracts/test/ExampleTxAllowListTest.sol
```diff
@@ -162,4 +164,140 @@ contract ExampleTxAllowListTest is AllowListTest {
...
+    assertTrue(!example.isEnabled(exampleAddress));
+    assertTrue(!example.isManager(managerAddress));
+
+    allowList.setManager(managerAddress);
+
+    assertTrue(manager.isManager(managerAddress));
+
+    manager.setEnabled(exampleAddress);
+    assertTrue(example.isEnabled(exampleAddress));
+  }
```
richardpringle, line 183: 👍

contracts/contracts/test/ExampleTxAllowListTest.sol
```diff
@@ -162,4 +164,140 @@ contract ExampleTxAllowListTest is AllowListTest {
...
+
+    allowList.setEnabled(exampleAddress);
+    allowList.setManager(managerAddress);
+
+    assertTrue(example.isEnabled(exampleAddress));
+    assertTrue(manager.isManager(managerAddress));
+
+    manager.revoke(exampleAddress);
+    assertTrue(!example.isEnabled(exampleAddress));
+  }
```
richardpringle, line 202: 👍

contracts/contracts/test/ExampleTxAllowListTest.sol
richardpringle, line 1: Tests look pretty good overall! 🥇

contracts/README.md
richardpringle, line 1: Edits like this should be landed in a separate PR.

```bash
git checkout main
git pull
git checkout -b readme-fixup
git checkout add-manager-role -- contracts/README.md
git commit
gh pr create
```

contracts/contracts/test/ExampleDeployerListTest.sol
richardpringle, line 1: Wait... should this be a separate change as well?

contracts/test/tx_allow_list.ts
```diff
@@ -61,11 +61,25 @@ describe("ExampleTxAllowList", function () {
 
   test("should allow admin to add allowed address as allowed through contract", "step_enableThroughContract")
 
-  test("should let allowed address deploy", "step_canDeploy")
+  test("should let allowed address to send tx", "step_canDeploy")
```
richardpringle, line 64:
```suggestion
  test("should let allowed-address send a tx", "step_canDeploy")
```

contracts/test/tx_allow_list.ts
```diff
@@ -61,11 +61,25 @@ describe("ExampleTxAllowList", function () {
...
 
   test("should not let allowed add another allowed", "step_onlyAdminCanEnable")
 
   test("should not let allowed to revoke admin", "step_onlyAdminCanRevoke")
 
   test("should let admin to revoke allowed", "step_adminCanRevoke")
+
+  test("should let manager to revoke allowed", "step_managerCanRevoke")
```
richardpringle, line 72:
```suggestion
  test("should let manager revoke allowed-address", "step_managerCanRevoke")
```

contracts/test/tx_allow_list.ts
```diff
@@ -61,11 +61,25 @@ describe("ExampleTxAllowList", function () {
...
+
+  test("should let manager to add allowed", "step_managerCanAllow")
```
richardpringle, line 74:
```suggestion
  test("should let manager add address to allow-list", "step_managerCanAllow")
```

contracts/test/tx_allow_list.ts
```diff
@@ -61,11 +61,25 @@ describe("ExampleTxAllowList", function () {
...
+
+  test("should not let manager to add admin", "step_managerCannotGrantAdmin")
+
+  test("should not let manager to revoke admin", "step_managerCannotRevokeAdmin")
+
+  test("should not let manager to revoke manager", "step_managerCannotRevokeManager")
+
+  test("should not let manager to add manager", "step_managerCannotGrantManager")
```
richardpringle, line 82:
```suggestion
  test("should not let manager add admin", "step_managerCannotGrantAdmin")

  test("should not let manager revoke admin", "step_managerCannotRevokeAdmin")

  test("should not let manager revoke manager", "step_managerCannotRevokeManager")

  test("should not let manager add manager", "step_managerCannotGrantManager")
```

params/network_upgrades.go
richardpringle, line 1: Seems like some of these changes could have been split into a separate PR. Not saying you should necessarily do this now as there have been a few reviews already, but it shouldn't be too much more work to split out some of the functionality after you've implemented it (but before people review your code). It makes reviews easier (this PR is massive) and makes the diff easier to inspect when looking back (after your changes have landed).

plugin/evm/vm_test.go
```diff
@@ -2967,28 +3078,6 @@ func TestSkipChainConfigCheckCompatible(t *testing.T) {
 	require.NoError(t, reinitVM.Shutdown(context.Background()))
 }
 
-func TestMandatoryUpgrades(t *testing.T) {
```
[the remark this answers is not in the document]
richardpringle: Change should be part of a separate PR

richardpringle (commented):
Wow... did you think your diff would be so big when you first took on this task? 

Nothing major in my comments. Just some suggestions for next time. Take a look at the precompile comments, they are nits though. There are also some grammar errors in the JS/TS tests, I would recommend applying those suggestions or something similar. 

Good job though! 🏆

ceyonur:
Closing in favor of https://github.com/ava-labs/subnet-evm/pull/801/

PR #765, opened 2023-07-31
Update warp readme

aaronbuchwald:
## Why this should be merged

This PR improves the README in `x/warp` and references the new Warp README in AvalancheGo instead of explaining everything in one place.

## How this works

Updates the Warp README

## How this was tested

n/a

## How is this documented

This PR is documentation

### DIFF
--- a/x/warp/README.md
+++ b/x/warp/README.md
@@ -11,82 +11,30 @@ It is intended to allow communication between arbitrary Custom Virtual Machines 
 
 ## How does Avalanche Warp Messaging Work
 
-Avalanche Warp Messaging relies on the Avalanche P-Chain to provide a read-only view of every Subnet's validator set. In Avalanche, the P-Chain is used to maintain the Primary Network's validator set, create new subnets and blockchains, and maintain the validator sets of each Subnet. As of the Banff Upgrade, Avalanche enables registering a BLS Public Key alongside a validator.
+Avalanche Warp Messaging uses BLS Multi-Signatures with Public-Key Aggregation where every Avalanche validator registers a public key alongside its NodeID on the Avalanche P-Chain.
 
-In order to be a validator of an Avalanche Subnet, a node must also validate the Avalanche Primary Network. This means each Subnet validator has read access to the P-Chain state.
+Every Subnet needs read access to the Avalanche P-Chain to provide weighted sets of BLS Public Keys that correspond to the validator sets of each Subnet on the Avalanche Network. Avalanche Warp Messaging provides a basic primitive for signing messages with an aggregation of signatures from a source Subnet, and the ability for any Subnet to verify that message was signed by a threshold of validators from the source subent.
 
-With just those two things:
-- Read access to all Subnet validator sets through the P-Chain
-- BLS Public Keys registered on the P-Chain
+For more details on Avalanche Warp Messaging, see the AvalancheGo [Warp README](https://github.com/ava-labs/avalanchego/blob/warp-readme/vms/platformvm/warp/README.md).
 
-We can build a generic Avalanche Warp Messaging Protocol. :point_down:
+## Integrating Avalanche Warp Messaging into the EVM
 
-### Subnet to Subnet
+### Flow of Sending / Receiving a Warp Message within the EVM
 
-The validator set of Subnet A can send a message on behalf of any blockchain on its Subnet.
+The Avalanche Warp Precompile enables this flow to send a message from blockchain A to blockchain B:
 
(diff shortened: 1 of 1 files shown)


martineckardt (commented):
Very good as well. I am missing information on best practices or a reference implementation for the encoding/decoding of complex data structures in the message

x/warp/README.md
```diff
@@ -181,7 +154,7 @@ The Warp Precompile was designed with the intention of minimizing the trusted co
 
 The Warp Precompile itself provides ONLY the following ability:
 
-send a verified message from a caller on blockchain A to a destination address on blockchain B
+- Send a verifiable message from a caller on blockchain A to a destination address on blockchain B
```
RodrigoVillar, line 157: Line 157 and line 163 seem to contradict each other.

In line 157, we are guaranteeing that the Warp Precompile sends a message from blockchain A to B. Therefore, one can logically deduce that the message will *eventually* be delivered. However, line 163 states that the Warp Precompile does not guarantee eventual message delivery.

x/warp/README.md
```diff
@@ -11,82 +11,30 @@ It is intended to allow communication between arbitrary Custom Virtual Machines
...
 
-The validator set of Subnet A can send a message on behalf of any blockchain on its Subnet.
+The Avalanche Warp Precompile enables this flow to send a message from blockchain A to blockchain B:
 
-For example, let's say that there are two Subnets: Subnet A and Subnet B each with a single Blockchain we'll call Blockchain A and Blockchain B.
+1. Call the Warp Precompile `sendWarpMessage` function with the arguments for the UnsignedMessage
+2. Warp Precompile emits an event / log containing the `UnsignedMessage` specified by the caller of `sendWarpMessage`
+3. Network accepts the block containing the `UnsignedMessage` in the log, so that validators are willing to sign the message
+4. An off-chain relayer queries the validators for their signatures of the message and aggregate the signatures to create a `SignedMessage`
+5. The off-chain relayer encodes the `SignedMessage` as the [predicate](#predicate-encoding) in of a transaction to deliver on blockchain B
```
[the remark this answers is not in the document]
aaronbuchwald: Yup `AccessList` was missing here. The predicate is encoded in the EIP-2930 access list of the transaction. Updating.

x/warp/README.md
```diff
@@ -142,6 +91,30 @@ Since the predicate is encoded into the [Transaction Access List](https://eips.e
...
 Therefore, we use the [Predicate Utils](../../../utils/predicate/README.md) package to encode the actual byte slice of size N into the access list.
 
+### Performance Optimization: C-Chain to Subnet
+
+To support C-Chain to Subnet communication, or more generally Primary Network to Subnet communication, we special case the C-Chain for two reasons:
+
+1. Every Subnet validator validates the C-Chain
+2. The Primary Network has the largest possible number of validators
+
+Since the Primary Network has the largest possible number of validators for any Subnet on Avalanche, it would also be the most expensive Subnet to verify Avalanche Warp Messages from (most signatures/public keys required to verify it). Luckily, we can do something much smarter.
```
darioush, line 101:
```suggestion
Since the Primary Network has the largest possible number of validators for any Subnet on Avalanche, it would also be the most expensive Subnet to verify Avalanche Warp Messages from (most signatures/public keys required to verify messages arriving from it if we use the described procedure as is). Luckily, we can do something much smarter.
```

x/warp/README.md
```diff
@@ -142,6 +91,30 @@ Since the predicate is encoded into the [Transaction Access List](https://eips.e
...
+
+When a Subnet receives a message from a blockchain on the Primary Network, we use the validator set of the receiving Subnet instead of the entire network when validating the message. This means that the C-Chain sending a message can be the exact same as Subnet to Subnet communciation.
```
darioush, line 103:
```suggestion
When a Subnet receives a message from a blockchain on the Primary Network, we use the validator set of the receiving Subnet instead of the entire network when validating the message to define a threshold of stake necessary for the receiving network to consider processing the message. This means that the C-Chain sending a message can be the exact same as Subnet to Subnet communication.
```

x/warp/README.md
```diff
@@ -142,6 +91,30 @@ Since the predicate is encoded into the [Transaction Access List](https://eips.e
...
+
+However, when Subnet B receives a message from the C-Chain, it changes the semantics to the following:
+
+1. Read the SourceChainID of the signed message (C-Chain)
```
darioush, line 107: Do we really only special case C-Chain or is it all of primary network?

x/warp/README.md
aaronbuchwald: Fixing the typo, but leaving the rest because I think it makes it more difficult to read.

PR #766, opened 2023-08-01
Warp contract load testing

darioush:
## Why this should be merged
Adds an end-to-end test that measures latency from issuance to acceptance for warp e2e messaging.

## How this works
Builds upon the load testing framework to support issuing and accepting warp messages.

## How this was tested
Runs in CI

## How is this documented
N/A

### TODOs:
[ ] use factored out GetCanonicalValidators when avalanchego is released
[ ] Add measure for gas/s (leaving this for a different PR)
[x] find better way to signal end of Tx generation from Tx issuer to relay
[x] account for possibility of duplicate BLS keys
[x] use weights instead of assuming equal
[x] Properly shutdown relay client goroutines
[x] Add "test type" flag and original EVM test back
[x] move triggerTx to before suite


tests/warp/warp_test.go
```diff
@@ -102,6 +105,22 @@ var _ = ginkgo.BeforeSuite(func() {
 		},
 	)
 	gomega.Expect(err).Should(gomega.BeNil())
+
+	// Issue transactions to activate the proposerVM fork on the receiving chain
```
aaronbuchwald, line 109: should changes to this test be a separate PR?

scripts/run_simulator.sh
```diff
@@ -31,12 +31,12 @@ run_simulator() {
     popd
     echo "running simulator from $PWD"
     ./cmd/simulator/simulator \
-        --endpoints=$RPC_ENDPOINTS \
+        "$@" \
```
aaronbuchwald, line 34: If we're going to use this, would it make sense to using only this instead of passing some of the parameters via `$@` and others set in the script?

cmd/simulator/txs/tx_generator.go
```diff
@@ -47,22 +47,22 @@ func GenerateTxSequences(ctx context.Context, generator CreateTx, client ethclie
...
 
-type txSequence struct {
-	txChan chan *types.Transaction
+type txSequence[T any] struct {
+	txChan chan T
 }
 
-func ConvertTxSliceToSequence(txs []*types.Transaction) TxSequence[*types.Transaction] {
-	txChan := make(chan *types.Transaction, len(txs))
+func ConvertTxSliceToSequence[T any](txs []T) TxSequence[T] {
```
aaronbuchwald, line 54: Nice

tests/warp/warp_test.go
darioush: https://github.com/ava-labs/subnet-evm/pull/804

cmd/simulator/load/awm_worker.go
```diff
@@ -0,0 +1,111 @@
...
+var _ txs.Worker[*AwmTx] = &awmWorker{}
+
+type AwmTx struct {
+	AwmID common.Hash // a hash of the payload used to uniquely identify this awm message
+	Tx    *types.Transaction
+}
+
+func (a *AwmTx) Hash() common.Hash {
+	return a.Tx.Hash() // this is used by the batch worker to track tx times
+}
```
aaronbuchwald, line 25: would it be simpler to embed the transaction if we need to expose `Hash() common.Hash` ?

cmd/simulator/load/awm_worker.go
```diff
@@ -0,0 +1,111 @@
...
+
+type awmWorker struct {
+	worker      *singleAddressTxWorker
+	onIssued    func(common.Hash)
+	onConfirmed func(common.Hash)
+	onClosed    func()
+}
```
aaronbuchwald, line 32: This doesn't seem specific to awm. Could we just add these callbacks into `singleAddressWorker` ?

cmd/simulator/load/awm_worker.go
```diff
@@ -0,0 +1,111 @@
...
+	start, ok := tt.issued[id]
+	if !ok {
+		panic("unexpected confirm " + id.Hex())
+	}
+	duration := time.Since(start)
+	tt.observer(duration.Seconds())
+
+	delete(tt.issued, id)
+	tt.checkDone()
+}
```
aaronbuchwald, line 95: how are these different from the metrics tracked here https://github.com/ava-labs/subnet-evm/blob/master/cmd/simulator/txs/agent.go#L128 ?

cmd/simulator/load/awm.go
```diff
@@ -0,0 +1,379 @@
...
+	threshold      uint64                      // threshold for quorum
+	signatures     <-chan warpSignature        // channel of signatures
+	signedMessages chan *avalancheWarp.Message // channel of signed messages
+	eg             *errgroup.Group             // tracks warp relay clients
+	cancelFn       context.CancelFunc          // invoking this stops the warp relay clients
+	done           <-chan struct{}             // channel to signal shutdown
+}
+
+func NewWarpRelay(
+	ctx context.Context, subnetA *runner.Subnet, thresholdNominator int,
```
aaronbuchwald, line 164: thresholdNumerator?

cmd/simulator/load/sequences.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/ava-labs/subnet-evm/cmd/simulator/config"
+	"github.com/ava-labs/subnet-evm/cmd/simulator/txs"
+	"github.com/ava-labs/subnet-evm/params"
+	"github.com/ethereum/go-ethereum/log"
+	"golang.org/x/exp/maps"
+)
+
+func GetWarpSendTxSequences(
+	ctx context.Context, config config.Config, chainID *big.Int,
+	pks []*ecdsa.PrivateKey, startingNonces []uint64,
```
aaronbuchwald, line 26: Can we update the function signatures so that they are either all on one line or each param is on its own line?

cmd/simulator/load/sequences.go
```diff
@@ -0,0 +1,130 @@
...
+) ([]txs.TxSequence[*AwmTx], error) {
+	bigGwei := big.NewInt(params.GWei)
+	gasTipCap := new(big.Int).Mul(bigGwei, big.NewInt(config.MaxTipCap))
+	gasFeeCap := new(big.Int).Mul(bigGwei, big.NewInt(config.MaxFeeCap))
+
+	subnetBStr := config.Subnets[1].BlockchainID.String()
```
aaronbuchwald, line 32: If we're only using `config.Subnets[1]` can we just pass in that subnet details instead of passing in the full config and assuming in this function that it will have the length we want?

aaronbuchwald, line 32: can we pass the arguments from `config` in directly instead of accessing some of the parameters and including an assumption about the length of `config.Subnets` ?

cmd/simulator/load/awm.go
```diff
@@ -0,0 +1,379 @@
...
+		if err != nil {
+			return nil, err
+		}
+		signer := types.LatestSignerForChainID(chainID)
+		tx, err := types.SignNewTx(key, signer, &types.DynamicFeeTx{
+			ChainID:   chainID,
+			Nonce:     nonce,
+			To:        &warp.Module.Address,
+			Gas:       200_000,
+			GasFeeCap: gasFeeCap,
```
anusha-ctrl, line 51: constant?

cmd/simulator/load/awm.go
```diff
@@ -0,0 +1,379 @@
...
+
+type warpRelayClient struct {
+	client     ethclient.Client
+	warpClient warpclient.WarpClient
+	aggregator chan<- warpSignature
+	nodeID     ids.NodeID
+}
+
+func (wr *warpRelayClient) doLoop(ctx context.Context) error {
+	log.Info("starting warp relay client", "nodeID", wr.nodeID)
```
anusha-ctrl, line 79: why do we call this doLoop

cmd/simulator/load/awm_worker.go
```diff
@@ -0,0 +1,127 @@
...
+		done:     make(chan struct{}),
+	}
+}
+
+// removeMethodID removes the first 4 bytes of data, which is the method id
+func removeMethodID(data []byte) []byte {
+	if len(data) < 4 {
+		log.Error("invalid warp message data", "data", common.Bytes2Hex(data))
+		panic("invalid warp message data")
+	}
```
anusha-ctrl, line 42: curious why do we panic over err here

cmd/simulator/load/awm_worker.go
```diff
@@ -0,0 +1,127 @@
...
+// getSendTxAwmID returns a unique identifier for the awm message contained in
+// the transaction. This is calculated by hashing the payload of the warp
+// message. If the transaction is not a well formed warp message, this panics.
+func getSendTxAwmID(tx *types.Transaction) common.Hash {
+	input := removeMethodID(tx.Data())
+	parsedInput, err := warp.UnpackSendWarpMessageInput(input)
+	if err != nil {
+		log.Error("failed to parse warp message input", "err", err)
+		panic(err)
+	}
```
anusha-ctrl, line 55: same

cmd/simulator/load/awm_worker.go
darioush: txTracker is used to track the tx across 2 different agents via the callbacks so a single agent cannot be used

scripts/run_simulator.sh
darioush: not opposed to this but I think it's cleaner as is.

cmd/simulator/load/awm_worker.go
darioush: passing err instead.

cmd/simulator/load/awm.go
darioush: I called it `doLoop` vs `Run` since it's not expected to be called by the user of the relayer and it will be initialized internally. Do you prefer `run` or a different name?

cmd/simulator/load/awm.go
darioush: I'll add a comment that this is just sufficient gas, I don't think we need to have a constant for every arbitrary gas amount.

cmd/simulator/config/flags.go
```diff
@@ -59,6 +70,14 @@ func BuildConfig(v *viper.Viper) (Config, error) {
 		Timeout:      v.GetDuration(TimeoutKey),
 		BatchSize:    v.GetUint64(BatchSizeKey),
 		MetricsPort:  v.GetUint64(MetricsPortKey),
+		TestType:     v.GetString(TestTypeKey),
+	}
+	if endpointsFile := v.GetString(EndpointsFileKey); endpointsFile != "" {
```
ceyonur, line 75: do we expect to have `end-points-file` and `end-points` flags at the same time? I feel like `end-points-file` should be sufficient itself.

cmd/simulator/load/loader.go
```diff
@@ -62,6 +62,81 @@ func ExecuteLoader(ctx context.Context, config config.Config) error {
...
+	agentBuilder := &transferTxAgentBuilder{}
+	return executeLoaderImpl(ctx, cfg, agentBuilder, m)
+}
+
+func warpTest(ctx context.Context, cfg config.Config, reg *prometheus.Registry) error {
+	if len(cfg.Subnets) != 2 {
+		return fmt.Errorf("expected 2 subnets, got %d", len(cfg.Subnets))
+	}
+
+	metricsA := metrics.NewMetrics("", reg)
```
ceyonur, line 101: `subnet_a_` for metrics name?

cmd/simulator/metrics/metrics.go
```diff
@@ -16,24 +18,29 @@ type Metrics struct {
...
-			Help:       "Individual Tx Confirmation Times for a Load Test",
-			Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
-		}),
-		IssuanceToConfirmationTxTimes: prometheus.NewSummary(prometheus.SummaryOpts{
-			Name:       "tx_issuance_to_confirmation_time",
-			Help:       "Individual Tx Issuance To Confirmation Times for a Load Test",
-			Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
-		}),
+		IssuanceTxTimes: NewSummary(
+			fmt.Sprintf("%stx_issuance_time", prefix),
```
ceyonur, line 33: shall we use `%s_tx_issuance_time` and remove appended `_` from prefixes? (i.e `subnet_a_` > `subnet_a`)

ceyonur (commented):
few nits

PR #773, opened 2023-08-02
Update block builder to enforce predicate one tx at a time

aaronbuchwald:
## Why this should be merged

This PR fixes a potential DoS vector during precompile predicate enforcement (no predicates exist yet, so this is not an active issue).

Previously, we called `enforcePredicates` on all pending transactions prior to starting to build a block. However, if there are N transactions in the mempool, then this would attempt to verify the predicate on all of them potentially unnecessarily. This changes predicate enforcement to happen within `commitTransaction` so that it only enforces the predicates as needed one at a time.

## How this works

as above

## How this was tested

CI

## How is this documented

n/a


miner/worker.go
```diff
@@ -241,6 +243,9 @@ func (w *worker) commitTransaction(env *environment, tx *types.Transaction, coin
 		gp   = env.gasPool.Gas()
 	)
 
+	if err := core.CheckPredicates(env.rules, env.predicateContext, tx); err != nil {
+		return nil, err
```
darioush, line 247:
```suggestion
		log.Debug("Transaction predicate failed verification in miner", "tx", tx.Hash(), "err", err)
		return nil, err
```

darioush: or could be Trace

darioush (approved):

PR #775, opened 2023-08-03
warp e2e: Single node primary network

darioush:
## Why this should be merged
Running multiple nodes only to process subnet transactions for local testing is not necessary,
it consumes resources and produces noisy logs.

## How this works
Passes some additional ANR options 

## How this was tested
CI + local

## How is this documented
N/A


darioush:
This didn't work and it seems it may be feasible via Maru's work in the avlanchego repo so going to stop attempting to do this for now.

PR #776, opened 2023-08-03
warp: Remove versiondb + re-order log processing in block.Accept

darioush:
## Why this should be merged
Allows subscribers to accepted logs to immediately query a validator for its warp signature

## How this works
Removes warpdb from the version db and re-orders operations in plugin/evm package's block.Accept

## How this was tested
e2e tests stop showing failed warp signature requests for accepted messages
CI (to check for regressions)

## How is this documented
N/A


aaronbuchwald (approved):

morrisettjohn (approved):
LGTM

PR #778, opened 2023-08-04
mark TestTransactionIndices as flaky

ceyonur:
Marks Flaky test as Flaky.


darioush (approved):
Should we apply the same change to coreth and have a ticket (if we don't already?)

anusha-ctrl (approved):

PR #780, opened 2023-08-07
Bump min go version

ceyonur:
bumps min go version

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -11,9 +11,10 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-go@v3
         with:
-          go-version: "1.19"
+          go-version: '~1.19.12'
+          check-latest: true
       - run: go mod download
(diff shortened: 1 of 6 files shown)


.github/workflows/bench.yml
```diff
@@ -11,9 +11,10 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-go@v3
         with:
-          go-version: "1.19"
+          go-version: '~1.19.12'
```
darioush, line 14: should we use `>=1.19.12`? What does `~` mean here?

ceyonur: It would allow `1.20.x` as well. `~` means `>= 1.9.12, < 1.20.0`

ceyonur: Copied this from avalanchego workflows btw.
https://github.com/ava-labs/avalanchego/blob/master/.github/workflows/build-and-test.yml#L23-L24

PR #784, opened 2023-08-10
Fix build script

tremblaythibaultl:
## Why this should be merged
To follow the documentation guide found [here](https://docs.avax.network/build/subnet/deploy/custom-vm-subnet).

## How this works
Specifies the version sourced at `scripts/versions.sh` as a build argument. 

## How this was tested
By following the documentation guide previously mentioned and obtaining a correct result.

## How is this documented
N/A


aaronbuchwald:
We removed this version flag since it was not necessary for Subnet-EVM and annoying to keep up to date. Since the build script does already include a commit, you could also make an additional commit on your fork of Subnet-EVM and observe the difference, but I think we should update the documentation rather than modifying the build script and adding back the flag.

Issue #781, opened 2023-08-09
Precompile Template Testing


ceyonur:
We need to test generated precompile files via precompile bind.  `bind_test.go` has a testing method to test golang bindings. We can create a similar one for precompile templates.

PR #788, opened 2023-08-14
add precompile bind tests

ceyonur:
## Why this should be merged

Adds precompile template tests   

Closes: #781 

## How this works

It generates precompile files in a temp dir with `PrecompileBind`. It also creates a test file for custom testing. Then it runs `go test` in all those generated precompile packages and custom tests. 

## How this was tested

This is testing itself.

## How is this documented

No need.


ceyonur:
Closing in favor of #792 #793

PR #792, opened 2023-08-15
Precompile bind cleanups

ceyonur:
Improves precompile templates:
* uses a more human friendly version for returning new types i.e  `*new(boolean)` vs `false` etc.
* exports some of the functions from bind package to in precompilebind package to verify ABI before attempting to generate the files.
* removes `abi` array from precompile gen tool and uses a single abi var as it's assumed/expected from the input.
* adds unpack methods for testing out the packers
* expands contract testing by adding fallback tests.


aaronbuchwald (approved):
LGTM

PR #798, opened 2023-08-17
Expand verify

ceyonur:
## Why this should be merged

Takes chainconfig to verify as a parameter so that custom verification depending on chain config (like network upgrades) can be done.


darioush (approved):

aaronbuchwald (approved):
LGTM

PR #801, opened 2023-08-18
Add manager allowlist role

ceyonur:
## Why this should be merged

Adds manager role

Closes: #345 

## How this works

Adds a new role that can change enabled addresses, permission to only modify enabled addresses. 

* Manager Role cannot modify other managers + admins
* Manager Role is not activated until DUpgrade
* Manager role is also enabled.
* Manager roles in upgrades/genesis are not applied until DUpgrade
  * config verification returns an error if configured timestamp is before the DUpgrade (T < D)
  * If allow list configuration is activated after DUpgrade (T >= D) and contains manager roles for addresses, those will be applied and addresses will become managers.
 * Existing precompiles will break because we have changed the `IsEnabled` signature to check if manager role is activated.
- [x] change precompile template 

## How this was tested

Added UTs

- [x] test locally with real network

## How is this documented

- [ ] Add documentation to docs repo


params/network_upgrades.go
```diff
@@ -4,14 +4,18 @@
...
 	"github.com/ava-labs/subnet-evm/utils"
 )
 
 var (
+	defaultActivationTime = time.Date(2020, time.December, 5, 5, 0, 0, 0, time.UTC)
+
 	LocalNetworkUpgrades = MandatoryNetworkUpgrades{
 		SubnetEVMTimestamp: utils.NewUint64(0),
-		DUpgradeTimestamp:  utils.NewUint64(0),
+		DUpgradeTimestamp:  utils.TimeToNewUint64(defaultActivationTime),
```
aaronbuchwald, line 18: can we change from handling it this way and updating the vm clock to only overriding mandatory network upgrades for specific networkIDs?

I think we could make this change a separate PR as well, since some tests are currently enabling the DUpgrade initially even when using genesisJSONSubnetEVM

precompile/allowlist/config.go
```diff
@@ -14,18 +14,26 @@ import (
...
+func (c *AllowListConfig) Configure(chainConfig precompileconfig.ChainConfig, precompileAddr common.Address, state contract.StateDB, blockContext contract.BlockContext) error {
 	for _, enabledAddr := range c.EnabledAddresses {
 		SetAllowListRole(state, precompileAddr, enabledAddr, EnabledRole)
 	}
 	for _, adminAddr := range c.AdminAddresses {
 		SetAllowListRole(state, precompileAddr, adminAddr, AdminRole)
 	}
+	// Verify() should have been called before Configure()
+	// so we know manager role is activated
+	if chainConfig.IsDUpgrade(blockContext.Timestamp()) {
```
aaronbuchwald, line 32: If this comment is true, then why do we need this check here?

precompile/allowlist/config.go
```diff
@@ -76,5 +85,28 @@ func (c *AllowListConfig) Verify(precompileconfig.ChainConfig) error {
 		addressMap[adminAddr] = AdminRole
 	}
 
+	if len(c.ManagerAddresses) != 0 && !upgrade.Disable && upgrade.Timestamp() != nil {
```
aaronbuchwald, line 88: do we need the disable check here? A disable config should not include manager addresses anyways, so I think we could remove `!upgrade.Disable`

precompile/allowlist/config.go
```diff
@@ -76,5 +85,28 @@ func (c *AllowListConfig) Verify(precompileconfig.ChainConfig) error {
...
+		// check that manager role activates at the timestamp
```
aaronbuchwald, line 89:
```suggestion
		// If the config attempts to activate a manager before the DUpgrade, fail verification
```

precompile/contracts/rewardmanager/contract_test.go
```diff
@@ -93,6 +96,118 @@ var (
...
+				return input
+			},
+			SuppliedGas: AllowFeeRecipientsGasCost,
+			ReadOnly:    false,
+			ExpectedRes: []byte{},
+			AfterHook: func(t testing.TB, state contract.StateDB) {
+				_, isFeeRecipients := GetStoredRewardAddress(state)
+				require.True(t, isFeeRecipients)
+			},
+		},
```
aaronbuchwald, line 116: just to confirm,  this should be an impossible case to hit because we ensure that a manager will not be present in the state prior to DUpgrade activation?

precompile/allowlist/allowlist.go
```diff
@@ -164,9 +170,14 @@ func CreateAllowListPrecompile(precompileAddr common.Address) contract.StatefulP
...
 	setEnabled := contract.NewStatefulPrecompileFunction(setEnabledSignature, createAllowListRoleSetter(precompileAddr, EnabledRole))
 	setNone := contract.NewStatefulPrecompileFunction(setNoneSignature, createAllowListRoleSetter(precompileAddr, NoRole))
 	read := contract.NewStatefulPrecompileFunction(readAllowListSignature, createReadAllowList(precompileAddr))
 
-	return []*contract.StatefulPrecompileFunction{setAdmin, setEnabled, setNone, read}
+	return []*contract.StatefulPrecompileFunction{setAdmin, setManager, setEnabled, setNone, read}
+}
+
+func isManagerRoleActivated(evm contract.AccessibleState) (bool, error) {
+	return evm.GetChainConfig().IsDUpgrade(evm.GetBlockContext().Timestamp()), nil
```
aaronbuchwald, line 182: does this need an error in the return type?

ceyonur: nope, changed.

precompile/contract/utils.go
```diff
@@ -88,3 +88,7 @@ func ParseABI(rawABI string) abi.ABI {
 
 	return parsed
 }
+
+func InvalidFunctionErr(selector []byte) error {
+	return fmt.Errorf("invalid function selector %#x", selector)
+}
```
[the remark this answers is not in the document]
ceyonur: removed

precompile/contract/contract.go
```diff
@@ -15,13 +15,26 @@ const (
 
 type RunStatefulPrecompileFunc func(accessibleState AccessibleState, caller common.Address, addr common.Address, input []byte, suppliedGas uint64, readOnly bool) (ret []byte, remainingGas uint64, err error)
 
+// ActivationFunc defines a function that is used to determine if a function is active
+// The first return value is whether or not the function is active
+type ActivationFunc func(AccessibleState) (bool, error)
```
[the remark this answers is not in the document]
ceyonur: merged

contracts/contracts/test/ExampleDeployerListTest.sol
```diff
@@ -12,6 +12,8 @@ import "./AllowListTest.sol";
 // The transactions are put together as steps of a complete test in contract_deployer_allow_list.ts.
 // TODO: a bunch of these tests have repeated code that should be combined
 contract ExampleDeployerListTest is AllowListTest {
+  address constant OTHER_ADDRESS = 0x0Fa8EA536Be85F32724D57A37758761B86416123;
```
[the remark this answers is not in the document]
ceyonur: we're adding those to tests (removing from example contracts)

params/network_upgrades.go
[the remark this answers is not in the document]
ceyonur: Giving a zero timestamp here means upgrade will be activated at genesis. I don't know if we actually want this or we want to activate those after the first block. Reverted this change.

precompile/allowlist/role.go
```diff
@@ -5,56 +5,74 @@ package allowlist
...
-// IsEnabled returns true if [s] indicates that it has permission to access the resource.
-func (s Role) IsEnabled() bool {
-	switch s {
-	case AdminRole, EnabledRole:
+func (r Role) CanModify(from, target Role) bool {
+	switch r {
+	case AdminRole:
 		return true
+	case ManagerRole:
+		return (from == EnabledRole && target == NoRole) || (from == NoRole && target == EnabledRole)
```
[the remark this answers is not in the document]
ceyonur: yea good catch.

precompile/allowlist/role_test.go
```diff
@@ -0,0 +1,125 @@
...
+		expected bool
+		from     Role
+		target   Role
+	}{
+		{
+			role:     ManagerRole,
+			expected: true,
+			from:     EnabledRole,
+			target:   NoRole,
+		},
```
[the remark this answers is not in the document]
ceyonur: added

precompile/allowlist/config.go
ceyonur: kept it since we already exposed block context to Configure. removed now.

params/network_upgrades.go
ceyonur: reverted for now.

precompile/allowlist/config.go
ceyonur: removed

precompile/allowlist/config.go
ceyonur: done

precompile/contracts/rewardmanager/contract_test.go
ceyonur: true. removed those before/after tests as they behave same.

plugin/evm/vm_test.go
```diff
@@ -2268,6 +2301,110 @@ func TestTxAllowListSuccessfulTx(t *testing.T) {
...
+		ctx,
+		dbManager,
+		genesisBytes,
+		[]byte(""),
+		[]byte(""),
+		issuer,
+		[]*commonEng.Fx{},
+		nil,
+	)
+	require.ErrorContains(t, err, "cannot add managers before DUpgrade")
```
darioush, line 2379: could we make this a defined error and use that (or at least its string representation) to avoid repeating this string?

plugin/evm/vm_test.go
```diff
@@ -2268,6 +2301,110 @@ func TestTxAllowListSuccessfulTx(t *testing.T) {
...
+	errs = vm.txPool.AddRemotesSync([]*types.Transaction{signedTx0})
+	require.NoError(t, errs[0])
+
+	// accept block to trigger upgrade
+	blk = issueAndAccept(t, issuer, vm)
+	newHead = <-newTxPoolHeadChan
+	require.Equal(t, newHead.Head.Hash(), common.Hash(blk.ID()))
+	block = blk.(*chain.BlockWrapper).Block.(*Block).ethBlock
+
+	txs = block.Transactions()
```
darioush, line 2321: seems this assignment is not used. maybe you meant to add some assertion?

plugin/evm/vm_test.go
```diff
@@ -2268,6 +2301,110 @@ func TestTxAllowListSuccessfulTx(t *testing.T) {
...
+		[]*commonEng.Fx{},
+		nil,
+	)
+	require.ErrorContains(t, err, "cannot add managers before DUpgrade")
+
+	// use an invalid upgrade now
+	dUpgradeTime := time.Unix(int64(*params.UnitTestNetworkUpgrades.DUpgradeTimestamp), 0)
+	upgradeConfig := &params.UpgradeConfig{
+		PrecompileUpgrades: []params.PrecompileUpgrade{
+			// re-enable the tx allowlist after DUpgrade to set the manager role
```
darioush, line 2385: it seems this comment doesn't match the next line (seems the timestamp is before DUpgrade), consider removing this comment and accepting the suggestion above.

plugin/evm/vm_test.go
```diff
@@ -2268,6 +2301,110 @@ func TestTxAllowListSuccessfulTx(t *testing.T) {
...
+		genesisBytes,
+		[]byte(""),
+		[]byte(""),
+		issuer,
+		[]*commonEng.Fx{},
+		nil,
+	)
+	require.ErrorContains(t, err, "cannot add managers before DUpgrade")
+
+	// use an invalid upgrade now
```
darioush, line 2381:
```suggestion
	// use an invalid upgrade now with managers set before DUpgrade
```

precompile/allowlist/test_allowlist_config.go
```diff
@@ -30,63 +32,182 @@ func mkConfigWithAllowList(module modules.Module, cfg *AllowListConfig) precompi
...
 		},
+		"invalid allow list config with same admin and manager in allowlist": {
+			Config: mkConfigWithAllowList(module, &AllowListConfig{
+				AdminAddresses:   []common.Address{TestAdminAddr},
+				ManagerAddresses: []common.Address{TestAdminAddr},
+				EnabledAddresses: nil,
+			}),
+			ExpectedError: "cannot set address as both admin and manager",
+		},
+		// this is because we don't allow the manager to have enabled permissions
```
darioush, line 91: nit: not sure this comment adds much value

darioush (approved):
only nits, so approving this PR

contracts/test/tx_allow_list.ts
```diff
@@ -68,4 +68,18 @@ describe("ExampleTxAllowList", function () {
...
+
+  test("should not let manager to add admin", "step_managerCannotGrantAdmin")
+
+  test("should not let manager to revoke admin", "step_managerCannotRevokeAdmin")
+
+  test("should not let manager to revoke manager", "step_managerCannotRevokeManager")
+
+  test("should not let manager to add manager", "step_managerCannotGrantManager")
+
+  test("should let manager to send tx", "step_managerCanDeploy")
```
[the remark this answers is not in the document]
ceyonur: renamed it. Testing deploys could be slightly misleading if we have activated tx allow list and contract deployer at the same time. But I think currently it's fine to keep them as they're.

precompile/allowlist/role.go
```diff
@@ -5,56 +5,74 @@ package allowlist
...
+// 3. Admin - allowed to both modify the allowlist and call the precompile
+// 4. Manager - allowed to add and remove only enabled addresses and also call the precompile. (only after DUpgrade)
+var (
+	// NoRole - this is equivalent to common.Hash{} and deletes the key from the DB when set.
+	NoRole = Role(common.BigToHash(common.Big0))
+	// EnabledRole - allowed to call the precompile.
+	EnabledRole = Role(common.BigToHash(common.Big1))
+	// Admin - allowed to both modify the allowlist and call the precompile.
+	AdminRole = Role(common.BigToHash(common.Big2))
+	// Manager - allowed to add and remove only enabled addresses, with being able to call the precompile (i.e enabled). Activated only after DUpgrade.
```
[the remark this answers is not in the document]
ceyonur: removed

precompile/contracts/deployerallowlist/config.go
```diff
@@ -20,11 +20,12 @@ type Config struct {
 
 // NewConfig returns a config for a network upgrade at [blockTimestamp] that enables
 // ContractDeployerAllowList with [admins] and [enableds] as members of the allowlist.
```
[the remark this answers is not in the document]
ceyonur: done

precompile/contracts/feemanager/config.go
```diff
@@ -23,11 +23,12 @@ type Config struct {
 // NewConfig returns a config for a network upgrade at [blockTimestamp] that enables
 // FeeManager with the given [admins] and [enableds] as members of the
```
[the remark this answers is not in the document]
ceyonur: done

precompile/contracts/rewardmanager/contract_test.go
```diff
@@ -93,6 +93,59 @@ var (
...
+			SuppliedGas: SetRewardAddressGasCost,
+			ReadOnly:    false,
+			ExpectedRes: []byte{},
+			AfterHook: func(t testing.TB, state contract.StateDB) {
+				address, isFeeRecipients := GetStoredRewardAddress(state)
+				require.Equal(t, testAddr, address)
+				require.False(t, isFeeRecipients)
+			},
+		},
+		"disable rewards from manager succeeds  ": {
```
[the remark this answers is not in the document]
ceyonur: done

plugin/evm/vm_test.go
ceyonur: good call

plugin/evm/vm_test.go
ceyonur: removed

plugin/evm/vm_test.go
ceyonur: removed

plugin/evm/vm_test.go
ceyonur: applied

params/network_upgrades.go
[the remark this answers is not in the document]
aaronbuchwald: Imo for the local network, we should activate the `DUpgrade` on local network upgrades at genesis when we activate on fuji/mainnet the same way we handle this for activating network upgrades on the local network on coreth rather than a non-zero default activation time.

contracts/test/tx_allow_list.ts
```diff
@@ -68,4 +68,18 @@ describe("ExampleTxAllowList", function () {
...
+
+  test("should not let manager to add admin", "step_managerCannotGrantAdmin")
+
+  test("should not let manager to revoke admin", "step_managerCannotRevokeAdmin")
+
+  test("should not let manager to revoke manager", "step_managerCannotRevokeManager")
+
+  test("should not let manager to add manager", "step_managerCannotGrantManager")
+
+  test("should let manager to deploy", "step_managerCanDeploy")
```
aaronbuchwald, line 84:
```suggestion
  test("should let manager to transact", "step_managerCanDeploy")
```

contracts/test/tx_allow_list.ts
```diff
@@ -68,4 +68,18 @@ describe("ExampleTxAllowList", function () {
   test("should not let allowed to revoke admin", "step_onlyAdminCanRevoke")
 
   test("should let admin to revoke allowed", "step_adminCanRevoke")
+
+  test("should let manager to revoke allowed", "step_managerCanRevoke")
```
aaronbuchwald, line 72: Can we put all of the tests in the same order here that they are declared on the `.sol` contract file?

go.mod
```diff
@@ -17,6 +17,7 @@ require (
 	github.com/fsnotify/fsnotify v1.6.0
 	github.com/gballet/go-libpcsclite v0.0.0-20191108122812-4678299bea08
 	github.com/go-cmd/cmd v1.4.1
+	github.com/golang/mock v1.4.4
```
aaronbuchwald, line 20: Can we make sure this is using uber's gomock library? I believe we merge this switch into master already so want to make sure we don't depend on both.

params/network_upgrades.go
```diff
@@ -23,6 +27,11 @@ var (
 		SubnetEVMTimestamp: utils.NewUint64(0),
 		// DUpgradeTimestamp: utils.NewUint64(0), // TODO: Uncomment and set this to the correct value
 	}
+
+	UnitTestNetworkUpgrades = MandatoryNetworkUpgrades{
```
aaronbuchwald, line 31: Instead of depending on a `defaultActivationTime` that we need to be aware of throughout the tests, can we change the behavior of overriding network upgrades to ONLY override the mandatory network upgrades for a predetermined set of networkIDs so that if our tests use a different networkID intentionally it is not overridden?

plugin/evm/vm.go
```diff
@@ -319,6 +319,17 @@ func (vm *VM) Initialize(
 		g.Config.FeeConfig = params.DefaultFeeConfig
 	}
 
+	// Apply upgradeBytes (if any) by unmarshalling them into [chainConfig.UpgradeConfig].
```
aaronbuchwald, line 322: Good catch

contracts/test/tx_allow_list.ts
ceyonur: replied here: https://github.com/ava-labs/subnet-evm/pull/801#discussion_r1309297349

If we want to send txs instead of deployments. it should be in it's own PR; as it would require a change to test contract.

contracts/test/tx_allow_list.ts
ceyonur: done (at least for these tests)

go.mod
ceyonur: good catch. weird, I thought I already removed that.

precompile/contracts/nativeminter/contract_test.go
```diff
@@ -58,6 +58,22 @@ var tests = map[string]testutils.PrecompileTest{
 			require.Equal(t, common.Big2, state.GetBalance(allowlist.TestEnabledAddr), "expected minted funds")
 		},
 	},
+	"mint funds from manager role succeeds ": {
```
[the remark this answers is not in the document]
ceyonur: done

scripts/mock.gen.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
```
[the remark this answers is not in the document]
ceyonur: yup done in: https://github.com/ava-labs/subnet-evm/pull/838

params/network_upgrades.go
ceyonur: done in :https://github.com/ava-labs/subnet-evm/pull/839

precompile/allowlist/allowlist.go
```diff
@@ -119,7 +117,15 @@ func createAllowListRoleSetter(precompileAddr common.Address, role Role) contrac
...
-		if !callerStatus.IsAdmin() {
+		if callerStatus == ManagerRole {
+			// Get current status.
+			// Before the manager role, we never checked the status of the address we are trying to modify.
+			// So we should keep the same behaviour by special casing this.
+			modifyStatus := GetAllowListStatus(stateDB, precompileAddr, modifyAddress)
+			if !callerStatus.CanModify(modifyStatus, role) {
+				return nil, remainingGas, fmt.Errorf("%w: modify address: %s, from role: %s, to role: %s", ErrManagerCannotModify, callerAddr, modifyStatus, role)
+			}
+		} else if !callerStatus.IsAdmin() {
```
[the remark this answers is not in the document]
aaronbuchwald: Hmm good point. Performing the extra read operation doesn't change the result if we don't use it, so this should be fine.

Another question here is whether we should charge for the extra read operation. If we do, then we need to keep this separation to charge gas for the extra read operation only when it occurs.

I'd be in favor of keeping the gas cost the same and switching to always use `CanModify` to keep the code as simple as possible. Otherwise, introducing `CanModify` to cover one case but not actually using it when the `callerStatus` is `Admin` makes this code more confusing to reason about what invariants we need to maintain vs. how they're actually used imo.

What do you think?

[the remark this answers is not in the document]
ceyonur: I agree with keeping the gas cost the same. If read operation doesn't matter that much I agree with keeping the code simple and clean.

aaronbuchwald (approved):

PR #807, opened 2023-08-24
Extended functionality of warp e2e tests

morrisettathena:
## Why this should be merged
Extends the functionality of warp e2e tests by calling the warp precompile contract through another intermediary contract.  This is a useful test, because it helps show any flaws that may occur from calling the precompile indirectly.

## How this works
This adds an additional ginkgo test suite to the warp e2e test which interacts with the warp precompile through another intermediary contract.  This intermediary contract calls the warp precompile contract, and then verifies that the outputs of the precompile are working as intended.


aaronbuchwald:
Closing in favor of https://github.com/ava-labs/subnet-evm/pull/836

Issue #735, opened 2023-07-18
Add Warp Verification Results to Block instead of using Predicate


aaronbuchwald:
# Problem: Processing Historical Blocks with Warp

Avalanche Warp Messaging leverages the P-Chain to provide read access to the validator set of every Subnet on Avalanche. To provide a canonical view for verifying Avalanche Warp Messages, each block is included within a `BlockContext` provided by the ProposerVM: https://github.com/ava-labs/avalanchego/blob/master/vms/proposervm/README.md.

The current implementation provides the `PChainHeight` used by the ProposerVM to validate the block. Warp message verification verifies all messages as being signed by a sufficient portion of stake of the validator set using the validator set provided by the P-Chain at that height.

However, the P-Chain does not and should not be required to provide permanent archival support (the ability to lookup the state at any point in its history) efficiently since this would drastically increase the requirements for validating the primary network and to validate warp messages.

The problem is that if a block occurs on Subnet-EVM at height 1000 that references P-Chain height 1100, then in the future when validators bootstrap the network they may not have an easy lookup to verify the warp messages were valid.

Therefore, in order to handle re-processing warp messages that depend on historical P-Chain states at some point in the future, we need to provide an alternative to a permanent archival P-Chain index.

## Current Implementation: Precompile Predicates

Currently, Subnet-EVM supports historical verification of Avalanche Warp Messages by including a predicate encoded in the access list of transactions. The predicate is verified before executing the transaction and the transaction is not considered valid to be included in a block if the predicate fails verification.

Since the predicate must have passed verification for the warp message to be included in a block, when we re-process historical blocks we can simply assume that the predicate was validated correctly by the validator set of the network when it was first processed and we don't need to deal at all with warp messages that failed verification.

Pros:
- Simple and easy to reason about within the EVM
- No need to perform warp message verification during the EVM's execution (we do it in advance)
- All warp messages are declared in the block. As a result, we don't need to execute the block to know what warp messages need to be verified throughout its execution. This means warp messages can be verified in parallel prior to entering the EVM's execution (or concurrently with EVM execution as a further optimization)

Cons:
- Potential DoS vector if Warp message verification is expensive
- complexity: which alternative is the simplest and easiest to reason about?

## Alternative: Stapling Results to Block

[...]

PR #811, opened 2023-08-24
Add DUpgrade extra data

aaronbuchwald:
This PR adds support for extra data past the dynamic fee window. This is intended to be used by https://github.com/ava-labs/subnet-evm/issues/735

### DIFF
--- a/consensus/dummy/consensus.go
+++ b/consensus/dummy/consensus.go
@@ -209,7 +209,7 @@ func (self *DummyEngine) verifyHeader(chain consensus.ChainHeaderReader, header 
 		if len(header.Extra) != expectedExtraDataSize {
 			return fmt.Errorf("expected extra-data field to be: %d, but found %d", expectedExtraDataSize, len(header.Extra))
 		}
-	case !config.IsSubnetEVM(header.Time):
+	default:
 		if uint64(len(header.Extra)) > params.MaximumExtraDataSize {
 			return fmt.Errorf("extra-data too long: %d > %d", len(header.Extra), params.MaximumExtraDataSize)
 		}
--- a/consensus/dummy/consensus.go
+++ b/consensus/dummy/consensus.go
@@ -156,7 +156,7 @@ func (self *DummyEngine) verifyHeaderGasFields(config *params.ChainConfig, heade
 	if err != nil {
 		return fmt.Errorf("failed to calculate base fee: %w", err)
 	}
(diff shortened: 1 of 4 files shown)


darioush (approved):

consensus/dummy/consensus.go
```diff
@@ -156,7 +156,7 @@ func (self *DummyEngine) verifyHeaderGasFields(config *params.ChainConfig, heade
 	if err != nil {
 		return fmt.Errorf("failed to calculate base fee: %w", err)
 	}
-	if !bytes.Equal(expectedRollupWindowBytes, header.Extra) {
+	if len(header.Extra) < len(expectedRollupWindowBytes) || !bytes.Equal(expectedRollupWindowBytes, header.Extra[:len(expectedRollupWindowBytes)]) {
```
ceyonur, line 159: should we do this check only for the DUpgrade?

consensus/dummy/dynamic_fees.go
```diff
@@ -23,15 +23,16 @@ func CalcBaseFee(config *params.ChainConfig, feeConfig commontype.FeeConfig, par
...
-	extraDataSize := params.ExtraDataSize
+	extraDataSize := params.DynamicFeeExtraDataSize
 
 	if !isSubnetEVM || parent.Number.Cmp(common.Big0) == 0 {
 		initialSlice := make([]byte, extraDataSize)
 		return initialSlice, feeConfig.MinBaseFee, nil
 	}
-	if len(parent.Extra) != extraDataSize {
-		return nil, nil, fmt.Errorf("expected length of parent extra data to be %d, but found %d", extraDataSize, len(parent.Extra))
+	if len(parent.Extra) < extraDataSize {
```
ceyonur, line 32: same comment about doing this only for the Dupgrade

aaronbuchwald: This is just loosening the condition.

Previously, it was required to be identical. Now it can be >= 80 bytes since it requires those 80 bytes for the dynamic fee window.

It's ok to loosen the condition here because we verify the exact match prior to DUpgrade activation in `verifyHeader`

consensus/dummy/consensus.go
aaronbuchwald: Previously we required that the byte array was identical, now we are loosening it to require that the first 80 bytes are identical to ensure that the dynamic fee window byte array is correct.

PR #812, opened 2023-08-24
add mocks for precompile tests

aaronbuchwald:
## Why this should be merged

- updates the AvalancheGo dependency to v1.10.9-rc.6 which includes the switch to using Uber's maintained version of gomock (abandoned by the golang team)
- switches to using the gomock library in precompile tests

## How this works

as above

## How this was tested

CI

## How is this documented

na


go.mod
```diff
@@ -4,8 +4,8 @@ go 1.19
 
 require (
 	github.com/VictoriaMetrics/fastcache v1.10.0
-	github.com/ava-labs/avalanche-network-runner v1.7.1
-	github.com/ava-labs/avalanchego v1.10.8
+	github.com/ava-labs/avalanche-network-runner v1.7.2-0.20230821200536-177d9acbae64
+	github.com/ava-labs/avalanchego v1.10.9-rc.6
```
ceyonur, line 8: why we used rc?

precompile/contract/interfaces.go
```diff
@@ -51,19 +51,24 @@ type AccessibleState interface {
...
+// ConfigurationBlockContext defines the interface required to configure a precompile.
+type ConfigurationBlockContext interface {
 	Number() *big.Int
 	Timestamp() uint64
 }
 
+// BlockContext defines the interface to provide information about the block to a precompile
+// during the execution of the EVM.
+type BlockContext interface {
+	ConfigurationBlockContext
```
ceyonur, line 63: I know we will add more functions to this interface. but still, is there any reason why we won't just keep expanding the old one?

go.mod
aaronbuchwald: I created this before the release was out. I'll bump this to v1.10.9.

precompile/contract/interfaces.go
aaronbuchwald: We use `*types.Block` as the `ConfigurationBlockContext` where only `Time()` and `Number()` are required. Separating this out into the interface required for configuring the precompile vs. requirements at execution time allow us to keep passing the block rather than constructing the block context there, which requires the `ChainContext` that is used to serve `GetBlockHashFn`. Specifically, constructing the `BlockContext` here https://github.com/ava-labs/subnet-evm/blob/master/core/genesis.go#L304 would require the `ChainContext`, so I opted to separate the block context required for configuration vs. execution because they serve slightly different purposes.

go.mod
aaronbuchwald: Once there is an ANR release compatible with v1.10.9, I'll update the dependency.

darioush (approved):

anusha-ctrl (approved):

Issue #818, opened 2023-08-26
Write Upgrade to Disk Regression: Precompile Activation 


aaronbuchwald:
# Bug Report

Subnet-EVM v0.5.3 introduced a bug where the UpgradeConfig would not be written to disk correctly. This would occur if the chain config on disk matched the new config (which excluded the serialization of the upgrade config) https://github.com/ava-labs/subnet-evm/blob/v0.5.3/core/genesis.go#L250.

If a new precompile upgrade activated using v0.5.3 after the node had already started, it would not write the PrecompileUpgrade to disk.

After a restart, if the chain had advanced past the activation of the upgrade it would then trigger an error such as:

```
error while creating new snowman vm rpc error: code = Unknown desc = mismatching cannot retroactively enable PrecompileUpgrade[0] in database (have timestamp 0, want timestamp 824650037512, rewindto timestamp 1692701999)
```

This error is normally a protection against allowing a subnet operator to try and apply a network upgrade after the chain has already advanced past that point, which would be invalid since the chain would have already accepted blocks using a different rule set.

# Remediation

If users encounter this issue, they can start their node by applying the config flag `skip-upgrade-check` to the chain config for Subnet-EVM: https://docs.avax.network/nodes/configure/chain-config-flags#skip-upgrade-check-bool. This will ignore the extra check allowing the node to start normally.

# Planned Fix

To fix this issue, v0.5.4 will remove the check here: https://github.com/ava-labs/subnet-evm/blob/master/core/genesis.go#L250 which just saves a single DB write on startup. Disabling this check will ensure that after the upgrade bytes are validated, they will be persisted to disk correctly.

PR #819, opened 2023-08-26
Fix persist upgrade bytes regression

aaronbuchwald:
This PR adds a regression test and fix for https://github.com/ava-labs/subnet-evm/issues/818.

### DIFF
--- a/core/genesis.go
+++ b/core/genesis.go
@@ -223,7 +223,6 @@ func SetupGenesisBlock(
 		rawdb.WriteChainConfig(db, stored, newcfg)
 		return newcfg, stored, nil
 	}
-	storedData, _ := json.Marshal(storedcfg)
 	// Check config compatibility and write the config. Compatibility errors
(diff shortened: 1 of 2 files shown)


darioush (approved):

core/genesis.go
```diff
@@ -246,10 +245,7 @@ func SetupGenesisBlock(
 			return newcfg, stored, compatErr
 		}
 	}
-	// Don't overwrite if the old is identical to the new
-	if newData, _ := json.Marshal(newcfg); !bytes.Equal(storedData, newData) {
-		rawdb.WriteChainConfig(db, stored, newcfg)
-	}
```
ceyonur, line 252: should we instead comment out these lines and add an extra warning why this breaks?

aaronbuchwald: Good call, adding a comment to explain the discrepancy with upstream. Not going to leave the commented out code though.

PR #820, opened 2023-08-28
Align blockchain.go trie metrics with coreth

darioush:
## Why this should be merged
Use the same calculation as coreth & more similar to upstream.
Current calculation is omitting reads from the total trie time

## How this works
Copies the code from coreth

## How this was tested
CI

## How is this documented
N/A


ceyonur (approved):

aaronbuchwald (approved):
LGTM

Issue #734, opened 2023-07-18
Avalanche Warp Signing Block Hashes


aaronbuchwald:
**Context and scope**
Avalanche Warp Messaging currently provides the functionality to send a point to point message from chain A address A to chain B address B with an arbitrary byte array payload as specified here: https://github.com/ava-labs/subnet-evm/blob/master/contracts/contracts/interfaces/IWarpMessenger.sol#L8.

This means that Avalanche Warp Messages will only be signed by subnet validators after an explicit request is performed on-chain.

In some cases, Subnets may want to be configured to support signing block hashes either every block, after an explicit request, or every N blocks.

Why should Subnet-EVM sign block hashes?

The block hash attests to the contents of an EVM block including a state root, receipt root, and transaction root. Combined with a signature from an Avalanche validator set, the signed block hash can be verified via Warp in order to verify the state or execution result on the chain that signed the warp message.

Many existing cross-chain communication protocols leverage a signed block hash in order to prove everything that they need between two different blockchains (ex. https://layerzero.gitbook.io/docs/ecosystem/oracle/develop-an-oracle).

**Discussion and alternatives**
One alternative is to use the currently implemented point to point Avalanche Warp Messaging system to deliver a block hash where it's needed.

To implement this, you would currently need to deploy a contract in order to use the `BLOCKHASH` opcode and request a signature from the validator set via the Warp precompile.

However, the `BLOCKHASH` opcode only supports retrieving block hashes between 1 and 256 blocks old, but does not support getting the current block hash. This is because when building a block, it is not known what the resulting block hash will be.

Therefore, a system built relying on the `BLOCKHASH` opcode would increase the latency of sending a block hash from chain A to chain B via Warp by one additional block confirmation time. A simple solution to avoid adding this additional latency is to provide first class support for signing block hashes directly as opposed to only signing point to point messages.

**Initial Design Idea**

1. Add a new message type to the `Warp` message codec: https://github.com/ava-labs/subnet-evm/blob/master/warp/payload/codec.go#L25 to include only a block hash:

```
type BlockHashPayload struct {
	BlockHash      ids.ID `serialize:"true"`

	bytes []byte
}
```

2. Add support for the new message type to `IWarpMessenger.sol` https://github.com/ava-labs/subnet-evm/blob/master/contracts/contracts/interfaces/IWarpMessenger.sol#L8 and add a new function `getVerifiedWarpBlockHash` to retrieve and return the new message type

[...]

PR #823, opened 2023-08-29
Warp stapling block context

aaronbuchwald:
## Why this should be merged

This PR addresses https://github.com/ava-labs/subnet-evm/issues/735 and https://github.com/ava-labs/subnet-evm/issues/734

## How this works

This PR creates a predicate results type to encode the per-transaction predicate results into the block header. This is intended to remove the possibility that predicate verification may fail and require dropping the transaction from the mempool/block, which creates a potential DoS vector as per: https://ethresear.ch/t/dos-vectors-in-account-abstraction-aa-or-validation-generalization-a-case-study-in-geth/7937.

Additionally, this PR adds a `BlockHashPayload` to allow signing a block hash instead of the existing point-to-point `AddressedPayload`.

To handle the `BlockHashPayload` this PR adds a new function `getVerifiedWarpBlockHash` to read the verified block hash and return it to the caller.

## How this was tested

- Adds new contract execution tests for the block hash payload function
- Adds new predicate test for combinations of verifying successful/failed warp messages within a single transaction
- Improve vm warp unit test (add testing of `traceBlock`)
- TODO: add unit predicate results serialization tests

## How is this documented

TODO: update Warp README


accounts/abi/bind/precompilebind/precompile_module_template.go
```diff
@@ -59,7 +59,7 @@ func (*configurator) MakeConfig() precompileconfig.Config {
 // This function is called by the EVM once per precompile contract activation.
 // You can use this function to set up your precompile contract's initial state,
 // by using the [cfg] config and [state] stateDB.
-func (*configurator) Configure(chainConfig precompileconfig.ChainConfig, cfg precompileconfig.Config, state contract.StateDB, blockContext contract.BlockContext) error {
+func (*configurator) Configure(chainConfig precompileconfig.ChainConfig, cfg precompileconfig.Config, state contract.StateDB, blockContext contract.ConfigurationBlockContext) error {
```
aaronbuchwald, line 62: Can be broken into a separate preparation PR if preferred

core/predicate_check.go
```diff
@@ -11,43 +11,43 @@ import (
...
-		predicateBytes := predicateutils.HashSliceToBytes(accessTuple.StorageKeys)
-		if err := predicater.VerifyPredicate(predicateContext, predicateBytes); err != nil {
-			return fmt.Errorf("predicate %s failed verification for tx %s: %w", address, tx.Hash(), err)
-		}
+
+		predicateArguments[address] = append(predicateArguments[address], predicateutils.HashSliceToBytes(accessTuple.StorageKeys))
+	}
+
+	for address, predicates := range predicateArguments {
+		predicate := rules.Predicates[address]
```
aaronbuchwald, line 46: nit: add comment noting that if `predicateArguments` was populated with this address, then the predicate must be present in `rules.Predicates`

core/state/statedb.go
```diff
@@ -109,7 +109,7 @@ type StateDB struct {
 	accessList *accessList
 	// Ordered storage slots to be used in predicate verification as set in the tx access list.
 	// Only set in PrepareAccessList, and un-modified through execution.
-	predicateStorageSlots map[common.Address][]byte
+	predicateStorageSlots map[common.Address][][]byte
```
aaronbuchwald, line 112: Allows multiple predicates across different access tuples instead of only a single access tuple.

core/state_processor.go
```diff
@@ -160,13 +160,12 @@ func applyTransaction(msg *Message, config *params.ChainConfig, gp *GasPool, sta
 // and uses the input parameters for its environment. It returns the receipt
 // for the transaction, gas used and an error if the transaction failed,
 // indicating the block was invalid.
-func ApplyTransaction(config *params.ChainConfig, bc ChainContext, author *common.Address, gp *GasPool, statedb *state.StateDB, header *types.Header, tx *types.Transaction, usedGas *uint64, cfg vm.Config) (*types.Receipt, error) {
+func ApplyTransaction(config *params.ChainConfig, bc ChainContext, blockContext vm.BlockContext, gp *GasPool, statedb *state.StateDB, header *types.Header, tx *types.Transaction, usedGas *uint64, cfg vm.Config) (*types.Receipt, error) {
```
aaronbuchwald, line 163: switch to passing in `vm.BlockContext` so that we can easily override this for executing a single transaction.

This is used specifically in the worker where we need to populate the `BlockContext` for the execution of a single transaction at a time. We don't want to re-serialize the predicate results and re-construct it within `ApplyTransaction` each time during `NewEVMBlockContext` so this code is modified to take in a `vm.BlockContext` so that we can set the transaction predicate results for each transaction before executing it.

core/vm/evm.go
```diff
@@ -110,6 +111,8 @@ type BlockContext struct {
 	Transfer TransferFunc
 	// GetHash returns the hash corresponding to n
 	GetHash GetHashFunc
+	// PredicateResults are the results of predicate verification available throughout the EVM's execution.
```
aaronbuchwald, line 114:
```suggestion
	// PredicateResults are the results of predicate verification available throughout the EVM's execution. PredicateResults may be nil if it is not encoded in the block's header.
```

plugin/evm/block.go
```diff
@@ -124,6 +130,21 @@ func (b *Block) handlePrecompileAccept(rules *params.Rules, sharedMemoryWriter *
 		}
 	}
 
+	// If Warp is enabled, add the block hash as an unsigned message to the warp backend.
```
aaronbuchwald, line 133: This is the simplest way to add the warp message into the warp backend. Alternatively, we can update the warp backend to provide it access to accepted blocks.

To make that change, we should add a separate handler for fetching signed block hashes so that they can be fetched by `blockHash` instead of the `UnsignedWarpMessageID`.

plugin/evm/block.go
```diff
@@ -247,10 +268,33 @@ func (b *Block) verify(predicateContext *precompileconfig.PredicateContext, writ
...
+	predicateResults := results.NewPredicateResults()
 	for _, tx := range b.ethBlock.Transactions() {
-		if err := core.CheckPredicates(rules, predicateContext, tx); err != nil {
+		results, err := core.CheckPredicates(rules, predicateContext, tx)
+		if err != nil {
+			return err
+		}
+		predicateResults.SetTxPredicateResults(tx.Hash(), results)
+	}
+	// TODO: document required gas constraints to ensure marshalling predicate results does not error
```
aaronbuchwald, line 286: TODO: create ticket

plugin/evm/syncervm_test.go
```diff
@@ -292,6 +293,12 @@ func createSyncServerAndClientVMs(t *testing.T, test syncTest) *syncVMSetup {
 	// configure [serverVM]
 	_, serverVM, _, serverAppSender := GenesisVM(t, true, genesisJSONSubnetEVM, "", "")
 	generateAndAcceptBlocks(t, serverVM, parentsToGet, func(i int, gen *core.BlockGen) {
+		b, err := results.NewPredicateResults().Bytes()
```
aaronbuchwald, line 296: if we move predicate result handling into the engine, this can be taken care of by `FinalizeAndAssemble` instead of requiring the extra modification here.

warp/payload/block_hash_payload.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"fmt"
+
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// BlockHashPayload includes the block hash
+type BlockHashPayload struct {
+	BlockHash common.Hash `serialize:"true"`
```
darioush, line 14: Would this message be more useful to applications if it included the block number as well?

warp/payload/block_hash_payload.go
```diff
@@ -0,0 +1,52 @@
...
+	bytes, err := c.Marshal(codecVersion, b)
+	if err != nil {
+		return fmt.Errorf("couldn't marshal warp addressed payload: %w", err)
+	}
+	b.bytes = bytes
+	return nil
+}
+
+// Bytes returns the binary representation of this payload. It assumes that the
+// payload is initialize from eitehr NewBlockHashPayload or ParseBlockHashPayload.
```
darioush, line 49:
```suggestion
// payload is initialized from either NewBlockHashPayload or ParseBlockHashPayload.
```

x/warp/contract.go
```diff
@@ -99,34 +112,168 @@ func getBlockchainID(accessibleState contract.AccessibleState, caller common.Add
...
+		return nil, remainingGas, fmt.Errorf("%w: %s", errInvalidWarpMsg, err)
+	}
+
+	blockHashPayload, err := warpPayload.ParseBlockHashPayload(warpMessage.UnsignedMessage.Payload)
+	if err != nil {
+		return nil, remainingGas, fmt.Errorf("%w: %s", errInvalidBlockHashPayload, err)
+	}
+	packedOutput, err := PackGetVerifiedWarpBlockHashOutput(GetVerifiedWarpBlockHashOutput{
+		WarpBlockHash: WarpBlockHash{
+			OriginChainID: common.Hash(warpMessage.SourceChainID),
```
darioush, line 197: nit: could be more consistent to also name this `SourceChainID`

x/warp/contract.go
```diff
@@ -99,34 +112,168 @@ func getBlockchainID(accessibleState contract.AccessibleState, caller common.Add
...
+
+func unpackWarpMessage(accessibleState contract.AccessibleState, input []byte) ([]byte, bool, error) {
+	warpIndex, err := UnpackGetVerifiedWarpMessageInput(input)
+	if err != nil {
+		return nil, false, err
+	}
+	if !warpIndex.IsUint64() {
+		return nil, false, fmt.Errorf("invalid index: %v", warpIndex)
+	}
+	warpIndexInt := int(warpIndex.Uint64())
```
darioush, line 254: is this conversion OK if it leads to a negative number here? I think we should check for IsInt64 and >= 0

x/warp/predicate_test.go
```diff
@@ -422,15 +421,12 @@ func TestWarpSignatureWeightsDefaultQuorumNumerator(t *testing.T) {
 		int(params.WarpQuorumDenominator) + 1,
 	} {
 		var (
-			predicateBytes       = createPredicate(numSigners)
-			expectedPredicateErr error
+			predicateBytes           = createPredicate(numSigners)
+			expectedPredicateResults = set.NewBits()
 		)
 		// If the number of signers is less than the params.WarpDefaultQuorumNumerator (67)
```
darioush, line 427: nit: could we update this comment similar to this one: https://github.com/ava-labs/subnet-evm/pull/823/files#diff-52b509c54180f89eb1daa8899e8c1404dcca86bee2bc984bba74961ff86c6762R523-R524

x/warp/config.go
```diff
@@ -118,7 +120,8 @@ func (c *Config) verifyWarpMessage(predicateContext *precompileconfig.PredicateC
 	// Verify the warp payload can be decoded to the expected type
 	_, err := warpPayload.ParseAddressedPayload(warpMsg.UnsignedMessage.Payload)
 	if err != nil {
-		return fmt.Errorf("%w: %s", errInvalidAddressedPayload, err)
+		log.Debug("failed to parse addressed payload", "msgID", warpMsg.ID(), "payload", warpMsg.UnsignedMessage.Payload, "err", err)
```
darioush, line 123: Why don't we need to check for the block hash payload as a possibility here?

precompile/precompileconfig/config.go
```diff
@@ -32,6 +32,14 @@ type Config interface {
 	Verify(ChainConfig) error
 }
 
+// PredicateContext is the context passed in to the ProposerPredicater interface to verify
```
darioush, line 35:
```suggestion
// PredicateContext is the context passed in to the Predicater interface to verify
```

plugin/evm/syncervm_test.go
darioush: I think that would be a nice change if possible.

plugin/evm/block.go
```diff
@@ -247,10 +268,33 @@ func (b *Block) verify(predicateContext *precompileconfig.PredicateContext, writ
...
+	predicateResultsBytes, err := predicateResults.Bytes()
+	if err != nil {
+		return fmt.Errorf("failed to marshal predicate results: %w", err)
+	}
+	headerPredicateResultsBytes := b.ethBlock.Extra()[params.DynamicFeeExtraDataSize:]
```
darioush, line 291: could we add a comment to why this is safe?

plugin/evm/block.go
darioush: I think this would be a good change for a follow-up PR, since it works as is.

x/warp/config.go
```diff
@@ -73,6 +74,7 @@ func (*Config) Key() string { return ConfigKey }
 
 // Verify tries to verify Config and returns an error accordingly.
 func (c *Config) Verify(precompileconfig.ChainConfig) error {
+	// TODO: return an error if Warp is enabled before DUpgrade
```
darioush, line 77: Could we add this in this PR? It seems like a good safety check.

core/predicate_check_test.go
```diff
@@ -14,150 +13,255 @@ import (
...
+				{
+					Address: addr2,
 					StorageKeys: []common.Hash{
 						{1},
 					},
 				},
 			}),
+			expectedRes: map[common.Address][]byte{
+				addr1: predicateResultBytes,
+				addr2: predicateResultBytes,
```
darioush, line 185: nit: may be nice to verify each predicate returns its corresponding result instead of returning both the same.

plugin/evm/vm_warp_test.go
```diff
@@ -314,4 +278,32 @@ func TestReceiveWarpMessage(t *testing.T) {
...
+	require.NoError(err)
+	unmarshalResults := make(map[string]interface{})
+	require.NoError(json.Unmarshal(blockTxTraceResultBytes, &unmarshalResults))
+	require.Equal(common.Bytes2Hex(expectedOutput), unmarshalResults["returnValue"])
+
+	txTraceResult, err := tracerAPI.TraceTransaction(context.Background(), getVerifiedWarpMessageTx.Hash(), nil)
+	require.NoError(err)
+	txTraceResultBytes, err := json.Marshal(txTraceResult)
+	require.NoError(err)
+	require.Equal(txTraceResultBytes, blockTxTraceResultBytes)
```
darioush, line 308: nit: consider comparing with `require.JSONEq`

plugin/evm/block.go
aaronbuchwald: 👍

plugin/evm/block.go
aaronbuchwald: https://github.com/ava-labs/subnet-evm/issues/824

plugin/evm/syncervm_test.go
aaronbuchwald: Looked into this and I think this would only be a worthwhile change if we also moved predicate verification into `Finalize` or at least into the engine for symmetry with `FinalizeAndAssemble`.

We could add engine callbacks in like with Coreth, but I generally don't think that was a great design, so planning to leave this as is in this PR.

core/evm.go
```diff
@@ -48,6 +51,29 @@ type ChainContext interface {
...
 // NewEVMBlockContext creates a new context for use in the EVM.
 func NewEVMBlockContext(header *types.Header, chain ChainContext, author *common.Address) vm.BlockContext {
+	var (
+		predicateResults *results.PredicateResults
+		err              error
+	)
+	if len(header.Extra) > params.DynamicFeeExtraDataSize {
+		predicateResults, err = results.ParsePredicateResults(header.Extra[params.DynamicFeeExtraDataSize:])
+		if err != nil {
+			log.Error("failed to parse predicate results creating new block context", "err", err, "extra", header.Extra)
```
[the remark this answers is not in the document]
aaronbuchwald: Adding a comment

warp/payload/block_hash_payload.go
aaronbuchwald: I considered the DevX tradeoff of including that information here and decided against including it since a user should be able to look up the height for a corresponding block height.

x/warp/config.go
aaronbuchwald: Dropping this. The contract execution will parse it and handle failure to parse as a failure.

x/warp/contract.go
aaronbuchwald: Will fix the conversion

x/warp/contract.go
aaronbuchwald: Ya that's a good idea

plugin/evm/block.go
aaronbuchwald: The length is checked in syntactic verification, but will add an extra check here to make it more obviously correct.

plugin/evm/vm_warp_test.go
aaronbuchwald: Updating

core/predicate_check_test.go
aaronbuchwald: Updating

x/warp/contract.go
aaronbuchwald: Nice catch, adding tests for maxuint64 and maxint64 as well

x/warp/config.go
aaronbuchwald: I planned to, but realized it was applied in this PR: https://github.com/ava-labs/subnet-evm/pull/801/files#diff-fd0c998db0ab5b6c9249e67a71921b09d0ea7180742e064beb890fda3176cfe7. I'll make a separate PR to apply this change to rebase this off of and check `DUpgrade` here.

aaronbuchwald: https://github.com/ava-labs/subnet-evm/pull/825/files

core/predicate_check.go
```diff
@@ -11,43 +11,45 @@ import (
...
-		if _, ok := precompileAddressChecks[address]; ok {
-			return fmt.Errorf("predicate %s failed verification for tx %s: specified %s in access list multiple times", address, tx.Hash(), address)
-		}
-		precompileAddressChecks[address] = struct{}{}
-		predicateBytes := predicateutils.HashSliceToBytes(accessTuple.StorageKeys)
-		if err := predicater.VerifyPredicate(predicateContext, predicateBytes); err != nil {
-			return fmt.Errorf("predicate %s failed verification for tx %s: %w", address, tx.Hash(), err)
-		}
+
+		predicateArguments[address] = append(predicateArguments[address], predicateutils.HashSliceToBytes(accessTuple.StorageKeys))
```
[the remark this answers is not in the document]
aaronbuchwald: This was so that the predicate can take in all of the byte slices for the transaction in one call rather than doing it one at a time and return the bit set as a byte slice.

Alternatively, we could switch back to `VerifyPredicate` taking in a single value and `CheckPredicates` assembling the bit set from those results. This would make the code a bit simpler at the cost of constraining the precompile predicates to return a true/false value instead of an arbitrary byte array.

core/evm.go
```diff
@@ -48,6 +51,32 @@ type ChainContext interface {
...
 // NewEVMBlockContext creates a new context for use in the EVM.
 func NewEVMBlockContext(header *types.Header, chain ChainContext, author *common.Address) vm.BlockContext {
+	var (
+		predicateResults *results.PredicateResults
+		err              error
+	)
+	if len(header.Extra) > params.DynamicFeeExtraDataSize {
+		// Prior to the DUpgrade, the VM enforces the extra data is smaller than or equal to this size.
+		// After the DUpgrade, the VM pre-verifies the extra data past the dynamic fee rollup window is
+		// valid.
```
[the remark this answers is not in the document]
aaronbuchwald: We could, it would just require changing the `ChainContext` here to also return the `ChainConfig`, which requires changing some further tests to make sure that `ChainContext` / `ChainConfig` are always populated where some tests do not currently ensure that.

Specifically, this introduces the new requirement that the `ChainContext` actually needs to be non-nil. Even though this seems like a reasonable assumption, `addTx` within `core/chain_makers.go` currently passes in `nil` instead of an actual blockchain instance, so this causes the existing tests using it to panic.

x/warp/config.go
aaronbuchwald: Merged

core/predicate_check.go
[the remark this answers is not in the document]
ceyonur: I feel like using a single value for `VerifyPredicate` could be simpler + cleaner (and costs less loops). It would be easier for 3rd party implementations as well.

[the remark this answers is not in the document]
aaronbuchwald: Could we make this a follow-on PR to try out the alternative?

This also means that we need to track the indices used by each precompile address to create that bitset, so I'm not sure if it will be cleaner.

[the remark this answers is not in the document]
ceyonur: Created https://github.com/ava-labs/subnet-evm/issues/844

darioush (approved):

core/state/statedb.go
```diff
@@ -1242,14 +1246,24 @@ func (s *StateDB) SlotInAccessList(addr common.Address, slot common.Hash) (addre
 	return s.accessList.Contains(addr, slot)
 }
 
+func (s *StateDB) GetTxHash() common.Hash {
+	return s.thash
+}
+
 // GetPredicateStorageSlots returns the storage slots associated with a given address, and whether or not
```
ceyonur, line 1253: add `index` to comments

precompile/results/predicate_results.go
```diff
@@ -0,0 +1,112 @@
...
+	if errs.Errored() {
+		panic(errs.Err)
+	}
+}
+
+// TxPredicateResults is a map of results for each precompile address to the resulting byte array.
+type TxPredicateResults map[common.Address][]byte
+
+// PredicateResults encodes the precompile predicate results included in a block on a per transaction basis.
+// PredicateResults is not thread-safe.
```
ceyonur, line 42: don't we need this to be thread-safe?

x/warp/contract.go
```diff
@@ -99,34 +111,162 @@ func getBlockchainID(accessibleState contract.AccessibleState, caller common.Add
...
+
+func unpackWarpMessage(accessibleState contract.AccessibleState, input []byte) ([]byte, bool) {
+	warpIndex, err := UnpackGetVerifiedWarpMessageInput(input)
+	if err != nil {
+		return nil, false
+	}
+	if !warpIndex.IsInt64() {
+		return nil, false
+	}
+	warpIndexInt := int(warpIndex.Int64())
```
ceyonur, line 250: paranoid 🤖: could someone issue a very big int (MaxInt32<x<MaxInt64) and cause this casting to be not lossless (in 32-bit systems)?

core/state/statedb.go
```diff
@@ -1242,14 +1246,24 @@ func (s *StateDB) SlotInAccessList(addr common.Address, slot common.Hash) (addre
...
 // predicates for transactions with access list addresses that match a precompile address.
-func (s *StateDB) GetPredicateStorageSlots(address common.Address) ([]byte, bool) {
-	storageSlots, exists := s.predicateStorageSlots[address]
-	return storageSlots, exists
+func (s *StateDB) GetPredicateStorageSlots(address common.Address, index int) ([]byte, bool) {
+	predicates, exists := s.predicateStorageSlots[address]
+	if !exists {
+		return nil, false
+	}
+	if index >= len(predicates) {
```
ceyonur, line 1263: should we check if index is negative (or simply require uint64 indexes here)? 

this is related to the comment about lossless conversion between int<>int64 below.

precompile/results/predicate_results.go
aaronbuchwald: We currently only modify this synchronously in the worker and during block verification. Is there a specific case you're concerned about?

x/warp/contract.go
aaronbuchwald: ya I'll update the input type to uint32 so that we don't have to do this at all.

core/state/statedb.go
aaronbuchwald: Added

ceyonur (approved):

PR #829, opened 2023-08-30
add loglevel to server conf to generate anr logfile

felipemadero:
## Why this should be merged
default loglevel setting for anr server conf is off, so no logfile is generated for anr operations, difficulting debugging from artifcats

## How this works
sets loglevel for anr server 

## How this was tested
e2e


aaronbuchwald (approved):
LGTM - nice catch

darioush (approved):

PR #830, opened 2023-08-31
x/warp: add warp message handler interface within warp contract

aaronbuchwald:
This PR adds a message handler interface to share more code between `getVerifiedWarpMessage` and `getVerifiedWarpBlockHash`.

This PR changes the handling of an invalid input from returning the message was invalid to returning an error, which is non-backwards compatible and updates the unit tests as a result.


x/warp/contract.go
```diff
@@ -288,18 +266,31 @@ func getVerifiedWarpMessage(accessibleState contract.AccessibleState, caller com
...
+		return nil, 0, fmt.Errorf("%w: %s", errInvalidWarpMsg, err)
+	}
+	res, err := handler.handleMessage(warpMessage)
+	if err != nil {
+		return nil, 0, err
 	}
+	return res, remainingGas, nil
+}
+
+type addressedHanlder struct{}
```
darioush, line 282: nit: could we rename this to `addressedPayloadHandler` for consistency? (or fix it to `addressedHandler`)

x/warp/contract.go
```diff
@@ -240,46 +231,33 @@ func UnpackGetVerifiedWarpMessageOutput(output []byte) (GetVerifiedWarpMessageOu
 	return outputStruct, err
 }
 
-func unpackWarpMessage(accessibleState contract.AccessibleState, input []byte) ([]byte, bool) {
+type messageHandler interface {
```
darioush, line 234: could we add interface assertions for the types implementing this?

x/warp/contract_test.go
```diff
@@ -416,13 +426,20 @@ func TestGetVerifiedWarpMessage(t *testing.T) {
 			},
 			SuppliedGas: GetVerifiedWarpMessageBaseCost,
 			ReadOnly:    false,
-			ExpectedRes: func() []byte {
-				res, err := PackGetVerifiedWarpMessageOutput(GetVerifiedWarpMessageOutput{Valid: false})
+			ExpectedErr: errInvalidIndexInput.Error(),
+		},
+		"get message invalid index": {
```
darioush, line 431:
```suggestion
		"get message invalid index input bytes": {
```

x/warp/contract_test.go
```diff
@@ -673,13 +700,20 @@ func TestGetVerifiedWarpBlockHash(t *testing.T) {
 			},
 			SuppliedGas: GetVerifiedWarpMessageBaseCost,
 			ReadOnly:    false,
-			ExpectedRes: func() []byte {
-				res, err := PackGetVerifiedWarpBlockHashOutput(GetVerifiedWarpBlockHashOutput{Valid: false})
+			ExpectedErr: errInvalidIndexInput.Error(),
+		},
+		"get message invalid index": {
```
darioush, line 705:
```suggestion
		"get message invalid index input bytes": {
```

darioush (approved):

x/warp/contract.go
```diff
@@ -240,46 +231,33 @@ func UnpackGetVerifiedWarpMessageOutput(output []byte) (GetVerifiedWarpMessageOu
...
-		return packedOutput, remainingGas, nil
+		return handler.packFailed(), remainingGas, nil
 	}
 
 	// Note: we charge for the size of the message during both predicate verification and each time the message is read during
 	// EVM execution because each execution incurs an additional read cost.
 	msgBytesGas, overflow := math.SafeMul(GasCostPerWarpMessageBytes, uint64(len(predicateBytes)))
 	if overflow {
-		return nil, remainingGas, vmerrs.ErrOutOfGas
+		return nil, 0, vmerrs.ErrOutOfGas
```
ceyonur, line 260: in template we don't follow this pattern, instead we return remaining gas. Should we follow this pattern here as well?

x/warp/contract.go
```diff
@@ -148,60 +167,32 @@ func UnpackGetVerifiedWarpBlockHashOutput(output []byte) (GetVerifiedWarpBlockHa
...
-	if !valid {
-		packedOutput, err := PackGetVerifiedWarpBlockHashOutput(GetVerifiedWarpBlockHashOutput{
-			Valid: false,
-		})
-		if err != nil {
-			return nil, remainingGas, err
-		}
-		return packedOutput, remainingGas, nil
-	}
+type blockHashHandler struct{}
```
ceyonur, line 170: would it make sense to move those into a message_handler file?

x/warp/contract.go
aaronbuchwald: Yup

x/warp/contract_test.go
aaronbuchwald: Updating

x/warp/contract.go
aaronbuchwald: Where in the template do we return `remainingGas` alongside `ErrOutOfGas` ? I don't seem to see it and think it seems weird to return non-zero gas with `ErrOutOfGas`. I think we may do this for all other errors?

x/warp/contract.go
aaronbuchwald: Yup tried this out and think it looks a lot better. Great call

x/warp/contract_warp_handler.go
```diff
@@ -0,0 +1,132 @@
...
+
+type messageHandler interface {
+	packFailed() []byte
+	handleMessage(msg *warp.Message) ([]byte, error)
+}
+
+func handleWarpMessage(accessibleState contract.AccessibleState, input []byte, remainingGas uint64, handler messageHandler) ([]byte, uint64, error) {
+	warpIndex, err := UnpackGetVerifiedWarpMessageInput(input)
+	if err != nil {
+		return nil, 0, fmt.Errorf("%w: %s", errInvalidIndexInput, err)
```
ceyonur, line 51: Follow up on https://github.com/ava-labs/subnet-evm/pull/830#discussion_r1316663749

I meant other errors like that: https://github.com/ava-labs/subnet-evm/blob/warp-stapling-block-context/accounts/abi/bind/precompilebind/precompile_contract_template.go#L243-L245

aaronbuchwald: Sounds good, I'll update this for the non `vmerrs.ErrOutOfGas` errors.

x/warp/contract_warp_handler.go
```diff
@@ -0,0 +1,128 @@
...
+	if err != nil {
+		return nil, remainingGas, fmt.Errorf("%w: %s", errInvalidPredicateBytes, err)
+	}
+	warpMessage, err := warp.ParseMessage(unpackedPredicateBytes)
+	if err != nil {
+		return nil, remainingGas, fmt.Errorf("%w: %s", errInvalidWarpMsg, err)
+	}
+	res, err := handler.handleMessage(warpMessage)
+	if err != nil {
+		return nil, 0, err
```
ceyonur, line 82: should it be `remainingGas`?

x/warp/contract_warp_handler.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	// Note: we charge for the size of the message during both predicate verification and each time the message is read during
+	// EVM execution because each execution incurs an additional read cost.
+	msgBytesGas, overflow := math.SafeMul(GasCostPerWarpMessageBytes, uint64(len(predicateBytes)))
+	if overflow {
+		return nil, 0, vmerrs.ErrOutOfGas
+	}
+	if remainingGas, err = contract.DeductGas(remainingGas, msgBytesGas); err != nil {
+		return nil, 0, err
```
ceyonur, line 68: should it be `remainingGas`?

aaronbuchwald: We use `0` in all other places where `contract.DeductGas` returns a non-nil error (which is always `vmerrs.ErrOutOfGas`)

x/warp/contract_warp_handler.go
aaronbuchwald: Updated

darioush (approved):

ceyonur (approved):

PR #838, opened 2023-09-04
add mock ci

ceyonur:
## Why this should be merged

We have added mock.gen.sh in the PR https://github.com/ava-labs/subnet-evm/pull/801. This one adds a CI job to ensure we're applying it correctly (as we do in avalanchego)

https://github.com/ava-labs/subnet-evm/pull/801#discussion_r1314280082

## How this was tested

CI


aaronbuchwald:
Can we make this a PR to master instead? I don't think this needs to be coupled to the manager allow list PR

aaronbuchwald (approved):

PR #839, opened 2023-09-04
override network upgrades

ceyonur:
## Why this should be merged

From comment:https://github.com/ava-labs/subnet-evm/pull/801#discussion_r1314279711

Adds a way to override enforced mandatory upgrades. Mandatory upgrades are still enforced for mainnet and fuji, but rest of the networks (unit tests and local) can be overridden. 

## How this works

We conditionally enforce.

## How this was tested

Expanded the current unit tests

## How is this documented

No need.

### DIFF
--- a/plugin/evm/vm.go
+++ b/plugin/evm/vm.go
@@ -16,6 +16,7 @@ import (
 	"time"
 
 	avalanchegoMetrics "github.com/ava-labs/avalanchego/api/metrics"
+	avalanchegoConstants "github.com/ava-labs/avalanchego/utils/constants"
 	"github.com/prometheus/client_golang/prometheus"
 
 	"github.com/ava-labs/subnet-evm/commontype"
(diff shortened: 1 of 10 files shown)


darioush (approved):

plugin/evm/vm.go
```diff
@@ -977,3 +987,18 @@ func attachEthService(handler *rpc.Server, apis []rpc.API, names []string) error
...
+	switch networkID {
+	case avalanchegoConstants.FujiID:
+		return params.FujiNetworkUpgrades, true
+	case avalanchegoConstants.MainnetID:
+		return params.MainnetNetworkUpgrades, true
+	case avalanchegoConstants.UnitTestID:
+		return params.UnitTestNetworkUpgrades, false
+	default:
+		return params.LocalNetworkUpgrades, false
+	}
```
aaronbuchwald, line 1003:
```suggestion
	case avalanchegoConstants.MainnetID:
		return params.MainnetNetworkUpgrades, true
	case avalanchegoConstants.FujiID:
		return params.FujiNetworkUpgrades, true
	case avalanchegoConstants.UnitTestID:
		return params.UnitTestNetworkUpgrades, false
	default:
		return params.LocalNetworkUpgrades, false
	}
```

aaronbuchwald (approved):

PR #843, opened 2023-09-06
add equality tests

ceyonur:
Adds Equality tests for our legacy Unpack/Pack methods vs ABI's Unpack/Pack methods (comes with precompile generator). The idea is that these should produce equal results (same packaged bytes and same unpacked structs), so that we can safely swap those methods with new ones.


aaronbuchwald:
Closing in favor of https://github.com/ava-labs/subnet-evm/pull/857 directly to master

PR #871, opened 2023-09-19
improve documentation; nits

danlaine:
## Why this should be merged

Improves documentation; style nits.

## How this works

Makes invariants more apparent.

## How this was tested

Existing UT.

## How is this documented

N/A


warp/backend.go
```diff
@@ -56,6 +57,7 @@ func NewBackend(snowCtx *snow.Context, db database.Database, cacheSize int) Back
 
 func (b *backend) Clear() error {
 	b.signatureCache.Flush()
+	b.messageCache.Flush()
```
danlaine, line 60: unrelated to documentation but I think this was a bug

warp/backend.go
```diff
@@ -16,27 +16,28 @@ import (
...
+
 // Backend tracks signature-eligible warp messages and provides an interface to fetch them.
 // The backend is also used to query for warp message signatures by the signature request handler.
 type Backend interface {
-	// AddMessage signs [unsignedMessage] and adds it to the warp backend database
+	// AddMessage stores [unsignedMessage] and this node's signature over it.
 	AddMessage(unsignedMessage *avalancheWarp.UnsignedMessage) error
 
-	// GetSignature returns the signature of the requested message hash.
+	// GetSignature returns this node's signature over the node with the given [messageHash].
```
darioush, line 29:
```suggestion
	// GetSignature returns this node's signature over the message with the given [messageHash].
```

darioush (approved):

warp/backend.go
```diff
@@ -16,27 +16,28 @@ import (
...
-
 const batchSize = ethdb.IdealBatchSize
 
+var _ Backend = &backend{}
+
 // Backend tracks signature-eligible warp messages and provides an interface to fetch them.
 // The backend is also used to query for warp message signatures by the signature request handler.
 type Backend interface {
-	// AddMessage signs [unsignedMessage] and adds it to the warp backend database
+	// AddMessage stores [unsignedMessage] and this node's signature over it.
```
aaronbuchwald, line 26: We only store the message, not the signature, so that if the node's BLS signer changes after a restart it will create a signature that matches the up to date BLS Signer.

danlaine: updated comment

aaronbuchwald: This comment still reads a bit funny to me. We store the signature in the cache, but but saying that we store the message and the signature over it and then clarifying the signature handling after seems a bit messy. Would you mind doing another pass?

warp/backend.go
```diff
@@ -15,27 +15,30 @@ import (
...
+	// Errors if the message wasn't added with AddMessage.
 	GetSignature(messageHash ids.ID) ([bls.SignatureLen]byte, error)
 
-	// GetMessage retrieves the [unsignedMessage] from the warp backend database if available
+	// GetMessage returns the message with the given [messageHash].
+	// Errors if the message wasn't added with AddMessage.
 	GetMessage(messageHash ids.ID) (*avalancheWarp.UnsignedMessage, error)
 
-	// Clear clears the entire db
+	// Clear clears the entire database, including persisted data.
```
aaronbuchwald, line 38: it's not clear to me why this comment is better than what was there previously. I don't think it's unclear that data written to the database is persisted to disk, but am guessing this is trying to say that we're clearing the caches as well?

warp/backend.go
```diff
@@ -15,27 +15,30 @@ import (
...
 	AddMessage(unsignedMessage *avalancheWarp.UnsignedMessage) error
 
-	// GetSignature returns the signature of the requested message hash.
+	// GetSignature returns this node's signature over the message with the given [messageHash].
+	// Errors if the message wasn't added with AddMessage.
 	GetSignature(messageHash ids.ID) ([bls.SignatureLen]byte, error)
 
-	// GetMessage retrieves the [unsignedMessage] from the warp backend database if available
+	// GetMessage returns the message with the given [messageHash].
+	// Errors if the message wasn't added with AddMessage.
```
aaronbuchwald, line 35: Is the added comment necessary? What case are we trying to clarify that would be unclear to someone using this package?

danlaine:
Moved `messageCache` clearing fix to https://github.com/ava-labs/subnet-evm/pull/918/files

PR #872, opened 2023-09-19
verify warp is not activated before dupgrade

ceyonur:
## Why this should be merged

Ensures Warp is only activatable with DUpgrade. 

Resolves a TODO. 

## How this works

Adds a timestamp verification for DUpgrade 

## How this was tested

Added unit test

## How is this documented

Not currently


x/warp/config.go
```diff
@@ -72,8 +73,15 @@ func NewDisableConfig(blockTimestamp *uint64) *Config {
 func (*Config) Key() string { return ConfigKey }
 
 // Verify tries to verify Config and returns an error accordingly.
-func (c *Config) Verify(precompileconfig.ChainConfig) error {
-	// TODO: return an error if Warp is enabled before DUpgrade
+func (c *Config) Verify(chainConfig precompileconfig.ChainConfig) error {
+	if c.Timestamp() != nil {
+		// If the config attempts to activate a manager before the DUpgrade, fail verification
```
aaronbuchwald, line 78:
```suggestion
		// If Warp attempts to activate before the DUpgrade, fail verification
```

aaronbuchwald (approved):

PR #874, opened 2023-09-19
Add contexts

danlaine:
## Why this should be merged

Right now, network requests don't honor context cancellation. This PR adds that.

## How this works

Pipe through and use contexts.

## How this was tested

New UT.

## How is this documented

Comments


joshua-kim (approved):

anusha-ctrl (approved):

StephenButtolph (commented):
I believe this will cause the response handlers to deadlock.

danlaine:
> I believe this will cause the response handlers to deadlock.

Made responseChan buffer length 1 to fix this

nytzuga (approved):

peer/network.go
```diff
@@ -190,7 +190,8 @@ func (n *network) sendAppRequest(nodeID ids.NodeID, request []byte, responseHand
 
 	// Send app request to [nodeID].
 	// On failure, release the slot from [activeAppRequests] and delete request from [outstandingRequestHandlers]
-	if err := n.appSender.SendAppRequest(context.TODO(), nodeIDs, requestID, request); err != nil {
+	// TODO: treat error from  n.appSender.SendAppRequest as fatal
```
aaronbuchwald, line 193: Can we remove these comments?

This is marked as "should be treated as fatal" in AvalancheGo, but there are other errors that could be returned here that should not be treated as fatal.

So... I think we will likely continue treating this as non-fatal here, since treating it as fatal would probably involve a special cased error or extra return value, which doesn't seem worth it.

If there's an alternative way you'd suggest treating this as fatal, then it would make sense to add the TODO, otherwise I think we should remove this.

peer/waiting_handler.go
```diff
@@ -34,5 +34,10 @@ func (w *waitingResponseHandler) OnFailure() error {
 
 // newWaitingResponseHandler returns new instance of the waitingResponseHandler
 func newWaitingResponseHandler() *waitingResponseHandler {
-	return &waitingResponseHandler{responseChan: make(chan []byte)}
+	return &waitingResponseHandler{
+		// Make buffer length 1 so that OnResponse can complete
+		// even if no goroutine is waiting on the channel (i.e.
+		// the context of a request is cancelled.)
+		responseChan: make(chan []byte, 1),
```
aaronbuchwald, line 41: Can we add a regression test to catch the error where the response handlers deadlock if we did not initialize the responseChan to size 1?

peer/network.go
danlaine: In avalanchego's `sync` package we have an error value for this case and check for it (see [here](https://github.com/ava-labs/avalanchego/blob/master/x/sync/client.go#L300) and [here](https://github.com/ava-labs/avalanchego/blob/master/x/sync/network_client.go#L266)).

I think that we should try to honor the invariant on the `AppSender`, but I don't think it's _strictly_ required, so if you don't want to do something like in avalanchego then I can remove these comments.

peer/waiting_handler.go
danlaine: Added for `TestAppRequestAnyOnCtxCancellation`. I can also add similar logic to other tests if you want -- lmk.

peer/network.go
aaronbuchwald: Can we remove this? I think that makes the code harder to follow

aaronbuchwald (approved):

PR #876, opened 2023-09-19
enforce warpIndex is <= MaxInt32

darioush:
## Why this should be merged
int is 32bits or 64bits

## How this works
Enforces the warpIndex to be not larger than MaxInt32 so casting to int cannot overflow if int is 32 bits

## How this was tested
CI

## How is this documented
N/A


ceyonur (approved):
In [commit](https://github.com/ava-labs/subnet-evm/pull/823/commits/e20b025264ce99343ec5c56057cc2a3ee6ac084b) we have changed the index type form `uint256` to `uint32` for the same concern; but I guess an explicit check is even better.

abi87 (approved):

PR #877, opened 2023-09-20
Add Extra Unit Test to Warp Aggregation Job 

anusha-ctrl:
## Why this should be merged
Context: Unit tests already cover parent context get canceled, before fetching signatures. In that case no signatures get fetched and we get an `avalancheWarp.ErrInsufficientWeight`

What about in the case that the context gets cancelled while signatures are being fetched? This unit test cancels parent context after a delay which should also cancel the child context as well. We are able to get 1 signature before the context gets cancelled.  All subsequent `signatureJobs` should return an err and we should still get an `avalancheWarp.ErrInsufficientWeight`  with a specific error message


warp/aggregator/aggregation_job_test.go
```diff
@@ -353,3 +354,56 @@ func TestAggregateThresholdSignaturesParentCtxCancels(t *testing.T) {
...
+			},
+		},
+		unsignedMsg,
+	)
+
+	// Specify error because it should have been able to grab one signature exactly but is still insufficient weight, so quorumNum * totalWeight > quorumDen * sigWeight
+	executeSignatureAggregationTest(t, signatureAggregationTest{
+		ctx:         ctx,
+		job:         aggregationJob,
+		expectedErr: errors.New("signature weight is insufficient: 40*500 > 100*100"),
```
darioush, line 407: could we make this a typed err so we can keep checking with ErrorIs vs. repeating the string?

warp/aggregator/aggregation_job_test.go
```diff
@@ -353,3 +354,56 @@ func TestAggregateThresholdSignaturesParentCtxCancels(t *testing.T) {
 		expectedErr: avalancheWarp.ErrInsufficientWeight,
 	})
 }
+
+func TestAggregateThresholdSignaturesParentCtxCancelsAfterDelay(t *testing.T) {
+	// Get one signature exactly and then parent context cancels
```
darioush, line 359: could we build this behavior deterministically instead of relying on the closeness of 1 second to 1001 ms (and also without slowing down the test for 1s)

anusha-ctrl: I can reduce the test length, but not sure how else I can deterministically cancel the parent context in the middle of the test without using timers. I don't exactly know when the test grabs a signature (unless we added channels to report that)

anusha-ctrl: Using a timer to block fetching signatures until it fires is the closest thing I have to approximating the function time.

anusha-ctrl:
closing in favor #936. I wrote this test prior to the refactor. It no longer applies.

PR #878, opened 2023-09-20
Add more unit tests for aggregator

anusha-ctrl:
## Why this should be merged
This new unit test is failing even though it shouldn't be.
Why does this test get 500/500 signature weight? Shouldn't it the child context be canceled by VerifyWeight after we hit 300/500 or 60/100 and therefore we shouldn't be able to fetch more signatures?''

Edit: Spun up 100 go routines. `signatureFetchCancel()`, aka the child ctx gets called exactly right. cancel propagates down to `fetchSignature and gets BLS signatures for about 80 go routines and the rest get handled by ctx.Done(). This makes me think we will fetch signatures from every node, even if a cancel triggers. 

#881  this should help the overfetching


aaronbuchwald:
Closing in favor of https://github.com/ava-labs/subnet-evm/pull/936

PR #881, opened 2023-09-20
Add Contexts Continued 

anusha-ctrl:
## Why this should be merged
This is an additional supplement to Dan's PR on the `add-contexts` branch. 
All Send functions in Networking now take in a context. 
If contexed cancelled prior to being called, we should return a ctx cancelled err. We also don't generate a new context on sends and instead pass the existing one through. 
## How this was tested
UT

### DIFF
--- a/peer/client.go
+++ b/peer/client.go
@@ -24,7 +24,7 @@ type NetworkClient interface {
 	// node version greater than or equal to minVersion.
 	// Returns response bytes, the ID of the chosen peer, and ErrRequestFailed if
 	// the request should be retried.
-	SendAppRequestAny(ctx context.Context, cominVersion *version.Application, request []byte) ([]byte, ids.NodeID, error)
+	SendAppRequestAny(ctx context.Context, minVersion *version.Application, request []byte) ([]byte, ids.NodeID, error)
 
 	// SendAppRequest synchronously sends request to the selected nodeID
 	// Returns response bytes, and ErrRequestFailed if the request should be retried.
--- a/peer/client.go
+++ b/peer/client.go
@@ -24,7 +24,7 @@ type NetworkClient interface {
 	// node version greater than or equal to minVersion.
 	// Returns response bytes, the ID of the chosen peer, and ErrRequestFailed if
 	// the request should be retried.
-	SendAppRequestAny(minVersion *version.Application, request []byte) ([]byte, ids.NodeID, error)
+	SendAppRequestAny(ctx context.Context, cominVersion *version.Application, request []byte) ([]byte, ids.NodeID, error)
 
 	// SendAppRequest synchronously sends request to the selected nodeID
 	// Returns response bytes, and ErrRequestFailed if the request should be retried.
@@ -32,7 +32,7 @@ type NetworkClient interface {
 
 	// SendCrossChainRequest sends a request to a specific blockchain running on this node.
 	// Returns response bytes, and ErrRequestFailed if the request failed.
-	SendCrossChainRequest(chainID ids.ID, request []byte) ([]byte, error)
+	SendCrossChainRequest(ctx context.Context, chainID ids.ID, request []byte) ([]byte, error)
 
 	// Gossip sends given gossip message to peers
 	Gossip(gossip []byte) error
@@ -60,24 +60,29 @@ func NewNetworkClient(network Network) NetworkClient {
 // node version greater than or equal to minVersion.
 // Returns response bytes, the ID of the chosen peer, and ErrRequestFailed if
 // the request should be retried.
-func (c *client) SendAppRequestAny(minVersion *version.Application, request []byte) ([]byte, ids.NodeID, error) {
+func (c *client) SendAppRequestAny(ctx context.Context, minVersion *version.Application, request []byte) ([]byte, ids.NodeID, error) {
 	waitingHandler := newWaitingResponseHandler()
-	nodeID, err := c.network.SendAppRequestAny(minVersion, request, waitingHandler)
+	nodeID, err := c.network.SendAppRequestAny(ctx, minVersion, request, waitingHandler)
 	if err != nil {
 		return nil, nodeID, err
 	}
-	response := <-waitingHandler.responseChan
-	if waitingHandler.failed {
-		return nil, nodeID, ErrRequestFailed
+
+	select {
+	case <-ctx.Done():
+		return nil, nodeID, ctx.Err()
+	case response := <-waitingHandler.responseChan:
+		if waitingHandler.failed {
+			return nil, nodeID, ErrRequestFailed
+		}
+		return response, nodeID, nil
 	}
-	return response, nodeID, nil
 }
 
 // SendAppRequest synchronously sends request to the specified nodeID
 // Returns response bytes and ErrRequestFailed if the request should be retried.
 func (c *client) SendAppRequest(ctx context.Context, nodeID ids.NodeID, request []byte) ([]byte, error) {
 	waitingHandler := newWaitingResponseHandler()
-	if err := c.network.SendAppRequest(nodeID, request, waitingHandler); err != nil {
+	if err := c.network.SendAppRequest(ctx, nodeID, request, waitingHandler); err != nil {
 		return nil, err
 	}
 
@@ -94,16 +99,20 @@ func (c *client) SendAppRequest(ctx context.Context, nodeID ids.NodeID, request 
 
 // SendCrossChainRequest synchronously sends request to the specified chainID
 // Returns response bytes and ErrRequestFailed if the request should be retried.
-func (c *client) SendCrossChainRequest(chainID ids.ID, request []byte) ([]byte, error) {
+func (c *client) SendCrossChainRequest(ctx context.Context, chainID ids.ID, request []byte) ([]byte, error) {
 	waitingHandler := newWaitingResponseHandler()
-	if err := c.network.SendCrossChainRequest(chainID, request, waitingHandler); err != nil {
+	if err := c.network.SendCrossChainRequest(ctx, chainID, request, waitingHandler); err != nil {
 		return nil, err
 	}
-	response := <-waitingHandler.responseChan
-	if waitingHandler.failed {
-		return nil, ErrRequestFailed
+	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	case response := <-waitingHandler.responseChan:
+		if waitingHandler.failed {
+			return nil, ErrRequestFailed
+		}
+		return response, nil
 	}
-	return response, nil
 }
 
 func (c *client) Gossip(gossip []byte) error {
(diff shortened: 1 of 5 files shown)


peer/client.go
```diff
@@ -24,15 +24,15 @@ type NetworkClient interface {
 	// node version greater than or equal to minVersion.
 	// Returns response bytes, the ID of the chosen peer, and ErrRequestFailed if
 	// the request should be retried.
-	SendAppRequestAny(minVersion *version.Application, request []byte) ([]byte, ids.NodeID, error)
+	SendAppRequestAny(ctx context.Context, cominVersion *version.Application, request []byte) ([]byte, ids.NodeID, error)
```
darioush, line 27:
```suggestion
	SendAppRequestAny(ctx context.Context, minVersion *version.Application, request []byte) ([]byte, ids.NodeID, error)
```

danlaine, line 27:
```suggestion
	SendAppRequestAny(ctx context.Context, minVersion *version.Application, request []byte) ([]byte, ids.NodeID, error)
```

peer/network_test.go
```diff
@@ -261,14 +309,74 @@ func TestAppRequestOnShutdown(t *testing.T) {
...
+
+	requestMessage := HelloRequest{Message: "this is a request"}
+	requestBytes, err := message.RequestToBytes(codecManager, requestMessage)
+	assert.NoError(t, err)
+
+	ctx, cancel := context.WithCancel(context.Background())
+	// cancel context prior to sending
+	cancel()
+	_, _, err = client.SendAppRequestAny(ctx, defaultPeerVersion, requestBytes)
+	assert.ErrorContains(t, err, "context canceled")
```
darioush, line 377: could we check for ErrorIs?

peer/network.go
```diff
@@ -190,7 +190,7 @@ func (n *network) sendAppRequest(nodeID ids.NodeID, request []byte, responseHand
 
 	// Send app request to [nodeID].
 	// On failure, release the slot from [activeAppRequests] and delete request from [outstandingRequestHandlers]
-	if err := n.appSender.SendAppRequest(context.TODO(), nodeIDs, requestID, request); err != nil {
+	if err := n.appSender.SendAppRequest(ctx, nodeIDs, requestID, request); err != nil {
```
danlaine, line 193: The `common.AppSender` interface says that a non-nil error should be treated as fatal but we currently don't honor that. Can you add a TODO for us to do that pls?

peer/network.go
```diff
@@ -203,7 +203,7 @@ func (n *network) sendAppRequest(nodeID ids.NodeID, request []byte, responseHand
 // SendCrossChainRequest sends request message bytes to specified chainID and adds [handler] to [outstandingRequestHandlers]
 // so that it can be invoked when the network receives either a response or failure message.
 // Returns an error if [appSender] is unable to make the request.
-func (n *network) SendCrossChainRequest(chainID ids.ID, request []byte, handler message.ResponseHandler) error {
+func (n *network) SendCrossChainRequest(ctx context.Context, chainID ids.ID, request []byte, handler message.ResponseHandler) error {
 	// Take a slot from total [activeCrossChainRequests] and block until a slot becomes available.
 	if err := n.activeCrossChainRequests.Acquire(context.Background(), 1); err != nil {
```
danlaine, line 208:
```suggestion
	if err := n.activeCrossChainRequests.Acquire(ctx, 1); err != nil {
```

danlaine: same for other `Acquire` calls

peer/network.go
```diff
@@ -222,7 +222,7 @@ func (n *network) SendCrossChainRequest(chainID ids.ID, request []byte, handler
 
 	// Send cross chain request to [chainID].
 	// On failure, release the slot from [activeCrossChainRequests] and delete request from [outstandingRequestHandlers].
-	if err := n.appSender.SendCrossChainAppRequest(context.TODO(), chainID, requestID, request); err != nil {
+	if err := n.appSender.SendCrossChainAppRequest(ctx, chainID, requestID, request); err != nil {
```
danlaine, line 225: Same regarding TODO on handling fatal error

peer/network_test.go
```diff
@@ -132,6 +132,54 @@ func TestRequestAnyRequestsRoutingAndResponse(t *testing.T) {
...
+	requestMessage := HelloRequest{Message: "this is a request"}
+	requestBytes, err := message.RequestToBytes(codecManager, requestMessage)
+	assert.NoError(t, err)
+
+	nodeID := ids.GenerateTestNodeID()
+	ctx, cancel := context.WithCancel(context.Background())
+	// cancel context prior to sending
+	cancel()
+	_, err = client.SendAppRequest(ctx, nodeID, requestBytes)
+	assert.ErrorContains(t, err, "context canceled")
```
danlaine, line 180: We should use `ErrorIs` to assert errors, not `ErrorContains`

peer/network_test.go
```diff
@@ -132,6 +132,54 @@ func TestRequestAnyRequestsRoutingAndResponse(t *testing.T) {
...
 
+func TestAppRequestOnCtxCancellation(t *testing.T) {
+	var net Network
+	codecManager := buildCodec(t, HelloRequest{}, HelloResponse{})
+	crossChainCodecManager := buildCodec(t, ExampleCrossChainRequest{}, ExampleCrossChainResponse{})
+
+	callNum := uint32(0)
+	senderWg := &sync.WaitGroup{}
+	sender := testAppSender{
+		sendAppRequestFn: func(nodes set.Set[ids.NodeID], requestID uint32, requestBytes []byte) error {
```
danlaine, line 143: Are these functions actually doing anything? Seems like they could be empty. The test passes when they are, and making them empty would allow some code removal.

peer/network_test.go
```diff
@@ -261,14 +309,74 @@ func TestAppRequestOnShutdown(t *testing.T) {
...
 
+func TestAppRequestAnyOnCtxCancellation(t *testing.T) {
+	var net Network
+	codecManager := buildCodec(t, HelloRequest{}, HelloResponse{})
+	crossChainCodecManager := buildCodec(t, ExampleCrossChainRequest{}, ExampleCrossChainResponse{})
+
+	callNum := uint32(0)
+	senderWg := &sync.WaitGroup{}
+	sender := testAppSender{
+		sendAppRequestFn: func(nodes set.Set[ids.NodeID], requestID uint32, requestBytes []byte) error {
```
danlaine, line 328: Same as above -- we can remove a bunch of this code

danlaine (changes requested):

peer/client.go
anusha-ctrl: thanks

anusha-ctrl: thanks

peer/network.go
anusha-ctrl: thanks

peer/network_test.go
anusha-ctrl: Not really, can remove

anusha-ctrl: Actually we need to define them since we use the functions in `SendAppRequest` etc

anusha-ctrl: Otherwise we would be accessing a nil pointer

peer/network.go
anusha-ctrl: thanks

peer/network_test.go
```diff
@@ -132,6 +132,54 @@ func TestRequestAnyRequestsRoutingAndResponse(t *testing.T) {
...
+	requestBytes, err := message.RequestToBytes(codecManager, requestMessage)
+	assert.NoError(t, err)
+
+	nodeID := ids.GenerateTestNodeID()
+	ctx, cancel := context.WithCancel(context.Background())
+	// cancel context prior to sending
+	cancel()
+	_, err = client.SendAppRequest(ctx, nodeID, requestBytes)
+	assert.ErrorIs(t, err, context.Canceled)
+}
```
danlaine, line 181:
```suggestion
func TestAppRequestOnCtxCancellation(t *testing.T) {
	codecManager := buildCodec(t, HelloRequest{}, HelloResponse{})
	crossChainCodecManager := buildCodec(t, ExampleCrossChainRequest{}, ExampleCrossChainResponse{})

	sender := testAppSender{
		sendAppRequestFn: func(set.Set[ids.NodeID], uint32, []byte) error {
			return nil
		},
		sendAppResponseFn: func(ids.NodeID, uint32, []byte) error {
			return nil
		},
	}

	net := NewNetwork(p2p.NewRouter(logging.NoLog{}, nil, prometheus.NewRegistry(), ""), sender, codecManager, crossChainCodecManager, ids.EmptyNodeID, 1, 1)
	net.SetRequestHandler(&HelloGreetingRequestHandler{codec: codecManager})
	client := NewNetworkClient(net)

	requestMessage := HelloRequest{Message: "this is a request"}
	requestBytes, err := message.RequestToBytes(codecManager, requestMessage)
	assert.NoError(t, err)

	nodeID := ids.GenerateTestNodeID()
	ctx, cancel := context.WithCancel(context.Background())
	// cancel context prior to sending
	cancel()
	_, err = client.SendAppRequest(ctx, nodeID, requestBytes)
	assert.ErrorIs(t, err, context.Canceled)
}
```

peer/network_test.go
```diff
@@ -261,14 +309,74 @@ func TestAppRequestOnShutdown(t *testing.T) {
...
+				atomic.AddUint32(&callNum, 1)
+			}()
+			return nil
+		},
+	}
+
+	net = NewNetwork(p2p.NewRouter(logging.NoLog{}, nil, prometheus.NewRegistry(), ""), sender, codecManager, crossChainCodecManager, ids.EmptyNodeID, 1, 1)
+	net.SetRequestHandler(&HelloGreetingRequestHandler{codec: codecManager})
+	client := NewNetworkClient(net)
+	for i := 0; i < 3; i++ {
```
danlaine, line 355: Should we just make one peer? I don't think having more really changes the test at all.

peer/network_test.go
```diff
@@ -261,14 +309,74 @@ func TestAppRequestOnShutdown(t *testing.T) {
...
+		assert.NoError(t,
+			net.Connected(
+				context.Background(),
+				ids.GenerateTestNodeID(),
+				&version.Application{
+					Major: 1,
+					Minor: 7,
+					Patch: i,
+				},
```
danlaine, line 364: Should we just use `version.CurrentApp`?

peer/network_test.go
```diff
@@ -261,14 +309,74 @@ func TestAppRequestOnShutdown(t *testing.T) {
...
+	sender := testAppSender{
+		sendAppRequestFn: func(nodes set.Set[ids.NodeID], requestID uint32, requestBytes []byte) error {
+			nodeID, _ := nodes.Pop()
+			senderWg.Add(1)
+			go func() {
+				defer senderWg.Done()
+				if err := net.AppRequest(context.Background(), nodeID, requestID, time.Now().Add(5*time.Second), requestBytes); err != nil {
+					panic(err)
+				}
+			}()
```
danlaine, line 336: Same as above, we don't actually need these callbacks to do anything. They can just return nil.

darioush (commented):
I think this is a good opportunity to clean up some of these tests with the suggestions Dan has mentioned. 
If we could further reduce the duplication or make the tests in this file shorter it would be an improvement.

danlaine (approved):

PR #883, opened 2023-09-20
Signature Aggregation Refactor

danlaine:
## Why this should be merged

Removes/simplifies code.

## How this works

* Change log levels from `Info` to `Debug`
* Rename `aggregator.NewAggregator` to `aggregator.New`
* Rename `SignatureBackend` to `SignatureGetter` and `FetchWarpSignature` to `GetSignature`
* Remove `signatureAggregationJob` and `signatureJob`. These types had one method, execute, which was called immediately after creation. I think it's cleaner to have the business logic in the aggregator itself.

## How this was tested

Rewrote unit tests.

## How is this documented

Comments -- interface didn't change.


warp/aggregator/aggregator.go
```diff
@@ -36,16 +62,133 @@ func (a *Aggregator) AggregateSignatures(ctx context.Context, unsignedMessage *a
...
+		return nil, fmt.Errorf("failed to get validator set: %w", err)
+	}
+	if len(validators) == 0 {
+		return nil, fmt.Errorf("%w (SubnetID: %s, Height: %d)", errNoValidators, a.subnetID, pChainHeight)
+	}
+
+	return aggregateSignatures(ctx, a.client, unsignedMessage, validators, quorumNum, totalWeight)
+}
+
+func aggregateSignatures(
```
danlaine, line 81: I thought it made sense to split `AggregateSignatures` into the part that gets the required state from `a` and the part that deals with actually acquiring and aggregating the signatures but I could also just combine these back into 1 function.

warp/aggregator/aggregator.go
```diff
@@ -36,16 +62,133 @@ func (a *Aggregator) AggregateSignatures(ctx context.Context, unsignedMessage *a
...
+
+			log.Info("Fetching warp signature",
+				"nodeID", nodeID,
+				"index", i,
+			)
+
+			signature, err := client.GetSignature(signatureFetchCtx, nodeID, unsignedMessage)
+			if err != nil {
+				log.Info("Failed to fetch warp signature",
+					"nodeID", nodeID,
```
anusha-ctrl, line 121: should this be a log.Error

warp/aggregator/aggregator.go
```diff
@@ -36,16 +62,133 @@ func (a *Aggregator) AggregateSignatures(ctx context.Context, unsignedMessage *a
...
+		var (
+			i         = i
+			validator = validator
+			nodeID    = validator.NodeIDs[0]
+		)
+		go func() {
+			defer wg.Done()
+
+			log.Info("Fetching warp signature",
+				"nodeID", nodeID,
```
anusha-ctrl, line 122: these logs will be really noisy, maybe move some of these to `log.Debug`

warp/aggregator/aggregator.go
danlaine: It's expected that we won't always be able to get a signature from a node -- this should actually be debug

warp/aggregator/network_signature_backend.go
```diff
@@ -30,11 +30,11 @@ type NetworkSigner struct {
...
 }
 
-// FetchWarpSignature attempts to fetch a BLS Signature of [unsignedWarpMessage] from [nodeID] until it succeeds or receives an invalid response
+// GetSignature attempts to fetch a BLS Signature of [unsignedWarpMessage] from [nodeID] until it succeeds or receives an invalid response
 //
 // Note: this function will continue attempting to fetch the signature from [nodeID] until it receives an invalid value or [ctx] is cancelled.
 // The caller is responsible to cancel [ctx] if it no longer needs to fetch this signature.
-func (s *NetworkSigner) FetchWarpSignature(ctx context.Context, nodeID ids.NodeID, unsignedWarpMessage *avalancheWarp.UnsignedMessage) (*bls.Signature, error) {
+func (s *NetworkSigner) GetSignature(ctx context.Context, nodeID ids.NodeID, unsignedWarpMessage *avalancheWarp.UnsignedMessage) (*bls.Signature, error) {
 	signatureReq := message.SignatureRequest{
```
anusha-ctrl, line 38: personally found this name change too similar to the  `backend.GetSignature`

warp/aggregator/aggregator_test.go
```diff
@@ -0,0 +1,334 @@
...
+				client.EXPECT().GetSignature(gomock.Any(), nodeID1, gomock.Any()).Return(nonVdrSig, nil)
+				client.EXPECT().GetSignature(gomock.Any(), nodeID2, gomock.Any()).Return(nil, errTest)
+				client.EXPECT().GetSignature(gomock.Any(), nodeID3, gomock.Any()).Return(sig3, nil)
+				return New(subnetID, state, client)
+			},
+			unsignedMsg:     unsignedMsg,
+			quorumNum:       30,
+			expectedSigners: []*avalancheWarp.Validator{vdr3},
+			expectedErr:     nil,
+		},
```
anusha-ctrl, line 294: nice tests, do we have some where the parent context cancels?

warp/aggregator/aggregator.go
ceyonur: I have no strong opinion but seems we can combine those two

warp/aggregator/aggregator.go
```diff
@@ -36,16 +70,134 @@ func (a *Aggregator) AggregateSignatures(ctx context.Context, unsignedMessage *a
...
+					"nodeID", nodeID,
+					"index", i,
+					"err", err,
+				)
+				return
+			}
+			log.Debug("Retrieved warp signature",
+				"nodeID", nodeID,
+				"index", i,
+				"signature", hexutil.Bytes(bls.SignatureToBytes(signature)),
```
ceyonur, line 139: can we assign this signature to some variable before calling this so we can use it below as well?

warp/aggregator/aggregator.go
```diff
@@ -36,16 +70,134 @@ func (a *Aggregator) AggregateSignatures(ctx context.Context, unsignedMessage *a
...
+
+			// If the signature weight meets the requested threshold, cancel signature fetching
+			if err := avalancheWarp.VerifyWeight(signatureWeight, totalWeight, quorumNum, params.WarpQuorumDenominator); err == nil {
+				log.Debug("Verify weight passed, exiting aggregation early",
+					"maxNeededQuorumNum", quorumNum,
+					"totalWeight", totalWeight,
+					"signatureWeight", signatureWeight,
+				)
+				signatureFetchCancel()
+			}
```
ceyonur, line 172: It seems that we double check this case once 
1- a go routine verifies this and cancels signature fetching. 
2- then the main go routine will again check this.

This also probably will try to fetch and verify signatures more than it actually requires: Sub go routines will be blocked at `signatureLock` for most of the active validators, meaning that they will perform this verification even if some other routine already verified before (and keep appending to already verified signature set). I wonder if splitting fetch + verification into two different routines would help this to not perform duplicate verifications. I.e
1- Spawn go routines for each validator and fetch signature
2- Fetched signatures can be fed to another channel 
3- Another goroutine could perform verification once it receives signature, without blocking fetching goroutines
4- When verification goroutine verifies the signature set it can cancel validator routines and returns with verified signatures.
5- Verification goroutine should wait until waitgroup is completed. If no validator goroutine left, return err. 

If verification goroutine successfully verifies, we wouldn't need to verify it again before aggregation.   

Alternatively we can try to exit from validator go routines after `signatureLock` is released (i.e with a check if `signatureFetchCancel` is called)

ceyonur (commented):
some comments for optimizing signature verification.

warp/aggregator/aggregator.go
```diff
@@ -36,16 +71,150 @@ func (a *Aggregator) AggregateSignatures(ctx context.Context, unsignedMessage *a
...
+	for i, validator := range validators {
+		var (
+			i         = i
+			validator = validator
+			// TODO: update from a single nodeID to the original slice and use extra nodeIDs as backup.
+			nodeID = validator.NodeIDs[0]
+		)
+		go func() {
+			log.Debug("Fetching warp signature",
+				"nodeID", nodeID,
```
darioush, line 119: I think these logs should include the message ID to be useful for debugging

warp/aggregator/aggregator.go
```diff
@@ -36,16 +71,150 @@ func (a *Aggregator) AggregateSignatures(ctx context.Context, unsignedMessage *a
...
+				)
+				signatureFetchResultChan <- nil
+				return
+			}
+
+			sigHex := hexutil.Bytes(bls.SignatureToBytes(signature))
+			log.Debug("Retrieved warp signature",
+				"nodeID", nodeID,
+				"index", i,
+				"signature", sigHex,
```
darioush, line 138: nit: not really sure we gain much by logging `sigHex`

warp/aggregator/aggregator.go
```diff
@@ -36,16 +71,150 @@ func (a *Aggregator) AggregateSignatures(ctx context.Context, unsignedMessage *a
...
+		index  int
+		weight uint64
+	}
+
+	// Create a child context to cancel signature fetching if we reach signature threshold.
+	signatureFetchCtx, signatureFetchCancel := context.WithCancel(ctx)
+	defer signatureFetchCancel()
+
+	// Fetch signatures from validators concurrently.
+	signatureFetchResultChan := make(chan *signatureFetchResult, len(validators))
```
darioush, line 109:
```suggestion
	signatureFetchResultChan := make(chan *signatureFetchResult)
```
I don't think this channel needs to be buffered?

warp/aggregator/aggregator.go
```diff
@@ -36,16 +71,150 @@ func (a *Aggregator) AggregateSignatures(ctx context.Context, unsignedMessage *a
...
+			)
+			signatureFetchCancel()
+			signaturesPassedThreshold = true
+			break
+		}
+	}
+
+	// If I failed to fetch sufficient signature stake, return an error
+	if !signaturesPassedThreshold {
+		return nil, errInsufficientWeight
```
darioush, line 196: should we use avalancheWarp.ErrInsufficientWeight here?

warp/aggregator/aggregator.go
```diff
@@ -36,16 +71,150 @@ func (a *Aggregator) AggregateSignatures(ctx context.Context, unsignedMessage *a
...
+		signatures = append(signatures, signatureFetchResult.sig)
+		signersBitset.Add(signatureFetchResult.index)
+		signaturesWeight += signatureFetchResult.weight
+		log.Debug("Updated weight",
+			"totalWeight", signaturesWeight,
+			"addedWeight", signatureFetchResult.weight,
+		)
+
+		// If the signature weight meets the requested threshold, cancel signature fetching
+		if err := avalancheWarp.VerifyWeight(signaturesWeight, totalWeight, quorumNum, params.WarpQuorumDenominator); err == nil {
```
darioush, line 182: I think we should check if err is avalancheWarp.ErrInsufficientWeight then continue else if err != nil log err or sthg so we can check unexpected errors

danlaine: `avalancheWarp.ErrInsufficientWeight` is the only error `VerifyWeight` can return so idk if we want to check for that error in particular. If people feel strongly I can change this but I don't think this would buy us anything

warp/aggregator/aggregator.go
danlaine: good catch

darioush (approved):
Overall lgtm, couple thoughts:
- whether we want to limit the concurrency of the requests (eg if we need 2/3 signatures maybe we should only request from 3/4 weight instead of everyone)
- use the validators’ weights to somehow optimize which validators to request from 
- should we add retries?

but all of these can be done in follow up PRs

ceyonur (approved):
Agreed with @darioush's comments, but I think we can unblock this PR and work on a separate one.
LGTM!

PR #886, opened 2023-09-20
add UT for miner core.ApplyTransaction err case

darioush:


darioush:
I think this test is too convoluted of a setup to add as is. Instead we improve the testability of the underlying code

PR #891, opened 2023-09-22
Update preparePredicateStorageSlots doc

minghinmatthewlam:
## Why this should be merged
Doc is outdated, we now support multiple access tuples with the same precompile address. 

## How this works
n/a

## How this was tested
n/a

## How is this documented
n/a

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -1197,8 +1197,9 @@ func (s *StateDB) Prepare(rules params.Rules, sender, coinbase common.Address, d
 }
 
 // preparePredicateStorageSlots populates the predicateStorageSlots field from the transaction's access list
-// Note: if an address is specified multiple times in the access list, the storage slots for that address are
-// appended to a list of predicate bytes, and every predicate in the list goes through verification.
+// Note: if an address is specified multiple times in the access list, each storage slot for that address is
+// appended to a slice of byte slices. Each byte slice represents a predicate, making it a slice of predicates
+// for each access list address, and every predicate in the slice goes through verification.
 func (s *StateDB) preparePredicateStorageSlots(rules params.Rules, list types.AccessList) {
 	s.predicateStorageSlots = make(map[common.Address][][]byte)
 	for _, el := range list {
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -1197,10 +1197,8 @@ func (s *StateDB) Prepare(rules params.Rules, sender, coinbase common.Address, d
 }
 
 // preparePredicateStorageSlots populates the predicateStorageSlots field from the transaction's access list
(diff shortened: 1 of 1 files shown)


darioush (approved):

core/state/statedb.go
```diff
@@ -1197,10 +1197,8 @@ func (s *StateDB) Prepare(rules params.Rules, sender, coinbase common.Address, d
 }
 
 // preparePredicateStorageSlots populates the predicateStorageSlots field from the transaction's access list
-// Note: if an address is specified multiple times in the access list, only the last storage slots provided
-// for it are used in predicates.
-// During predicate verification, we require that a precompile address is only specififed in the access list
-// once to avoid a situation where we verify multiple predicate and only expose data from the last one.
+// Note: if an address is specified multiple times in the access list, the storage slots for that address are
+// appended to a list of predicate bytes, and every predicate in the list goes through verification.
```
aaronbuchwald, line 1201: This comment is not very clear. Appended to a list of predicate bytes is not very clear / specific to explain that it is type `[][]byte`. I think we should try to make this more clear before merging.

core/state/statedb.go
```diff
@@ -1197,10 +1197,9 @@ func (s *StateDB) Prepare(rules params.Rules, sender, coinbase common.Address, d
...
 
 // preparePredicateStorageSlots populates the predicateStorageSlots field from the transaction's access list
-// Note: if an address is specified multiple times in the access list, only the last storage slots provided
-// for it are used in predicates.
-// During predicate verification, we require that a precompile address is only specififed in the access list
-// once to avoid a situation where we verify multiple predicate and only expose data from the last one.
+// Note: if an address is specified multiple times in the access list, each storage slot for that address is
+// appended to a slice of byte slices. Each byte slice represents a predicate, making it a slice of predicates
+// for each access list address, and every predicate in the slice goes through verification.
 func (s *StateDB) preparePredicateStorageSlots(rules params.Rules, list types.AccessList) {
```
minghinmatthewlam, line 1203: noticed that the predicate check https://github.com/ava-labs/subnet-evm/blob/master/core/predicate_check.go#L35C47-L35C47 composing of `predicateArguments` is pretty similar to `preparePredicateStorageSlots`, not sure if any use in consolidating the two.

core/state/statedb.go
minghinmatthewlam: updated to be more specific to use of slice of byte slices.

core/state/statedb.go
aaronbuchwald: Yup great find, I think this would make sense to add as a helper in https://github.com/ava-labs/subnet-evm/tree/master/utils/predicate. Do you want to make a PR to refactor this so that these two both call that helper?

aaronbuchwald (approved):

PR #895, opened 2023-09-25
SetTxPredicateResults: Consistent overwriting

darioush:
## Why this should be merged
More consistent behavior for SetTxPredicateResults. This also matches the existing comment which suggests present results will be overrided.

## How this works
Removes an previous entries from the map if an empty entry is provided for the same Tx hash.

## How this was tested
CI

## How is this documented
Updated comments


StephenButtolph (approved):

StephenButtolph:
Should we add a regression test?

ceyonur (approved):

PR #896, opened 2023-09-25
add err return to NewEVMBlockContext

darioush:


darioush:
close since we merged the alternative solution

PR #897, opened 2023-09-25
NewEVMBlockContext: more explicit err path

darioush:

### DIFF
--- a/core/evm.go
+++ b/core/evm.go
@@ -51,20 +51,18 @@ type ChainContext interface {
 
 // NewEVMBlockContext creates a new context for use in the EVM.
 func NewEVMBlockContext(header *types.Header, chain ChainContext, author *common.Address) vm.BlockContext {
-	var (
-		predicateResults *results.PredicateResults
-		err              error
-	)
-	if len(header.Extra) > params.DynamicFeeExtraDataSize {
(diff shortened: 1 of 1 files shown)


core/evm.go
```diff
@@ -51,20 +51,18 @@ type ChainContext interface {
...
 	}
 
+	// Prior to the DUpgrade, the VM enforces the extra data is smaller than or
+	// equal to this size. After the DUpgrade, the VM pre-verifies the extra
+	// data past the dynamic fee rollup window is valid.
+	predicateResults, err := results.ParsePredicateResults(header.Extra[params.DynamicFeeExtraDataSize:])
+	if err != nil {
+		log.Error("failed to parse predicate results creating new block context", "err", err, "extra", header.Extra)
+		return newEVMBlockContext(header, chain, author, nil)
+	}
```
aaronbuchwald, line 65:
```suggestion
	if err != nil {
		log.Error("failed to parse predicate results creating new block context", "err", err, "extra", header.Extra)
		// As mentioned above, we pre-verify the extra data to ensure this never happens.
		// If we hit an error, construct a new block context rather than use a potentially half initialized value
		// as defense in depth.
		return newEVMBlockContext(header, chain, author, nil)
	}
```

aaronbuchwald (commented):
Requested one change

ceyonur (approved):

PR #901, opened 2023-09-25
Invalidate block when missing predicate context

StephenButtolph:
## Why this should be merged

Verification of a predicate requires a the predicate context. If the predicate context is missing, we should invalidate the whole block rather than marking the result of the predicate as failed.

## How this works

Moves the check out of the warp package and into the core package.

## How this was tested

CI

## How is this documented

With a comment

### DIFF
--- a/core/predicate_check.go
+++ b/core/predicate_check.go
@@ -4,6 +4,7 @@
 package core
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/ava-labs/subnet-evm/core/types"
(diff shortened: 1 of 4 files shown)


ceyonur (approved):

darioush (approved):

dhrubabasu (approved):

core/predicate_check.go
```diff
@@ -35,6 +38,16 @@ func CheckPredicates(rules params.Rules, predicateContext *precompileconfig.Pred
 	// Prepare the predicate storage slots from the transaction's access list
 	predicateArguments := predicateutils.PreparePredicateStorageSlots(rules, tx.AccessList())
 
+	// If there are no predicates to verify, we must terminate early because the
+	// proposervm block context isn't guaranteed to be populated.
+	if len(predicateArguments) == 0 {
+		return predicateResults, nil
+	}
```
aaronbuchwald, line 45:
```suggestion
	// If there are no predicates to verify, return early and skip requiring the proposervm block
	// context to be populated.
```

saying we must terminate early is a bit confusing to me here, I think it's better to explicitly call out the check we're skipping by returning early here.

PR #906, opened 2023-09-26
add get predicate result bytes helpers

### LINKED ISSUE #905: Add a helper method to fetch predicate bytes from extra data
We use offset that does not have any reference to index of predicate bytes and has to be known by heart. We should add an helper method to fetch predicate bytes from extraData/header and remove all these hardcoded accesses.
https://github.com/ava-labs/subnet-evm/blob/master/plugin/evm/block.go#L295

ceyonur:
Closes #905

### DIFF
--- a/core/evm.go
+++ b/core/evm.go
@@ -32,8 +32,8 @@ import (
 	"github.com/ava-labs/subnet-evm/consensus"
 	"github.com/ava-labs/subnet-evm/core/types"
 	"github.com/ava-labs/subnet-evm/core/vm"
-	"github.com/ava-labs/subnet-evm/params"
 	"github.com/ava-labs/subnet-evm/precompile/results"
+	"github.com/ava-labs/subnet-evm/utils/predicate"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/log"
 	//"github.com/ethereum/go-ethereum/log"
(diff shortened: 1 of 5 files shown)


utils/predicate/predicate_bytes.go
```diff
@@ -66,3 +67,20 @@ func BytesToHashSlice(b []byte) []common.Hash {
...
 }
+
+// GetPredicateResultBytes returns the predicate result bytes from the extra data.
+// If the extra data does not contain predicate result bytes, an error is returned.
+func GetPredicateResultBytes(extraData []byte) ([]byte, error) {
+	// Prior to the DUpgrade, the VM enforces the extra data is smaller than or equal to this size.
+	// After the DUpgrade, the VM pre-verifies the extra data past the dynamic fee rollup window is
+	// valid.
+	if len(extraData) < params.DynamicFeeExtraDataSize {
+		return nil, fmt.Errorf("header extra data too short for predicate verification found: %d, required: %d", len(extraData), params.DynamicFeeExtraDataSize)
```
darioush, line 78: should we make this a typed error?

utils/predicate/predicate_bytes.go
```diff
@@ -66,3 +67,20 @@ func BytesToHashSlice(b []byte) []common.Hash {
...
+	}
+	return extraData[params.DynamicFeeExtraDataSize:], nil
+}
+
+// HasPredicateResultBytes returns true if the extra data contains predicate result bytes.
+func HasPredicateResultBytes(extraData []byte) bool {
```
darioush, line 84: seems unused? not against adding this as an exported function for later use though

core/evm.go
```diff
@@ -51,15 +51,9 @@ type ChainContext interface {
...
-		predicateResults *results.PredicateResults
-		err              error
-	)
-	if len(header.Extra) > params.DynamicFeeExtraDataSize {
-		// Prior to the DUpgrade, the VM enforces the extra data is smaller than or equal to this size.
-		// After the DUpgrade, the VM pre-verifies the extra data past the dynamic fee rollup window is
-		// valid.
-		predicateResults, err = results.ParsePredicateResults(header.Extra[params.DynamicFeeExtraDataSize:])
+	var predicateResults *results.PredicateResults
+	if predicateBytes, err := predicate.GetPredicateResultBytes(header.Extra); err == nil {
```
darioush, line 55: nit: maybe add a comment like 
```
// Parse predicate results if they are present (ie GetPredicateResultBytes returns err == nil)
```

darioush (approved):

core/evm.go
ceyonur: added

utils/predicate/predicate_bytes.go
ceyonur: let's remove it for now.

utils/predicate/predicate_bytes.go
ceyonur: done

aaronbuchwald (approved):

PR #907, opened 2023-09-26
move predicate package and merge with predicate utils

ceyonur:
Merges two predicate packages and moves to a single package. It seemed to me that having a predicate package under precompiles might not be the best fit as they're not actually related.


plugin/evm/block.go
```diff
@@ -18,8 +18,8 @@ import (
 	"github.com/ava-labs/subnet-evm/core/types"
 	"github.com/ava-labs/subnet-evm/params"
 	"github.com/ava-labs/subnet-evm/precompile/precompileconfig"
-	"github.com/ava-labs/subnet-evm/precompile/results"
-	"github.com/ava-labs/subnet-evm/utils/predicate"
+	"github.com/ava-labs/subnet-evm/predicate"
+	predicate1 "github.com/ava-labs/subnet-evm/predicate"
```
darioush, line 22: could we fix this?

ceyonur: fixed

core/evm.go
```diff
@@ -51,10 +50,10 @@ type ChainContext interface {
 
 // NewEVMBlockContext creates a new context for use in the EVM.
 func NewEVMBlockContext(header *types.Header, chain ChainContext, author *common.Address) vm.BlockContext {
-	var predicateResults *results.PredicateResults
+	var predicateResults *predicate.PredicateResults
```
aaronbuchwald, line 53: Could we change the type from `PredicateResults` to `Results` so it doesn't stutter?

aaronbuchwald (approved):

params/config.go
```diff
@@ -386,11 +386,11 @@ func (c *ChainConfig) IsDUpgrade(time uint64) bool {
 }
 
 func (r *Rules) PredicatesExist() bool {
```
darioush, line 388: should this be PredicatersExist?

params/config.go
```diff
@@ -719,9 +719,9 @@ type Rules struct {
 	// Note: none of these addresses should conflict with the address space used by
 	// any existing precompiles.
 	ActivePrecompiles map[common.Address]precompileconfig.Config
-	// Predicates maps addresses to stateful precompile predicate functions
+	// Predicaters maps addresses to stateful precompile predicate functions
```
darioush, line 722:
```suggestion
	// Predicaters maps addresses to stateful precompile Predicaters
```

precompile/contract/interfaces.go
```diff
@@ -64,7 +64,7 @@ type BlockContext interface {
 	ConfigurationBlockContext
 	// GetPredicateResuls returns an arbitrary byte array result of verifying the predicates
```
darioush, line 65:
```suggestion
	// GetResults returns an arbitrary byte array result of verifying the predicates
```

core/predicate_check.go
```diff
@@ -31,12 +31,12 @@ func CheckPredicates(rules params.Rules, predicateContext *precompileconfig.Pred
...
 	predicateResults := make(map[common.Address][]byte)
 	// Short circuit early if there are no precompile predicates to verify
-	if len(rules.Predicates) == 0 {
+	if len(rules.Predicaters) == 0 {
 		return predicateResults, nil
 	}
 
 	// Prepare the predicate storage slots from the transaction's access list
-	predicateArguments := predicateutils.PreparePredicateStorageSlots(rules, tx.AccessList())
+	predicateArguments := predicate.PreparePredicateStorageSlots(rules, tx.AccessList())
```
darioush, line 39: should we also remove "Predicate" from this fn name?

tests/warp/warp_test.go
```diff
@@ -181,7 +181,6 @@ var _ = ginkgo.Describe("[Warp]", ginkgo.Ordered, func() {
 		log.Info("Creating ethclient for blockchainB", "wsURI", chainBWSURI)
 		chainBWSClient, err = ethclient.Dial(chainBWSURI)
 		gomega.Expect(err).Should(gomega.BeNil())
-
```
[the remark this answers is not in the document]
ceyonur: weird I'd expected to be caught by current linter as we use `whitespace` in enabled rules.

predicate/predicate_results.go
```diff
@@ -70,17 +70,17 @@ func ParsePredicateResults(b []byte) (*PredicateResults, error) {
...
 	if !ok {
 		return nil
 	}
 	return txResults[address]
 }
 
-// SetTxPredicateResults sets the predicate results for the given [txHash]. Overrides results if present.
-func (p *PredicateResults) SetTxPredicateResults(txHash common.Hash, txResults TxPredicateResults) {
+// SetTxResults sets the predicate results for the given [txHash]. Overrides results if present.
+func (p *Results) SetTxResults(txHash common.Hash, txResults TxResults) {
```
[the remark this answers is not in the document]
ceyonur: yup

core/vm/evm.go
```diff
@@ -133,11 +133,11 @@ func (b *BlockContext) Timestamp() uint64 {
 	return b.Time
 }
 
-func (b *BlockContext) GetPredicateResults(txHash common.Hash, address common.Address) []byte {
+func (b *BlockContext) GetResults(txHash common.Hash, address common.Address) []byte {
```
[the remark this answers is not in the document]
ceyonur: good catch!

params/config.go
ceyonur: changed

core/predicate_check.go
ceyonur: you mean from `CheckPredicates`? I guess it's fine. Alternatively we can rename it to something like `RunPredicaters` etc.

aaronbuchwald (approved):

PR #908, opened 2023-09-27
Consolidate predicate prepare storage slots function

minghinmatthewlam:
## Why this should be merged
In `predicate_check` and `state_db` we were duplicating code for preparing the predicate storage slots from a transaction's access list. This PR consolidates them to a `util` function to make sure they're in sync and reduces duplication.

## How this works
- Creates separate file in utils `predicate_slots.go`
- Create utils `PreparePredicateStorageSlots` function
- Use util function for previous places.

## How this was tested
n/a

## How is this documented
comments in code


aaronbuchwald (approved):
LGTM - nice change, thanks for putting this up!

ceyonur (approved):

PR #913, opened 2023-09-27
Fixed config type cast

abi87:
## Why this should be merged
A faulty type cast that won't log correctly. This PR fixes it.

## How this works
Just log the right structs

## How this was tested
CI

### DIFF
--- a/precompile/contracts/txallowlist/module.go
+++ b/precompile/contracts/txallowlist/module.go
@@ -43,7 +43,7 @@ func (*configurator) MakeConfig() precompileconfig.Config {
 func (*configurator) Configure(chainConfig precompileconfig.ChainConfig, cfg precompileconfig.Config, state contract.StateDB, blockContext contract.ConfigurationBlockContext) error {
 	config, ok := cfg.(*Config)
 	if !ok {
-		return fmt.Errorf("incorrect config %T: %v", config, config)
+		return fmt.Errorf("expected config type %T, got %T: %v", &Config{}, cfg, cfg)
 	}
 	return config.AllowListConfig.Configure(chainConfig, ContractAddress, state, blockContext)
 }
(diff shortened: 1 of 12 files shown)


x/warp/module.go
```diff
@@ -48,9 +48,8 @@ func (*configurator) MakeConfig() precompileconfig.Config {
 
 // Configure is a no-op for warp since it does not need to store any information in the state
 func (*configurator) Configure(chainConfig precompileconfig.ChainConfig, cfg precompileconfig.Config, state contract.StateDB, _ contract.ConfigurationBlockContext) error {
-	config, ok := cfg.(*Config)
-	if !ok {
-		return fmt.Errorf("incorrect config %T: %v", config, config)
+	if _, ok := cfg.(*Config); !ok {
+		return fmt.Errorf("expected config type %T, got %T: %v", &Config{}, cfg, cfg)
```
ceyonur, line 52: We do have same issue in other precompiles + template. Could you do a search + replace to fix others as well?

ceyonur (changes requested):
We should fix other places as well.

x/warp/module.go
abi87: Done.

ceyonur (approved):

precompile/contracts/txallowlist/module.go
```diff
@@ -43,7 +43,7 @@ func (*configurator) MakeConfig() precompileconfig.Config {
 func (*configurator) Configure(chainConfig precompileconfig.ChainConfig, cfg precompileconfig.Config, state contract.StateDB, blockContext contract.ConfigurationBlockContext) error {
 	config, ok := cfg.(*Config)
```
ceyonur, line 44: hmm we should probably replace `config` with `_` as we don't use it (to fix UTs)

darioush (approved):

Issue #868, opened 2023-09-19
More explicit support for `getVerifiedWarpMessage` as Anycast vs. Explicit Destination


aaronbuchwald:
The Warp precompile currently supports `getVerifiedWarpMessage`, which will verify that a warp message was signed by a threshold of the source blockchainID's validator set. The address-to-address payload type specifies a `destinationChainID` as well, and the current implementation does not require that the `destinationChainID` matches the actual `blockchainID` of the blockchain receiving this message.

The motivation for this was to support Anycast messages in a lightweight way and defer verification of the `destinationChainID` to the caller.

This ticket is a feature request to support a more explicit "safe" version of `getVerifiedWarpMessage` that does this check for you and separates the Anycast and explicit destination handler functions.

Specifically, this proposes that we break it down into two functions:

- `getVerifiedWarpMessage` which verifies the `destinationChainID` matches the receiving chain's `blockchainID`
- `getAnyCastVerifiedWarpMessage` which verifies the `destinationChainID` is exactly the `AnycastChainID` defined in AvalancheGo (or alternatively keeps the current behavior of deferring to the caller to allow any `blockchainID` is specified for the `destinationChainID`

michaelkaplan13:
This is an interesting idea, though my initial feeling that it is best to make as few assumptions as possible on how messages will be used by developers. 

For instance, it's imaginable that a project would want to be able to receive warp messages sent to a specific chain that is neither their own chain ID or the anycast chain ID. It could be used for reporting messages sent or other uses cases aside from "anycasting". It'd be odd to have to use `getAnyCastVerifiedWarpMessage` in that case. 

100% agree that it needs to be very clear that the `destinationChainID` for messages returned by `getVerifiedWarpMessage` can have any value and is not necessarily the current chain ID. Ultimately though, I think this may result in more code that needs to be maintained to offer the same functionality.

michaelkaplan13:
Trying to think how else an additional guard rail around the `destinationChainID` could be implemented here, which is what I think the real goal is.

One option could be to add an `allowOtherChainIDs` boolean parameter to `getVerifiedWarpMessage`. It would need to be explicitly set to `true` to allow chain IDs that are not the chain ID of the current chain. Protocol contracts such as `TeleporterMessenger` would set it to `false` rather than having to veryify the `destinationChainID` in their own logic.

richardpringle:
[...]

PR #920, opened 2023-09-28
Remove destinationAddress and destinationChainID from Warp Message

nytzuga:
## Why this should be merged

Remove destinationChainID and destinationAddress from the payload. Other libraries can implement their own mechanism to make the messages specific instead of anycast. More context in https://github.com/ava-labs/subnet-evm/issues/868

## How this works

Remove any reference to destinationChainID and destinationAddress

## How this was tested

Unit tests and pipeline

## How is this documented

Internal documentation will be updated to reflect changes https://github.com/ava-labs/subnet-evm/issues/868


contracts/contracts/interfaces/IWarpMessenger.sol
```diff
@@ -19,12 +17,7 @@ struct WarpBlockHash {
 }
 
 interface IWarpMessenger {
-    event SendWarpMessage(
-        bytes32 indexed destinationChainID,
-        address indexed destinationAddress,
-        address indexed sender,
-        bytes message
-    );
+    event SendWarpMessage(address indexed sender, bytes message);
```
aaronbuchwald, line 20: note: since we removed two index fields with this change, we could include the unsigned warp message ID if we want to.

This should be a separate change if we want it though. cc @michaelkaplan13 @ceyonur

warp/payload/README.md
```diff
@@ -13,10 +13,6 @@ AddressedPayload:
 +---------------------+----------+-------------------+
 |       sourceAddress : [20]byte |          20 bytes |
 +---------------------+----------+-------------------+
-|  destinationChainID : [32]byte |          32 bytes |
-+---------------------+----------+-------------------+
-|  destinationAddress : [20]byte |          20 bytes |
-+---------------------+----------+-------------------+
```
aaronbuchwald, line 19: Can we update the total size as well since these fields have been removed?

aaronbuchwald (changes requested):
Could you update the total size of the addressed payload in the README now that these two fields are removed? After that, this change LGTM

tests/warp/warp_test.go
```diff
@@ -200,9 +200,7 @@ var _ = ginkgo.Describe("[Warp]", ginkgo.Ordered, func() {
 		gomega.Expect(err).Should(gomega.BeNil())
 
 		packedInput, err := warp.PackSendWarpMessage(warp.SendWarpMessageInput{
-			DestinationChainID: common.Hash(blockchainIDB),
-			DestinationAddress: fundedAddress,
-			Payload:            payload,
+			Payload: payload,
```
darioush, line 203: Should we remove SendWarpMessageInput as a struct since it's just []byte for payload now?

darioush (approved):
I think the code change looks good.
The change itself makes it such that applications must emit logs for destination chain and address if indexing is desired. I think we could make it clear in the readme that the typical application will need to encode the destination chain and contract on sending then verify this information as the first step in processing.

tests/warp/warp_test.go
nytzuga: I think this will make it future proof, but I also like the idea of just passing `[]byte`. It's ultimately your call @aaronbuchwald

contracts/contracts/interfaces/IWarpMessenger.sol
ceyonur: Yup that's what I thought after seeing this lol. It would be a great change!

contracts/contracts/ExampleWarp.sol
```diff
@@ -9,60 +9,42 @@ contract ExampleWarp {
     IWarpMessenger warp = IWarpMessenger(WARP_ADDRESS);
 
     // sendWarpMessage sends a warp message to the specified destination chain and address pair containing the payload
```
ceyonur, line 11:
```suggestion
    // sendWarpMessage sends a warp message containing the payload
```

tests/warp/warp_test.go
aaronbuchwald: Yup, we should make it just `[]byte`. Could you re-generate the stateful precompile using the precompile gen tool and replace the packers/unpackers that are impacted by this change? One of them will be making this change to `[]byte`

contracts/contracts/interfaces/IWarpMessenger.sol
michaelkaplan13: I think that'd be a great idea. To confirm, that's the ID used for querying validators for their signature of a specific message right?

x/warp/contract.go
```diff
@@ -191,18 +184,18 @@ func getVerifiedWarpMessage(accessibleState contract.AccessibleState, caller com
...
 }
 
-// UnpackSendWarpMessageInput attempts to unpack [input] as SendWarpMessageInput
+// UnpackSendWarpMessageInput attempts to unpack [input] as []byte
 // assumes that [input] does not include selector (omits first 4 func signature bytes)
-func UnpackSendWarpMessageInput(input []byte) (SendWarpMessageInput, error) {
-	inputStruct := SendWarpMessageInput{}
+func UnpackSendWarpMessageInput(input []byte) ([]byte, error) {
+	inputStruct := []byte{}
 	err := WarpABI.UnpackInputIntoInterface(&inputStruct, "sendWarpMessage", input)
```
ceyonur, line 191: We should be using `UnpackInput` rather than `UnpackInputIntoInterface`. See the comment: https://github.com/ava-labs/subnet-evm/blob/warp-remove-destinationChainID-destinationAddress/accounts/abi/abi.go#L216-L218

Generator should be generating something like:
```
func UnpackSendWarpMessageInput(input []byte) ([]byte, error) {
	res, err := ContractABI.UnpackInput("sendWarpMessage", input)
	if err != nil {
		return []byte{}, err
	}
	unpacked := *abi.ConvertType(res[0], new([]byte)).(*[]byte)
	return unpacked, nil
}
```

ceyonur (changes requested):
Could we regenerate the precompile to see if we need to modify packers/unpackers for changed inputs?

darioush (approved):

aaronbuchwald (approved):

Issue #925, opened 2023-09-29
Return Message ID from `SendWarpMessage`


ceyonur:
Currently we don't return anything from `SendWarpMessage` function as an output. This seems a less weird if it successfully sends a message. We can return the message ID.

Also after #920, we can put this message ID in event for an easier filter for subscribers. Without #920 we won't be able to put this to topics because maximum 3 (+ event ID) topic are supported.

PR #930, opened 2023-10-03
Return message ID

ceyonur:
## Why this should be merged

This adds unsigned message ID as the both output to `SendWarpMessage` and indexed topic to the event.

Closes: #925 

## How this was tested

Added both UT + hardhat tests

## How is this documented

Updated Readme and interface.


x/warp/contract.go
```diff
@@ -264,13 +282,18 @@ func sendWarpMessage(accessibleState contract.AccessibleState, caller common.Add
...
 	)
 
+	packed, err := PackSendWarpMessageOutput(common.Hash(unsignedWarpMessage.ID()))
+	if err != nil {
+		return nil, remainingGas, err
+	}
+
 	// Return an empty output and the remaining gas
-	return []byte{}, remainingGas, nil
+	return packed, remainingGas, nil
```
anusha-ctrl, line 291: nit: Adjust comment on 290 as output is no longer empty.

x/warp/contract.go
```diff
@@ -264,13 +282,18 @@ func sendWarpMessage(accessibleState contract.AccessibleState, caller common.Add
 		accessibleState.GetBlockContext().Number().Uint64(),
 	)
 
+	packed, err := PackSendWarpMessageOutput(common.Hash(unsignedWarpMessage.ID()))
+	if err != nil {
```
anusha-ctrl, line 286: curious as to why we now return the packed message output? is this for verification purposes for testing? can we also update the function description?

ceyonur: we have changed the function signature to return messageID as output.

contracts/contracts/interfaces/IWarpMessenger.sol
```diff
@@ -6,46 +6,46 @@
 pragma solidity ^0.8.0;
 
 struct WarpMessage {
-    bytes32 sourceChainID;
-    address originSenderAddress;
-    bytes payload;
+  bytes32 sourceChainID;
+  address originSenderAddress;
+  bytes payload;
```
aaronbuchwald, line 11: can we revert back to the original indentation throughout this file?

aaronbuchwald (changes requested):

contracts/contracts/interfaces/IWarpMessenger.sol
ceyonur: We use 2 spaces for tab (specified in .prettierrc). So this is the corrected format. I also corrected spaces in ExampleWarp.sol

anusha-ctrl (approved):

PR #933, opened 2023-10-03
t8ntool: tiny bugfix for difficulty field

tactical-retreat:
## Why this should be merged

Pretty straightforward bugfix. 

## How this was tested

Tests pass before and after.

### DIFF
--- a/cmd/evm/internal/t8ntool/block.go
+++ b/cmd/evm/internal/t8ntool/block.go
@@ -156,7 +156,7 @@ func (i *bbInput) ToBlock() *types.Block {
 	if i.Header.Nonce != nil {
 		header.Nonce = *i.Header.Nonce
 	}
-	if header.Difficulty != nil {
+	if i.Header.Difficulty != nil {
 		header.Difficulty = i.Header.Difficulty
 	}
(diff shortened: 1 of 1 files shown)


cmd/evm/internal/t8ntool/block.go
```diff
@@ -156,7 +156,7 @@ func (i *bbInput) ToBlock() *types.Block {
 	if i.Header.Nonce != nil {
 		header.Nonce = *i.Header.Nonce
 	}
-	if header.Difficulty != nil {
+	if i.Header.Difficulty != nil {
```
aaronbuchwald, line 159: nice catch, looks like this mistake was copied from upstream: https://github.com/ethereum/go-ethereum/blob/master/cmd/evm/internal/t8ntool/block.go#L161

Do you want to make a PR to geth to get credit for your work upstream as well?

aaronbuchwald (approved):

cmd/evm/internal/t8ntool/block.go
tactical-retreat: Hah, I didn't even think of that. Will send them a PR too, thanks.

ceyonur (approved):

Issue #894, opened 2023-09-25
SendWarpMessage Event Message Field is not being packed correctly


ceyonur:
**Describe the bug**
`Message` field with arbitrary byte type in `SendWarpMessage` needs to be packed correctly to support web3 tooling.

**To Reproduce**
Run a network with `run_warp_network.sh`
Run tests with :
```
EXPECTED_UNSIGNED_MESSAGE="0000000030399eb4b63147034a240984ebd48d34283169bf1f6f30023a353a6b75e4e5a99fa600000003010203" RPC_URI="http://127.0.0.1:60257/ext/bc/7LmkQsZzTUFdSGquBiP7sKG97AVCLn8mUa7hAP7pDLpctPbT6/rpc" DESTINATION_CHAIN_ID="1" PAYLOAD="010203" npx hardhat test test/warp.ts --network local
```

or just simply subscribe to the event via web3 with 
```
  contracts.list.IWarpMessenger.events
    .SendWarpMessage(
      {
        fromBlock: "latest",
      },
      function (error, event) {
        console.log(event);
      }
    )
    .on("connected", function (subscriptionId) {
      console.log(subscriptionId);
    })
    .on("data", function (event) {
      console.log(event); // same results as the optional callback above
    })
    .on("changed", function (event) {
      // remove event from local database
    })
    .on("error", function (error, receipt) {
      // If the transaction was rejected by the network with a receipt, the second parameter will be the receipt.
      console.log(error);
    });
 ```

**Expected behavior**
Should be able to decode the event and run tests according to the expected parameters. 

Or in web3 case, should log the subscribed event.

**Logs**
For hardhat tests:
```
  ExampleWarp
Sending warp message to chain: e902a9a86640bfdb1cd0e36c0cc982b83e5765fad5f6bbe6abdcce7b5ae7d7c7, address: 0x0550000000000000000000000000000000000000 with payload 010203
    1) contract should be to send warp message


  0 passing (5s)
  1 failing

  1) ExampleWarp
       contract should be to send warp message:
     Error: overflow [ See: https://links.ethers.org/v5-errors-NUMERIC_FAULT-overflow ] (fault="overflow", operation="toNumber", value="550033034597317853516956781285697674262213869476469076103517423188", code=NUMERIC_FAULT, version=bignumber/5.7.0)
      at Logger.makeError (node_modules/@ethersproject/logger/src.ts/index.ts:269:28)
[...]

PR #935, opened 2023-10-04
add hardhat tests for warp

ceyonur:
Adds hardhat contract tests for warp messenger. CI is expected to fail since we have discovered a bug with these tests. It will be fixed with #894


contracts/test/warp.ts
```diff
@@ -0,0 +1,46 @@
...
+    await expect(contract.sendWarpMessage(payloadHex))
+      .to.emit(contract, 'SendWarpMessage')
+      .withArgs(senderAddress, unsignedMessageHex);
+  })
+
+  it("should be able to fetch correct blockchain ID", async function () {
+    let sourceID = process.env["SOURCE_CHAIN_ID"];
+    let sourceIDHex = "0x" + sourceID.toString().padStart(32, "0");
+    expect(ethers.utils.isHexString(sourceIDHex)).to.be.true;
+  })
```
aaronbuchwald, line 45: Do we need a HardHat test that is only checking the value of an environment variable set by the caller?

contracts/test/warp.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  it("contract should be to send warp message", async function () {
+    let payload = process.env["PAYLOAD"];
+    let expectedUnsignedMessage = process.env["EXPECTED_UNSIGNED_MESSAGE"];
+    let payloadHex = "0x" + payload.toString()
+    expect(ethers.utils.isHexString(payloadHex)).to.be.true;
+    let unsignedMessageHex = "0x" + expectedUnsignedMessage.toString()
+    expect(ethers.utils.isHexString(unsignedMessageHex)).to.be.true;
+
+    console.log(`Sending warp message with payload ${payloadHex}, unsigned message ${unsignedMessageHex}`);
```
aaronbuchwald, line 34: If we need to pass a hex string in anyways, can we reduce the amount of JavaScript code that we need by setting the environment variable to the exact string that we need?

Could we also move all of the env variables to be extracted as constants at the top of the file instead of extracting them within the individual `it` nodes?

tests/warp/warp_test.go
```diff
@@ -375,4 +366,106 @@ var _ = ginkgo.Describe("[Warp]", ginkgo.Ordered, func() {
...
+	gomega.Expect(err).Should(gomega.BeNil())
+	log.Info("Running test command", "cmd", cmd.String())
+
+	out, err := cmd.CombinedOutput()
+	fmt.Printf("\nCombined output:\n\n%s\n", string(out))
+	gomega.Expect(err).Should(gomega.BeNil())
+}
+
+func toRPCURI(uri string, blockchainID string) string {
+	return fmt.Sprintf("%s/ext/bc/%s/rpc", uri, blockchainID)
```
aaronbuchwald, line 470: Can we use `utils.RunHardHatTests` here instead? It looks like the only difference between the two is logs, which we can move outside of this function if it's important to include extra logs that specify this is warp related.

tests/warp/warp_test.go
```diff
@@ -375,4 +366,106 @@ var _ = ginkgo.Describe("[Warp]", ginkgo.Ordered, func() {
...
+			for {
+				block, err := client.BlockByNumber(ctx, nil)
+				gomega.Expect(err).Should(gomega.BeNil())
+				if block.NumberU64() >= newHead.Number.Uint64() {
+					log.Info("client accepted the block containing SendWarpMessage", "client", i, "height", block.NumberU64())
+					break
+				}
+			}
+		}
+	})
```
aaronbuchwald, line 446: Where is the next stage this is referring to, should we get rid of this if there is no later stage in favor of checking the logs in both HardHat and golang on a single node?

tests/warp/warp_test.go
```diff
@@ -39,9 +43,20 @@ import (
...
+	unsignedWarpMessageID          ids.ID
+	signedWarpMsg                  *avalancheWarp.Message
+	blockchainIDA, blockchainIDB   ids.ID
+	chainAURIs, chainBURIs         []string
+	chainAWSClient, chainBWSClient ethclient.Client
+	chainID                        = big.NewInt(99999)
+	fundedKey                      *ecdsa.PrivateKey
+	fundedAddress                  common.Address
+	payload                        = []byte{1, 2, 3}
+	txSigner                       = types.LatestSignerForChainID(chainID)
```
aaronbuchwald, line 59: Does moving around these variables need to be part of this PR? If not, could we separate it out?

aaronbuchwald (changes requested):

tests/warp/warp_test.go
ceyonur: https://github.com/ava-labs/subnet-evm/pull/940/files

tests/warp/warp_test.go
ceyonur: cleaned it up

tests/warp/warp_test.go
ceyonur: `RunHardHatTests` uses a default port (9650) whereas we use a custom one in these tests. Added a new helper method and reduced duplicate code.

contracts/test/warp.ts
ceyonur: done

contracts/test/warp.ts
ceyonur: certainly no... added the actual test.

tests/utils/command.go
```diff
@@ -95,9 +95,12 @@ func RegisterNodeRun() {
 // [execPath] is the path where the test command is executed
 func RunHardhatTests(ctx context.Context, blockchainID string, execPath string, testPath string) {
 	chainURI := GetDefaultChainURI(blockchainID)
+	RunHardhatTestsCustomURI(ctx, chainURI, execPath, testPath)
```
aaronbuchwald, line 98: ❤️

contracts/test/warp.ts
aaronbuchwald: Can we move this to the top of the file as opposed to in the `before` node?

contracts/test/warp.ts
```diff
@@ -0,0 +1,49 @@
...
+  let owner: SignerWithAddress
+  let contract: Contract
+  before(async function () {
+    senderAddress = process.env["SENDER_ADDRESS"];
+    owner = await ethers.getSigner(senderAddress);
+    contract = await ethers.getContractAt("IWarpMessenger", WARP_ADDRESS, owner)
+
+    payload = process.env["PAYLOAD"];
+    expectedUnsignedMessage = process.env["EXPECTED_UNSIGNED_MESSAGE"];
+    sourceID = process.env["SOURCE_CHAIN_ID"];
```
aaronbuchwald, line 30: Can we move these variables to be set on lines 12-16 instead of in the before suite?

contracts/test/warp.ts
```diff
@@ -0,0 +1,49 @@
...
+  });
+
+  it("contract should be to send warp message", async function () {
+    expect(ethers.utils.isHexString(payload)).to.be.true;
+    expect(ethers.utils.isHexString(expectedUnsignedMessage)).to.be.true;
+
```
aaronbuchwald, line 36: Can we get rid of these since we're setting them ourselves in a different part of the test and logging them below?

if this were one test, I don't think we'd add a check that we had correctly constructed a test variable correctly. This is definitely a little different, but would prefer that we remove these checks.

aaronbuchwald (approved):

aaronbuchwald (approved):

contracts/test/warp.ts
```diff
@@ -0,0 +1,40 @@
...
+const WARP_ADDRESS = "0x0200000000000000000000000000000000000005";
+let senderAddress =  process.env["SENDER_ADDRESS"];
+// Expected to be a hex string
+let payload = process.env["PAYLOAD"];
+let expectedUnsignedMessage = process.env["EXPECTED_UNSIGNED_MESSAGE"];
+let sourceID = process.env["SOURCE_CHAIN_ID"];
+
+describe("IWarpMessenger", function () {
+  this.timeout("30s")
```
anusha-ctrl, line 19: might be worth setting up a global mocha config file to specify timeouts. Multiple tests are using the 30 second timeout.

anusha-ctrl (approved):

anusha-ctrl (approved):

aaronbuchwald (approved):

PR #936, opened 2023-10-04
Add Unit Test for Test Aggregate Signatures

anusha-ctrl:
## Why this should be merged

This unit test simulates the scenario where the context is canceled midway through signature fetching. 

All signature fetching go routines prior to cancellation should be able to get a signature. 

All signature fetching go routines after cancellation should return with context cancelled error and no signature. 

## How this works
The result should have an aggregate signature from only one validator.
## How this was tested
Only adds a UT. No actual change to any prod code. 
## How is this documented
UT is commented.

### DIFF
--- a/warp/aggregator/aggregator_test.go
+++ b/warp/aggregator/aggregator_test.go
@@ -76,20 +76,22 @@ func TestAggregateSignatures(t *testing.T) {
 	}
 
 	type test struct {
-		name            string
-		contextFunc     func() context.Context
-		aggregatorFunc  func(*gomock.Controller) *Aggregator
-		unsignedMsg     *avalancheWarp.UnsignedMessage
-		quorumNum       uint64
-		expectedSigners []*avalancheWarp.Validator
-		expectedErr     error
+		name                  string
+		contextWithCancelFunc func() (context.Context, context.CancelFunc)
+		aggregatorFunc        func(*gomock.Controller, context.CancelFunc) *Aggregator
+		unsignedMsg           *avalancheWarp.UnsignedMessage
+		quorumNum             uint64
+		expectedSigners       []*avalancheWarp.Validator
+		expectedErr           error
 	}
 
 	tests := []test{
 		{
-			name:        "can't get height",
-			contextFunc: context.Background,
-			aggregatorFunc: func(ctrl *gomock.Controller) *Aggregator {
+			name: "can't get height",
+			contextWithCancelFunc: func() (context.Context, context.CancelFunc) {
+				return context.Background(), nil
+			},
+			aggregatorFunc: func(ctrl *gomock.Controller, _ context.CancelFunc) *Aggregator {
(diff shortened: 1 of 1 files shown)


warp/aggregator/aggregator_test.go
```diff
@@ -390,9 +477,10 @@ func TestAggregateSignatures(t *testing.T) {
 			ctrl := gomock.NewController(t)
 			require := require.New(t)
 
-			a := tt.aggregatorFunc(ctrl)
+			ctx, cancel := tt.contextWithCancelFunc()
+			a := tt.aggregatorFunc(ctrl, cancel)
```
darioush, line 481: should we add a check like:
`if cancel != nil { defer cancel() }` to avoid leaking goroutine or is it guaranteed that `aggregatorFunc` calls cancel?

warp/aggregator/aggregator_test.go
```diff
@@ -354,9 +378,72 @@ func TestAggregateSignatures(t *testing.T) {
...
+				)
+				client.EXPECT().GetSignature(gomock.Any(), nodeID2, gomock.Any()).DoAndReturn(
+					func(ctx context.Context, _ ids.NodeID, _ *avalancheWarp.UnsignedMessage) (*bls.Signature, error) {
+						select {
+						case <-ctx.Done():
+							err := ctx.Err()
+							require.ErrorIs(t, err, context.Canceled)
+							return nil, err
+						default:
+							// We should not be able to get this signature since context cancelled in another go routine
```
darioush, line 416: if this is the case then is it necessary to add this expected call to the mock?
if it's necessary, could we factor out the common code here (eg calling `client.EXPECT()` in a loop)

warp/aggregator/aggregator_test.go
anusha-ctrl: Currently in all the places where cancel is not nil, aggregatorFunc calls `cancel`. 
I can put this check in place for safety

warp/aggregator/aggregator_test.go
anusha-ctrl: I see, I can remove the default case then. I just thought it would be more clear that this case would NOT happen.

warp/aggregator/aggregator_test.go
```diff
@@ -390,9 +469,14 @@ func TestAggregateSignatures(t *testing.T) {
 			ctrl := gomock.NewController(t)
 			require := require.New(t)
 
-			a := tt.aggregatorFunc(ctrl)
+			ctx, cancel := tt.contextWithCancelFunc()
+			// Guarantees that cancel is called preventing memory leak
```
darioush, line 473:
```suggestion
			// Guarantees that cancel is called preventing goroutine leak
```

darioush (approved):

aaronbuchwald (approved):

PR #942, opened 2023-10-05
warp: use bitset in predicate results to encode indexes of failures instead of success

darioush:
## Why this should be merged
Warp messages are encoded in the transaction's access list as a predicate, and the result of verifying the predicate with calling VerifyPredicate (success, fail) is stored the header as an encoded bitset. If we expect most of the transactions to represent successes (as opposed to txs that pay gas but reference invalid warp messages in the predicate), we can use 0 to encode success and 1 to encode failure.
This change has the advantage of using the empty byte slice to represent a valid transaction that only specifies valid warp messages.

The current implementation does the opposite: 1 for success and 0 for failure, which adds one or more bytes in the header for valid txs that only specify valid warp messages.

## How this works
Changes the logic to add 1 to the bitset on error instead of success

## How this was tested
CI + additional UT

## How is this documented
N/A (implementation change)

### DIFF
--- a/x/warp/config.go
+++ b/x/warp/config.go
@@ -197,12 +197,13 @@ func (c *Config) verifyPredicate(predicateContext *precompileconfig.PredicateCon
 	return c.verifyWarpMessage(predicateContext, warpMessage)
 }
 
-// VerifyPredicate verifies the predicate represents a valid signed and properly formatted Avalanche Warp Message.
+// VerifyPredicate computes indices of predicates that failed verification as a bitset then returns the result
+// as a byte slice.
 func (c *Config) VerifyPredicate(predicateContext *precompileconfig.PredicateContext, predicates [][]byte) []byte {
 	resultBitSet := set.NewBits()
 
 	for predicateIndex, predicateBytes := range predicates {
-		if c.verifyPredicate(predicateContext, predicateBytes) {
+		if !c.verifyPredicate(predicateContext, predicateBytes) {
 			resultBitSet.Add(predicateIndex)
 		}
 	}
--- a/x/warp/contract_test.go
+++ b/x/warp/contract_test.go
@@ -201,9 +201,7 @@ func TestGetVerifiedWarpMessage(t *testing.T) {
 			Caller: callerAddr,
 			InputFn: func(t testing.TB) []byte {
 				input, err := PackGetVerifiedWarpMessage(1)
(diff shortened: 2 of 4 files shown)


x/warp/contract_test.go
```diff
@@ -252,11 +254,36 @@ func TestGetVerifiedWarpMessage(t *testing.T) {
 				return res
 			}(),
 		},
+		"get message failure non-zero index": {
```
darioush, line 257: seems we should add a case like this regardless of this change.

ceyonur (approved):

x/warp/contract_test.go
```diff
@@ -252,11 +254,36 @@ func TestGetVerifiedWarpMessage(t *testing.T) {
...
+			Caller: callerAddr,
+			InputFn: func(t testing.TB) []byte {
+				input, err := PackGetVerifiedWarpMessage(1)
+				if err != nil {
+					t.Fatal(err)
+				}
```
anusha-ctrl, line 263: curious as to why we don't use `require.NoError` here?

x/warp/config.go
```diff
@@ -197,12 +197,13 @@ func (c *Config) verifyPredicate(predicateContext *precompileconfig.PredicateCon
...
 
-// VerifyPredicate verifies the predicate represents a valid signed and properly formatted Avalanche Warp Message.
+// VerifyPredicate computes indices of predicates that failed verification as a bitset then returns the result
+// as a byte slice.
 func (c *Config) VerifyPredicate(predicateContext *precompileconfig.PredicateContext, predicates [][]byte) []byte {
 	resultBitSet := set.NewBits()
 
 	for predicateIndex, predicateBytes := range predicates {
-		if c.verifyPredicate(predicateContext, predicateBytes) {
+		if !c.verifyPredicate(predicateContext, predicateBytes) {
```
anusha-ctrl, line 206: can we add a comment to explain this change here? a 1 is added to the bitset to describe failure instead of 0? or an empty bitset represents all successes?

anusha-ctrl (approved):

x/warp/config.go
darioush: There's a comment for the function that explains this behavior, so I don't think it's useful to replicate it from a couple lines above here.

anusha-ctrl (approved):

PR #943, opened 2023-10-06
predicate: document predicate results size cap

### LINKED ISSUE #824: Document Warp Precompile Result Size Limitations
https://github.com/ava-labs/subnet-evm/pull/823 encodes precompile results into the block header using the AvalancheGo codec, which adds a constraint that the total size of those results must be small enough to be marshaled by the codec. Additionally, precompiles should ensure that they do not increase the worst case block size by returning large values that will be encoded in the block header.

The Warp precompile charges a base cost per message and each message is encoded into a bitset, which makes it extremely unlikely that it will come anywhere close to exceeding the max size of the codec or increasing the worst case block size.

Nevertheless, as per https://github.com/ava-labs/subnet-evm/pull/823#discussion_r1309226730 we should document this tradeoff and set reasonable expectations about the size of predicate results to be encoded in the block.

aaronbuchwald:
This PR

- moves the `precompile/results/README.md` file -> `predicate/Results.md` since the relevant code was moved into this package.
- renames `predicate/README.md` to `predicate/Predicate.md`.
- adds documentation for the maximum size of `Results` to `predicate/Predicate.md`

Resolves #824


predicate/Results.md
```diff
@@ -112,3 +112,15 @@ TxPredicateResults
 // bytes
 0x01, 0x02, 0x03
 ```
+
+### Maximum Size
+
+Results has a maximum size of 1MB enforced by the codec. The maximum actual size depends on how much data the available Precompile predicates may put into the results / the gas cost they charge and the block gas limit.
```
darioush, line 118:
```suggestion
Results has a maximum size of 1MB enforced by the codec. The actual size depends on how much data the Precompile predicates may put into the results, the gas cost they charge, and the block gas limit.
```

predicate/Results.md
```diff
@@ -112,3 +112,15 @@ TxPredicateResults
...
+
+The Results maximum size should comfortably exceed the maximum value that could happen in practice, so that a correct block builder will not attempt to build a block and fail to marshal the predicate results using the codec.
+
+We make this easy to reason about by assigning a minimum gas cost to the `PredicateGas` function of precompiles. In the case of Warp, the minimum gas cost is set to 200k gas, which can lead to at most 32 additional bytes being included in Results.
+
+The additional bytes come from the transaction hash (32 bytes), length of tx predicate results (4 bytes), the precompile address (20 bytes), length of the bytes result (4 bytes), and the additional byte in the results bitset (1 byte). This results in 200k gas contributing a maximum of 61 additional bytes to Result.
```
darioush, line 124: I thought we are changing to indices instead of hashes?

aaronbuchwald: It is currently using hashes, so this documentation is correct as is. I don't have a strong opinion on switching to index offsets.

ceyonur (approved):

anusha-ctrl (approved):

PR #945, opened 2023-10-06
x/warp: update PredicateGas to require warp message contains a valid payload

aaronbuchwald:
This PR updates `PredicateGas` of the warp precompile to parse the payload included in a warp message and return an error if it does not match a valid payload format defined by AvalancheGo.

### DIFF
--- a/x/warp/config.go
+++ b/x/warp/config.go
@@ -30,7 +30,7 @@ var (
 	errOverflowSignersGasCost  = errors.New("overflow calculating warp signers gas cost")
 	errInvalidPredicateBytes   = errors.New("cannot unpack predicate bytes")
 	errInvalidWarpMsg          = errors.New("cannot unpack warp message")
-	errInvalidWarpMsgPayoad    = errors.New("cannot unpack warp message payload")
+	errInvalidWarpMsgPayload   = errors.New("cannot unpack warp message payload")
 	errInvalidAddressedPayload = errors.New("cannot unpack addressed payload")
 	errInvalidBlockHashPayload = errors.New("cannot unpack block hash payload")
 	errCannotGetNumSigners     = errors.New("cannot fetch num signers from warp message")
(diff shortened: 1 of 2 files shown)


x/warp/config.go
```diff
@@ -29,6 +30,7 @@ var (
 	errOverflowSignersGasCost  = errors.New("overflow calculating warp signers gas cost")
 	errInvalidPredicateBytes   = errors.New("cannot unpack predicate bytes")
 	errInvalidWarpMsg          = errors.New("cannot unpack warp message")
+	errInvalidWarpMsgPayoad    = errors.New("cannot unpack warp message payload")
```
ceyonur, line 33:
```suggestion
	errInvalidWarpMsgPayload    = errors.New("cannot unpack warp message payload")
```

x/warp/config.go
```diff
@@ -166,6 +170,10 @@ func (c *Config) PredicateGas(predicateBytes []byte) (uint64, error) {
 	if err != nil {
 		return 0, fmt.Errorf("%w: %s", errInvalidWarpMsg, err)
 	}
+	_, err = payload.Parse(warpMessage.Payload)
+	if err != nil {
+		return 0, fmt.Errorf("%w: %s", errInvalidWarpMsgPayoad, err)
+	}
```
ceyonur, line 176: The only returned from `IntrinsicGas` is `ErrGasUintOverflow`. I wonder if we should do this verification (and the `err` returned from `ParseMessage`) in elsewhere. I don't think payload has anyhow related with gas calculation. 

I'm a bit in favor of using gas "pessimistically" in almost every case to prevent any possible DoS attacks. Alternative to moving parse verification to elsewhere, we can define a base gas cost for predicates and return that in case of errors other than overflows.

aaronbuchwald: This is executed in the tx pool, so it's already run "pessimistically".

This seemed like the most convenient place to put it since it's already used where needed and invalidates the transaction. Where else would you suggest putting it?

ceyonur (approved):

Issue #948, opened 2023-10-10
Add Event detection & generation to precompile generator


ceyonur:
Events can be defined In ABI and requires it's own special Pack/Unpack methods. In ABI it's not possible which function emits which event, so it's not possible to actually scaffold the code that will emit the event (`AddLog()`) but we can at least generate Pack/Unpack methods along with event structs in Golang and a proper comment for how to use those.

PR #949, opened 2023-10-10
add event packing/unpacking to contract template

ceyonur:
## Why this should be merged

Adds event pack/unpack generation to the contract template. 

Closes: #948

## How this works

It detects if there is a defined event type in ABI and generates proper pack/unpack methods with comments how to emit an event.

* Packing returns topics + packed unindexed data bytes in two separate outputs.
* Unpacking only works for unindexed data and is only generated when there is unindexed data defined in the event. It's not possible to unpack topics as they are 32-bytes hashes and thus not lossless. 
* Data is always represented as a struct, with correct golang type fields. 

## How this was tested

- [x] Local testing
- [x] ABI testing

## How is this documented

- [x] Comments in the code 
- [ ] Document in docs repo


accounts/abi/bind/precompilebind/precompile_contract_template.go
```diff
@@ -107,6 +107,24 @@ var (
 	}
 {{- end}}
 
+{{range .Contract.Events}}
+	{{$event := .}}
+	{{$createdStruct := false}}
+	{{- range .Normalized.Inputs}}
+	{{- if not .Indexed}}
+	{{- if not $createdStruct}}
```
abi87, line 115: nit: should this be further indented?

accounts/abi/bind/precompilebind/precompile_contract_template.go
ceyonur, line 1: Alternatively we can move events to a separate `events.go` file, as we don't know where events will be used in contracts when we generate them. So the contract code be cleaner.

accounts/abi/bind/precompilebind/precompile_contract_template.go
```diff
@@ -107,6 +107,24 @@ var (
 	}
 {{- end}}
 
+{{range .Contract.Events}}
+	{{$event := .}}
+	{{$createdStruct := false}}
+	{{- range .Normalized.Inputs}}
+	{{- if not .Indexed}}
```
abi87, line 114: nit: I feel using an early return would be more readable. Something like
```
{{- if .Indexed}}
	{{ continue }}
{{- end}}
```
would remove one of the `{- end}` within the loop.

accounts/abi/bind/precompilebind/precompile_contract_template.go
```diff
@@ -137,6 +155,60 @@ func Set{{.Contract.Type}}AllowListStatus(stateDB contract.StateDB, address comm
...
+topics, data, err := Pack{{.Normalized.Name}}Event(
+	{{- range .Normalized.Inputs}}
+	{{- if .Indexed}}
+	{{ decapitalise .Name}},
+	{{- end}}
+	{{- end}}
+	{{- if $hasData}}data,{{- end}}
+)
+if err != nil {
+	return nil, remainingGas, err
```
abi87, line 188: nit: `remainingGas`?

accounts/abi/bind/precompilebind/precompile_contract_template.go
```diff
@@ -137,6 +155,60 @@ func Set{{.Contract.Type}}AllowListStatus(stateDB contract.StateDB, address comm
...
+}
+accessibleState.GetStateDB().AddLog(
+	ContractAddress,
+	topics,
+	{{if $hasData}}data{{else}}[]byte{} // no undindexed data{{end}},
```
abi87, line 193: nit, typo: `undindexed` --> `unindexed`?

abi87 (changes requested):
nits

accounts/abi/bind/precompilebind/precompile_contract_template.go
ceyonur: this is a comment how to use events in precompile functions. Remaining gas is defined in functions. This is to indicate they should be using a remaining gas (deducted at first line in precompile functions)  rather than returning full or 0 gas.

abi87 (dismissed):
lgtm

accounts/abi/bind/precompilebind/precompile_event_template.go
```diff
@@ -0,0 +1,91 @@
...
+	if err != nil {
+		return nil, remainingGas, err
+	}
+	accessibleState.GetStateDB().AddLog(
+		ContractAddress,
+		topics,
+		{{if $createdDataStruct}}data{{else}}[]byte{} // no unindexed data{{end}},
+		accessibleState.GetBlockContext().Number().Uint64(),
+	)
+	*/
```
aaronbuchwald, line 76: Since this will auto-generate for each event, can we change how we document this so that it doesn't contain a repetitive comment of how to use events?

aaronbuchwald (changes requested):

accounts/abi/bind/precompilebind/precompile_bind.go
```diff
@@ -167,19 +192,38 @@ func verifyABI(abiData string) error {
 	if len(evmABI.Methods) == 0 {
 		return errors.New("no ABI methods found")
 	}
+
+	for _, event := range evmABI.Events {
+		if event.Anonymous {
+			return fmt.Errorf("%w: %s", errNoAnonymousEvent, event.Name)
+		}
+		eventNames := set.NewSet[string](len(event.Inputs))
```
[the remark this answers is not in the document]
ceyonur: reverted

accounts/abi/bind/precompilebind/precompile_event_template.go
ceyonur: yup changed it

aaronbuchwald (dismissed):

accounts/abi/bind/bind.go
```diff
@@ -571,3 +572,7 @@ func hasStruct(t abi.Type) bool {
 func mkList(args ...interface{}) []interface{} {
 	return args
 }
+
+func add(a, b int) int {
+	return a + b
+}
```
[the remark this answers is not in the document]
ceyonur: yes https://github.com/ava-labs/subnet-evm/blob/precompile-events-template/accounts/abi/bind/precompilebind/precompile_event_template.go#L64

core/vm/interface.go
```diff
@@ -86,7 +86,7 @@ type StateDB interface {
 	Snapshot() int
 
 	AddLog(addr common.Address, topics []common.Hash, data []byte, blockNumber uint64)
-	GetLogData() [][]byte
+	GetLogData() ([][]common.Hash, [][]byte)
```
abi87, line 89: nit: should we name parameters like other functions here?
`GetLogData() (topics [][]common.Hash, logData [][]byte)`

precompile/allowlist/allowlist.go
```diff
@@ -170,14 +170,10 @@ func CreateAllowListPrecompile(precompileAddr common.Address) contract.StatefulP
 
 func CreateAllowListFunctions(precompileAddr common.Address) []*contract.StatefulPrecompileFunction {
 	setAdmin := contract.NewStatefulPrecompileFunction(setAdminSignature, createAllowListRoleSetter(precompileAddr, AdminRole))
-	setManager := contract.NewStatefulPrecompileFunctionWithActivator(setManagerSignature, createAllowListRoleSetter(precompileAddr, ManagerRole), isManagerRoleActivated)
+	setManager := contract.NewStatefulPrecompileFunctionWithActivator(setManagerSignature, createAllowListRoleSetter(precompileAddr, ManagerRole), contract.IsDUpgradeActivated)
```
abi87, line 173: nit: not sure about subnet-evm, but in avalanchego we renamed DUpgrade to DurangoUpgrade (with extended name). Wonder if we should do the same here?

precompile/contract/interfaces.go
```diff
@@ -33,7 +33,7 @@ type StateDB interface {
 	Exist(common.Address) bool
 
 	AddLog(addr common.Address, topics []common.Hash, data []byte, blockNumber uint64)
-	GetLogData() [][]byte
+	GetLogData() ([][]common.Hash, [][]byte)
```
abi87, line 36: nit again consider naming parameters (topics and logData maybe)

precompile/testutils/test_precompile.go
```diff
@@ -49,7 +49,7 @@ type PrecompileTest struct {
 	ExpectedErr string
 	// ChainConfig is the chain config to use for the precompile's block context
 	// If nil, the default chain config will be used.
-	ChainConfig precompileconfig.ChainConfig
+	ChainConfigFn func(*gomock.Controller) precompileconfig.ChainConfig
```
abi87, line 52: <3

abi87 (approved):
just nits

precompile/allowlist/allowlist.go
ceyonur: I will do it before the release

core/vm/interface.go
ceyonur: added

precompile/contract/interfaces.go
ceyonur: added

precompile/testutils/test_precompile.go
```diff
@@ -49,7 +49,7 @@ type PrecompileTest struct {
 	ExpectedErr string
 	// ChainConfig is the chain config to use for the precompile's block context
```
darioush, line 50:
```suggestion
	// ChainConfigFn returns the chain config to use for the precompile's block context
```

precompile/contract/utils.go
```diff
@@ -18,6 +18,13 @@ import (
 const (
 	WriteGasCostPerSlot = 20_000
 	ReadGasCostPerSlot  = 5_000
+
+	// Per LOG operation.
+	LogGas uint64 = 375 // from params/protocol_params.go
+	// Gas cost of single topic of the LOG. Should be multiplied by the number of topics.
+	LogTopicGas uint64 = 375 // from params/protocol_params.go
+	// Per byte cost in a LOG operation's data. Should be multiplied by the byte size of the data.
+	LogDataGas uint64 = 8 // from params/protocol_params.go
```
darioush, line 27: I assume this is because of the import loop issue.
I'd like to spend some time later to figure out if there's a better way to handle the config unmarshal, which I believe to be the source of the import loop issue.

accounts/abi/bind/precompilebind/precompile_event_template.go
```diff
@@ -0,0 +1,122 @@
...
+		{{- end}}
+		{{- if not $createdDataStruct}}
+			{{$createdDataStruct = true}}
+			// {{$contract.Type}}{{$event.Normalized.Name}} represents a {{$event.Normalized.Name}} non-indexed event data raised by the {{$contract.Type}} contract.
+			type {{$event.Normalized.Name}}EventData struct {
+		{{- end}}
+		{{capitalise .Name}} {{bindtype .Type $structs}}
+	{{- end}}
+	{{- if $createdDataStruct}}
+		}
```
darioush, line 75:
```suggestion
			}
```

darioush (approved):

precompile/contract/utils.go
ceyonur: yup. params package is a really thin ice and can get cyclic with almost everything.

accounts/abi/bind/bind.go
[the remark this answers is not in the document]
aaronbuchwald: Ah I searched for it incorrectly

aaronbuchwald (approved):

darioush (approved):

Issue #729, opened 2023-07-12
Off-Chain Avalanche Warp Messages


aaronbuchwald:
# Off-Chain Avalanche Warp Messages: Signing out of Band

Subnet-EVM should support signing Avalanche Warp Messages out of band. 

Signing messages out of band will enable the validator set of a subnet collaborating to sign a message that does not correspond to an on-chain event. This can be used in order to implement upgradeable smart contracts where the Subnet's validator set replaces the typical multi-sig in order to authorize an update.

## Problem

If you are deploying a critical smart contract on-chain, then you have the choice between leaving that contract as immutable and upgradeable.

If you choose to make it immutable, then handling any discovered vulnerabilities in the contract may be extremely difficult.

If you choose to make it upgradeable, then you need to decide who/what has the ability to perform that upgrade, which creates a hazardous point of failure.

Therefore, authorizing a dynamic validator set to sign an out of band Avalanche Warp Message offers an alternative. There are two key advantages to this approach:

1. base upgradeability off of a dynamic subnet validator set instead of a multisig
2. support deploying upgradeable contracts at the same address across multiple chains with a unified upgrade mechanism

## Solution

Subnet-EVM should add an option to its chain config https://github.com/ava-labs/subnet-evm/blob/master/plugin/evm/config.go#L83 to create an allow list of messages that the validator should be willing to sign from the warp backend: https://github.com/ava-labs/subnet-evm/blob/master/warp/backend.go.

On startup the VM should:

1. Read the list of messages included in the chain config
2. Parse each message as an Avalanche Warp Message to ensure that it's valid (fatal if an invalid message is specified)
3. Load the list of messages in-memory into the Warp Backend (should we write them to the database? leaning towards no)
4. Modify `GetSignature(...)` to check the in-memory map for an eligible warp message, so that it will sign these as well

### Testing

We should add a unit test to ensure that adding a valid/invalid message to the VM's chain config works as expected

aaronbuchwald:
Follow on to this should be to create a demo upgrading OpenZeppelin upgradeable contracts via an Off-Chain Avalanche Warp Message.

ceyonur:
We have already merged this [PR](https://github.com/ava-labs/subnet-evm/pull/1028)  Are there still remaining items for this issue? If not can we close this?

aaronbuchwald:
Warp Upgradeable contracts have now been implemented via https://github.com/ava-labs/teleporter/pull/404

PR #959, opened 2023-10-17
Off Chain Avalanche Warp Messages

anusha-ctrl:
## Why this should be merged
Issue: https://github.com/ava-labs/subnet-evm/issues/729
On warp backend initialization, off-chain avalanche warp messages are parsed and added directly to the off-chain message map
If the off-chain warp message is invalid, the VM should not initialize.

`GetMessageSignature` should check the off-chain message map and return the signed message given the messageID.

`GetBlockSignature`  checks the off-chain block map and returns the signed message given the blockID

## How this was tested
UT


warp/backend.go
```diff
@@ -81,6 +84,11 @@ func (b *backend) AddMessage(unsignedMessage *avalancheWarp.UnsignedMessage) err
 	return nil
 }
 
+func (b *backend) CacheMessage(unsignedMessage *avalancheWarp.UnsignedMessage) {
```
ceyonur, line 87: can we rename this to `PutMessage`?

plugin/evm/config.go
```diff
@@ -191,6 +191,11 @@ type Config struct {
 	// Database Settings
 	InspectDatabase bool `json:"inspect-database"` // Inspects the database on startup if enabled.
 
+	// OutOfBandWarpMessagesBytes represent off-chain warp messages (in bytes) a validator is willing to sign
+	// These messages are not tied to on-chain events and are particularly useful for implementing a unified
+	// upgrade mechanism for critical smart contracts on the chain.
+	OutOfBandWarpMessagesBytes [][]byte `json:"out-of-band-warp-messages"`
```
ceyonur, line 197: nit: 
hmm this feels like:
image
Can we name it `OffChainWarpMessagesBytes`?

plugin/evm/vm.go
```diff
@@ -146,14 +147,15 @@ var (
...
-	errInvalidHeaderPredicateResults = errors.New("invalid header predicate results")
+	errEmptyBlock                       = errors.New("empty block")
+	errUnsupportedFXs                   = errors.New("unsupported feature extensions")
+	errInvalidBlock                     = errors.New("invalid block")
+	errInvalidNonce                     = errors.New("invalid nonce")
+	errUnclesUnsupported                = errors.New("uncles unsupported")
+	errNilBaseFeeSubnetEVM              = errors.New("nil base fee is invalid after subnetEVM")
+	errNilBlockGasCostSubnetEVM         = errors.New("nil blockGasCost is invalid after subnetEVM")
+	errInvalidHeaderPredicateResults    = errors.New("invalid header predicate results")
+	errInvalidOutOfBandWarpMessageBytes = errors.New("invalid warp message bytes")
```
ceyonur, line 158:
```suggestion
	errInvalidOutOfBandWarpMessageBytes = errors.New("invalid off-chain warp message bytes")
```

warp/backend.go
```diff
@@ -81,6 +84,11 @@ func (b *backend) AddMessage(unsignedMessage *avalancheWarp.UnsignedMessage) err
...
+	b.messageCache.Put(unsignedMessage.ID(), unsignedMessage)
+	log.Debug("Adding warp message to cache", "messageID", unsignedMessage.ID())
+}
+
 func (b *backend) GetSignature(messageID ids.ID) ([bls.SignatureLen]byte, error) {
```
ceyonur, line 92: I guess we don't need to modify this, despite the issue description? https://github.com/ava-labs/subnet-evm/issues/729:
`Modify GetSignature(...) to check the in-memory map for an eligible warp message, so that it will sign these as well`

warp/backend.go
anusha-ctrl: sure

plugin/evm/config.go
anusha-ctrl: sure

warp/backend.go
anusha-ctrl: Yes, GetSignature first checks the signature cache, then calls `GetMessage` which checks the message cache and returns immediately. There was no need to modify it.

plugin/evm/config.go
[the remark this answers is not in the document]
anusha-ctrl: I made the corresponding change. The JSON file would look something like this.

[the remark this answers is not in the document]
anusha-ctrl: https://go.dev/play/p/HLjPIz1vNum
I create an easy example to play with here.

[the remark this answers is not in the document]
ceyonur: hmm not sure if this is the best way. I guess we need a json unmarshaller if we want to present more readable option for the users. Or maybe hexadecimals?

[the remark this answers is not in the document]
anusha-ctrl: changed to hex 
``` json
"off-chain-warp-messages": ["00000000000a63636861696e7465737400000000000000000000000000000000000000000000000000077061796c6f6164"]
```

plugin/evm/vm.go
```diff
@@ -481,6 +484,19 @@ func (vm *VM) Initialize(
...
+	for _, message := range vm.config.OffChainWarpMessages {
+		messageByte, err := hex.DecodeString(message)
+		if err != nil {
+			return errInvalidOffChainWarpMessage
+		}
+		unsignedMessage, err := avalancheWarp.ParseUnsignedMessage(messageByte)
+		if err != nil {
+			return errInvalidOffChainWarpMessage
+		}
+		vm.warpBackend.PutMessage(unsignedMessage)
```
darioush, line 497: Should we add an info log here for the node validator to know the message has been added as expected?

plugin/evm/vm_test.go
```diff
@@ -3332,3 +3333,101 @@ func TestSignatureRequestsToVM(t *testing.T) {
...
+			signature, err := vm.warpBackend.GetSignature(unsignedWarpMessage.ID())
+			require.NoError(err)
+			require.Equal(([bls.SignatureLen]byte)(expectedSignature), signature)
+		}
+	}
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			// Each iteration runs the TestOffChainWarpMessages function so vm can shut down properly
+			TestOffChainWarpMessages(test)
```
darioush, line 3430: could we rename to  `testOffChainWarpMessages(test)`?

darioush (changes requested):
I sort of dislike that we are putting the off-chain messages into an LRU.
It seems the behavior of whether the node would be willing to sign could depend on how many other warp messages are sent, and possibly on the usage of the signatureCache as well. Could we have a separate map for the off-chain messages if they are only stored in-memory?

plugin/evm/vm.go
```diff
@@ -493,7 +493,8 @@ func (vm *VM) Initialize(
 		if err != nil {
 			return errInvalidOffChainWarpMessage
 		}
-		vm.warpBackend.PutMessage(unsignedMessage)
+		vm.warpBackend.AddOffChainMessage(unsignedMessage)
+		log.Info("off-chain warp message added successfully")
```
darioush, line 497: could we log the message as hex or a hash of it? (alternatively we could remove this log and make the log in `AddOffChainMessage` info)

darioush (dismissed):
lgtm, just nit about the log.

plugin/evm/vm.go
```diff
@@ -480,6 +483,20 @@ func (vm *VM) Initialize(
...
+	// parse and add valid off-chain warp messages to warp backend
+	for _, message := range vm.config.OffChainWarpMessages {
+		messageByte, err := hex.DecodeString(message)
+		if err != nil {
+			return errInvalidOffChainWarpMessage
+		}
+		unsignedMessage, err := avalancheWarp.ParseUnsignedMessage(messageByte)
+		if err != nil {
+			return errInvalidOffChainWarpMessage
+		}
```
aaronbuchwald, line 495: Can we change from doing this here to adding an argument to the constructor of the warp backend that takes in a slice of byte arrays to add as the set of hardcoded off-chain warp messages?

Inside the warp backend, we should also parse the unsigned message and parse the underlying payload as well to ensure that it's valid.

plugin/evm/config.go
```diff
@@ -191,6 +191,11 @@ type Config struct {
 	// Database Settings
 	InspectDatabase bool `json:"inspect-database"` // Inspects the database on startup if enabled.
 
+	// OffChainWarpMessagesBytes represent off-chain warp messages (represented in hex) a validator is willing to sign.
+	// These messages are not tied to on-chain events and are particularly useful for implementing a unified
+	// upgrade mechanism for critical smart contracts on the chain.
```
aaronbuchwald, line 196:
```suggestion
	// OffChainWarpMessagesBytes represent off-chain warp messages (represented in hex) a validator is willing to sign.
	// These messages are not tied to on-chain events.
```

I don't think we should specify the part about upgrading smart contracts on chain in the comment here. It is not enough detail to fully explain how they would be used, so I think it's better to omit the comment.

aaronbuchwald: Can we also switch to using `hexutil.Bytes` rather than a string so that we don't need to manually decode it?

plugin/evm/vm.go
aaronbuchwald: Let's move this log into the warp backend and specify the full bytes as hex in addition to the unsigned messageID so that the off chain messages are easily observable

warp/backend.go
```diff
@@ -161,6 +171,10 @@ func (b *backend) GetMessage(messageID ids.ID) (*avalancheWarp.UnsignedMessage,
 		return message, nil
 	}
 
+	if message, ok := b.offChainMessageMap[messageID]; ok {
+		return message, nil
+	}
+
```
aaronbuchwald, line 177: This doesn't work for messages containing a block hash payload. Could we add a regression test to make sure that both message types are handled correctly?

aaronbuchwald (changes requested):

plugin/evm/config.go
anusha-ctrl: done

warp/backend.go
anusha-ctrl: done

warp/backend.go
```diff
@@ -56,11 +59,13 @@ type backend struct {
...
 	blockSignatureCache   *cache.LRU[ids.ID, [bls.SignatureLen]byte]
 	messageCache          *cache.LRU[ids.ID, *avalancheWarp.UnsignedMessage]
+	offChainMessageMap    map[ids.ID]*avalancheWarp.UnsignedMessage
 }
 
 // NewBackend creates a new Backend, and initializes the signature cache and message tracking database.
-func NewBackend(networkID uint32, sourceChainID ids.ID, warpSigner avalancheWarp.Signer, blockClient BlockClient, db database.Database, cacheSize int) Backend {
-	return &backend{
+// It returns an error if [offChainMessages] is invalid.
+func NewBackend(networkID uint32, sourceChainID ids.ID, warpSigner avalancheWarp.Signer, blockClient BlockClient, db database.Database, cacheSize int, offChainMessages []hexutil.Bytes) (Backend, error) {
```
aaronbuchwald, line 67:
```suggestion
func NewBackend(networkID uint32, sourceChainID ids.ID, warpSigner avalancheWarp.Signer, blockClient BlockClient, db database.Database, cacheSize int, offChainMessages [][]byte) (Backend, error) {
```

Let's convert from `[]hexutil.Bytes` to `[][]byte` in `vm.Initialize(...)` instead of passing in the type alias `hexutil.Bytes` here

plugin/evm/vm_test.go
```diff
@@ -3333,6 +3335,134 @@ func TestMessageSignatureRequestsToVM(t *testing.T) {
...
+			require.NoError(err)
+			expectedSignature, err := vm.ctx.WarpSigner.Sign(unsignedWarpMessage)
+			require.NoError(err)
+
+			// On warp backend initialization off-chain warp messages were put into the off-chain message map
+			signature, err := vm.warpBackend.GetMessageSignature(unsignedWarpMessage.ID())
+			require.NoError(err)
+			require.Equal(([bls.SignatureLen]byte)(expectedSignature), signature)
+		}
+	}
```
aaronbuchwald, line 3456: Since block payloads are served from a different handler, we need to parse the payload and add a type switch to check which of the two functions it should call. Additionally, calling `GetMessageSignature` on a `blockID` of a block hash payload should not return a signature.

warp/backend_test.go
```diff
@@ -175,3 +182,101 @@ func TestZeroSizedCache(t *testing.T) {
...
+	blockHashPayload, err := payload.NewHash(blkID)
+	require.NoError(err)
+	unsignedMessage, err := avalancheWarp.NewUnsignedMessage(networkID, sourceChainID, blockHashPayload.Bytes())
+	require.NoError(err)
+	expectedSig, err := warpSigner.Sign(unsignedMessage)
+	require.NoError(err)
+
+	offChainMessages := []hexutil.Bytes{hexutil.Bytes(unsignedMessage.Bytes())}
+	backend, err := NewBackend(networkID, sourceChainID, warpSigner, testVM, db, 500, offChainMessages)
+	require.NoError(err)
```
aaronbuchwald, line 267: Can we remove from this test `GetBlockF` returning this `blkID` as accepted? Since that blockID is reported as accepted by the VM this test will pass even if the unsigned message is not passed in via `offChainMessages`.

warp/backend.go
```diff
@@ -174,3 +197,39 @@ func (b *backend) GetMessage(messageID ids.ID) (*avalancheWarp.UnsignedMessage,
...
+	}
+
+	id, err := parsePayload(unsignedMessage)
+	if err != nil {
+		return err
+	}
+
+	log.Info("Adding warp message to off-chain message map", "messageID", unsignedMessage.ID(), "decrypted warp message", hex.EncodeToString(unsignedMessage.Bytes()))
+	b.offChainMessageMap[id] = unsignedMessage
+	return nil
```
aaronbuchwald, line 215: Since there are two separate handlers for hash payloads vs. messages, we should create a separate map for the two different types. If a client mistakenly sends a request to the block hash handler with a messageID specified, we should not return a signature over a warp message containing an addressed call payload, since that is an unexpected response for that handler.

warp/backend.go
```diff
@@ -174,3 +197,39 @@ func (b *backend) GetMessage(messageID ids.ID) (*avalancheWarp.UnsignedMessage,
...
+		id = blockHashPayload.Hash
+	} else {
+		_, err = payload.Parse(unsignedMessage.Payload)
+		if err != nil {
+			return ids.Empty, fmt.Errorf("%w: %s", warp.ErrInvalidWarpMsgPayload, err)
+		}
+	}
+
+	return id, nil
+}
```
aaronbuchwald, line 235: Can we use `payload.Parse` directly and then have a type switch on the returned value to check which off chain messages map it needs to be placed in?

This should be pretty short and is only used once, so I think it would make sense to inline as well.

aaronbuchwald (changes requested):

warp/backend.go
anusha-ctrl: done

warp/backend.go
anusha-ctrl: got it

warp/backend.go
anusha-ctrl: yup

warp/backend_test.go
anusha-ctrl: Hm, I see. Should we skip block checks (valid and accepted) if its an offchain message then?

anusha-ctrl: If I remove the above it will fail the block checks in `GetBlockSignature`

aaronbuchwald: Off chain messages is meant to allow the node to force sign an individual message, so if we perform the check that it corresponds to an actual accepted block, then it would have already served the same request regardless of if it was populated in the off chain messages config.

anusha-ctrl: I see. Got it thanks

PR #961, opened 2023-10-19
Precompile events - Reward Manager

abi87:
## Why this should be merged

Adds event to allow list functions. 

## How this works

Adds new Events to ABI and regenerated with the precompile generation tool.

`event RewardAddressChanged(address indexed sender, address indexed oldRewardAddress, address indexed newRewardAddress);`

`event FeeRecipientsAllowed(address indexed sender);`

`event RewardsDisabled(address indexed sender);`

Increases `AllowFeeRecipients` total gas cost by 1125 (base event gas, 2 topics gas)

Increases `SetRewardAddress` total gas cost by 6875 (base event gas, reading old reward address from storage, 4 topics gas)

Increases `DisableRewards` total gas cost by 1125 (base event gas, 2 topics gas)

## How this was tested

Added UT and hardhat tests.

## How is this documented


precompile/contracts/rewardmanager/contract.go
```diff
@@ -25,7 +25,11 @@ const (
 	AreFeeRecipientsAllowedGasCost uint64 = allowlist.ReadAllowListGasCost
 	CurrentRewardAddressGasCost    uint64 = allowlist.ReadAllowListGasCost
 	DisableRewardsGasCost          uint64 = (contract.WriteGasCostPerSlot) + allowlist.ReadAllowListGasCost // write 1 slot + read allow list
-	SetRewardAddressGasCost        uint64 = (contract.WriteGasCostPerSlot) + allowlist.ReadAllowListGasCost // write 1 slot + read allow list
+
+	DuplicatedLogGas      uint64 = 375 // TODO: duplicated from params/protocol_params.go to avoid import cycle in tests.
+	DuplicatedLogTopicGas uint64 = 375 // TODO: duplicated from params/protocol_params.go to avoid import cycle in tests.
```
abi87, line 30: @ceyonur temporary code duplication to avoid import cycle in tests. Not sure how to solve this.

anusha-ctrl: We can move `LogGas` and `LogTopicGas` to a new package. Both packages that need the constants can import?

anusha-ctrl: I think its worth pulling out `precompile_config_test.go` out of params and into a new package. It also might be worth moving the specific precompile tests out of `config.test.go`. Best to keep params/protocol_params.go as synced to geth. @ceyonur knows best though

precompile/contracts/rewardmanager/contract_test.go
```diff
@@ -341,3 +345,116 @@ func TestRewardManagerRun(t *testing.T) {
...
+	expectedTopic := []common.Hash{
+		RewardManagerABI.Events["RewardAddress"].ID,
+		allowlist.TestAdminAddr.Hash(),
+		rewardAddress.Hash(),
+	}
+
+	allLogs := baseState.(*state.StateDB).Logs()
+	require.Len(allLogs, 1)
+	require.Equal(expectedTopic, allLogs[0].Topics)
+}
```
anusha-ctrl, line 388: Check sad path too? if setRewardAddress fails, we should not expect event
?

contracts/contracts/interfaces/IRewardManager.sol
```diff
@@ -3,6 +3,12 @@ pragma solidity ^0.8.0;
 import "./IAllowList.sol";
 
 interface IRewardManager is IAllowList {
+  event RewardAddress(address indexed sender, address indexed reward);
+
+  event AllowFeeRecipients(address indexed sender);
+
+  event DisableRewards(address indexed sender);
```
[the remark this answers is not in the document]
abi87: Done

contracts/contracts/interfaces/IRewardManager.sol
```diff
@@ -3,6 +3,12 @@ pragma solidity ^0.8.0;
 import "./IAllowList.sol";
 
 interface IRewardManager is IAllowList {
+  event RewardAddress(address indexed sender, address indexed reward);
+
+  event AllowFeeRecipients(address indexed sender);
```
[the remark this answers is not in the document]
abi87: Done

contracts/contracts/interfaces/IRewardManager.sol
```diff
@@ -3,6 +3,12 @@ pragma solidity ^0.8.0;
 import "./IAllowList.sol";
 
 interface IRewardManager is IAllowList {
+  event RewardAddress(address indexed sender, address indexed reward);
```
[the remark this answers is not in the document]
abi87: Done

precompile/contracts/rewardmanager/event.go
```diff
@@ -0,0 +1,59 @@
...
+if err != nil {
+	return nil, remainingGas, err
+}
+accessibleState.GetStateDB().AddLog(
+	ContractAddress,
+	topics,
+	data,
+	accessibleState.GetBlockContext().Number().Uint64(),
+)
+*/
```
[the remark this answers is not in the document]
abi87: Done

precompile/contracts/rewardmanager/contract.go
```diff
@@ -101,6 +104,18 @@ func allowFeeRecipients(accessibleState contract.AccessibleState, caller common.
...
+	topics, data, err := PackAllowFeeRecipientsEvent(caller)
+	if err != nil {
+		return nil, remainingGas, err
+	}
+	accessibleState.GetStateDB().AddLog(
+		ContractAddress,
+		topics,
+		data,
+		accessibleState.GetBlockContext().Number().Uint64(),
+	)
```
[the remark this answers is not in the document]
abi87: Indeed, fixed over all log invocations

precompile/contracts/rewardmanager/contract_test.go
```diff
@@ -341,3 +345,116 @@ func TestRewardManagerRun(t *testing.T) {
...
+
+func TestAllowFeeRecipientsLogging(t *testing.T) {
```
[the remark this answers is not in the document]
abi87: Done

precompile/contracts/rewardmanager/contract_test.go
```diff
@@ -341,3 +345,116 @@ func TestRewardManagerRun(t *testing.T) {
...
+		RewardManagerABI.Events["AllowFeeRecipients"].ID,
+		allowlist.TestAdminAddr.Hash(),
+	}
+
+	allLogs := baseState.(*state.StateDB).Logs()
+	require.Len(allLogs, 1)
+	require.Equal(expectedTopic, allLogs[0].Topics)
+}
+
+func TestDisableRewardsLogging(t *testing.T) {
```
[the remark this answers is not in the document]
abi87: Done

precompile/contracts/rewardmanager/contract_test.go
```diff
@@ -341,3 +345,116 @@ func TestRewardManagerRun(t *testing.T) {
 func BenchmarkRewardManager(b *testing.B) {
 	allowlist.BenchPrecompileWithAllowList(b, Module, state.NewTestStateDB, tests)
 }
+
+func TestSetRewardAddressLogging(t *testing.T) {
```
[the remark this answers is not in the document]
abi87: Done

precompile/contracts/rewardmanager/contract.go
```diff
@@ -21,11 +21,14 @@ import (
...
+	DuplicatedLogTopicGas uint64 = 375 // TODO: duplicated from params/protocol_params.go to avoid import cycle in tests.
+
+	AllowFeeRecipientsGasCost      uint64 = DuplicatedLogGas + DuplicatedLogTopicGas + contract.WriteGasCostPerSlot + allowlist.ReadAllowListGasCost // logging cost + write 1 slot + read allow list
 	AreFeeRecipientsAllowedGasCost uint64 = allowlist.ReadAllowListGasCost
 	CurrentRewardAddressGasCost    uint64 = allowlist.ReadAllowListGasCost
-	DisableRewardsGasCost          uint64 = (contract.WriteGasCostPerSlot) + allowlist.ReadAllowListGasCost // write 1 slot + read allow list
-	SetRewardAddressGasCost        uint64 = (contract.WriteGasCostPerSlot) + allowlist.ReadAllowListGasCost // write 1 slot + read allow list
+	DisableRewardsGasCost          uint64 = (contract.WriteGasCostPerSlot) + allowlist.ReadAllowListGasCost                                            // write 1 slot + read allow list
+	SetRewardAddressGasCost        uint64 = DuplicatedLogGas + 2*DuplicatedLogTopicGas + contract.WriteGasCostPerSlot + allowlist.ReadAllowListGasCost // logging cost + write 1 slot + read allow list
```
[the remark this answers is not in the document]
abi87: Not sure I fully follow the topic/data distinction but I have introduce the fork to duly handle the change in gas cost

precompile/contracts/rewardmanager/contract_test.go
abi87: @anusha-ctrl not sure what the sad path is here? Log serialization can really fail only if there is bug. I introduced a test to show no logs are stored pre-fork. Not sure what else should I cover, except for paying pre-fork gas cost for ops involving logs?

precompile/contract/utils.go
```diff
@@ -18,6 +18,9 @@ import (
 const (
 	WriteGasCostPerSlot = 20_000
 	ReadGasCostPerSlot  = 5_000
+
+	DuplicatedLogGas      uint64 = 375 // TODO: duplicated from params/protocol_params.go to avoid import cycle in tests.
+	DuplicatedLogTopicGas uint64 = 375 // TODO: duplicated from params/protocol_params.go to avoid import cycle in tests.
```
abi87, line 23: @ceyonur still I don't get how to avoid code duplication here.

precompile/contracts/rewardmanager/contract.go
```diff
@@ -21,14 +21,17 @@ import (
 )
 
 const (
-	DuplicatedLogGas      uint64 = 375 // TODO: duplicated from params/protocol_params.go to avoid import cycle in tests.
-	DuplicatedLogTopicGas uint64 = 375 // TODO: duplicated from params/protocol_params.go to avoid import cycle in tests.
+	AllowFeeRecipientsGasCost         uint64 = contract.WriteGasCostPerSlot + allowlist.ReadAllowListGasCost // write 1 slot + read allow list
+	UpgradedAllowFeeRecipientsGasCost uint64 = contract.LogTopicGas + contract.WriteGasCostPerSlot + allowlist.ReadAllowListGasCost
```
ceyonur, line 25: I think `PreDUpgrade...` and `PostDUpgrade...` prefixes were good (unless we name it `PreDFork` etc)

precompile/contracts/rewardmanager/contract.go
```diff
@@ -101,6 +117,20 @@ func allowFeeRecipients(accessibleState contract.AccessibleState, caller common.
...
+		topics, data, err := PackFeeRecipientsAllowedEvent(caller)
+		if err != nil {
+			return nil, remainingGas, err
+		}
+		stateDB.AddLog(
+			ContractAddress,
+			topics,
+			data,
+			blkCtx.Number().Uint64(),
+		)
```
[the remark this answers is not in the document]
aaronbuchwald: Hmm since this event has one fixed size argument we should know the number of topics and total size in advance.

imo we should ideally charge before calling pack since packing itself is a variable amount of work based on the arguments that need to be packed into the event. That being said, I'd also agree that it would be good to defer charging that gas to this if branch to make that charge as close to the event packing work as possible.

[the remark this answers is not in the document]
aaronbuchwald: For this PR, since it is handling a fixed size event, I think it makes sense to charge that fixed amount here when it's time to pack the event

precompile/contracts/rewardmanager/contract_test.go
```diff
@@ -67,29 +71,79 @@ var (
...
-			SuppliedGas: AllowFeeRecipientsGasCost,
+			SuppliedGas: UpgradedAllowFeeRecipientsGasCost,
 			ReadOnly:    false,
 			ExpectedRes: []byte{},
 			AfterHook: func(t testing.TB, state contract.StateDB) {
 				_, isFeeRecipients := GetStoredRewardAddress(state)
 				require.True(t, isFeeRecipients)
 			},
 		},
+		"no log set fee recipients if D fork is not active": {
```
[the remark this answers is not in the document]
abi87: Done

precompile/contracts/rewardmanager/contract.go
abi87: Done

contracts/contracts/interfaces/IRewardManager.sol
```diff
@@ -3,6 +3,12 @@ pragma solidity ^0.8.0;
 import "./IAllowList.sol";
 
 interface IRewardManager is IAllowList {
+  event RewardAddressChanged(address indexed sender, address indexed rewardAddress);
+
+  event FeeRecipientsAllowed(address indexed sender);
+
+  event RewardsDisabled(address indexed sender);
```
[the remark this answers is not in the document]
abi87: Done

precompile/contract/utils.go
```diff
@@ -18,6 +18,9 @@ import (
 const (
 	WriteGasCostPerSlot = 20_000
 	ReadGasCostPerSlot  = 5_000
+
+	LogGas      uint64 = 375 // from params/protocol_params.go
+	LogTopicGas uint64 = 375 // from params/protocol_params.go
```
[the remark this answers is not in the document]
abi87: Done

contracts/contracts/interfaces/IRewardManager.sol
```diff
@@ -3,6 +3,15 @@ pragma solidity ^0.8.0;
...
 
 interface IRewardManager is IAllowList {
+  // RewardAddressChanged is the event logged whenever reward address is modified
+  event RewardAddressChanged(address indexed sender, address indexed rewardAddress);
+
+  // FeeRecipientsAllowed is the event logged whenever fee recipient is modified
+  event FeeRecipientsAllowed(address indexed sender);
+
+  // RewardsDisabled is the event logged whenever rewards are disabled
+  event RewardsDisabled(address indexed sender);
```
ceyonur, line 13: I'm not sure if `sender` addresses would be very helpful to be an indexed topic in logs. Maybe we should omit emitting those. cc @ava-labs/platform-evm @abi87

ceyonur: We decided to include them

precompile/contracts/rewardmanager/contract.go
```diff
@@ -97,12 +97,24 @@ func allowFeeRecipients(accessibleState contract.AccessibleState, caller common.
 	}
 	// allow list code ends here.
 
-	// this function does not return an output, leave this one as is
+	if contract.IsDUpgradeActivated(accessibleState) {
```
abi87, line 100: so in the end we're forced to remove gas in two steps

precompile/contracts/rewardmanager/contract_test.go
```diff
@@ -67,29 +71,78 @@ var (
 
 				return input
 			},
-			SuppliedGas: AllowFeeRecipientsGasCost,
+			SuppliedGas: AllowFeeRecipientsGasCost + FeeRecipientsAllowedEventGasCost,
```
abi87, line 74: q: should we check that this gas is strictly needed? I.e. that if we provide some more gas, remaining gas would be equale to the extra specified?

abi87 (commented):
can't approve since I opened the PR. Still LGTM

precompile/contracts/rewardmanager/contract_test.go
ceyonur: We require exact gas amount supplied in tests. It requires remainingGas == 0  here https://github.com/ava-labs/subnet-evm/blob/fix-abi-get-inputs/precompile/testutils/test_precompile.go#L77

precompile/contracts/rewardmanager/contract_test.go
```diff
@@ -164,11 +269,60 @@ var (
...
+			ReadOnly:    false,
+			ExpectedRes: []byte{},
+			AfterHook: func(t testing.TB, stateDB contract.StateDB) {
+				// Check logs are not stored in state
+				topics, data := stateDB.GetLogData()
+				require.Len(t, topics, 0)
+				require.Len(t, data, 0)
+			},
+		},
+		"log disable rewards if DUpgrade is active": {
```
aaronbuchwald, line 298: For this pair of tests and the other new ones, aren't these creating duplicate tests since there were tests for the pre DUpgrade behavior before (which have been modified to include the event, but not including a check for the event) ?

Could we consolidate these tests and make sure that the names are clear? I think something like `disable rewards pre / post DUpgrade` would make it a bit more clear and also make it more obvious if we've accidentally created duplicates

aaronbuchwald (changes requested):

precompile/contracts/rewardmanager/config.go
```diff
@@ -47,7 +48,7 @@ func (i *InitialRewardConfig) Configure(state contract.StateDB) error {
 		DisableFeeRewards(state)
 	} else {
 		// set reward address
-		return StoreRewardAddress(state, i.RewardAddress)
+		StoreRewardAddress(state, i.RewardAddress)
```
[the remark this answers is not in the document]
ceyonur: We still preserve the check but not in `StoreRewardAddress` directly. I moved the check before emitting logs [here](https://github.com/ava-labs/subnet-evm/blob/precompile-events-set-reward-address/precompile/contracts/rewardmanager/contract.go#L220-L222). Also in the `Configure` we already check this case [here](https://github.com/ava-labs/subnet-evm/blob/precompile-events-set-reward-address/precompile/contracts/rewardmanager/config.go#L45).

precompile/contracts/rewardmanager/contract.go
```diff
@@ -209,14 +216,31 @@ func setRewardAddress(accessibleState contract.AccessibleState, caller common.Ad
 	}
 	// allow list code ends here.
 
-	if err := StoreRewardAddress(stateDB, inputStruct); err != nil {
-		return nil, remainingGas, err
```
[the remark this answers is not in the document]
ceyonur: https://github.com/ava-labs/subnet-evm/pull/961#discussion_r1428079063

precompile/contracts/rewardmanager/event.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"github.com/ava-labs/subnet-evm/precompile/contract"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+const (
+	FeeRecipientsAllowedEventGasCost = contract.LogGas + contract.LogTopicGas*1                               // 1 indexed topic
+	RewardAddressChangedEventGasCost = contract.LogGas + contract.LogTopicGas*2 + contract.ReadGasCostPerSlot // 1 indexed topic + reading oldRewardAddress from state
+	RewardsDisabledEventGasCost      = contract.LogGas + contract.LogTopicGas*1                               // 1 indexed topic
```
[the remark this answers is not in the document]
ceyonur: I thought that was what LOG0 for (to include only topic but charge no cost). but I totally forgot about anoynmous events.

precompile/contracts/rewardmanager/contract_test.go
ceyonur: By default all tests run with DUpgrade activated. I think this is the default behaviour in most of our tests (to run with latest chain config). If that's the case I don't think we can rename every one of them with post DUpgrade. 

The reason I duplicated the tests and named them with pre/post Dupgrade was to follow append-like fashion to add tests rather than change targets of existing tests. 

The question here should be: If there was something wrong with events, do we want to fail the existing tests as well or just the new ones? BTW I'm totally fine to add duplicated test setup (unless they also test the same outcome).

aaronbuchwald: I see, I hadn't thought of it that way.

Curious what @marun recommends for testing best practices as well.

I'll defer to you guys on the best way to handle these go-forward tests.

ceyonur: I think most of the tests are upgraded through `TestChainConfig` which activates the latest upgrade. But we don't remove/deprecate any of tests that target specific upgrades.

The case here is simply having two tests that have two exact same setups and steps, but testing different outcomes. One of them tests if the caller is allowed to call the function, the other tests if the outcome of the call has correct outputs.

contracts/test/reward_manager.ts
```diff
@@ -3,10 +3,16 @@
 
 import { ethers } from "hardhat"
 import { test } from "./utils"
+import { SignerWithAddress } from "@nomiclabs/hardhat-ethers/signers";
+import { expect } from "chai";
+import {
+  Contract,
+} from "ethers"
```
darioush, line 10:
```suggestion
import { Contract } from "ethers"
```

darioush (dismissed):

precompile/contracts/rewardmanager/contract_test.go
aaronbuchwald: Yes, the VM must be able to re-process any historical block that was processed with a previous rule set.

ceyonur: I refactored those by adding new assert helper functions for events and used them in existing tests.

aaronbuchwald (approved):

PR #967, opened 2023-10-23
 Precompile events - Fee Config

abi87:
## Why this should be merged

Adds event to set fee config function. 

## How this works

Adds new Event to ABI and regenerated with the precompile generation tool.

`event FeeConfigChanged(address indexed sender, FeeConfig oldFeeConfig, FeeConfig newFeeConfig);`

Increases total gas cost by 45221 (base event gas, reading old fee config from statedb, 2 topics gas, 512-byte unindexed data)

## How this was tested

Added UT and hardhat tests.

## How is this documented


precompile/contracts/feemanager/contract.go
```diff
@@ -242,6 +249,34 @@ func setFeeConfig(accessibleState contract.AccessibleState, caller common.Addres
...
+		topics, data, err := PackFeeConfigChangedEvent(
+			caller,
+			oldConfig,
+			feeConfig,
+		)
+		if err != nil {
+			return nil, remainingGas, err
+		}
+
+		logDataGasCost := uint64(len(data)) * contract.LogDataGas
```
aaronbuchwald, line 267: Could we subtract `GetFeeConfigGasCost` at the top here and move the entire calculation of the gas cost due to the produced log into one place, so that it's not split up?

precompile/contracts/feemanager/contract_test.go
```diff
@@ -258,3 +332,14 @@ func TestFeeManager(t *testing.T) {
...
 }
+
+func getGasCost(oldFeeConfig commontype.FeeConfig, feeConfig commontype.FeeConfig) uint64 {
+	// address doesn't matter for gas cost
+	_, data, err := PackFeeConfigChangedEvent(common.Address{}, oldFeeConfig, feeConfig)
+	if err != nil {
+		panic(err)
+	}
+
+	logGasCost := GetFeeConfigGasCost + FeeConfigChangedEventBaseGasCost + contract.LogDataGas*uint64(len(data))
```
aaronbuchwald, line 343: All of the fields of the event are a fixed size, so this should be a constant value regardless of the inputs

precompile/contracts/feemanager/contract_test.go
```diff
@@ -248,6 +250,78 @@ var (
 			ReadOnly:    false,
 			ExpectedErr: vmerrs.ErrOutOfGas.Error(),
 		},
+		"set config doesn't log before DUpgrade": {
```
aaronbuchwald, line 253: same comment from other event PRs, doesn't creating a before and after + modifying the previously existing test create a duplicate of the same case?

precompile/contracts/feemanager/event.go
```diff
@@ -0,0 +1,68 @@
...
+type ChangeFeeConfigEventData struct {
+	GasLimit                 *big.Int
+	TargetBlockRate          *big.Int
+	MinBaseFee               *big.Int
+	TargetGas                *big.Int
+	BaseFeeChangeDenominator *big.Int
+	MinBlockGasCost          *big.Int
+	MaxBlockGasCost          *big.Int
+	BlockGasCostStep         *big.Int
+}
```
aaronbuchwald, line 27: Could we add a comment to explain why we need the intermediate type between `commontype.FeeConfig` and `ChangeFeeConfigEventData` ?

aaronbuchwald (changes requested):

precompile/contracts/feemanager/event.go
ceyonur: I think we can just unexport it and return commontype.FeeConfig instead.

precompile/contracts/feemanager/contract_test.go
ceyonur: yep, changed them

precompile/contracts/feemanager/contract.go
ceyonur: moved

precompile/contracts/feemanager/contract_test.go
ceyonur: I moved after hook to existing test.

contracts/test/fee_manager.ts
```diff
@@ -4,6 +4,10 @@
 import { expect } from "chai"
 import { ethers } from "hardhat"
 import { test } from "./utils"
+import { SignerWithAddress } from "@nomiclabs/hardhat-ethers/signers";
+import {
+  Contract,
+} from "ethers"
```
darioush, line 10:
```suggestion
import { Contract } from "ethers"
```

contracts/test/fee_manager.ts
```diff
@@ -53,3 +57,66 @@ describe("ExampleFeeManager", function () {
     "step_lowerMinFeeByOne",
   ])
 })
+const C_FEES = {
```
darioush, line 60:
```suggestion

const C_FEES = {
```

precompile/contracts/feemanager/contract_test.go
```diff
@@ -67,12 +77,27 @@ var (
...
+				topics := logsTopics[0]
+				require.Len(t, topics, 2)
+				require.Equal(t, FeeManagerABI.Events["FeeConfigChanged"].ID, topics[0])
+				require.Equal(t, allowlist.TestEnabledAddr.Hash(), topics[1])
+
+				logData := logsData[0]
+				oldFeeConfig, resFeeConfig, err := UnpackFeeConfigChangedEventData(logData)
+				require.NoError(t, err)
+				require.True(t, zeroFeeConfig.Equal(&oldFeeConfig), "expected %v, got %v", zeroFeeConfig, oldFeeConfig)
+				require.True(t, testFeeConfig.Equal(&resFeeConfig), "expected %v, got %v", testFeeConfig, resFeeConfig)
```
darioush, line 100: could we abstract this to a helper function instead of repeating it?

ceyonur: good idea 👍

darioush (approved):

precompile/allowlist/allowlist.go
```diff
@@ -165,11 +165,26 @@ func CreateAllowListPrecompile(precompileAddr common.Address) contract.StatefulP
...
+		case ManagerRole.GetSetterFunctionName():
+			fn = contract.NewStatefulPrecompileFunctionWithActivator(method.ID, createAllowListRoleSetter(precompileAddr, ManagerRole), contract.IsDUpgradeActivated)
+		default:
+			panic(fmt.Sprintf("unexpected method name: %s", name))
+		}
+		functions = append(functions, fn)
+	}
 
-	return []*contract.StatefulPrecompileFunction{setAdmin, setManager, setEnabled, setNone, read}
+	return functions
```
aaronbuchwald, line 189: What's the reason for this change?

ceyonur: you might be looking at a merge commit.

aaronbuchwald (approved):

PR #968, opened 2023-10-25
Precompile events allow list

abi87:
## Why this should be merged

Adds event to allow list functions. 

## How this works

Adds new Event to ABI and regenerated with the precompile generation tool.

`event RoleSet(uint256 indexed role, address indexed account, address indexed sender, uint256 oldRole);`

Increases total gas cost by 2131 (base event gas, 4 topics gas, 32-byte unindexed data)

## How this was tested

Added UT and hardhat tests.

## How is this documented


contracts/contracts/interfaces/IAllowList.sol
```diff
@@ -2,6 +2,11 @@
 pragma solidity ^0.8.0;
 
 interface IAllowList {
+  event AdminAdded(address indexed sender, address indexed account);
+  event EnabledAdded(address indexed sender, address indexed account);
+  event ManagerAdded(address indexed sender, address indexed account);
+  event RoleRemoved(address indexed sender, address indexed account);
```
aaronbuchwald, line 8: I think the event should include this information (and in this order of priority with respect to whether it's indexed).

1. modified account
2. new role
3. account that caused the modification
4. previous role

Since that's 4 fields, one of them would need to be captured in the function name or we'd need to make at least one non-indexed.

I'd lean towards making it a single event and leaving the previous role un-indexed since it's unlikely imo a user would need to search all of the allow list modifications that changed the status from manager -> X.

precompile/allowlist/event.go
```diff
@@ -0,0 +1,53 @@
...
+		return PackAdminAddedEvent(sender, account)
+	case ManagerRole:
+		return PackManagerAddedEvent(sender, account)
+	case EnabledRole:
+		return PackEnabledAddedEvent(sender, account)
+	case NoRole:
+		return PackRoleRemovedEvent(sender, account)
+	default:
+		panic("invalid role")
+	}
```
aaronbuchwald, line 52: can we return an error instead of panic-ing here? I'd prefer defense in depth from something that should never happen to a panic that will shut down the whole chain.

precompile/allowlist/test_allowlist.go
```diff
@@ -539,6 +539,191 @@ func AllowListTests(t testing.TB, module modules.Module) map[string]testutils.Pr
 				require.Equal(t, EnabledRole, GetAllowListStatus(state, contractAddress, TestNoRoleAddr))
 			},
 		},
+		"allowList admin event doesn't log if D fork is not active": {
```
aaronbuchwald, line 542: Can we adjust these names?

I think a better naming convention is set above with the address of the caller, the role being set, and then adding pre-DUpgrade

For example, this one could be "admin set admin pre-DUpgrade"

contracts/test/contract_deployer_allow_list.ts
```diff
@@ -48,3 +53,40 @@ describe("ExampleDeployerList", function () {
...
+    await expect(contract.setAdmin(testAddress))
+      .to.emit(contract, 'AdminAdded')
+      .withArgs(owner.address, testAddress)
+  })
+
+  it("should emit manager address added event", async function () {
+    let testAddress = "0x0222000000000000000000000000000000000002"
+    await expect(contract.setManager(testAddress))
+      .to.emit(contract, 'ManagerAdded')
+      .withArgs(owner.address, testAddress)
```
aaronbuchwald, line 76: These assertions are sick ❤️ , so easy

aaronbuchwald (changes requested):

precompile/allowlist/test_allowlist.go
ceyonur: done

contracts/contracts/interfaces/IAllowList.sol
ceyonur: done

contracts/test/contract_deployer_allow_list.ts
```diff
@@ -2,7 +2,12 @@
 // See the file LICENSE for licensing terms.
 
 import { ethers } from "hardhat"
-import { test } from "./utils"
+import { Roles, test } from "./utils"
+import { SignerWithAddress } from "@nomiclabs/hardhat-ethers/signers";
+import { expect } from "chai";
+import {
+  Contract,
+} from "ethers"
```
darioush, line 10:
```suggestion
import { Contract } from "ethers"
```

contracts/test/tx_allow_list.ts
```diff
@@ -2,7 +2,12 @@
 // See the file LICENSE for licensing terms.
 
 import { ethers } from "hardhat"
-import { test } from "./utils"
+import { Roles, test } from "./utils"
+import { SignerWithAddress } from "@nomiclabs/hardhat-ethers/signers";
+import { expect } from "chai";
+import {
+  Contract,
+} from "ethers"
```
darioush, line 10:
```suggestion
import { Contract } from "ethers"
```

precompile/allowlist/role.go
```diff
@@ -77,18 +80,18 @@ func (r Role) Hash() common.Hash {
...
+		return "setManager", nil
 	case EnabledRole:
-		return "setEnabled"
+		return "setEnabled", nil
 	case NoRole:
-		return "setNone"
+		return "setNone", nil
 	default:
-		panic("unknown role")
+		return "", ErrInvalidRole
```
darioush, line 94: seems this err is ignored if it is returned?

precompile/allowlist/test_allowlist.go
```diff
@@ -276,16 +324,28 @@ func AllowListTests(t testing.TB, module modules.Module) map[string]testutils.Pr
...
+				require.Len(t, logsTopics, 1)
+				require.Len(t, logsData, 1)
+				topics := logsTopics[0]
+				require.Len(t, topics, 4)
+				require.Equal(t, AllowListABI.Events["RoleSet"].ID, topics[0])
+				require.Equal(t, NoRole.Hash(), topics[1])
+				require.Equal(t, TestNoRoleAddr.Hash(), topics[2])
+				require.Equal(t, TestManagerAddr.Hash(), topics[3])
+				data := logsData[0]
+				require.Equal(t, NoRole.Bytes(), data)
```
darioush, line 345: could we abstract this to a helper function?

precompile/allowlist/role.go
ceyonur: we handle the err in case we call it with an arbitrary role. I.e calls like role.GetSetterFunctionName without a defined Role type Admin/Enabled etc.

darioush (dismissed):

contracts/contracts/interfaces/IAllowList.sol
```diff
@@ -2,10 +2,7 @@
 pragma solidity ^0.8.0;
 
 interface IAllowList {
-  event AdminAdded(address indexed sender, address indexed account);
-  event EnabledAdded(address indexed sender, address indexed account);
-  event ManagerAdded(address indexed sender, address indexed account);
-  event RoleRemoved(address indexed sender, address indexed account);
+  event RoleSet(uint256 indexed role, address indexed account, address indexed sender, uint256 oldRole);
```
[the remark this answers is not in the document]
aaronbuchwald: We should update the name for this throughout imo

precompile/allowlist/event.go
```diff
@@ -5,49 +5,33 @@
...
-		return PackEnabledAddedEvent(sender, account)
-	case NoRole:
-		return PackRoleRemovedEvent(sender, account)
-	default:
-		panic("invalid role")
+// UnpackRoleSetEventData attempts to unpack non-indexed [dataBytes].
+func UnpackRoleSetEventData(dataBytes []byte) (Role, error) {
+	eventData := struct {
+		OldRole *big.Int
+	}{}
```
aaronbuchwald, line 31: Is this not auto-generated by the precompile template?

precompile/allowlist/allowlist.go
```diff
@@ -181,11 +187,24 @@ func CreateAllowListPrecompile(precompileAddr common.Address) contract.StatefulP
...
+		} else if noRoleFnName, _ := NoRole.GetSetterFunctionName(); name == noRoleFnName {
+			fn = contract.NewStatefulPrecompileFunction(method.ID, createAllowListRoleSetter(precompileAddr, NoRole))
+		} else if managerFnName, _ := ManagerRole.GetSetterFunctionName(); name == managerFnName {
+			fn = contract.NewStatefulPrecompileFunctionWithActivator(method.ID, createAllowListRoleSetter(precompileAddr, ManagerRole), contract.IsDUpgradeActivated)
+		} else {
+			panic(fmt.Sprintf("unexpected method name: %s", name))
+		}
+		functions = append(functions, fn)
+	}
+	return functions
```
aaronbuchwald, line 209: what's the reason for this change?

aaronbuchwald (changes requested):

precompile/allowlist/event.go
ceyonur: it was generated but IMO it's better UX to convert these inline to use higher-level structs (like roles etc).

precompile/allowlist/allowlist.go
ceyonur: We introduced ABI to allow list. We also return error from `GetSetterFunctionName`.

aaronbuchwald (approved):

PR #981, opened 2023-11-02
Trie stress test

nytzuga:
## Why this should be merged

Write a benchmark with a smart contract that will do a lot of inserts in the TrieDB to test and measure our improvements.

## How this works

A smart contract has been created that inserts N number of elements in each transaction. The main goal is to effortlessly to make the EVM do a lot of inserts in the triedb and measure their performance.


core/bench_contract_test.go
```diff
@@ -0,0 +1,87 @@
...
+	stressBinStr string
+	//go:embed TrieStressTest.abi
+	stressABIStr string
+)
+
+func BenchmarkTrie(t *testing.B) {
+	benchInsertChain(t, true, generateTx(100, 5000))
+}
+
+func generateTx(blocks int, elements int64) func(int, *BlockGen) {
```
aaronbuchwald, line 51: Could we support a function signature like:

```
func generateTx(numContracts int, numElementsPerContract int64) func(int, *BlockGen)
```

where `numContracts` is the number of contracts to deploy and `numElementsPerContract is th number of elements to use as a parameter to `writeValues`.

Then we can leave it to the caller to ensure that the parameters they use do not exceed the gas limit (we could also support setting a different gas limit for those tests if we want to as well)

core/bench_contract_test.go
```diff
@@ -0,0 +1,87 @@
...
+	return func(i int, gen *BlockGen) {
+		block := gen.PrevBlock(i - 1)
+		gas := block.GasLimit()
+
+		if !deployedContract {
+			nonce := gen.TxNonce(benchRootAddr)
+			tx, _ := types.SignTx(types.NewContractCreation(nonce, big.NewInt(0), gasCreation, gasPrice, common.FromHex(stressBinStr)), signer, testKey)
+			sender, _ := types.Sender(signer, tx)
+			gen.AddTx(tx)
+			contractAddr = crypto.CreateAddress(sender, nonce)
```
aaronbuchwald, line 73: I think we can move the address calculation prior to this function ie. pre-calculate the addresses in a slice of size `numContracts`.

Then we can also adjust this block to look like:

```
if !deployedContract {
    // deploy the contracts
    return
}

// issue txs to interact with already deployed contracts
```

in golang it's very common to try and return early in the simple case and put more complex logic below. One example of this is in error handling ie. `if err != nil { return err }`, but the same idea applies for other control flows as well and usually leads to easier to read code.

core/bench_contract_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+		for {
+			gas -= gasTx
+			if gas < gasTx {
+				break
+			}
+			tx, _ := types.SignTx(types.NewTransaction(gen.TxNonce(benchRootAddr), contractAddr, big.NewInt(0), gasTx, gasPrice, txPayload), signer, testKey)
+			gen.AddTx(tx)
+		}
+	}
```
aaronbuchwald, line 86: I'd prefer to remove the gas calculation in favor of leaving it to the caller to ensure that the parameters they specify don't exceed the gas limit.

In my opinion, it's simpler to do it this way because it removes a bit of logic that is not necessary for the test and it's very reasonable to make that the responsibility of the caller imo.

core/bench_contract_test.go
```diff
@@ -0,0 +1,87 @@
...
+func BenchmarkTrie(t *testing.B) {
+	benchInsertChain(t, true, stressTestTrieDb(100, 100, 5000))
+}
+
+func stressTestTrieDb(numContracts int, callsPerBlock int, elements int64) func(int, *BlockGen) {
+	contractAddr := make([]common.Address, numContracts)
+	contractTxs := make([]*types.Transaction, numContracts)
+
+	gasPrice := big.NewInt(225000000000)
+	gasTx := uint64(22000)
```
aaronbuchwald, line 56: If you specify this amount of gas, then regardless of how many elements are requested to be modified once it hits this limit it will hit `ErrOutOfGas` and revert.

Could we change the logic here to make sure that it sanity checks it has enough gas to execute the requested number of elements?

core/bench_contract_test.go
```diff
@@ -0,0 +1,87 @@
...
+				gen.AddTx(contractTxs[deployedContracts])
+			}
+			return
+		}
+
+		for e := 0; e < callsPerBlock; e++ {
+			tx, _ := types.SignTx(types.NewTransaction(gen.TxNonce(benchRootAddr), contractAddr[i%deployedContracts], big.NewInt(0), gasTx, gasPrice, txPayload), signer, testKey)
+			gen.AddTx(tx)
+		}
+	}
```
aaronbuchwald, line 86: could we make this fail loudly instead of quietly if it fails to deploy the contracts or hits err out of gas when executing transactions?

aaronbuchwald (changes requested):

core/bench_contract_test.go
nytzuga: AddTx and addTx [are very loud](https://github.com/ava-labs/subnet-evm/blob/8beb366405008ee0f40337069e7bf124f8e23140/core/chain_makers.go#L115-L117). Should I make a variant that returns the error instead of panicking?

core/bench_contract_test.go
```diff
@@ -0,0 +1,177 @@
...
+					fmt.Printf("\nBlock: %d\nReceipt: %+v\n", i, receipt)
+					panic(err)
+				}
+				gas -= gasCreation
+			}
+			return
+		}
+
+		for e := 0; e < callsPerBlock; e++ {
+			tx, err := types.SignTx(types.NewTransaction(gen.TxNonce(benchRootAddr), contractAddr[i%deployedContracts], big.NewInt(0), gasTxLimit, gasPrice, txPayload), signer, testKey)
```
ceyonur, line 167: if we use only a single contractaddr per block, why do we pre-deploy *all* of them before this codeblock?

core/bench_contract_test.go
```diff
@@ -0,0 +1,177 @@
...
+		"writeValues",
+		big.NewInt(elements),
+	)
+
+	return func(i int, gen *BlockGen) {
+		if len(contractTxs) != deployedContracts {
+			block := gen.PrevBlock(i - 1)
+			gas := block.GasLimit()
+			for ; deployedContracts < len(contractTxs) && gasCreation < gas; deployedContracts++ {
+				if receipt, err := gen.AddTxOrFail(contractTxs[deployedContracts]); err != nil {
```
ceyonur, line 157: if we're panicking here why we added the new `AddTxOrFail` function?

core/bench_contract_test.go
```diff
@@ -0,0 +1,177 @@
...
+	contractAddr := make([]common.Address, numContracts)
+	contractTxs := make([]*types.Transaction, numContracts)
+
+	gasPrice := big.NewInt(225000000000)
+	gasCreation := uint64(258000)
+	deployedContracts := 0
+
+	for i := 0; i < numContracts; i++ {
+		nonce := uint64(i)
+		tx, _ := types.SignTx(types.NewContractCreation(nonce, big.NewInt(0), gasCreation, gasPrice, common.FromHex(stressBinStr)), signer, testKey)
```
ceyonur, line 140: seems `NewContractCreation` is deprecated

core/bench_contract_test.go
ceyonur, line 167: q: why we use `benchRootAddr` for nonce?

core/bench_contract_test.go
```diff
@@ -0,0 +1,177 @@
...
+				}
+				gas -= gasCreation
+			}
+			return
+		}
+
+		for e := 0; e < callsPerBlock; e++ {
+			tx, err := types.SignTx(types.NewTransaction(gen.TxNonce(benchRootAddr), contractAddr[i%deployedContracts], big.NewInt(0), gasTxLimit, gasPrice, txPayload), signer, testKey)
+			if err != nil {
+				panic(err)
```
ceyonur, line 169: can we use `require.NoError(t, err)` instead of panicking?

core/bench_contract_test.go
ceyonur, line 167: `NewTransaction` is deprecated

core/bench_contract_test.go
```diff
@@ -0,0 +1,177 @@
...
+	txPayload, _ := stressABI.Pack(
+		"writeValues",
+		big.NewInt(elements),
+	)
+
+	return func(i int, gen *BlockGen) {
+		if len(contractTxs) != deployedContracts {
+			block := gen.PrevBlock(i - 1)
+			gas := block.GasLimit()
+			for ; deployedContracts < len(contractTxs) && gasCreation < gas; deployedContracts++ {
```
ceyonur, line 156: Wouldn't this silently skip deploying contracts once it exceeds gas limit? I think `AddTxOrFail` will return once we exceed limit, so this check should not be required.

core/bench_contract_test.go
```diff
@@ -0,0 +1,177 @@
...
+				if receipt, err := gen.AddTxOrFail(contractTxs[deployedContracts]); err != nil {
+					fmt.Printf("\nBlock: %d\nReceipt: %+v\n", i, receipt)
+					panic(err)
+				}
+				gas -= gasCreation
+			}
+			return
```
ceyonur, line 163: q: Why do we return here?

core/bench_contract_test.go
nytzuga: @ceyonur I think benchRootAddr is the one signing these transactions. Most of this is boilerplate from other benchmarks. I can make another's keys to sign transactions, it would make no difference imho.

nytzuga: @ceyonur what should I use instead?

core/bench_contract_test.go
nytzuga: No, it will be executed next time this lamba/function is called.

core/bench_contract_test.go
nytzuga: To split the function into two parts, the first bit is deploying all the contracts, using as many blocks as needed. Once that stage is done, it will start calling these contracts. Does that make sense @ceyonur ?

contracts/contracts/TrieStressTest.sol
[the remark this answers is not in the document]
nytzuga: That's the source code of the smart contract. I think it should be included but it is your call. @ceyonur . What should be done?

core/bench_contract_test.go
nytzuga: I will improve the logic to call smart contracts in a better way. The idea is to stress test the trie structure.

nytzuga: Done in 380e198a

core/bench_contract_test.go
nytzuga: Is it better now @aaronbuchwald? The gas calculation is to split the contract creations into multiple blocks, but no other gas calculation is being used.

core/bench_contract_test.go
nytzuga: We have a better error reporting system now, it will panic/error if ErrOutGas is returned. Is that ok @aaronbuchwald ?

contracts/contracts/TrieStressTest.sol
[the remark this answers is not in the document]
ceyonur: This file is not being directly used (compiled files like `.abi` and `.bin` used) in tests. This really doesn't fit the purpose of `contracts` folder which is to provide examples/libraries for other smart contracts on chain. So we should either remove this, or if we want to keep it we should move all of these to a subpackage under `core` like `core/trietest` (assuming we don't need any unexported stuff from `core` package)

core/bench_contract_test.go
ceyonur: We should use `benchRootKey`. Luckily these two were defined for same addresses, otherwise this would not work: 
https://github.com/ava-labs/subnet-evm/blob/trie-stress-test/core/bench_test.go#L88
https://github.com/ava-labs/subnet-evm/blob/trie-stress-test/core/state_processor_test.go#L51

ceyonur: https://github.com/ava-labs/subnet-evm/blob/trie-stress-test/core/types/tx_legacy.go#L47

core/bench_contract_test.go
ceyonur: hmm so we deploy them in different blocks.

core/bench_contract_test.go
```diff
@@ -0,0 +1,173 @@
...
+
+var (
+	//go:embed TrieStressTest.bin
+	stressBinStr string
+	//go:embed TrieStressTest.abi
+	stressABIStr string
+)
+
+func BenchmarkTrie(t *testing.B) {
+	benchInsertChain(t, true, stressTestTrieDb(t, 100, 6, 50, 1202102))
```
ceyonur, line 52: what is the target we want to benchmark here? This currently inserts `N` number of blocks but these seems static (numContracts, callsPerBlock, elements). Should these be depending on `N`?

ceyonur: or do we just want to see how total writes impact the benchmark (N * callsPerBlock * elements?)

core/bench_contract_test.go
ceyonur: I see, yea I think it makes sense if we need all of them before starting to write.

core/bench_contract_test.go
nytzuga: It is up to the callee to decide how many instances of their contract they need to test the feature they are testing. This callback function will ensure the deployment of all the contracts needed, even if that means doing so in different blocks. Once all the contracts are deployed, the real benchmarking begins. At that stage, every call should be finished in one block.

core/bench_contract_test.go
nytzuga: Yes, as described above, this function has two stages, the first one is preparing the benchmark (deploying the contract) and the other part does the actual things we are benchmarking.

I think in the future `benchInsertChain` should be extended to take *two* callbacks (or we create our own flavor with this feature), the first one that is not part of the benchmark should prepare the blockchain and set to the expected state, the other function will be kicked in later, and will be benchmarked.

core/bench_contract_test.go
nytzuga: We want to deploy N smart contract instances (100), Call 6 different contracts per block (and measure their performance). Each call will append 50 new elements (this logic is part of the solidity code). The last argument is how much gas should be used per each call.

aaronbuchwald (approved):

core/bench_contract_test.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	stressABI := contract.ParseABI(stressABIStr)
+	txPayload, _ := stressABI.Pack(
+		"writeValues",
+		big.NewInt(elements),
+	)
+
+	return func(i int, gen *BlockGen) {
+		if len(contractTxs) != deployedContracts {
```
ceyonur, line 82: can we add a comment here about this will first deploy all contracts before we actually start writing?

core/bench_contract_test.go
ceyonur: So we don't want these to be anyhow dynamic with benchmark value (`N`)? 
I'm a bit confused at "We want to deploy N smart contract instances (100)". This won't be same with `t.N` which is the benchmark iteration value. I.e BenchmarkTrie-8 BenchmarkTrie-16 where N=8, 16 etc.

ceyonur:
I just want to make sure we're on the same page with benchmark target, this looked a bit static for me.
Also there is one unresolved convo [here](https://github.com/ava-labs/subnet-evm/pull/981#discussion_r1401823453) Other than that LGTM

core/bench_contract_test.go
nytzuga: @ceyonur not really.  The idea is to make changes to the underlying library and see how the benchmark changes with commit changes.

contracts/contracts/TrieStressTest.sol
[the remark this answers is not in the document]
nytzuga: I'm moving to their folder inside core as suggested.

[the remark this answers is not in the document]
ceyonur: I meant moving the test as well to the same subpackage.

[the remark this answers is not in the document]
nytzuga: @ceyonur functions inside the tests files cannot be exported unfortunately. Unless a bigger refactor is involved this test must be inside `core` /cc @aaronbuchwald

ceyonur (dismissed):

core/trie_stress_bench_test.go
```diff
@@ -0,0 +1,116 @@
...
+				To:       &contractAddr[contractId],
+				Value:    big.NewInt(0),
+				Gas:      gasTxLimit,
+				GasPrice: gasPrice,
+				Data:     txPayload,
+			}), signer, testKey)
+			require.NoError(err)
+			gen.AddTx(tx)
+			require.Equal(gen.receipts[len(gen.receipts)-1].Status, types.ReceiptStatusSuccessful, "Execution of last transaction failed")
+			require.NoError(err)
```
aaronbuchwald, line 113:
```suggestion
```

this is a duplicate on the same error value

aaronbuchwald (dismissed):

ceyonur (approved):

aaronbuchwald (approved):

PR #984, opened 2023-11-02
warp: add details to accepted message log

darioush:
## Why this should be merged
Adds more information to the log emitted when a block sending warp messages are accepted.

## How this works
Passes through blockNumber, blockHash.

## How this was tested
CI

## How is this documented
N/A

### DIFF
--- a/precompile/precompileconfig/config.go
+++ b/precompile/precompileconfig/config.go
@@ -75,7 +75,7 @@ type AcceptContext struct {
 // will not maintain backwards compatibility of this interface and your code should not
 // rely on this. Designed for use only by precompiles that ship with subnet-evm.
 type Accepter interface {
-	Accept(acceptCtx *AcceptContext, txHash common.Hash, logIndex int, topics []common.Hash, logData []byte) error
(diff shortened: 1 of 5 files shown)


aaronbuchwald (approved):

precompile/precompileconfig/config.go
```diff
@@ -75,7 +75,7 @@ type AcceptContext struct {
 // will not maintain backwards compatibility of this interface and your code should not
 // rely on this. Designed for use only by precompiles that ship with subnet-evm.
 type Accepter interface {
-	Accept(acceptCtx *AcceptContext, txHash common.Hash, logIndex int, topics []common.Hash, logData []byte) error
+	Accept(acceptCtx *AcceptContext, blockHash common.Hash, blockNumber uint64, txHash common.Hash, logIndex int, topics []common.Hash, logData []byte) error
```
ceyonur, line 78: I wish we could have imported types.Log but seems it is cyclic :/

ceyonur (approved):
The interface function got a bit bulky, but I guess we have an cyclic import with types package. If this function gets any bigger, we should create a LogContext interface in the future.

PR #989, opened 2023-11-07
remove unnecessary avalanchego version from script

ceyonur:
## Why this should be merged

Cleans duplicate variable in scripts

## How this works

Merges two variable.

## How this was tested

E2E tests should cover this.

## How is this documented

No need


darioush (approved):

aaronbuchwald (dismissed):

Dockerfile
```diff
@@ -1,8 +1,8 @@
 # syntax=docker/dockerfile:experimental
 
 # ============= Setting up base Stage ================
-# Set required AVALANCHE_VERSION parameter in build image script
-ARG AVALANCHE_VERSION
+# Set required AVALANCHEGO_VERSION parameter in build image script
```
darioush, line 4: I think we should go the other way around here, let's keep the Dockerfile name unchanged since it may be referenced externally.

ceyonur: yup good idea

darioush (approved):

PR #995, opened 2023-11-14
Cleanup warp related code

aaronbuchwald:
- Remove unnecessary pointer to rules type from `handlePrecompilesAccept`
- Migrates warp message handler unit tests from `vm_test.go` -> `vm_warp_test.go`
- Removes check that duration after responding to a warp signature request is greater than zero (fails on Windows)


plugin/evm/block.go
```diff
@@ -69,7 +69,7 @@ func (b *Block) Accept(context.Context) error {
 	// precompiles are committed atomically with the vm's lastAcceptedKey.
 	rules := b.vm.chainConfig.AvalancheRules(b.ethBlock.Number(), b.ethBlock.Timestamp())
 	sharedMemoryWriter := NewSharedMemoryWriter()
-	if err := b.handlePrecompileAccept(&rules, sharedMemoryWriter); err != nil {
+	if err := b.handlePrecompileAccept(rules, sharedMemoryWriter); err != nil {
```
darioush, line 72: this is a struct so if we pass it without a ptr it will create an unneeded copy, not sure why we would prefer that?

darioush (commented):
other changes look good

plugin/evm/block.go
aaronbuchwald: This was just for consistency, since we pass by value elsewhere. Looking around, it looks like we've introduced this ourselves and there's only one place where it's passed this way in geth in the new exported `LookupInstructionSet`.

Will revert

ceyonur (approved):

PR #1002, opened 2023-11-18
add WaitForResult

ceyonur:
Adds a helper method to remove duplicated code.


darioush (approved):

darioush:
cc @joshua-kim : not sure if this cleanup applies to the networking code in avalanchego as well, thought I'd mention you here

joshua-kim:
> cc @joshua-kim : not sure if this cleanup applies to the networking code in avalanchego as well, thought I'd mention you here

It shouldn't, long term we will deprecate all this code to use the `p2p.Client`

PR #1008, opened 2023-11-28
Add subnetID arg to warp APIs

aaronbuchwald:
This PR adds a `subnetIDStr` argument to the Warp APIs, so that the caller can specify what subnetID's validator set the API should attempt to fetch validator signatures from.

This is used in the case of the C-Chain, so that signatures from subnetA can be fetched when aggregating signatures for a message that needs to be sent to subnetA.

### DIFF
--- a/tests/warp/warp_test.go
+++ b/tests/warp/warp_test.go
@@ -329,7 +329,7 @@ var _ = ginkgo.Describe("[Warp]", ginkgo.Ordered, func() {
 		gomega.Expect(err).Should(gomega.BeNil())
 
 		// Specify WarpQuorumDenominator to retrieve signatures from every validator
-		signedWarpMessageBytes, err := client.GetMessageAggregateSignature(ctx, unsignedWarpMessageID, params.WarpQuorumDenominator)
+		signedWarpMessageBytes, err := client.GetMessageAggregateSignature(ctx, unsignedWarpMessageID, params.WarpQuorumDenominator, "")
 		gomega.Expect(err).Should(gomega.BeNil())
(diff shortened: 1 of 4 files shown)


warp/backend.go
```diff
@@ -109,7 +109,7 @@ func (b *backend) GetMessageSignature(messageID ids.ID) ([bls.SignatureLen]byte,
 
 	unsignedMessage, err := b.GetMessage(messageID)
 	if err != nil {
-		return [bls.SignatureLen]byte{}, fmt.Errorf("failed to get warp message %s from db: %w", messageID.String(), err)
+		return [bls.SignatureLen]byte{}, err
```
aaronbuchwald, line 112: This previously resulted in a double-wrapped error with the same message twice

tests/warp/warp_test.go
```diff
@@ -329,7 +329,7 @@ var _ = ginkgo.Describe("[Warp]", ginkgo.Ordered, func() {
 		gomega.Expect(err).Should(gomega.BeNil())
 
 		// Specify WarpQuorumDenominator to retrieve signatures from every validator
-		signedWarpMessageBytes, err := client.GetMessageAggregateSignature(ctx, unsignedWarpMessageID, params.WarpQuorumDenominator)
+		signedWarpMessageBytes, err := client.GetMessageAggregateSignature(ctx, unsignedWarpMessageID, params.WarpQuorumDenominator, "")
```
aaronbuchwald, line 332: Passing in `""` is a no-op here

darioush (approved):

PR #1009, opened 2023-11-28
cmd/simulator: refactor simulator to support warp load test

aaronbuchwald:
This PR refactors the load simulator to better support async tx sequence generation for warp load testing.

It also better separates some of the load simulator code (still significant room for improvement).

Specific changes:
- add tx worker version that confirms transactions by receipt instead of nonce
- add confirm reached tip helper function to confirm all workers have observed the latest tip (as of calling the function)

### DIFF
--- a/cmd/simulator/load/loader.go
+++ b/cmd/simulator/load/loader.go
@@ -31,6 +31,10 @@ const (
 	MetricsEndpoint = "/metrics" // Endpoint for the Prometheus Metrics Server
 )
 
+// Loader executes a series of worker/tx sequence pairs.
+// Each worker/txSequence pair issues [batchSize] transactions, confirms all
+// of them as accepted, and then moves to the next batch until the txSequence
+// is exhausted.
 type Loader[T txs.THash] struct {
 	clients     []txs.Worker[T]
 	txSequences []txs.TxSequence[T]
--- a/cmd/simulator/load/loader.go
+++ b/cmd/simulator/load/loader.go
@@ -6,16 +6,13 @@ package load
 import (
 	"context"
 	"crypto/ecdsa"
-	"errors"
 	"fmt"
-	"io/ioutil"
 	"math/big"
-	"net/http"
 	"os"
 	"os/signal"
 	"strconv"
-	"strings"
 	"syscall"
+	"time"
 
 	"github.com/ava-labs/subnet-evm/cmd/simulator/config"
 	"github.com/ava-labs/subnet-evm/cmd/simulator/key"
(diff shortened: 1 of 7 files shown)


cmd/simulator/load/worker.go
```diff
@@ -44,14 +45,49 @@ func NewSingleAddressTxWorker(ctx context.Context, client ethclient.Client, addr
...
 	for {
+		// Update the worker's accepted nonce, so we can check on the next iteration
+		// if the transaction has been accepted.
+		acceptedNonce, err := tw.client.NonceAt(ctx, tw.address, nil)
+		if err != nil {
+			return fmt.Errorf("failed to await tx %s nonce %d: %w", tx.Hash(), txNonce, err)
+		}
+		tw.acceptedNonce = acceptedNonce
+
+		log.Info("trying to confirm tx", "txHash", tx.Hash(), "txNonce", txNonce, "acceptedNonce", tw.acceptedNonce)
```
aaronbuchwald, line 90:
```suggestion
		log.Debug("confirming tx", "txHash", tx.Hash(), "txNonce", txNonce, "acceptedNonce", tw.acceptedNonce)
```

cmd/simulator/load/loader.go
```diff
@@ -27,15 +24,101 @@ import (
...
 )
 
+type Loader[T txs.THash] struct {
+	clients     []txs.Worker[T]
+	txSequences []txs.TxSequence[T]
+	batchSize   uint64
+	metrics     *metrics.Metrics
+}
+
+func New[T txs.THash](
```
darioush, line 41: NewLoader maybe?

cmd/simulator/txs/agent.go
```diff
@@ -75,12 +76,6 @@ func (a issueNAgent[T]) Execute(ctx context.Context) error {
 		totalConfirmedTime time.Duration
 	)
 
-	defer func() {
-		if err := a.worker.Close(ctx); err != nil {
```
darioush, line 79: where are the workers closed now?

cmd/simulator/metrics/metrics.go
```diff
@@ -40,3 +57,77 @@ func NewMetrics(reg prometheus.Registerer) *Metrics {
...
+		if err := server.ListenAndServe(); !errors.Is(err, http.ErrServerClosed) {
+			log.Error("Metrics server error: %v", err)
+		}
+	}()
+
+	return ms
+}
+
+func (ms *MetricsServer) Shutdown() {
+	ms.stopOnce.Do(func() {
```
darioush, line 109: nit: I don't think stopOnce is needed since calling cancel and reading from stopCh can both be repeated.

I think we can simplify the shutdown, but no need to block the PR for this.

cmd/simulator/load/loader.go
```diff
@@ -27,15 +24,101 @@ import (
...
-	"github.com/prometheus/client_golang/prometheus"
-	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"golang.org/x/sync/errgroup"
 )
 
 const (
 	MetricsEndpoint = "/metrics" // Endpoint for the Prometheus Metrics Server
 )
 
+type Loader[T txs.THash] struct {
```
darioush, line 34: Could we add a comment for Loader?

cmd/simulator/load/loader.go
aaronbuchwald: We've started explicitly removing names like this that repeat the package name, so I'd prefer to continue that trend and use `New` here so that the caller uses `load.New(...)` instead of `load.NewLoader(...)`

cmd/simulator/metrics/metrics.go
aaronbuchwald: Will remove

cmd/simulator/txs/agent.go
aaronbuchwald: The workers are no longer closed. It wasn't strictly necessary to close them before, so it's now up to the creator of those clients to call `Close` if desired.

cmd/simulator/load/loader.go
aaronbuchwald: Good call

darioush (approved):

ceyonur (approved):

PR #1015, opened 2023-12-01
remove unnecessary assignment

ceyonur:
## Why this should be merged

Removes unnecessary assignment

## How this works

Removes unnecessary assignment

## How this was tested

Existing tests should cover this


darioush (approved):
nice

Issue #1020, opened 2023-12-04
Create custom serialization for external structs


nytzuga:
## **Context and scope**

As part of #758 and #268, it is needed to support external structures. Our [codec uses reflection and serializes only fields which have the  `serialize:"true"` tag.](https://github.com/ava-labs/avalanchego/blob/7623ffd4be915a5185c9ed5e11fa9be15a6e1f00/codec/reflectcodec/struct_fielder.go#L15-L25) 

That is problematic for external structs, particularly the `big.Int` struct, which is part of the data to be exchanged between nodes or to calculate the upgrade config change. Hashing or exchanging configurations require serialization and `big.Int` is ignored since they don't have any `serialize:"true"` for their inner members.

## **Discussion and alternatives**

There are two alternatives proposed in this issue.

### `serialize:"all"`

The easiest approach is introducing yet another patch to our codec, extending it to support `serialize:"all"`. 

For context, several improvements have already been made to enhance the codec to support the serialization of the config ([nullable pointers](https://github.com/ava-labs/avalanchego/pull/2171), [deterministic map serialization](https://github.com/ava-labs/avalanchego/pull/2171)).

This new tag will be another way of serializing fields (alongside with `serialize:"true"`, `serialize:"true,nullable"`). This tag will signal the codec to serialize all inner fields as if they all have the `serialize:"true"` tag in each field and inner fields.

This change will leverage a lot of the existing code in the codec, and it is backward compatible.

### Wrap the `big.Int`

The alternative solution would be to have our version of `big.Int`, owned by the package. This type will be converted to a property `big.Int` for usage after deserialization.

This solution would require zero changes in the codec but will require changes in all the codec inside subnet-EVM.  This solution would have to be replicated everywhere in the configurations where `big.Int` and any external struct is being read.

**Open questions**

Is there any other alternative?

ceyonur:
There are few other options:

* Adding a custom marshaller/encoder support so that we would not need to expand the codec everytime with new types.
* Creating our own codec in subnet-evm.
* Create intermediate types like in gen_genesis.go

ceyonur:
Some more ideas:

### Custom Handler In Codec

This involves adding a handler that would check if a certain interface is implemented and call custom serialize/deserialize method. This involves a change to the codec. 

In this case we should decide which interface to be handled in codec. Possibilities are `encoding.MarshalBinary`, `encoding.MarshalText`,  `gob.Encode` or a new custom interface . 

[...]

PR #1022, opened 2023-12-05
Rewrote handshake encoding to use a deterministic JSON instead

nytzuga:
## Why this should be merged

Golang uses the order defined in the structures for the ordering of their fields unless it is serializing an unnamed struct / an `interface {}`. In the latter case, the fields are going to be sorted alphabetically by their keys.


plugin/evm/message/handshake/upgrade_config_message.go
```diff
@@ -36,45 +24,32 @@ func (u *UpgradeConfigMessage) ID() common.Hash {
...
+	return &upgradeConfig, nil
+}
 
-	return &params.UpgradeConfig{
-		OptionalNetworkUpgrades: config.OptionalNetworkUpgrades,
-		StateUpgrades:           config.StateUpgrades,
-		PrecompileUpgrades:      PrecompileUpgrades,
-	}, nil
+// Encodes any object to JSON with a deterministic output. All the keys, even
+// into inner objects, are sorted alphabetically.
```
patrick-ogrady, line 41: Please link docs/code from the stdlib that explain why this is correct. I found this documentation in `encoding/json`, for example:
```
Map values encode as JSON objects. The map's key type must either be a string, an integer type, or implement encoding.TextMarshaler. The map keys are sorted and used as JSON object keys by applying the following rules, subject to the UTF-8 coercion described for string values above:
```

patrick-ogrady: The most important line seems to be this one (inside the `mapEncoder`: https://cs.opensource.google/go/go/+/refs/tags/go1.21.5:src/encoding/json/encode.go;l=751)

plugin/evm/message/handshake/upgrade_config_message.go
```diff
@@ -36,45 +24,32 @@ func (u *UpgradeConfigMessage) ID() common.Hash {
...
-// UpgradeConfigToNetworkMessage).
+// Attempts to parse a params.UpgradeConfig from a []byte
 //
 // The function returns a reference of *params.UpgradeConfig
-func NewUpgradeConfigMessageFromBytes(bytes []byte) (*params.UpgradeConfig, error) {
-	var config networkUpgradeConfigMessage
-	version, err := Codec.Unmarshal(bytes, &config)
+func UpgradeConfigFromBytes(bytes []byte) (*params.UpgradeConfig, error) {
+	var upgradeConfig params.UpgradeConfig
+	err := json.Unmarshal(bytes, &upgradeConfig)
```
patrick-ogrady, line 32: We should comment that this only works for JSON upgrade bytes. There is no requirement that JSON is used here generally.

plugin/evm/message/handshake/upgrade_config_message.go
patrick-ogrady: All in all, this is an interesting property to rely on as it goes above what is required in the JSON spec: https://www.ietf.org/rfc/rfc4627.txt

patrick-ogrady: its pretty crazy they do this by default instead of putting it behind a config flag (because there is an obvious performance penalty)

patrick-ogrady: (I don't think your implementation is wrong to be clear, just potentially brittle)

plugin/evm/message/handshake/upgrade_config_message.go
nytzuga: I agree @patrick-ogrady, this was just a quick prototype. The alternatives are described in #1020 or relying in an external library such as [json-canonicalization](https://github.com/cyberphone/json-canonicalization).

For me the ideal solution would be to extend our codec to allow the serialization of structs not owned by the package.

nytzuga: There are many other options to calculate a hash, but I've been told to avoid reflection at all costs.

nytzuga:
I think this is closed in favour of https://github.com/ava-labs/subnet-evm/issues/1020 or any other approach as part of  https://github.com/ava-labs/subnet-evm/pull/762 (less ideal)

PR #1023, opened 2023-12-06
Fix abi get inputs

ceyonur:
## Why this should be merged

Fixes the enforced strict mode in Unpack decoder, which prevents using extra padded bytes in inputs.

More details: https://docs.soliditylang.org/en/latest/abi-spec.html#strict-encoding-mode

## How this works

Adds a flag to toggle the strict mode. It is by default enabled, and DUpgrade disables it (strict mode is off in DUpgrade) 

## How this was tested

Added unit tests

## How is this documented

Should be documented with Dupgrade changes


accounts/abi/abi_extra_test.go
```diff
@@ -36,13 +36,36 @@ func TestUnpackInputIntoInterface(t *testing.T) {
...
 	require.NoError(t, err)
 
 	// Verify unpacked values match input
 	require.Equal(t, v.Amount, input.Amount)
 	require.EqualValues(t, v.Amount, input.Amount)
 	require.True(t, bytes.Equal(v.Memo, input.Memo))
+
+	// add 32-bytes extra padding to data
+	// this should work because it is divisible by 32
+	data = append(data, make([]byte, 32)...)
```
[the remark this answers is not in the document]
ceyonur: done

x/warp/contract.go
```diff
@@ -149,7 +150,9 @@ func getVerifiedWarpBlockHash(accessibleState contract.AccessibleState, caller c
 // UnpackGetVerifiedWarpMessageInput attempts to unpack [input] into the uint32 type argument
 // assumes that [input] does not include selector (omits first 4 func signature bytes)
 func UnpackGetVerifiedWarpMessageInput(input []byte) (uint32, error) {
-	res, err := WarpABI.UnpackInput("getVerifiedWarpMessage", input)
+	// We don't use strict mode here because it was disabled with the DUpgrade.
+	// And Warp is will be deployed after the DUpgrade.
```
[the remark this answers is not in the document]
ceyonur: done

precompile/contracts/feemanager/contract_test.go
```diff
@@ -20,6 +20,17 @@ import (
...
+	regressionFeeConfig = commontype.FeeConfig{
+		GasLimit:                 big.NewInt(25000000),
+		TargetBlockRate:          2,
+		MinBaseFee:               big.NewInt(5000000000),
+		TargetGas:                big.NewInt(75000000),
+		BaseFeeChangeDenominator: big.NewInt(84),
+		MinBlockGasCost:          big.NewInt(0),
+		MaxBlockGasCost:          big.NewInt(2000000000000000000),
+		BlockGasCostStep:         big.NewInt(1000000000000000000),
+	}
```
darioush, line 33: should we add some assertion that the regression bytes is the FeeConfig + some padding or derive the regression bytes by adding padding to the serialized regressionFeeConfig?
Also, do the padding bytes have to be 0 or do we allow any as this PR suggests?

accounts/abi/abi_extra_test.go
```diff
@@ -31,18 +31,69 @@ func TestUnpackInputIntoInterface(t *testing.T) {
...
-	err = abi.UnpackInputIntoInterface(&v, "receive", data[4:]) // skips 4 byte selector
+	abi, err = JSON(strings.NewReader(TEST_ABI))
 	require.NoError(t, err)
 
-	// Verify unpacked values match input
-	require.Equal(t, v.Amount, input.Amount)
-	require.EqualValues(t, v.Amount, input.Amount)
-	require.True(t, bytes.Equal(v.Memo, input.Memo))
+	for _, test := range []struct {
+		Name                   string
```
darioush, line 42: style nit: I think we should use unexported names when possible.

darioush (approved):

precompile/contracts/feemanager/contract_test.go
ceyonur: I derived the feeconfig from bytes and confirmed with the DFK that it was the requested config. So I don't think we need to continuously assert that these are equal. 

Yes it can be other than 0. But I think the common case is 0.

accounts/abi/abi_extra_test.go
ceyonur: done

darioush (approved):

PR #1040, opened 2024-01-03
Add ChainConfig to ethclient

cam-schultz:
## Why this should be merged
Adds `ChainConfig` to `ethclient.Client` so that it can be called by an external Go application. No namespace changes are required.

## How this works
See above.

## How this was tested
Local testing using an external Go application that calls the method via the client.

## How is this documented
N/A


ceyonur (approved):
lgtm

ethclient/ethclient.go
```diff
@@ -141,6 +143,16 @@ func (ec *client) Client() *rpc.Client {
 
 // Blockchain Access
 
+// ChainConfig retrieves the current chain config.
+func (ec *client) ChainConfig(ctx context.Context) (*params.ChainConfigWithUpgradesJSON, error) {
+	var result *params.ChainConfigWithUpgradesJSON
```
aaronbuchwald, line 148: Can we return `*params.ChainConfig` instead of `*params.ChainConfigWithUpgradesJSON` since it's meant to be an intermediate type for marshalling?

cam-schultz: Sure thing. Should we also return `*params.ChainConfig` interally? https://github.com/ava-labs/subnet-evm/blob/master/internal/ethapi/api.go#L145

ceyonur: It's also used in `getChainConfig` API and it also returns `ChainConfigWithUpgradesJSON`. we should keep them same.

cam-schultz: Fixed now.

ethclient/ethclient.go
```diff
@@ -141,6 +143,16 @@ func (ec *client) Client() *rpc.Client {
 
 // Blockchain Access
 
+// ChainConfig retrieves the current chain config.
+func (ec *client) ChainConfig(ctx context.Context) (*params.ChainConfig, error) {
```
ceyonur, line 147: I meant returning `ChainConfigWithUpgradesJSON` instead of  `ChainConfig` [here](https://github.com/ava-labs/subnet-evm/pull/1040#discussion_r1442719438). Sorry for the confusion.

internal/ethapi/api.go
```diff
@@ -141,8 +141,8 @@ func (s *EthereumAPI) Syncing() (interface{}, error) {
 	return false, nil
 }
 
-func (s *BlockChainAPI) GetChainConfig(ctx context.Context) *params.ChainConfigWithUpgradesJSON {
-	return s.b.ChainConfig().ToWithUpgradesJSON()
+func (s *BlockChainAPI) GetChainConfig(ctx context.Context) *params.ChainConfig {
```
ceyonur, line 144: If we think it's more convenient to return this without Upgrades, we should add a new API method for `GetUpgrades`. Otherwise we should revert this change and use  `ChainConfigWithUpgradesJSON` in both client and API function. 
cc @aaronbuchwald

ceyonur (approved):

darioush (approved):

PR #1043, opened 2024-01-05
fix load test

patrick-ogrady:
TODO: add `NoLocals: false` + Increase `GlobalSlots` config for the load test to ensure nothing gets dropped (still racy even with faster bloom)


core/txpool/txpool.go
```diff
@@ -207,10 +207,10 @@ var DefaultConfig = Config{
 	PriceLimit: 1,
 	PriceBump:  10,
 
-	AccountSlots: 16,
-	GlobalSlots:  4096 + 1024, // urgent + floating queue capacity with 4:1 ratio
+	AccountSlots: 102400,
```
patrick-ogrady, line 210: You might be able to achieve the same by using locals in this load test?

tests/utils/runner/network_manager.go
```diff
@@ -66,12 +66,14 @@ func NewDefaultANRConfig() ANRConfig {
...
 		GlobalNodeConfig: `{
+			"log-level":"info",
 			"log-display-level":"info",
 			"proposervm-use-current-height":true
 		}`,
 		GlobalCChainConfig: `{
 			"warp-api-enabled": true,
-			"log-level": "debug"
+			"log-level": "debug",
+			"log-display-level": "debug"
```
aaronbuchwald, line 76:
```suggestion
			"log-level": "debug",
```

There is no `log-display-level` for the C-Chain specific config

tests/utils/runner/network_manager.go
```diff
@@ -66,12 +66,14 @@ func NewDefaultANRConfig() ANRConfig {
 		AvalancheGoExecPath: os.ExpandEnv("$GOPATH/src/github.com/ava-labs/avalanchego/build/avalanchego"),
 		PluginDir:           os.ExpandEnv("$GOPATH/src/github.com/ava-labs/avalanchego/build/plugins"),
 		GlobalNodeConfig: `{
+			"log-level":"info",
```
aaronbuchwald, line 69: Should we remove `log-display-level` if we are going to specify `log-level`? Also, not sure why we were originally specifying `log-display-level` here (although it looks like I wrote it, I think I copied it from somewhere).

Most likely this was to log to stdout without writing to the log directory. I'd assume if anything we'd want more specific logs written to stdout, so that we can inspect them on test failure without crowding out the output from the network runner, which pipes the logs from stdout (as I understand it).

aaronbuchwald:
Looks like this needs to bump the `compatibility.json` file

plugin/evm/tx_gossip_test.go
```diff
@@ -82,9 +82,9 @@ func TestEthTxGossip(t *testing.T) {
 	}
 
 	// Ask the VM for any new transactions. We should get nothing at first.
-	emptyBloomFilter, err := gossip.NewBloomFilter(txGossipBloomMaxItems, txGossipBloomFalsePositiveRate)
+	emptyBloomFilter, err := gossip.NewBloomFilter(txGossipBloomMinTargetElements, txGossipBloomTargetFalsePositiveRate, txGossipBloomResetFalsePositiveRate)
 	require.NoError(err)
-	emptyBloomFilterBytes, _, err := emptyBloomFilter.Marshal()
+	emptyBloomFilterBytes, _ := emptyBloomFilter.Marshal()
 	require.NoError(err)
```
darioush, line 88: should be removed: 
```suggestion
```

darioush (approved):

tests/utils/runner/network_manager.go
ceyonur: removed

tests/utils/runner/network_manager.go
ceyonur: removed

PR #1048, opened 2024-01-19
feat(plugin/evm): make tx pool lifetime configurable

gyuho:
## Why this should be merged

Currently, tx-pool lifetime is hard-coded to 3-hour and not configurable.

## How this works

Make it configurable via flag `tx-pool-lifetime`.

## How this was tested

Local.

## How is this documented

Flag.


ceyonur (approved):

gyuho:
Will revisit.

PR #1049, opened 2024-01-21
Fix bloom reset

ceyonur:
## Why this should be merged

There is a bug that prevents pending txs to be added to bloom filter after reset.

## How this works

Change variables from pendingTxs to iterated txs 

## How this was tested

Added UT

## How is this documented

No need

### DIFF
--- a/plugin/evm/gossip_test.go
+++ b/plugin/evm/gossip_test.go
@@ -94,7 +94,9 @@ func setupPoolWithConfig(t *testing.T, config *params.ChainConfig, fundedAddress
 	}
 	chain, err := core.NewBlockChain(diskdb, core.DefaultCacheConfig, gspec, engine, vm.Config{}, common.Hash{}, false)
 	require.NoError(t, err)
-	pool := txpool.NewTxPool(txpool.DefaultConfig, config, chain)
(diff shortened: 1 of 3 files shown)


plugin/evm/gossip_test.go
```diff
@@ -24,3 +36,67 @@ func TestGossipEthTxMarshaller(t *testing.T) {
...
+	}
+
+	// Notify VM about second eth txs batch
+	batch2 := ethTxs[5:]
+	errs = txPool.AddRemotesSync(batch2)
+	for _, err := range errs {
+		require.NoError(err, "failed adding subnet-evm tx to remote mempool")
+	}
+	time.Sleep(50 * time.Millisecond)
+	cancel()
```
darioush, line 80: is there a way to avoid this pattern, since it may lead to flakes?

darioush (approved):

plugin/evm/gossip_test.go
ceyonur: I intended to use VM, but ethTxPool was exported/not possible to tweak maxCount with bloom parameters. I just increased it to 1 sec to be sure, if this flakes we can try to change the actual ethTxPool structure.

PR #1061, opened 2024-01-23
`ci`: Add shellcheck step to lint job

marun:
# #lintallthethings!

We lint code, why not shell scripts? In addition to simplifying the avoidance of common shell errors, the [shellcheck wiki](https://www.shellcheck.net/wiki) has lots of documentation for all the error types it checks for.

### Testing

In addition to CI, local testing on arm64 linux and macos verified the logic for downloading and using a local (vs one in the path) shellcheck.

Related: [AvalancheGo PR](https://github.com/ava-labs/avalanchego/pull/2650)


joshua-kim (dismissed):

cmd/evm/transition-test.sh
ceyonur, line 1: this script is fetched from upstream repo. I wonder if we should exclude the spellcheck for any upstream script.

metrics/validate.sh
ceyonur, line 1: same comment about upstream scripts.

ceyonur (commented):
some scripts are from upstream repo, maybe we should exclude them from spellcheck. other changes LGTM.

ceyonur (commented):
some scripts are from upstream repo, maybe we should exclude them from spellcheck. other changes LGTM.

darioush (approved):

ceyonur (approved):

PR #1067, opened 2024-01-25
Align with coreth

darioush:
## Why this should be merged
This PR aligns cosmetic and minor differences with coreth. Non-cosmetic changes are highlighted with PR comments.

## How this works
Takes code from coreth, using upstream go-ethereum as reference for style / wspace where applicable.

## How this was tested
CI

## How is this documented
N/A


eth/state_accessor.go
```diff
@@ -68,6 +68,7 @@ var noopReleaser = tracers.StateReleaseFunc(func() {})
 //     provided, it would be preferable to start from a fresh state, if we have it
 //     on disk.
 func (eth *Ethereum) StateAtBlock(ctx context.Context, block *types.Block, reexec uint64, base *state.StateDB, readOnly bool, preferDisk bool) (statedb *state.StateDB, release tracers.StateReleaseFunc, err error) {
+	reexec = 0 // Do not support re-executing historical blocks to grab state
```
darioush, line 71: This changes API behavior following a change we introduced in coreth

eth/tracers/internal/tracetest/calltrace_test.go
```diff
@@ -45,9 +45,6 @@ import (
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/common/math"
 	"github.com/ethereum/go-ethereum/rlp"
-
-	// Force-load native, to trigger registration
-	_ "github.com/ava-labs/subnet-evm/eth/tracers/native"
```
darioush, line 50: This is not present in coreth or geth, so I don't believe it's needed

plugin/evm/gossiper.go
```diff
@@ -474,6 +477,8 @@ func (h *GossipHandler) HandleEthTxs(nodeID ids.NodeID, msg message.EthTxsGossip
 			)
 			if err == txpool.ErrAlreadyKnown {
 				h.stats.IncEthTxsGossipReceivedKnown()
+			} else {
+				h.stats.IncEthTxsGossipReceivedError()
```
darioush, line 481: This is a new stat in subnet-evm, but reduces the diff so I figured to take it here.

core/bench_test.go
```diff
@@ -238,7 +218,7 @@ func makeChainForBench(db ethdb.Database, full bool, count uint64) {
 	var hash common.Hash
 	for n := uint64(0); n < count; n++ {
 		header := &types.Header{
-			Coinbase:    constants.BlackholeAddr,
+			Coinbase:    common.Address{},
```
ceyonur, line 221: hmm this might fail benchs as we do have a coinbase verification in subnet-evm.

darioush: The benchmarks are currently failing due to another issue (passing genesis), which I suggest we patch along with the update from upstream.
Since we only call `chain.GetHeaderByNumber` it actually doesn't verify the coinbase

accounts/abi/bind/backends/simulated.go
```diff
@@ -115,10 +115,8 @@ func NewSimulatedBackendWithDatabase(database ethdb.Database, alloc core.Genesis
 		Alloc:    alloc,
 	}
 	cacheConfig := &core.CacheConfig{}
-	blockchain, err := core.NewBlockChain(database, cacheConfig, &genesis, dummy.NewCoinbaseFaker(), vm.Config{}, common.Hash{}, false)
-	if err != nil {
-		panic(fmt.Sprintf("failed to create simulated blockchain: %v", err))
-	}
+	blockchain, _ := core.NewBlockChain(database, cacheConfig, &genesis, dummy.NewCoinbaseFaker(), vm.Config{}, common.Hash{}, false)
```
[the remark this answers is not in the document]
darioush: it's probably not okay to ignore the error, but it's the way the code is in upstream.

PR #1068, opened 2024-01-25
Add error handling to Warp

minghinmatthewlam:
## Why this should be merged
Additional warp error handling

## How this works
- When initializing warp off chain messages, we check whether the `SourceChainID` and `NetworkID` are correct.
- For signature requests through `AppRequest` if the node does not have the Warp message, it'll return an empty signature without an error, and ends up with log message saying `failed to decompress signature`, which is misleading implying getting the signature was okay and that the node did find the Warp message.

## How this was tested
n/a

## How is this documented
n/a


warp/backend.go
```diff
@@ -94,6 +94,14 @@ func (b *backend) initOffChainMessages(offchainMessages [][]byte) error {
 			return fmt.Errorf("%w at index %d: %w", errParsingOffChainMessage, i, err)
 		}
 
+		if unsignedMsg.NetworkID != b.networkID {
+			return fmt.Errorf("%w at index %d: network ID mismatch", avalancheWarp.ErrWrongNetworkID, i)
+		}
+
+		if unsignedMsg.SourceChainID != b.sourceChainID {
+			return fmt.Errorf("%w at index %d: source chain ID mismatch", avalancheWarp.ErrWrongSourceChainID, i)
```
ceyonur, line 102: I think we can drop the second err description
```suggestion
			return fmt.Errorf("%w at index %d", avalancheWarp.ErrWrongSourceChainID, i)
```

warp/backend.go
```diff
@@ -94,6 +94,14 @@ func (b *backend) initOffChainMessages(offchainMessages [][]byte) error {
 			return fmt.Errorf("%w at index %d: %w", errParsingOffChainMessage, i, err)
 		}
 
+		if unsignedMsg.NetworkID != b.networkID {
+			return fmt.Errorf("%w at index %d: network ID mismatch", avalancheWarp.ErrWrongNetworkID, i)
```
ceyonur, line 98:
```suggestion
			return fmt.Errorf("%w at index %d", avalancheWarp.ErrWrongNetworkID, i)
```

darioush (approved):

PR #1078, opened 2024-02-02
Add Warp Audits

JacobEverly:
## Why this should be merged
Add Audits for warp


audits/Avalanche Warp Messaging Audit Report OpenZeppelin.pdf
ceyonur, line 1: Can we move these to `docs/audits`? Otherwise LGTM.

ceyonur (changes requested):
Can we move these to `docs/audits`? Otherwise LGTM.

ceyonur (approved):

PR #1079, opened 2024-02-02
Use typed error when reverting in EthereumAPI

cam-schultz:
## Why this should be merged
When calling a contract via the `EthereumAPI`, there's no type-safe way to check if an error is due to a revert or something else. The client could of course check for the string "execution reverted", but that runs the risk of becoming mismatched. The proposed change uses an error var that's defined once, eliminating that risk.

## How this works
Uses the existing error `vmerrs.ErrExecutionReverted` so that the client can use `errors.Is` on the returned value. The returned error string is unchanged.

See https://go.dev/play/p/XuGfk6VdmJ5 for an example. I can also add a unit test.

## How this was tested
CI, above Go playground testing

## How is this documented
N/A


cam-schultz:
Will propose this change in `go-ethereum` directly.

PR #1084, opened 2024-02-06
Add line of stars to comment header

darioush:
## Why this should be merged
Makes all the avalanche headers 10 lines long, so it's easy to handle them in a script

## How this works
Adds comments

## How this was tested
CI

## How is this documented
N/A


ceyonur (approved):

ceyonur:
Should we put up this into linter rules?

PR #1086, opened 2024-02-07
eth/filters, eth/api: update rpc numbers in unfinalized query mode

darioush:
## Why this should be merged
Seems in geth the rpc numbers (when resolved to real blocks) semantically follow a pattern of
Finalized <= Safe <= Latest <= Pending

Since we don't have Pending, we can map Latest/Pending to the chain's head in unfinalized query mode
Without unfinalized query mode, we can use last accepted for all.

## How this works
Updates semantics as above
Aligns code with upstream & adds FinalizedBlockNumber as an alias for AcceptedBlockNumber (esp in tests)

## How this was tested
CI

## How is this documented
not yet

### DIFF
--- a/eth/api.go
+++ b/eth/api.go
@@ -197,7 +197,11 @@ func (api *DebugAPI) DumpBlock(blockNr rpc.BlockNumber) (state.Dump, error) {
 	}
 	var header *types.Header
 	if blockNr.IsAccepted() {
-		header = api.eth.LastAcceptedBlock().Header()
+		if blockNr.IsLatest() && api.eth.APIBackend.IsAllowUnfinalizedQueries() {
+			header = api.eth.blockchain.CurrentHeader()
+		} else {
+			header = api.eth.LastAcceptedBlock().Header()
+		}
 	} else {
 		block := api.eth.blockchain.GetBlockByNumber(uint64(blockNr))
 		if block == nil {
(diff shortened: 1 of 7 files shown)


rpc/types.go
```diff
@@ -72,11 +72,12 @@ type jsonWriter interface {
...
 
 const (
-	SafeBlockNumber     = BlockNumber(-4)
-	AcceptedBlockNumber = BlockNumber(-3)
-	LatestBlockNumber   = BlockNumber(-2)
-	PendingBlockNumber  = BlockNumber(-1)
-	EarliestBlockNumber = BlockNumber(0)
+	SafeBlockNumber      = BlockNumber(-4)
+	FinalizedBlockNumber = BlockNumber(-3)
+	AcceptedBlockNumber  = BlockNumber(-3)
```
ceyonur, line 77: do we want these to be same?

ceyonur: Should we just remove AcceptedBlockNumber and use Finalized?

eth/api.go
```diff
@@ -197,7 +197,11 @@ func (api *DebugAPI) DumpBlock(blockNr rpc.BlockNumber) (state.Dump, error) {
 	}
 	var header *types.Header
 	if blockNr.IsAccepted() {
-		header = api.eth.LastAcceptedBlock().Header()
+		if blockNr.IsLatest() && api.eth.APIBackend.IsAllowUnfinalizedQueries() {
```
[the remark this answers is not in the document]
darioush: I think this amount of duplication is fine, since I don't think we should add information about the backend or configs to types/rpc

rpc/types.go
darioush: I also thought about that, but I think we could leave that to a different PR, mostly I added FinalizedBlockNumber back to allow tests to be the same as they are in upstream

ceyonur (approved):

PR #1090, opened 2024-02-11
Geth v1.12.2

darioush:
## Why this should be merged
Updates shared code with go-ethereum

- Had to apply some changes to coreth (even as indirect dependency) to get it to compile https://github.com/ava-labs/coreth/tree/geth-v1.12.2-compile
- TODO (future PR): enable periodic update for eip 4844 fee
- TODO (future PR): enable blob txs in gossip

## How this works
Takes recent changes from the upstream repository, v1.12.0..v1.12.2. Steps:

1. Add upstream as a remote and fetch tags eg:
- git remote add ethereum git@github.com:ethereum/go-ethereum.git
- git fetch ethereum --tags
- (I didn't set up a config to fetch the tags with a prefix, but we can do that if we prefer)
2. run ./scripts/format_as_upstream
- This will rename the repository as "ethereum/go-ethereum"
3. git pull ethereum v1.12.0 -s ours --allow-unrelated-histories
4. git pull ethereum v1.12.2
5. Resolve merge conflicts & update code
6. ./scripts/build_test.sh (make sure tests pass locally)
7. ./scripts/format_as_fork.sh v1.12.2
- Note the version specified as an argument should be the newer version.
- (unfortunately this script has a small bug, which I didn't fix yet, so it will not rename these comments https://github.com/ava-labs/subnet-evm/pull/1090/commits/d08dfe28e1abb83b2288f6baa6fd193626b52ddf)

## How this was tested
CI, manual testing pending

## How is this documented
Not yet


plugin/evm/vm_test.go
```diff
@@ -2082,7 +2082,7 @@ func TestBuildSubnetEVMBlock(t *testing.T) {
 		}
 		txs[i] = signedTx
 	}
-	errs := vm.txPool.AddRemotes(txs)
+	errs := vm.txPool.AddRemotesSync(txs)
```
darioush, line 2085: Not sure if sync/async matters here, used sync to avoid creating a wrapper

plugin/evm/vm_test.go
```diff
@@ -2644,7 +2644,7 @@ func TestFeeManagerChangeFee(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	err = vm.txPool.AddRemote(signedTx2)
+	err = vm.txPool.AddRemotesSync([]*types.Transaction{signedTx2})[0]
```
darioush, line 2647: same as above

core/txpool/legacypool/legacypool.go
```diff
@@ -0,0 +1,2148 @@
...
+	return pool.all.GetLocal(hash) != nil
+}
+
+// RemoveTx removes a single transaction from the queue, moving all subsequent
+// transactions back to the future queue.
+func (pool *LegacyPool) RemoveTx(hash common.Hash) {
+	pool.mu.Lock()
+	defer pool.mu.Unlock()
+
+	// XXX: check that passing true to unreserve is correct
```
darioush, line 1187: We should verify this

ceyonur (commented):
Mostly LGTM, just few questions.

I also put up a nit PR here: https://github.com/ava-labs/subnet-evm/pull/1099

core/txpool/legacypool/legacypool.go
darioush: turns out to be unused.

core/state/pruner/pruner.go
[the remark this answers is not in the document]
darioush: added

core/txpool/blobpool/blobpool.go
```diff
@@ -0,0 +1,1649 @@
...
+//     fee cap and data fee cap, there's no singular parameter to create a total
+//     price ordering on. What's more, since the base fee and blob fee can move
+//     independently of one another, there's no pre-defined way to combine them
+//     into a stable order either. This leads to a multi-dimensional problem to
+//     solve after every block.
+//
+//   - The first observation is that comparing 1559 base fees or 4844 blob fees
+//     needs to happen in the context of their dynamism. Since these fees jump
+//     up or down in ~1.125 multipliers (at max) across blocks, comparing fees
+//     in two transactions should be based on log1.125(fee) to eliminate noise.
```
aaronbuchwald, line 262: Is this true that they will increase at max of 1.125 of the current base fee? I think that may be incorrect for EIP-1559

aaronbuchwald: cc @abi87 who just pointed this out to me today

abi87: It seems to me it was the original intention (see https://github.com/ethereum/EIPs/issues/1559) but that is not strictly true anymore in https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md nor in the implementation.
To be clear, I understand fee changes are still capped, but not of 1/changeDemon. changeDemon seems a smoothing parameter to control fee dynamics.

PR #1099, opened 2024-02-19
Geth v1.12.2 nits

ceyonur:
Mostly cosmetic changes to reduce diff between upstream


core/state/state_object.go
```diff
@@ -72,11 +71,6 @@ func (s Storage) Copy() Storage {
 // - Account values as well as storages can be accessed and modified through the object.
 // - Finally, call commit to return the changes of storage trie and update account data.
 type stateObject struct {
-	// dataLock protects the [data] field to prevent a race condition
-	// in the transaction pool tests. TODO remove after re-implementing
-	// tx pool to be synchronous.
-	dataLock sync.RWMutex
```
darioush, line 78: Are we sure this is safe to remove?

cmd/utils/flags.go
```diff
@@ -1,89 +0,0 @@
...
-		}
-		// Mark the flag if it's set
-		if ctx.IsSet(flag.Names()[0]) {
-			set = append(set, "--"+name)
-		}
-	}
-	if len(set) > 1 {
-		Fatalf("Flags %v can't be used at the same time", strings.Join(set, ", "))
-	}
-}
```
darioush, line 89: I added this so we can avoid copying the entire cmd/utils package when doing the geth update.
This is because the entire cmd/utils package imports a lot of geth code.
So I think it's better to leave this in place with only the 2 functions we use from this package.

params/config.go
```diff
@@ -778,9 +695,9 @@ func (c *ChainConfig) rules(num *big.Int, timestamp uint64) Rules {
 	}
 }
 
-// AvalancheRules returns the Avalanche modified rules to support Avalanche
+// Rules returns the Avalanche modified rules to support Avalanche
 // network upgrades
-func (c *ChainConfig) AvalancheRules(blockNum *big.Int, timestamp uint64) Rules {
+func (c *ChainConfig) Rules(blockNum *big.Int, timestamp uint64) Rules {
```
darioush, line 700: if we rename this method to `Rules`, like it is in upstream, should we also take the same bool `isMerge` parameter?
Honestly it's kinda weird that they have the isMerge parameter, but I think if the name matches the signature should also match, otherwise perhaps it's better to leave it as AvalancheRules.

ethclient/subnetevmclient/subnet_evm_client.go
```diff
@@ -105,6 +105,11 @@ func (ec *Client) GetProof(ctx context.Context, account common.Address, keys []s
 		StorageProof []storageResult `json:"storageProof"`
 	}
 
+	// Avoid keys being 'null'.
+	if keys == nil {
```
darioush, line 109: Why is this added?

darioush: nvm I see it's from upstream.

core/state/state_object.go
ceyonur: I think these were fixed way before with https://github.com/ava-labs/coreth/commit/8e32d4371af9e7e8d12c64451bf0dcefc14d4e88#diff-64a06d0cea30e12d6001fb322e583930c1cf3879bca4e8019fa89499ed9f7de4

Also seems like this was only an issue in tests. cc @aaronbuchwald

ceyonur: I also added `-race` flag to UTs to catch these.

params/config.go
ceyonur: IMO we should not use that `isMerge` param. but I also slightly think it is probably better to use same `Rules` function. I can revert that if you think we should be using `AvalancheRules` instead.

Issue #1112, opened 2024-03-04
[code sync] Update code to go-ethereum 1.13.2


darioush:
This issue represents the work necessary to take upstream changes from v1.12.2 to v.1.13.2

PR #1116, opened 2024-03-06
Geth v1.13.2

darioush:
## Why this should be merged
https://github.com/ava-labs/subnet-evm/issues/1112

## How this works
Takes code from upstream

## How this was tested
[x] CI
[ ] Testnet

## How is this documented
TODO: update deprecated flag


core/blockchain.go
```diff
@@ -464,6 +506,17 @@ func (bc *BlockChain) dispatchTxUnindexer() {
 		return
 	}
 	defer sub.Unsubscribe()
+	log.Info("Initialized transaction unindexer", "limit", txLookupLimit)
+
+	// XXX
```
darioush, line 511: not sure why the test fails when I uncomment the below.
Not sure we need to take this change either.

core/blockchain.go
```diff
@@ -59,6 +61,8 @@ import (
 )
 
 var (
+	chainInfoGauge = metrics.NewRegisteredGaugeInfo("chain/info", nil)
```
darioush, line 64: not sure we need this either

trie/triedb/pathdb/layertree.go
```diff
@@ -142,7 +142,32 @@ func (tree *layerTree) cap(root common.Hash, layers int) error {
 			return err
 		}
 		// Replace the entire layer tree with the flat base
-		tree.layers = map[common.Hash]layer{base.rootHash(): base}
+		// tree.layers = map[common.Hash]layer{base.rootHash(): base}
+		// XXX
```
darioush, line 146: Seems geth adopts a different semantics for commit than what we rely on in blockchain.go
I implemented this version which allows uncommitted layers to remain on top instead of flattening everything.

miner/worker.go
```diff
@@ -95,10 +98,11 @@ type worker struct {
...
 	// Subscriptions
-	mux      *event.TypeMux // TODO replace
-	mu       sync.RWMutex   // The lock used to protect the coinbase and extra fields
-	coinbase common.Address
-	clock    *mockable.Clock // Allows us mock the clock for testing
+	mux        *event.TypeMux // TODO replace
+	mu         sync.RWMutex   // The lock used to protect the coinbase and extra fields
+	coinbase   common.Address
+	clock      *mockable.Clock // Allows us mock the clock for testing
+	beaconRoot *common.Hash    // XXX: this is not set anywhere
```
darioush, line 105: I think we should support setting beaconRoot somehow during block building, but not sure what we would put in it.

metrics/prometheus/prometheus.go
```diff
@@ -73,6 +73,8 @@ func (g gatherer) Gather() ([]*dto.MetricFamily, error) {
 					},
 				}},
 			})
+		case metrics.GaugeInfo:
+			// XXX
```
darioush, line 77: not sure what to put here, not sure we need this either, see the comment for core/blockchain.go.

core/state/statedb.go
```diff
@@ -1088,59 +1014,125 @@ func (s *StateDB) clearJournalAndRefund() {
...
-// deleteStorage iterates the storage trie belongs to the account and mark all
-// slots inside as deleted.
-func (s *StateDB) deleteStorage(addr common.Address, addrHash common.Hash, root common.Hash) (bool, map[common.Hash][]byte, *trienode.NodeSet, error) {
-	start := time.Now()
+// fastDeleteStorage is the function that efficiently deletes the storage trie
+// of a specific account. It leverages the associated state snapshot for fast
+// storage iteration and constructs trie node deletion markers by creating
+// stack trie with iterated slots.
+func (s *StateDB) fastDeleteStorage(addrHash common.Hash, root common.Hash) (bool, common.StorageSize, map[common.Hash][]byte, *trienode.NodeSet, error) {
+	iter, _ := s.snap.StorageIterator(addrHash, common.Hash{}) // XXX
```
darioush, line 1022: don't think any test hits this path, should verify

core/blockchain_repair_test.go
```diff
@@ -544,17 +563,29 @@ func testRepair(t *testing.T, tt *rewindTest, snapshots bool) {
...
 	var sideblocks types.Blocks
 	if tt.sidechainBlocks > 0 {
-		sideblocks, _, _ = GenerateChain(gspec.Config, gspec.ToBlock(), engine, rawdb.NewMemoryDatabase(), tt.sidechainBlocks, 10, func(i int, b *BlockGen) {
+		genDb := rawdb.NewMemoryDatabase()
+		gspec.MustCommit(genDb, trie.NewDatabase(genDb, nil))
+		sideblocks, _, err = GenerateChain(gspec.Config, gspec.ToBlock(), engine, genDb, tt.sidechainBlocks, 10, func(i int, b *BlockGen) {
 			b.SetCoinbase(common.Address{0x01})
+			tx, err := types.SignTx(types.NewTransaction(b.TxNonce(addr1), common.Address{0x01}, big.NewInt(10000), params.TxGas, common.Big1, nil), signer, key1)
+			require.NoError(err)
+			b.AddTx(tx)
```
darioush, line 572: Had to make these changes since in geth, the roots are different because of the coinbase payment.
Here we just keep on with the empty trie and that doesn't work for the pathdb (doesn't expect repeated roots)

core/types/block.go
```diff
@@ -99,6 +99,9 @@ type Header struct {
 
 	// ExcessBlobGas was added by EIP-4844 and is ignored in legacy headers.
 	ExcessBlobGas *uint64 `json:"excessBlobGas" rlp:"optional"`
+
+	// ParentBeaconRoot was added by EIP-4788 and is ignored in legacy headers.
+	ParentBeaconRoot *common.Hash `json:"parentBeaconBlockRoot" rlp:"optional"`
```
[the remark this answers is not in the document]
darioush: Yeah I was thinking it could be used for something like that. We can delete it until the next fork so I suggest we leave it for now, even if its empty. We did omit withdrawals but I'm not sure that was the best idea for compatibility.

core/state/statedb.go
darioush: now covered by TestDeleteStorage

trie/triedb/hashdb/database.go
```diff
@@ -736,7 +736,6 @@ func (db *Database) Size() (common.StorageSize, common.StorageSize) {
 func (db *Database) Close() error {
 	if db.cleans != nil {
 		db.cleans.Reset()
-		db.cleans = nil
```
ceyonur: why did we remove this?

darioush: It causes a race in testing. Applied to upstream as well: https://github.com/ethereum/go-ethereum/pull/29309

darioush:
replaced by https://github.com/ava-labs/subnet-evm/pull/1141

PR #1123, opened 2024-03-14
Genesis chainconfig cleanup

ceyonur:
## Why this should be merged

1- This adds a new upgrade config `networkUpgradeOverrides` to override any avalanche network upgrade. With this upgrade config and `skip-upgrade-check` flag set to true, one can modify the timestamp of activated upgrade. This will come handy if a subnet has missed the upgrade (failed to update nodes) and as a result their chain has become unsyncable with newer versions. In this case `networkUpgradeOverrides` can be used to set a later time for upgrades. I.e:
`upgrade.json`:
```
{
  "networkUpgradeOverrides": {
    "durangoTimestamp": 1710415083
  }
}
```
2- Default Network Upgrades are introduced. `nil` and `blank` values in network upgrades (eth forks + avalanche upgrades) will now default to values. Meaning that it won't be possible to deactivate an existing upgrade via `nil` values. 

3- This PR also removes `OptionalNetworkUpgrades` since now we can modify all NetworkUpgrades.

## How this was tested

Unit tests + local tests

## How is this documented

it will be documented in docs repo.

### DIFF
--- a/core/state_processor.go
+++ b/core/state_processor.go
@@ -27,6 +27,7 @@
 package core
 
 import (
+	"encoding/json"
 	"fmt"
 	"math/big"
 
@@ -197,7 +198,15 @@ func ApplyPrecompileActivations(c *params.ChainConfig, parentTimestamp *uint64, 
 				// since Suicide will be committed after the reconfiguration.
 				statedb.Finalise(true)
 			} else {
-				log.Info("Activating new precompile", "name", module.ConfigKey, "config", activatingConfig)
+				var printIntf interface{}
+				marshalled, err := json.Marshal(activatingConfig)
+				if err == nil {
+					printIntf = string(marshalled)
+				} else {
+					printIntf = activatingConfig
+				}
+
+				log.Info("Activating new precompile", "name", module.ConfigKey, "config", printIntf)
 				// Set the nonce of the precompile's address (as is done when a contract is created) to ensure
 				// that it is marked as non-empty and will not be cleaned up when the statedb is finalized.
 				statedb.SetNonce(module.Address, 1)
--- a/params/config.go
+++ b/params/config.go
@@ -477,12 +477,12 @@ func checkForks(forks []fork, blockFork bool) error {
 			// Next one must be higher number
 			if lastFork.timestamp == nil && cur.timestamp != nil {
 				return fmt.Errorf("unsupported fork ordering: %v not enabled, but %v enabled at %v",
-					lastFork.name, cur.name, cur.timestamp)
+					lastFork.name, cur.name, *cur.timestamp)
 			}
 			if lastFork.timestamp != nil && cur.timestamp != nil {
 				if *lastFork.timestamp > *cur.timestamp {
 					return fmt.Errorf("unsupported fork ordering: %v enabled at %v, but %v enabled at %v",
-						lastFork.name, lastFork.timestamp, cur.name, cur.timestamp)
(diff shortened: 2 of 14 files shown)


params/config.go
```diff
@@ -263,6 +257,42 @@ func (c *ChainConfig) Description() string {
 	return banner
 }
 
+func (c *ChainConfig) SetNetworkUpgradeDefaults() {
```
ceyonur, line 260: This will remove the capability of giving `nil` values for upgrades/forks to disable them. I don't think the current codebase supports deactivating any of those, but if we want to maintain the "full" backward-compatibility we can expand the `*uint64/*bigInt` types to some struct with which we can deduct information about `nil` vs `value` vs `blank` in json files. This would require a bigger change in the code and would increase the diff between the upstream.

params/network_upgrades.go
```diff
@@ -4,63 +4,98 @@
...
 		return newTimestampCompatError("Durango fork block timestamp", m.DurangoTimestamp, newcfg.DurangoTimestamp)
 	}
-	if isForkTimestampIncompatible(m.CancunTime, newcfg.CancunTime, time) {
-		return newTimestampCompatError("Cancun fork block timestamp", m.CancunTime, m.CancunTime)
-	}
 	return nil
 }
 
-func (m *MandatoryNetworkUpgrades) mandatoryForkOrder() []fork {
+func (m *NetworkUpgrades) forkOrder() []fork {
```
darioush, line 40: nit: should we rename receivers to `n`?

core/state_processor.go
```diff
@@ -197,7 +198,15 @@ func ApplyPrecompileActivations(c *params.ChainConfig, parentTimestamp *uint64,
...
-				log.Info("Activating new precompile", "name", module.ConfigKey, "config", activatingConfig)
+				var printIntf interface{}
+				marshalled, err := json.Marshal(activatingConfig)
+				if err == nil {
+					printIntf = string(marshalled)
+				} else {
+					printIntf = activatingConfig
+				}
+
+				log.Info("Activating new precompile", "name", module.ConfigKey, "config", printIntf)
```
darioush, line 209: should we require a `.String()` on the config instead?

plugin/evm/vm_test.go
```diff
@@ -2819,7 +2819,7 @@ func TestRewardManagerPrecompileSetRewardAddress(t *testing.T) {
 	data, err := rewardmanager.PackSetRewardAddress(testAddr)
 	require.NoError(t, err)
 
-	gas := 21000 + 240 + rewardmanager.SetRewardAddressGasCost // 21000 for tx, 240 for tx data
+	gas := 21000 + 240 + rewardmanager.SetRewardAddressGasCost + rewardmanager.RewardAddressChangedEventGasCost // 21000 for tx, 240 for tx data
```
darioush, line 2822: why are these changed?

ceyonur: because VMs will now have non-nil upgrades (if default is defined). This was running with only `subnetEVM` upgrade (Durango was nil in chainconfig) but since now we apply defaults for nils, this test is upgraded to Durango fork. In Durango precompiles have increased costs (because of events)

core/state_processor.go
ceyonur: I thought about it but that would require a change to all precompiles and i don't if we there is any gain with custom string configs vs plain marshalling them to json.

darioush (approved):

PR #1130, opened 2024-03-22
bump avago v1.11.3

ceyonur:
## Why this should be merged

Bumps avalanchego to v1.11.3

## How this works

Bumps avalanchego to v1.11.3

## How this was tested

CI tests


darioush (dismissed):

plugin/evm/vm.go
```diff
@@ -701,6 +701,7 @@ func (vm *VM) initBlockBuilding() error {
 		ethTxPushGossiper, err = gossip.NewPushGossiper[*GossipEthTx](
 			ethTxGossipMarshaller,
 			ethTxPool,
+			vm.validators,
 			ethTxGossipClient,
 			ethTxGossipMetrics,
 			pushGossipParams,
```
StephenButtolph, line 707: Should we be adding the Stake percentage into the BranchingFactor here?

warp/aggregator/aggregator_test.go
```diff
@@ -23,7 +23,7 @@ func newValidator(t testing.TB, weight uint64) (*bls.SecretKey, *avalancheWarp.V
 	pk := bls.PublicFromSecretKey(sk)
 	return sk, &avalancheWarp.Validator{
 		PublicKey:      pk,
-		PublicKeyBytes: bls.PublicKeyToBytes(pk),
+		PublicKeyBytes: bls.PublicKeyToUncompressedBytes(pk),
```
StephenButtolph, line 26: I think this would align with the prior `PublicKeyToBytes`
```suggestion
		PublicKeyBytes: bls.PublicKeyToCompressedBytes(pk),
```

plugin/evm/vm.go
ceyonur: done

warp/aggregator/aggregator_test.go
ceyonur: applied

darioush (approved):

StephenButtolph (approved):
lgtm

PR #1131, opened 2024-03-25
Tx indexing fix

ceyonur:


core/test_blockchain.go
```diff
@@ -1647,3 +1649,49 @@ func TestStatefulPrecompiles(t *testing.T, create func(db ethdb.Database, gspec
...
+	} else {
+		tailValue = *expectedTail
+
+		require.Eventually(
+			func() bool {
+				stored := *rawdb.ReadTxIndexTail(db)
+				return tailValue == stored
+			},
+			10*time.Second, 100*time.Millisecond, "expected tail to be %d eventually", tailValue)
+	}
```
darioush, line 1678: nit
```suggestion
	if expectedTail == nil {
		require.Nil(rawdb.ReadTxIndexTail(db))
	} else {
		tailValue := *expectedTail

		require.Eventually(
			func() bool {
				stored := *rawdb.ReadTxIndexTail(db)
				return tailValue == stored
			},
			10*time.Second, 100*time.Millisecond, "expected tail to be %d eventually", tailValue)
	}
```

core/test_blockchain.go
```diff
@@ -1647,3 +1649,49 @@ func TestStatefulPrecompiles(t *testing.T, create func(db ethdb.Database, gspec
...
 	checkBlockChainState(t, blockchain, gspec, chainDB, create, checkState)
 }
+
+func CheckTxIndices(t *testing.T, expectedTail *uint64, head uint64, db ethdb.Database, allowNilBlocks bool) {
+	var tailValue uint64
+	if expectedTail == nil {
+		tailValue = 0
+	} else {
+		tailValue = *expectedTail
+	}
```
darioush, line 1659: nit
```suggestion
	if expectedTail != nil {
		tailValue = *expectedTail
	}
```

plugin/evm/syncervm_test.go
```diff
@@ -456,6 +480,15 @@ func testSyncerVM(t *testing.T, vmSetup *syncVMSetup, test syncTest) {
 		}
 	})
 
+	if syncerVM.ethConfig.TxLookupLimit != 0 {
+		tail := syncerVM.blockChain.LastAcceptedBlock().NumberU64() - syncerVM.ethConfig.TxLookupLimit + 1
+
+		syncerVM.blockChain.DrainAcceptorQueue()
+		time.Sleep(50 * time.Millisecond) // Wait for indices initialisation
```
darioush, line 487: is this still needed with the `require.Eventually` added to the check fn?

plugin/evm/syncervm_test.go
```diff
@@ -434,6 +447,17 @@ func testSyncerVM(t *testing.T, vmSetup *syncVMSetup, test syncTest) {
 	require.True(syncerVM.blockChain.HasState(syncerVM.blockChain.LastAcceptedBlock().Root()), "unavailable state for last accepted block")
 	assertSyncPerformedHeights(t, syncerVM.chaindb, map[uint64]struct{}{retrievedSummary.Height(): {}})
 
+	lastNumber := syncerVM.blockChain.LastAcceptedBlock().NumberU64()
+	// check the last block is indexed
+	block := rawdb.ReadBlock(syncerVM.chaindb, rawdb.ReadCanonicalHash(syncerVM.chaindb, lastNumber), lastNumber)
+	if block.Transactions().Len() == 0 {
+		return
+	}
```
darioush, line 455: why do we return here?

core/rawdb/accessors_state_sync.go
```diff
@@ -176,3 +176,19 @@ func NewSyncPerformedIterator(db ethdb.Iteratee) ethdb.Iterator {
 func UnpackSyncPerformedKey(key []byte) uint64 {
 	return binary.BigEndian.Uint64(key[len(syncPerformedPrefix):])
 }
+
+// GetLatestSyncPerformed returns the latest block number state synced performed to.
+func GetLatestSyncPerformed(db ethdb.Iteratee) uint64 {
+	// XXX: assuming this is not sorted
```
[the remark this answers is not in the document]
ceyonur: hmm if they're sorted we probably won't even need to unpack all keys, but only the last key.

Since this is checked in startup only we can leave this as it is if we even have a slightest doubt about their ordering.

core/test_blockchain.go
```diff
@@ -1647,3 +1649,49 @@ func TestStatefulPrecompiles(t *testing.T, create func(db ethdb.Database, gspec
...
+	checkTxIndicesHelper(t, expectedTail, tailValue, head, head, db, allowNilBlocks)
+}
+
+func checkTxIndicesHelper(t *testing.T, expectedTail *uint64, indexedFrom uint64, indexedTo uint64, head uint64, db ethdb.Database, allowNilBlocks bool) {
```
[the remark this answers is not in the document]
ceyonur: added

plugin/evm/syncervm_test.go
ceyonur: ah good catch, no we should not.

plugin/evm/syncervm_test.go
ceyonur: hmm weird, removed.

core/rawdb/accessors_state_sync.go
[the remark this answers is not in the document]
darioush: The order is maintained by the db, let's just remove the `XXX` comment then?

darioush (dismissed):

darioush (approved):

PR #1135, opened 2024-03-28
E upgrade config

ceyonur:
## Why this should be merged

* Adds E Upgrade Config
* Maps Cancun activation to EUpgrade activation
* Changes default activation timestamps for non-production chains as well
  * 0 and nil timestamps default to Avalanche activations, however non-production chains were able to skip this behaviour. Now they will also default when 0 or nil provided.

## How this was tested

UTs + local testing

## How is this documented

Tooling and docs should reflect that genesis timestamps must be non-zero


core/state_transition.go
```diff
@@ -299,7 +299,7 @@ func (st *StateTransition) buyGas() error {
 		balanceCheck = balanceCheck.Mul(balanceCheck, st.msg.GasFeeCap)
 		balanceCheck.Add(balanceCheck, st.msg.Value)
 	}
-	if st.evm.ChainConfig().IsCancun(st.evm.Context.BlockNumber, st.evm.Context.Time) {
+	if st.evm.ChainConfig().IsCancun(st.evm.Context.Time) {
```
darioush, line 302: I think we should leave the block number parameter so future changes will apply with fewer merge conflicts from upstream.

core/genesis.go
```diff
@@ -213,8 +213,9 @@ func SetupGenesisBlock(
 	}
 	// Check whether the genesis block is already written.
 	hash := genesis.ToBlock().Hash()
+	hdr := genesis.ToBlock().Header()
 	if hash != stored {
-		return genesis.Config, common.Hash{}, &GenesisMismatchError{stored, hash}
+		return genesis.Config, common.Hash{}, &GenesisMismatchError{stored, hdr.Hash()}
```
darioush, line 218: is `hdr.Hash()` different than hash? what causes that?

plugin/evm/vm_upgrade_bytes_test.go
```diff
@@ -249,7 +256,7 @@ func TestVMStateUpgrade(t *testing.T) {
 		Code:          upgradedCode,
 	}
 
-	upgradeTimestamp := time.Unix(10, 0) // arbitrary timestamp to perform the network upgrade
+	upgradeTimestamp := time.Now() // arbitrary timestamp to perform the network upgrade
```
darioush, line 259: nit: should we specify the required upgrade (`DefaultGenesisTime`) + some arbitrary amount instead of time.Now?

ceyonur: yup changed it

darioush (approved):

PR #1136, opened 2024-03-28
move network upgrade code to file

ceyonur:
## Why this should be merged

Moves network upgrade code to it's own file. 

Fixes the Durango timestamp = 0 issue in https://github.com/ava-labs/avalanchego/issues/2877

## How this works

0 timestamps now defaults to the network defaults

## How this was tested

Locally tested

### DIFF
--- a/params/network_upgrades.go
+++ b/params/network_upgrades.go
@@ -6,19 +6,27 @@ package params
 import (
 	"fmt"
 	"reflect"
-	"strconv"
 
 	"github.com/ava-labs/avalanchego/version"
 	"github.com/ava-labs/subnet-evm/utils"
 )
 
+var (
+	errCannotBeNil = fmt.Errorf("timestamp cannot be nil")
+
+	TestNetworkUpgrades = NetworkUpgrades{
+		SubnetEVMTimestamp: utils.NewUint64(0),
+		DurangoTimestamp:   utils.NewUint64(0),
+	}
+)
+
 // NetworkUpgrades contains timestamps that enable network upgrades.
 // Avalanche specific network upgrades are also included here.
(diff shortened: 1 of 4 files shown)


params/network_upgrades.go
```diff
@@ -48,22 +56,27 @@ func (n *NetworkUpgrades) forkOrder() []fork {
 // This overrides deactivating the network upgrade by providing a timestamp of nil value.
 func (n *NetworkUpgrades) setDefaults(networkID uint32) {
 	defaults := getDefaultNetworkUpgrades(networkID)
-	if n.SubnetEVMTimestamp == nil {
+	// If the network upgrade is not set, set it to the default value.
+	// If the network upgrade is set to 0, we also treat it as nil and set it default.
+	// This is because in prior versions, upgrades were not modifiable and were directly set to their default values.
+	// Most of the tools and configurations just provide these as 0, so it is safer to treat 0 as nil and set to default
+	// to prevent premature activations of the network upgrades.
```
darioush, line 63:
```suggestion
	// to prevent premature activations of the network upgrades for live networks.
```

darioush: somehow specifying this only applies to testnet / mainnet networks, not local / testing

darioush (dismissed):

params/network_upgrades.go
ceyonur: this is because local + testing networks are activated at timestamp 0 already (there is no local or testing definition in https://github.com/ava-labs/avalanchego/blob/master/version/constants.go#L139-L146)

darioush (approved):

darioush (approved):

PR #1137, opened 2024-03-28
`ci`: Skip monitoring if secrets are not present

marun:
Metrics and logs can only be collected if the required secrets are present, but they will not be available to jobs executing against PRs for forks.


ceyonur (approved):

darioush (approved):

.github/workflows/tests.yml
```diff
@@ -127,18 +127,21 @@ jobs:
         shell: bash
         run: ./scripts/build.sh /tmp/e2e-test/avalanchego/plugins/srEXiWaHuhNyGwPUi444Tu47ZEDwxTWrbQiuD7FmgSAQ6X7Dy
       - name: Start prometheus
+        if: secrets.PROMETHEUS_ID != ''
```
ceyonur, line 130: hmm this file is not working https://github.com/ava-labs/subnet-evm/actions/runs/8471478509/workflow
I think it should be ${{ secrets.PROMETHEUS_ID }}

ceyonur (changes requested):
requesting changes in case auto-merge triggers somehow before fixing this

ceyonur (approved):

darioush (approved):

PR #1146, opened 2024-04-10
Geth v1.13.8

darioush:
## Why this should be merged
Update the code base to take improvements and fixes from upstream.

## How this works
Takes code from upstream

## How this was tested
[X] CI
[x] Testnet

## How is this documented
N/A


eth/ethconfig/config.go
```diff
@@ -56,9 +55,9 @@ var DefaultConfig = NewDefaultConfig()
 
 func NewDefaultConfig() Config {
 	return Config{
-		NetworkId:                 1,
+		// XXX: is this desirable 1 -> 0?
+		NetworkId:                 0, // enable auto configuration of networkID == chainID
```
darioush, line 59: Should we take this change or leave it as before?

rpc/websocket_test.go
```diff
@@ -125,6 +125,66 @@ func TestWebsocketLargeCall(t *testing.T) {
...
+			}
+		}
+	}
+	ptr := func(v int64) *int64 { return &v }
+
+	testLimit(ptr(-1)) // Should be ignored (use default)
+	testLimit(ptr(0))  // Should be ignored (use default)
+	testLimit(nil)     // Should be ignored (use default)
+	testLimit(ptr(200))
+	testLimit(ptr(wsDefaultReadLimit + 1024))
```
darioush, line 185: I reduced this from upstream (wsDefaultReadLimit * 2) since the pong message would time out when run with the race detector in CI.

log/handler.go
```diff
@@ -0,0 +1,196 @@
...
+		if attr.Value.Kind() == slog.KindTime {
+			if logfmt {
+				return slog.String("t", attr.Value.Time().Format(timeFormat))
+			} else {
+				return slog.Attr{Key: "t", Value: attr.Value}
+			}
+		}
+	case slog.LevelKey:
+		if l, ok := attr.Value.Any().(slog.Level); ok {
+			attr = slog.Any("level", LevelString(l))
```
darioush, line 166: changed this from upstream's `lvl` for compatibility with before

log/handler.go
```diff
@@ -0,0 +1,196 @@
...
+func builtinReplaceJSON(_ []string, attr slog.Attr) slog.Attr {
+	return builtinReplace(nil, attr, false)
+}
+
+func builtinReplace(_ []string, attr slog.Attr, logfmt bool) slog.Attr {
+	switch attr.Key {
+	case slog.TimeKey:
+		if attr.Value.Kind() == slog.KindTime {
+			if logfmt {
+				return slog.String("t", attr.Value.Time().Format(timeFormat))
```
darioush, line 159: note that this uses "t" for the timestamp of the log, previously we used "timestamp", maybe we should change it.

log/handler.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+// JSONHandler returns a handler which prints records in JSON format.
+func JSONHandler(wr io.Writer) slog.Handler {
+	return slog.NewJSONHandler(wr, &slog.HandlerOptions{
+		ReplaceAttr: builtinReplaceJSON,
+	})
+}
+
+func JSONHandlerWithLevel(wr io.Writer, level slog.Leveler) slog.Handler {
```
darioush, line 120: added this compared to upstream

log/handler.go
```diff
@@ -0,0 +1,196 @@
...
+// Example:
+//
+//	[DBUG] [May 16 20:58:45] remove route ns=haproxy addr=127.0.0.1:50002
+func NewTerminalHandler(wr io.Writer, useColor bool) *TerminalHandler {
+	return NewTerminalHandlerWithLevel(wr, levelMaxVerbosity, useColor)
+}
+
+// NewTerminalHandlerWithLevel returns the same handler as NewTerminalHandler but only outputs
+// records which are less than or equal to the specified verbosity level.
+func NewTerminalHandlerWithLevel(wr io.Writer, lvl slog.Leveler, useColor bool) *TerminalHandler {
```
darioush, line 70: changed this from Level -> Leveler

log/format.go
```diff
@@ -95,348 +30,180 @@ type TerminalStringer interface {
...
+	}
+	if buf == nil {
+		buf = make([]byte, 0, 30+termMsgJust)
+	}
+	b := bytes.NewBuffer(buf)
+
+	// Note the timestamp is moved before the log level compared to upstream
+	b.WriteString("[")
+	writeTimeTermFormat(b, r.Time)
+	b.WriteString("] ")
```
darioush, line 60: note this change as well (this would be the most difficult to use the upstream code for)

log/format.go
```diff
@@ -95,348 +30,180 @@ type TerminalStringer interface {
...
-			align := int(locationLength.Load())
-			if align < len(location) {
-				align = len(location)
-				locationLength.Store(uint32(align))
-			}
-			padding := strings.Repeat(" ", align-len(location))
+	// Prefix is added compared to upstream
+	if h.Prefix != nil {
+		b.WriteString(h.Prefix(r))
+	}
```
darioush, line 73: added a Prefix function (but we could probably alter the message instead if preferred)

internal/ethapi/api.go
```diff
@@ -1153,142 +1155,46 @@ func (s *BlockChainAPI) Call(ctx context.Context, args TransactionArgs, blockNrO
...
 	}
 
-	// We first execute the transaction at the highest allowable gas limit, since if this fails we
-	// can return error immediately.
-	failed, result, err := executeEstimate(ctx, b, args, state.Copy(), header, gasCap, hi)
+	// If the user has not specified a gas limit, use the block gas limit
+	if args.Gas == nil {
+		args.Gas = new(hexutil.Uint64)
+		*args.Gas = hexutil.Uint64(header.GasLimit)
+	}
```
ceyonur, line 1184: I think this was intentionally left out, so that globalGasCap can be used in `ToMessage`

ceyonur (commented):
Left few comments and opened a PR with some nits:
https://github.com/ava-labs/subnet-evm/pull/1164

internal/ethapi/api.go
ceyonur: it appears that there was a bug, and this is a fix. resolving now.

plugin/evm/gossip.go
```diff
@@ -121,7 +121,7 @@ func (g *GossipEthTxPool) IsSubscribed() bool {
 }
 
 func (g *GossipEthTxPool) Subscribe(ctx context.Context) {
-	sub := g.mempool.SubscribeNewTxsEvent(g.pendingTxs)
+	sub := g.mempool.SubscribeTransactions(g.pendingTxs, false) // XXX: passing reorgs = false
```
darioush, line 124: I wonder if passing false for reorgs is correct here.

ceyonur (approved):

ceyonur (approved):

PR #1151, opened 2024-04-15
fix unindexor in state sync

darioush:

### DIFF
--- a/core/blockchain_test.go
+++ b/core/blockchain_test.go
@@ -8,7 +8,6 @@ import (
 	"math/big"
 	"os"
 	"testing"
-	"time"
 
 	"github.com/ava-labs/subnet-evm/consensus/dummy"
 	"github.com/ava-labs/subnet-evm/core/rawdb"
@@ -697,6 +696,7 @@ func TestTransactionSkipIndexing(t *testing.T) {
 	require.NoError(err)
 	currentBlockNumber := chain.CurrentBlock().Number.Uint64()
 	checkTxIndicesHelper(t, nil, currentBlockNumber+1, currentBlockNumber+1, currentBlockNumber, chainDB, false) // check all indices has been skipped
+	chain.Stop()
 
 	// test2: specify lookuplimit with tx index skipping enabled. Blocks should not be indexed but tail should be updated.
 	conf.TxLookupLimit = 2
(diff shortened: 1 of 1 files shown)


core/blockchain.go
```diff
@@ -483,6 +484,7 @@ func (bc *BlockChain) maintainTxIndex() {
 	if head := bc.CurrentBlock(); head != nil && head.Number.Uint64() > txLookupLimit {
 		done = make(chan struct{})
 		tail := rawdb.ReadTxIndexTail(bc.db)
+		bc.wg.Add(1)
```
darioush, line 487: alternatively I think we can check `if done != nil { ... }` like its done in `bc.quit`, but also `wg` seems fine.

core/blockchain.go
```diff
@@ -428,7 +428,20 @@ func NewBlockChain(
...
+				headCh = make(chan ChainEvent, 1) // Buffered to avoid locking up the event feed
+				sub    = bc.SubscribeChainAcceptedEvent(headCh)
+			)
+			if sub == nil {
+				log.Warn("could not create chain accepted subscription to unindex txs")
+				return
+			}
+			defer sub.Unsubscribe()
+
+			bc.maintainTxIndex(headCh)
```
ceyonur, line 443: why did we move the subscription out?

darioush: if not, it is possible (but unlikely) that the chain will accept the 1st block before the call to `chain.Accept`.
Actually I think these should be moved out of the goroutine as well

core/blockchain.go
```diff
@@ -428,7 +428,20 @@ func NewBlockChain(
 	// Start tx indexer/unindexer if required.
 	if bc.cacheConfig.TxLookupLimit != 0 {
 		bc.wg.Add(1)
-		go bc.maintainTxIndex()
+		go func() {
+			defer bc.wg.Done()
+			var (
+				headCh = make(chan ChainEvent, 1) // Buffered to avoid locking up the event feed
+				sub    = bc.SubscribeChainAcceptedEvent(headCh)
+			)
```
darioush, line 436:
```suggestion
		var (
				headCh = make(chan ChainEvent, 1) // Buffered to avoid locking up the event feed
				sub    = bc.SubscribeChainAcceptedEvent(headCh)
		)
		go func() {
			defer bc.wg.Done()
```

core/blockchain.go
```diff
@@ -428,7 +428,20 @@ func NewBlockChain(
 	// Start tx indexer/unindexer if required.
 	if bc.cacheConfig.TxLookupLimit != 0 {
 		bc.wg.Add(1)
-		go bc.maintainTxIndex()
+		var (
+				headCh = make(chan ChainEvent, 1) // Buffered to avoid locking up the event feed
+				sub    = bc.SubscribeChainAcceptedEvent(headCh)
```
darioush, line 433:
```suggestion
			headCh = make(chan ChainEvent, 1) // Buffered to avoid locking up the event feed
			sub    = bc.SubscribeChainAcceptedEvent(headCh)
```

PR #1157, opened 2024-04-18
Geth sync nits

ceyonur:
Few nits/cleanups after latest geth sync (v1.13.2)

### DIFF
--- a/tests/utils/tmpnet.go
+++ b/tests/utils/tmpnet.go
@@ -15,8 +15,9 @@ import (
 )
 
 var DefaultChainConfig = tmpnet.FlagsMap{
-	"log-level":        "debug",
-	"warp-api-enabled": true,
+	"log-level":         "debug",
+	"warp-api-enabled":  true,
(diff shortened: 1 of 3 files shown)


tests/utils/tmpnet.go
```diff
@@ -15,8 +15,9 @@ import (
 )
 
 var DefaultChainConfig = tmpnet.FlagsMap{
-	"log-level":        "debug",
-	"warp-api-enabled": true,
+	"log-level":         "debug",
+	"warp-api-enabled":  true,
+	"local-txs-enabled": true,
```
darioush, line 20: should we enable this only in the load tests?

darioush (dismissed):

tests/utils/tmpnet.go
ceyonur: I think it makes sense to enable them for warp tests as well (since it has a load test in warp tests)

JacobEverly (approved):

PR #1164, opened 2024-05-06
resync

ceyonur:
Resync

### DIFF
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -1176,12 +1176,6 @@ func DoEstimateGas(ctx context.Context, b Backend, args TransactionArgs, blockNr
 		State:      state,
 		ErrorRatio: estimateGasErrorRatio,
 	}
-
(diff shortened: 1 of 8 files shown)


internal/ethapi/api.go
```diff
@@ -1176,12 +1176,6 @@ func DoEstimateGas(ctx context.Context, b Backend, args TransactionArgs, blockNr
 		State:      state,
 		ErrorRatio: estimateGasErrorRatio,
 	}
-
-	// If the user has not specified a gas limit, use the block gas limit
-	if args.Gas == nil {
-		args.Gas = new(hexutil.Uint64)
-		*args.Gas = hexutil.Uint64(header.GasLimit)
-	}
```
darioush, line 1184: see https://github.com/ethereum/go-ethereum/pull/29509
Everything else looks good.

ceyonur: reapplied the fix

PR #1166, opened 2024-05-07
[antithesis] Add initial test setup

marun:
## Why this should be merged
This PR adds the scaffolding required to build images for an antithesis test setup for subnet-evm.

## How this works

Adds scripts, dockerfiles and code that comprise the antithesis test setup and its tooling.

## How was this tested
New CI job checking workload sanity and construction of the images.

**NOTE:** Due to the limitations inherent to github actions, validation of image publication and the triggering of antithesis workloads can only be performed once this PR is merged.

### Pre-merge
 - [x] Merge https://github.com/ava-labs/avalanchego/pull/3305
 - [x] Update subnet-evm to a compatible hash/tag that includes 3305

### Post-merge
  - [ ] Require the new CI workflow
  - [ ] Manually test the image build and test trigger
  - [ ] Add a schedule

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,8 @@ go 1.21.12
 
 require (
 	github.com/VictoriaMetrics/fastcache v1.12.1
-	github.com/ava-labs/avalanchego v1.11.11-0.20240813203340-ab83fb41528d
+	github.com/antithesishq/antithesis-sdk-go v0.3.8
+	github.com/ava-labs/avalanchego v1.11.11-0.20240819192939-df91c2f4ab99
 	github.com/cespare/cp v0.1.0
 	github.com/crate-crypto/go-ipa v0.0.0-20231025140028-3c0104f4b233
 	github.com/davecgh/go-spew v1.1.1
@@ -66,12 +67,16 @@ require (
 	github.com/cockroachdb/pebble v0.0.0-20230928194634-aa077af62593 // indirect
 	github.com/cockroachdb/redact v1.1.3 // indirect
 	github.com/cockroachdb/tokenbucket v0.0.0-20230807174530-cc333fc44b06 // indirect
+	github.com/compose-spec/compose-go v1.20.2 // indirect
 	github.com/consensys/bavard v0.1.13 // indirect
 	github.com/consensys/gnark-crypto v0.12.1 // indirect
 	github.com/cpuguy83/go-md2man/v2 v2.0.2 // indirect
 	github.com/crate-crypto/go-kzg-4844 v0.7.0 // indirect
 	github.com/decred/dcrd/dcrec/secp256k1/v4 v4.1.0 // indirect
+	github.com/distribution/reference v0.5.0 // indirect
 	github.com/dlclark/regexp2 v1.7.0 // indirect
+	github.com/docker/go-connections v0.4.0 // indirect
+	github.com/docker/go-units v0.5.0 // indirect
 	github.com/ethereum/c-kzg-4844 v0.4.0 // indirect
 	github.com/getsentry/sentry-go v0.18.0 // indirect
 	github.com/go-logr/logr v1.4.1 // indirect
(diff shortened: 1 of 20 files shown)


tests/antithesis/main.go
```diff
@@ -0,0 +1,106 @@
...
+			key:     key,
+			addrs:   set.Of(key.Address()),
+			uris:    c.URIs,
+		}
+	}
+
+	for _, w := range workloads[1:] {
+		go w.run(ctx)
+	}
+	genesisWorkload.run(ctx)
```
joshua-kim, line 72: Is the only reason we do this to block while this is running? Feel like we could just use a wg here to do the same thing without having to do the `workloads[1:]` trick:

```
        wg := &sync.WaitGroup{}
	for _, w := range workloads {
		go w.run(ctx, wg)
                wg.Add(1)
	}
        wg.Wait()
```

joshua-kim (dismissed):

tests/antithesis/main.go
joshua-kim: Ah didn't realize this is copied.

tests/antithesis/main.go
```diff
@@ -0,0 +1,106 @@
...
+	key     *secp256k1.PrivateKey
+	addrs   set.Set[ids.ShortID]
+	uris    []string
+}
+
+func (w *workload) run(ctx context.Context) {
+	timer := time.NewTimer(0)
+	if !timer.Stop() {
+		<-timer.C
+	}
```
darioush, line 87: what does this do? is it to initialize the timer so it can be called with Reset?

darioush (dismissed):

tests/antithesis/main.go
StephenButtolph: This is something we've used a couple of times in avalanchego (might make sense to make it a helper).

This is creating a stopped timer so that there is no entry on the `C` channel (and there isn't one scheduled to be added).

This means that after calling `Reset` there will be no events on the channel until the timer fires (at which point there will be exactly one event sent to the channel).

It enables re-using the timer across loop iterations without needing to have the first loop iteration perform any `== nil` checks to initialize the first invocation.

go.mod
```diff
@@ -4,7 +4,8 @@ go 1.21.12
 
 require (
 	github.com/VictoriaMetrics/fastcache v1.12.1
-	github.com/ava-labs/avalanchego v1.11.11-0.20240729205337-a0f7e422bb84
+	github.com/antithesishq/antithesis-sdk-go v0.3.8
```
darioush, line 7: this seems to add some dependencies, not sure if the binary size matters.
perhaps it can be its own binary?

tests/antithesis/main.go
```diff
@@ -0,0 +1,201 @@
...
+	})
+
+	// TODO(marun) What should this value be?
+	txAmount := uint64(10000)
+	for {
+		// TODO(marun) Exercise a wider variety of transactions
+		recipientEthAddress := crypto.PubkeyToAddress(w.key.PublicKey)
+		err := transferFunds(ctx, w.client, w.key, recipientEthAddress, txAmount)
+		if err != nil {
+			log.Printf("failed to transfer funds: %s", err)
```
darioush, line 137: should this fail the test?

darioush (approved):

go.mod
darioush: I was wondering if it is cleaner to make a separate go.mod for the antithesis binary, but I am not sure if that adds too much complexity for what it brings.

tests/antithesis/main.go
darioush: Should we add some comment about this?

darioush (approved):

darioush (approved):

darioush:
This seems fine to me, also let's get @ceyonur 's approval

ceyonur (approved):
lgtm. Do we need to add any of these as required status checks in branch settings?

PR #1167, opened 2024-05-08
fix space in log levels

ceyonur:
Fixes level spacing problems like
`[05-07|12:25:21.446] DEBUG<hBRSYikeL9fzeioygWcgDKFtEnaGsLS5VFxRFNxUiDpan79rg Chain> plugin/evm/gossip.go:138 shutting down subscription
`

### DIFF
--- a/log/logger.go
+++ b/log/logger.go
@@ -71,13 +71,13 @@ func LevelAlignedString(l slog.Level) string {
 	case slog.LevelDebug:
 		return "DEBUG"
 	case slog.LevelInfo:
-		return "INFO "
+		return "INFO"
 	case slog.LevelWarn:
-		return "WARN "
+		return "WARN"
 	case slog.LevelError:
 		return "ERROR"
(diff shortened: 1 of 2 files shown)


log/logger.go
```diff
@@ -71,13 +71,13 @@ func LevelAlignedString(l slog.Level) string {
 	case slog.LevelDebug:
 		return "DEBUG"
 	case slog.LevelInfo:
-		return "INFO "
```
darioush, line 74: should we keep these spaces? I think they were put to keep the logs aligned

darioush (approved):

log/logger.go
ceyonur: yes but we don't do that for avalanche logs, I think it's better we align them?

darioush: keeping them aligned is the upstream behavior as well so I suggest we keep it

Issue #1050, opened 2024-01-22
Replace context.TODOs 


ceyonur:
There are some contexts left with `TODO()`, we should be using cancellable/timed contexts. From comment:
https://github.com/ava-labs/subnet-evm/pull/1031/files#r1462043218

PR #1183, opened 2024-05-22
Fix contexts

ceyonur:
## Why this should be merged

Replaced context.TODOs

Fixes: #1050 

## How this was tested

Local UT/e2e tests

## How is this documented

No need


darioush (approved):

darioush (approved):

core/blockchain.go
```diff
@@ -1891,7 +1892,7 @@ func (bc *BlockChain) initSnapshot(b *types.Header) {
 // it reaches a block with a state committed to the database. reprocessState does not use
 // snapshots since the disk layer for snapshots will most likely be above the last committed
 // state that reprocessing will start from.
-func (bc *BlockChain) reprocessState(current *types.Block, reexec uint64) error {
+func (bc *BlockChain) reprocessState(ctx context.Context, current *types.Block, reexec uint64) error {
```
tsachiherman, line 1895: I really live the addition of an expiration context here. Two concerns:
1. In the case of a context expiration, should we clear out the `current` block so that this method would be atomic ?
2. Is there any state issue with partial execution of this method ? i.e. would terminating the execution in the middle would bring the node to some undesired state ?

tsachiherman (commented):
I have the same comment, which would apply at different locations - but with the same concern:
- terminating a method mid-flight might leave the database in a different state compared to it's existing behavior.
- I can't really tell if that would be an issue or not. Some code might need to be written differently to support that. ( i.e. perform a single commit instead of multiple commits ).
- If there is such a place where it would be hard to break a for-loop, I'd suggest placing the context checking both before and after the loop. That's not ideal, but it wouldn't cause any regression.

core/blockchain.go
ceyonur: I think contexts used to gracefully shutdown are safe to use, but we decided not to merge this PR as contexts from VM init can be closed from parent thread when `SetState` is called.

PR #1193, opened 2024-05-28
bump js dependencies

ceyonur:
## Why this should be merged

Updated npm package to 1.2.1 with following updates:

* Solidity versions up-to 0.8.24 (the version before cancun is enabled by default)
* Hardhat to latest available (2.22.4)
  * this transitively updates ethers5 to ethers6 (includes breaking changes) 
* Replaced hardhat-ethers and hardhat-waffle with hardhat-toolbox
* types/node to 20.12.12
* avalanchejs to 4.0.5
* Node version to 20.13
* npm version to 7

Also removed allowing of failures in ds-test asset failures (asserted failures should always return a failure to tests)

## How this was tested

Locally & e2e

## How is this documented

Nothing changes in terms of how contracts works (just test changes)

### DIFF
--- a/contracts/.gitignore
+++ b/contracts/.gitignore
@@ -144,3 +144,7 @@ dist
 .pnp.*
 
 local_rpc.json
+
+# TypeChain files
+/typechain
+/typechain-types
--- a/contracts/contracts/AllowList.sol
+++ b/contracts/contracts/AllowList.sol
@@ -1,5 +1,5 @@
 //SPDX-License-Identifier: MIT
-pragma solidity ^0.8.0;
+pragma solidity ^0.8.24;
 
 import "./interfaces/IAllowList.sol";
 import "@openzeppelin/contracts/access/Ownable.sol";
--- a/contracts/test/utils.ts
+++ b/contracts/test/utils.ts
@@ -1,5 +1,5 @@
 import { ethers } from "hardhat"
-import { CallOverrides } from "ethers"
+import { Overrides } from "ethers"
 import assert from "assert"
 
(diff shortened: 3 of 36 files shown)


contracts/test/utils.ts
```diff
@@ -117,11 +119,7 @@ const buildTestFn = (fnNameOrObject: FnNameOrObject, overrides = {}, debug = fal
         console.log('')
       }
 
-      if (fn.shouldFail) {
-        assert(failed, `${fn.method} should have failed`)
-      } else {
-        assert(!failed, `${fn.method} failed`)
-      }
```
ceyonur, line 124: I think if any of those ds-test assertions fail, the test should fail. Assertions should be placed such that they will not fail.

contracts/contracts/AllowList.sol
```diff
@@ -1,5 +1,5 @@
 //SPDX-License-Identifier: MIT
-pragma solidity ^0.8.0;
+pragma solidity ^0.8.24;
```
ceyonur, line 2: I've explicitly defined shangai version in hardhat config, but contracts should explicitly use <0.8.24 (cancun activation)

contracts/.gitignore
```diff
@@ -144,3 +144,7 @@ dist
 .pnp.*
 
 local_rpc.json
+
+# TypeChain files
+/typechain
+/typechain-types
```
ceyonur, line 150: it's suggested that these files should not be checked-out (since they can be generated)

darioush (approved):

PR #1195, opened 2024-05-29
Fix overflow in state sync eta

darioush:
## Why this should be merged
This calculation has the potential to overflow:
	now.Sub(t.triesStartTime) * time.Duration(t.triesRemaining)

## How this works
`time.Duration(t.triesRemaining/t.triesSynced)` is less likely to overflow as the ratio decreases as the time passed goes on.

## How this was tested
CI

## How is this documented
N/A

### DIFF
--- a/sync/statesync/trie_sync_stats.go
+++ b/sync/statesync/trie_sync_stats.go
@@ -132,7 +132,7 @@ func (t *trieSyncStats) updateETA(sinceUpdate time.Duration, now time.Time) time
 	}
 
 	triesTime := now.Sub(t.triesStartTime) * time.Duration(t.triesRemaining/t.triesSynced)
-	eta := leafsTime + triesTime // TODO: should we use max instead of sum?
+	eta := max(leafsTime, triesTime)
 	log.Info(
 		"state sync: syncing storage tries",
 		"triesRemaining", t.triesRemaining,
--- a/sync/statesync/trie_sync_stats.go
+++ b/sync/statesync/trie_sync_stats.go
@@ -114,7 +114,7 @@ func (t *trieSyncStats) trieDone(root common.Hash) {
 // updateETA calculates and logs and ETA based on the number of leafs
 // currently in progress and the number of tries remaining.
 // assumes lock is held.
-func (t *trieSyncStats) updateETA(sinceUpdate time.Duration, now time.Time) {
+func (t *trieSyncStats) updateETA(sinceUpdate time.Duration, now time.Time) time.Duration {
 	leafsRate := float64(t.leafsSinceUpdate) / sinceUpdate.Seconds()
 	if t.leafsRate == nil {
 		t.leafsRate = utils_math.NewAverager(leafsRate, leafRateHalfLife, now)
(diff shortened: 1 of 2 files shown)


sync/statesync/trie_sync_stats.go
```diff
@@ -128,15 +128,17 @@ func (t *trieSyncStats) updateETA(sinceUpdate time.Duration, now time.Time) {
 		// provide a separate ETA for the account trie syncing step since we
 		// don't know the total number of storage tries yet.
 		log.Info("state sync: syncing account trie", "ETA", roundETA(leafsTime))
-		return
+		return leafsTime
 	}
 
-	triesTime := now.Sub(t.triesStartTime) * time.Duration(t.triesRemaining) / time.Duration(t.triesSynced)
+	triesTime := now.Sub(t.triesStartTime) * time.Duration(t.triesRemaining/t.triesSynced)
```
tsachiherman, line 134: I'm not sure that this is the right calculation for triesTime.
```golang
total-time-so-far = now.Sub(t.triesStartTime)
```
during that time, we have completed `(t.triesSynced-t.triesRemaining)` tries.
that means that the time per trie is
```golang
time-per-trie = total-time-so-far /  (t.triesSynced-t.triesRemaining)
```
and the remaining time is `t.triesRemaining * time-per-trie`.

expanding that:
```golang
triesTime = t.triesRemaining * now.Sub(t.triesStartTime) / (t.triesSynced-t.triesRemaining)
```

sync/statesync/trie_sync_stats.go
```diff
@@ -128,15 +128,17 @@ func (t *trieSyncStats) updateETA(sinceUpdate time.Duration, now time.Time) {
...
+	eta := leafsTime + triesTime // TODO: should we use max instead of sum?
```
tsachiherman, line 135: I think that using max here would be providing more realistic timelines.

sync/statesync/trie_sync_stats.go
darioush: > during that time, we have completed (t.triesSynced-t.triesRemaining) tries.
we've completed `t.triesSynced` tries.

tsachiherman: ohh - so, in that case
```golang
triesTime := (now.Sub(t.triesStartTime) * time.Duration(t.triesRemaining)) / time.Duration(t.triesSynced)
```
since we want to get some reasonable precision.. otherwise, `t.triesRemaining/t.triesSynced` would be very lossy.

darioush: this re-introduces the overflow

StephenButtolph: Can we implement this in a common location? [timer.EstimateETA](https://github.com/ava-labs/avalanchego/blob/master/utils/timer/eta.go#L23) seems like it can probably do this (and shouldn't have the overflow issue). If that method doesn't work for this can we make one that does?

StephenButtolph (approved):

ceyonur (approved):

PR #1198, opened 2024-05-30
[testing] Remove dep on gomega

marun:
## Why this should be merged

Previously gomega was replaced by testify so that only a single assertion library was used across all types of testing. The replacement was incomplete, though, since registration of a fail handler for gomega was still being performed and somehow the  `tests/utils` path was missed entirely. This changeset removes the use of gomega entirely.

## How this was tested

CI


darioush (approved):
looks great, thanks

Issue #1200, opened 2024-06-01
Panic while syncing


heki-ru:
Was syncing from scratch DFK subnet and it's crashed

```
{
    "state-sync-enabled": true,
    "state-sync-min-blocks": 86400,
    "validatorOnly": true,
    "metrics-expensive-enabled": false
}
```

```
Ubuntu 24.04
Linux 6.8.0-31-generic #31-Ubuntu SMP PREEMPT_DYNAMIC Sat Apr 20 00:40:06 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
subnet-evm v0.6.5
avalanchego v1.11.6
```

[q2aTwKuyzgs8pynF7UXBZCU7DejbZbZ6EUyHr3JQzYgwNPUPi.log](https://github.com/user-attachments/files/15523410/q2aTwKuyzgs8pynF7UXBZCU7DejbZbZ6EUyHr3JQzYgwNPUPi.log)

```
[06-01|05:46:30.315] INFO <q2aTwKuyzgs8pynF7UXBZCU7DejbZbZ6EUyHr3JQzYgwNPUPi Chain> sync/statesync/trie_sync_stats.go:137 state sync: syncing storage tries        triesRemaining=169  ETA=1h12m
[06-01|05:47:30.399] INFO <q2aTwKuyzgs8pynF7UXBZCU7DejbZbZ6EUyHr3JQzYgwNPUPi Chain> sync/statesync/trie_sync_stats.go:137 state sync: syncing storage tries        triesRemaining=157  ETA=1h10m
[06-01|05:48:30.470] INFO <q2aTwKuyzgs8pynF7UXBZCU7DejbZbZ6EUyHr3JQzYgwNPUPi Chain> sync/statesync/trie_sync_stats.go:137 state sync: syncing storage tries        triesRemaining=157  ETA=1h7m
[06-01|05:49:30.525] INFO <q2aTwKuyzgs8pynF7UXBZCU7DejbZbZ6EUyHr3JQzYgwNPUPi Chain> sync/statesync/trie_sync_stats.go:137 state sync: syncing storage tries        triesRemaining=45   ETA=59m
panic: send on closed channel

goroutine 321 [running]:
github.com/ava-labs/subnet-evm/sync/statesync.(*trieToSync).createSegments(0xc000769340, 0x4)
	/home/runner/work/subnet-evm/subnet-evm/subnet-evm/sync/statesync/trie_segments.go:312 +0x2c7
[...]

PR #1201, opened 2024-06-03
resolve race in closing sync segments channel

darioush:
## Why this should be merged
Fixes https://github.com/ava-labs/subnet-evm/issues/1200

## How this works
Closes the sync channel after the last storage trie has finished (or if there is no storage tries).
This avoids writing to a closed channel when a storage trie splits into segments:
https://github.com/ava-labs/subnet-evm/blob/master/sync/statesync/trie_segments.go#L311-L313

but in the existing behavior the storageTrieProducer could have previously closed the channel.

## How this was tested
CI

## How is this documented
N/A

### DIFF
--- a/sync/statesync/state_syncer.go
+++ b/sync/statesync/state_syncer.go
@@ -254,11 +254,8 @@ func (t *stateSync) addTrieInProgress(root common.Hash, trie *trieToSync) {
 	t.triesInProgress[root] = trie
 }
 
-// removeTrieInProgress removes root from the set of tracked
-// tries in progress and notifies the storage root producer
-// so it can continue in case it was paused due to the
-// maximum number of tries in progress being previously reached.
-// returns the number of tries in progress after the removal.
+// removeTrieInProgress removes root from the set of tracked tries in progress
+// and returns the number of tries in progress after the removal.
 func (t *stateSync) removeTrieInProgress(root common.Hash) (int, error) {
 	t.lock.Lock()
 	defer t.lock.Unlock()
--- a/sync/statesync/state_syncer.go
+++ b/sync/statesync/state_syncer.go
@@ -56,6 +56,7 @@ type stateSync struct {
 
 	// track completion and progress of work
 	mainTrieDone       chan struct{}
+	storageTriesDone   chan struct{}
 	triesInProgressSem chan struct{}
 	done               chan error
 	stats              *trieSyncStats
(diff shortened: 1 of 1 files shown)


sync/statesync/state_syncer.go
```diff
@@ -167,30 +182,36 @@ func (t *stateSync) storageTrieProducer(ctx context.Context) error {
...
-				return err
-			}
-			t.addTrieInProgress(root, storageTrie)
-			storageTrie.startSyncing() // start syncing after tracking the trie as in progress
+		// acquire semaphore (to keep number of tries in progress limited)
+		select {
+		case t.triesInProgressSem <- struct{}{}:
+		case <-ctx.Done():
+			return ctx.Err()
+		}
```
ceyonur, line 195: does that mean we will continue if `root == (common.Hash{}) && more`?

sync/statesync/state_syncer.go
```diff
@@ -167,30 +182,36 @@ func (t *stateSync) storageTrieProducer(ctx context.Context) error {
...
+			return err
 		}
-		// if there are no more storage tries, close
-		// the task queue and exit the producer.
+		t.addTrieInProgress(root, storageTrie)
+		if !more {
+			close(t.storageTriesDone)
+		}
+		// start syncing after tracking the trie as in progress
+		storageTrie.startSyncing()
```
ceyonur, line 212: any reason why we did not call this right after `addTrieInProgress`?

sync/statesync/state_syncer.go
darioush: it cannot be the case that `root == (common.Hash{}) && more`

sync/statesync/state_syncer.go
darioush: this is because `startSyncing` runs concurrent to the current thread, so we should establish the expected invariant that if all pending tries are done and the `storageTriesDone` channel is closed before starting the process.

otherwise there could be an (unlikely) race condition where the trie completes syncing but the `storageTriesDone` channel is still open.

ceyonur: hmm but closing the `storageTriesDone` channel also means we close `t.segments` [here](https://github.com/ava-labs/subnet-evm/blob/sync-race/sync/statesync/state_syncer.go#L128), right? and then `startSycing` will panic with same error? is there any other invariant?

darioush: that code is called when the storage trie finishes syncing, and only if there is no other pending ones (and no new ones to be added).

ceyonur (approved):

PR #1209, opened 2024-06-10
chore: make function comments match function names

piguagua:
## Why this should be merged

make function comments match function names


ceyonur (approved):
LGTM!
could you also open the same PR for https://github.com/ava-labs/coreth ?
Thanks.

darioush (approved):

PR #1213, opened 2024-06-17
truncate err data logs

ceyonur:

### DIFF
--- a/rpc/handler.go
+++ b/rpc/handler.go
@@ -29,6 +29,7 @@ package rpc
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"reflect"
 	"strconv"
 	"strings"
@@ -39,6 +40,8 @@ import (
 	"github.com/ethereum/go-ethereum/log"
 	"golang.org/x/time/rate"
 )
+
+const truncateLimit = 1024
 
 // handler handles JSON-RPC messages. There is one handler per connection. Note that
 // handler is not safe for concurrent use. Message handling never blocks indefinitely
@@ -571,7 +574,20 @@ func (h *handler) handleCallMsg(ctx *callProc, msg *jsonrpcMessage) *jsonrpcMess
 		if resp.Error != nil {
 			ctx = append(ctx, "err", resp.Error.Message)
 			if resp.Error.Data != nil {
-				ctx = append(ctx, "errdata", resp.Error.Data)
+				// If the log level is debug, log the full error data. Otherwise, truncate it.
+				if h.log.Enabled(context.Background(), log.LvlDebug) {
+					ctx = append(ctx, "errdata", resp.Error.Data)
+				} else {
+					errDataStr := fmt.Sprintf("%v", resp.Error.Data)
+					// Truncate the error data if it is too long. Otherwise, preserve the original data.
+					if len(errDataStr) > truncateLimit {
+						remaining := len(errDataStr) - truncateLimit
+						errDataStr = errDataStr[:truncateLimit] + "...(truncated remaining " + strconv.Itoa(remaining) + " chars)"
+						ctx = append(ctx, "errdata", errDataStr)
+					} else {
+						ctx = append(ctx, "errdata", resp.Error.Data)
+					}
+				}
 			}
 			h.log.Info("Served "+msg.Method, ctx...)
 		} else {
--- a/rpc/handler.go
+++ b/rpc/handler.go
@@ -41,7 +41,7 @@ import (
 	"golang.org/x/time/rate"
 )
 
-const truncateLimit = 1024
+const callErrorLogTruncateLimit = 1024
 
 // handler handles JSON-RPC messages. There is one handler per connection. Note that
 // handler is not safe for concurrent use. Message handling never blocks indefinitely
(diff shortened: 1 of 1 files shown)


rpc/handler.go
```diff
@@ -40,6 +41,8 @@ import (
 	"golang.org/x/time/rate"
 )
 
+const truncateLimit = 1024
```
ceyonur, line 44: this should be around 4 kb, seemed a reasonable limit to me. open for negotiations :)

rpc/handler.go
```diff
@@ -571,7 +574,20 @@ func (h *handler) handleCallMsg(ctx *callProc, msg *jsonrpcMessage) *jsonrpcMess
...
+				// If the log level is debug, log the full error data. Otherwise, truncate it.
+				if h.log.Enabled(context.Background(), log.LvlDebug) {
+					ctx = append(ctx, "errdata", resp.Error.Data)
+				} else {
+					errDataStr := fmt.Sprintf("%v", resp.Error.Data)
+					// Truncate the error data if it is too long. Otherwise, preserve the original data.
+					if len(errDataStr) > truncateLimit {
+						remaining := len(errDataStr) - truncateLimit
+						errDataStr = errDataStr[:truncateLimit] + "...(truncated remaining " + strconv.Itoa(remaining) + " chars)"
+						ctx = append(ctx, "errdata", errDataStr)
```
ceyonur, line 586: this would eventually convert the err data into string (it's actually an {}interface) but I think that was a more straightforward approach than using `reflect` package.

rpc/handler.go
ARR4N: (style) I think that given the distance from where it's used, the name should be more informative; perhaps `callErrorLogTruncateLimit`. Alternatively, just declare it closer to where it's used.

[See style guide](https://google.github.io/styleguide/go/decisions#:~:text=The%20general%20rule%20of%20thumb%20is%20that%20the%20length%20of%20a%20name%20should%20be%20proportional%20to%20the%20size%20of%20its%20scope)

rpc/handler.go
```diff
@@ -571,7 +574,20 @@ func (h *handler) handleCallMsg(ctx *callProc, msg *jsonrpcMessage) *jsonrpcMess
 		if resp.Error != nil {
 			ctx = append(ctx, "err", resp.Error.Message)
```
ARR4N, line 575: (optional) I appreciate that this is pre-existing code, but the `ctx` variable is idiomatically reserved for a `context.Context`. Seeing it being appended to as a slice really confused me; it's also shadowing a method argument. Perhaps `logCtx` or (my preference because this isn't a context) `logData`?

rpc/handler.go
ARR4N: That `interface{}` is quite annoying!

I think it's clearer to use another `fmt.Sprintf` here:

```go
errDataStr = fmt.Sprintf("%s... (truncated remaining %d chars)", errDataStr[:truncateLimit], remaining)
```

rpc/handler.go
```diff
@@ -571,7 +574,20 @@ func (h *handler) handleCallMsg(ctx *callProc, msg *jsonrpcMessage) *jsonrpcMess
...
 			if resp.Error.Data != nil {
-				ctx = append(ctx, "errdata", resp.Error.Data)
+				// If the log level is debug, log the full error data. Otherwise, truncate it.
+				if h.log.Enabled(context.Background(), log.LvlDebug) {
```
ARR4N, line 578: `context.Background()` is reserved for entrypoint functions like `main()`, `init()` or tests. Ideally the `context.Context` would be accepted as the first argument of the function/method, but I appreciate that that may result in a lot of refactoring just to get the plumbing working. In this case, `context.TODO()` is more appropriate than `Background()`.

[See style guide](https://google.github.io/styleguide/go/decisions#:~:text=It%20is%20very%20rare%20for%20code%20in%20the%20middle%20of%20a%20callchain%20to%20require%20creating%20a%20base%20context%20of%20its%20own%20using%20context.Background())

ARR4N (approved):
Approving early despite requested changes just so I don't block you.

rpc/handler.go
ceyonur: I'm not sure if `TODO` would be suitable here as `context.Background()` was explicitly used to check and write [here](https://github.com/ava-labs/subnet-evm/blob/master/log/logger.go#L177-L191) . I believe the current implementation of enabled does not even respect `ctx` parameter at all [here](https://github.com/ava-labs/subnet-evm/blob/master/log/handler.go#L87-L90) (not even in slog's default handlers); so we can even keep it nil.

rpc/handler.go
ceyonur: renamed 👍

rpc/handler.go
ceyonur: yup makes sense

ceyonur:
> I think it would be nicer to:
> 
>     * Apply the fix to the logger itself, where all string / variable length outputs are truncated at a certain limit
> 
>     * Make the limit configurable and apply a default

I looked at the code, unfortunately we don't own JSON & Logfmt handlers, and `ReplaceAttr` for those handlers do not take any usable parameters so we can start truncating them in replace functions (`builtinReplace`). The closest (and probably that does not require a huge refactor) place is the upstream's logger's [`Write`](https://github.com/ethereum/go-ethereum/blob/master/log/logger.go#L160-L174). 

I can create a PR to upstream to get their feedbacks on this.

rpc/handler.go
ceyonur: yup it's definetly confusing, changed to logCtx

ceyonur: I did not want to drift too much away from `ctx` (dunno why it's named this way though)

darioush (approved):

PR #1215, opened 2024-06-18
feat: `gethclone` binary

ARR4N:


scripts/lint_allowed_geth_imports.sh
```diff
@@ -8,7 +8,7 @@ set -o pipefail
 # 1. Recursively search through all go files for any lines that include a direct import from go-ethereum
 # 2. Sort the unique results
 # #. Print out the difference between the search results and the list of specified allowed package imports from geth.
-extra_imports=$(grep -r --include='*.go' --exclude-dir='simulator' '"github.com/ethereum/go-ethereum/.*"' -o -h | sort -u | comm -23 - ./scripts/geth-allowed-packages.txt)
+extra_imports=$(grep -Pr --include='*.go' --exclude-dir='simulator' '^\s+"github.com/ethereum/go-ethereum/.*"' -o -h | tr -d '[:blank:]' | sort -u | comm -23 - ./scripts/geth-allowed-packages.txt)
```
ARR4N, line 11: There was a false positive on `const geth = "github.com/ethereum/go-ethereum/"` so I updated the check to only match with (possible) whitespace but no other prefix.

ARR4N:
> Interesting. For mechanical changes like imports or struct modifications this seems like a very promising approach. Might this approach also be appropriate for more complex fork maintenance (i.e. maintaining our changes at the AST level rather than as git commits)?

I think it would form part of a broader strategy. Changes like you describe, as well as deletions and other "simple" modifications can be easily implemented as an AST change. Additions of entire methods/functions should just be done as new files already in the destination module (i.e. regular Go code). More complex code changes will be addressed as we encounter them.

Cloned+transformed files should probably still be checked in so others can import the package, but with pre-merge checks to confirm that they're up to date with `gethclone` output (e.g. with [check-clean-branch.sh](https://github.com/ava-labs/subnet-evm/blob/5567b8904853741255ad732a4643016ad93281e7/.github/workflows/tests.yml#L255)).

Advantages of maintaining the changes as AST and in-situ files:

- Explicit record of the differences compared to `geth`
- Robust (hopefully) to refactoring of `geth`

The primary disadvantage is the complexity of writing an AST patch although I plan on writing some generators; that would be used something like:

```go
patches.AddStructField(geth("core/types"), "StateAccount", &ast.Field{..."IsMultiCoin"...})
```

x/gethclone/main.go
```diff
@@ -0,0 +1,196 @@
...
+	outputGoMod string
+
+	// Internal
+	outputModule *modfile.Module
+	astPatches   astpatch.PatchRegistry
+
+	processed map[string]bool
+}
+
+const geth = "github.com/ethereum/go-ethereum"
```
tsachiherman, line 61: maybe consider moving this into a flag while retaining this value as the default value ?

tsachiherman (commented):
I think that I would prefer to have the `config` struct defined in it's own file and leave the `main.go` to be the light-weight entry point for the executable.

ARR4N:
> No argument that the current approach of merging changes without shared history is not ideal, but how would you compare/contrast the AST approach proposed in this PR with maintenance of a proper fork with shared git history for which sets of persistent and transient carry patches are maintained?

Context for others as I didn't know about transient/carry patches: https://github.com/openshift/kubernetes/blob/master/REBASE.openshift.md

ARR4N:
@tsachiherman:
> I think that I would prefer to have the `config` struct defined in it's own file and leave the `main.go` to be the light-weight entry point for the executable.

Moved into `gethclone.go`

tsachiherman (approved):

PR #1223, opened 2024-06-26
add example staking manager

ceyonur:


contracts/contracts/interfaces/staking-manager.md
```diff
@@ -0,0 +1,109 @@
...
+
+Register validator is the process of adding a validator to the subnet.The contract initiates this with a `registerValidator` function that takes following inputs:
+
+- Subnet ID (bytes32): The ID of the subnet where the validator is being registered
+- Node ID (bytes32): The ID of the node that is being registered as a validator
+- Amount (uint64): The staked amount that will determine the weight of the validator
+- expirationTime (uint64): The time when this register message will be expired
+- ed25519 signature (bytes): The signature of the message signed by the node's BLS key
+
+The messageID of the request can be crafted with `sha256(subnetID, nodeID, amount, expirationTime, signature)`. This should be same with the one that will be crafted for P-chain.
```
ceyonur, line 35: This is important

contracts/contracts/interfaces/staking-manager.md
```diff
@@ -0,0 +1,109 @@
...
+
+The contract should verify these:
+
+- (Optional) the staking amount is greater than the minimum staking amount
+- (Optional) The node is not already registered as a validator. This will be handled by the P-Chain though.
```
ceyonur, line 40: This will be handled by P-Chain anyway, but I think we have the info to check this in contract

contracts/contracts/interfaces/staking-manager.md
```diff
@@ -0,0 +1,109 @@
...
+- (Optional) Expiry timestamp is not in the past
```
ceyonur, line 41: This seems reasonable but again subnet chain and p-chain might not be in sync.

contracts/contracts/interfaces/staking-manager.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Uptime Based Rewards
+
+The manager contract should reward based on uptimes of validators. VMs should be able to track precise uptimes of validators. The manager contract should be able to query the uptime of a validator and reward based on the uptime.
+
+TODO: Design currently is not finalized. We need to discuss how to track uptimes and how to report them to the manager contract. Some ideas:
+
+1- If a validator becomes ineligible permanetly due to low uptime, a warp message should be sent to the manager contract.
+2- Once a validator is removed, a warp message that indicates the uptime is enough will be sent to the contract.
+3- A message will be aggregated upon a request from the manager contract (e.g via an event), then the message will be sent to the manager contract.
```
ceyonur, line 104: I feel like if the manager contract can signal for aggregation of a message which target a uptime specified by the contract; it can implement many different use cases. I.e contract can ask about >%80 uptime message to be signed and if so it can use it to rewarding. In another case it might ask for <%50 uptime and if signed can slash. A simply dynamic request for >%x itself is very helpful for contracts to determine their min uptime.

contracts/contracts/interfaces/staking-manager.md
```diff
@@ -0,0 +1,116 @@
...
+- Stop accruing rewards for the validator
+
+### receiveRemoveValidator
+
+Once the `SetSubnetValidatorWeightTx` is accepted, a warp message `InvalidValidatorRegisterMessage` will be relayed to the contract with `messageID` of the request. The message represents a messageID will not possibly become a valid validator. This essentialy means the message was not valid in the first place, or the validator has removed from the set. The InvalidValidatorRegisterMessage can either be a result of `removeValidator` from the contract or it can be originated directly from P-chain via `ExitValidatorSetTx`. In either case this ultimately tells the contract that a validator was evicted. See related ACP-77 section [here](https://github.com/avalanche-foundation/ACPs/tree/main/ACPs/77-reinventing-subnets#exitvalidatorsettx).
+
+The contract should verify these in `receiveRemoveValidator`:
+
+- Verified warp message
+- The typecheck on the warp message (== InvalidValidatorRegisterMessage)
```
darioush, line 93: nit: I dislike the name of this message `InvalidValidatorRegisterMessage`, mainly because "Invalid....Message" seems weird as part of an expected flow.

Should we rename it to `ValidatorUnregistered` or `ValidatorNotRegistered` (or perhaps `Validation`)?
Not sure if it is too late for this

contracts/contracts/interfaces/staking-manager.md
ceyonur: Pseudocode :
```
  constant  threshold = 50

function receiveRegisterMessageInvalid(uint32 messageIndex) external {
  ...
  if (validator.redeemedAt != 0) {
    uint256 reward = calculateReward(
      validator.weight,
      validator.startedAt,
      validator.redeemedAt
    );
    uptimeID = calculateID(validationID, threshold, reward)
    lockedRewards[uptimeID] = {validator.rewardAddress, reward}
    emit UptimeRequest(uptimeID, threshold, validationID, validator.nodeID)
   ...
}

function receiveUptime(uint32 messageIndex) external{
  (WarpMessage memory warpMessage, bool success) = WARP_MESSENGER.getVerifiedWarpMessage(messageIndex);
  require(success, "ExampleValidatorManager: invalid warp message");
  uptimeMessage = decode(warpMessage)
  uptimeID = uptimeMessage.ID
  reward = lockedRewards[uptimeID] 
  delete lockedRewards[uptimeID] 
  stakingToken.transfer(reward.address, reward.amount)
  ...
}
```

darioush: I prefer this flexible alternative, do we plan to support only >= and < operators?

ceyonur: In most of the cases I don't think we even need to support "<". Most of the time it will be enough to check if a validator has enough uptime. However I also don't think it would be too complicated to support "<". I think these both can cover almost all use cases.

contracts/contracts/ExampleStakingManager.sol
```diff
@@ -0,0 +1,228 @@
...
+contract ExampleStakingManager is Ownable {
+  IWarpMessenger public constant WARP_MESSENGER = IWarpMessenger(0x0200000000000000000000000000000000000005);
+  uint64 constant MAX_UINT64 = type(uint64).max;
+
+  IERC20 public immutable stakingToken;
+  IERC20 public immutable rewardsToken;
+
+  uint256 public minStakingDuration; // Minimun duration for staking in seconds.
+  uint256 public minStakingAmount; // Minimum amount for staking.
+  bool public stakingEnabled; // Start time of the staking period.s
```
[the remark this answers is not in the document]
ceyonur: yup corrected

contracts/contracts/ExampleStakingManager.sol
```diff
@@ -0,0 +1,228 @@
...
+
+  // TODO: validation vs delegation should be weighted differently for rewards
+  // Function to calculate the user's reward.
+  function calculateReward(uint256 amount, uint256 startedAt, uint256 finishedAt) internal view returns (uint256) {
+    if (finishedAt <= startedAt || !stakingEnabled || finishedAt - startedAt < minStakingDuration) {
+      return 0;
+    }
+
+    uint256 stakingTime = finishedAt - startedAt;
+    return (stakingTime * rewardRate * amount) / (1 ether); // Assuming rewardRate is scaled appropriately
```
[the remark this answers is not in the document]
ceyonur: I don't think this reward calculation is even something we would use in production.

contracts/contracts/interfaces/staking-manager.md
```diff
@@ -0,0 +1,116 @@
...
+- (Optional) Validator being removed after minimum staking period
+
+After verification contract should perform the following:
+
+- Send a warp message to P-Chain for `SetSubnetValidatorWeightTx{messageID: relatedRegisterValidatorMessageID, nonce: MaxUint64, Weight: 0, }` transaction
+- Stop accruing rewards for the validator
+
+### receiveRemoveValidator
+
+Once the `SetSubnetValidatorWeightTx` is accepted, a warp message `InvalidValidatorRegisterMessage` will be relayed to the contract with `messageID` of the request. The message represents a messageID will not possibly become a valid validator. This essentialy means the message was not valid in the first place, or the validator has removed from the set. The InvalidValidatorRegisterMessage can either be a result of `removeValidator` from the contract or it can be originated directly from P-chain via `ExitValidatorSetTx`. In either case this ultimately tells the contract that a validator was evicted. See related ACP-77 section [here](https://github.com/avalanche-foundation/ACPs/tree/main/ACPs/77-reinventing-subnets#exitvalidatorsettx).
```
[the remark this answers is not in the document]
ceyonur: yes. I think following sentence reflects that, but if you think it's not clear lmk.
`This essentialy means the message was not valid in the first place, or the validator has removed from the set.`

contracts/contracts/interfaces/staking-manager.md
```diff
@@ -0,0 +1,116 @@
...
+
+- Verified warp message
+- The typecheck on the warp message (== InvalidValidatorRegisterMessage)
+- messageID is not empty
+
+At this point the contract should unlock all funds that were locked for the validator and issue locked rewards. If the validator was not removed by the contract itself, the contract should either give partial rewards or no rewards at all. This is because we don't know when the validator was removed.
+
+Rewards should not be unlocked or sent to the staker at this point. This is because we need to first perform a uptime check to determine the rewards.
+
+TODO: should we give a partial reward in case the validator got removed from P-chain (balance drained)? Possibly by timestamping the `InvalidValidatorRegisterMessage`.
```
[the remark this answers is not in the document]
ceyonur: yes. We have discussed about `ExitValidatorSetTx`. It won't result into a signable `InvalidValidatorRegisterMessage`, rather it will mark the validator `inactive` and not actually remove from the set. I think we can stop increasing uptime once they got `inactive`.

contracts/contracts/interfaces/staking-manager.md
```diff
@@ -0,0 +1,116 @@
...
+
+TODO: should we add a cooldown period for validator removal/readd?
```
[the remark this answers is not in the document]
ceyonur: I was thinking if it would be DoSable if someone removes and readds itself too quickly. but I think that should be good. We should however still need to add a cooldown for partial redeeming.

contracts/contracts/interfaces/staking-manager.md
ceyonur: yea i don't like it very much either. I'm not quite sure how to reflect "the message was invalid" + "the message won't be valid anymore. Maybe `DefunctMessage`?

contracts/contracts/interfaces/staking-manager.md
```diff
@@ -0,0 +1,126 @@
...
+
+- Verified warp message
+- The typecheck on the warp message (== ValidatorRegisteredMessage)
+- messageID is not empty
+- The message is not already used
+- (?) There is a pending request for this messageID
+- (?) The nodeID in the pending request is not a validator already
+- (?) Message is from P-Chain
+
+TODO: we might not want last 3 checks (with ?) to minimize reverts. The contract should be able to handle these cases without reverting. This is because the timing of the messages is not guaranteed and the contract should be able to handle out-of-order messages.
```
geoff-vball, line 62: @michaelkaplan13 

- If there isn't a pending request for this message ID, doesn't that mean we've been sent something invalid?
- The nodeID in the pending request being a validator already seems fine, we may send multiple requests to register.
- How would we handle a message not from the P-Chain here? Shouldn't that revert, as a malicious subnet could send a similar message?

ceyonur: 1- I guess yes, sending a msg from another subnet should be mostly covered before msg arrives here. But still we'd assume that it's not possible + we assume those pending requests get never deleted. Otherwise one can replay those messages. 
2- in theory we can send many requests, but P-chain won't accept them if there is already a validator with that nodeID. We should definetly let multiple requests to be sent in case there is no validator (to prevent blocking that nodeID), but I was not %100 sure if we should let those if there is a known validator for that nodeID. Mostly to prevent replay attacks. 
3- as told in 1 I was assuming the message would not possibly arrive at this point/not even signed. 

Looking at this again, it now seems that we should probably add all possible checks.

Issue #1230, opened 2024-07-09
Make AWM Primary Network Optimization configurable


martineckardt:
**Context and scope**
When running a Subnet with all validators in partial sync mode, it will not be able to verify incoming messages from the primary network. The optimization will try to gather signatures over the C-Chain's message from the Subnet's validator set. Since these do not sync the C-Chain they won't be able to sign the message. 

We should make this optimization configurable.

**Discussion and alternatives**
TBD

**Open questions**
TBD

michaelkaplan13:
FWIW, this optimization will need to be removed all together in the E upgrade in order to allow for subnets with only subnet-only validators to verify messages from the C-Chain. This will also need to be accounted for in the `awm-relayer` so that it is able to determine which validators to query for BLS signatures.

aaronbuchwald:
It could still be configurable in the E Upgrade if subnets opt in to validating the C-Chain. I'm not opposed to removing it completely in the E Upgrade if that's the easiest path, but it's worth evaluating the performance of aggregating a signature from >= 67% of the primary network before making that decision.

PR #1232, opened 2024-07-09
warp: add "requirePrimaryNetworkSigners" precompile config

darioush:
## Why this should be merged
https://github.com/ava-labs/subnet-evm/issues/1230

## How this works
Adds a precompile config option where "requirePrimaryNetworkSigners" can be added at warp activation (or de-activated and then re-activated with this flag by existing networks).

Note we can't use a "per node" config, since when the optimization is applied the validator bitset is modified.
Basically the existing rules do not allow for the primary network's validators to sign for themselves but expects only the subnet validator set. Since the entire validator set is replaced, this includes expectations for the bitset, etc.

## How this was tested
CI:
1. Expands UT for TestWarpMessageFromPrimaryNetwork (in precompile/contracts/warp/predicate_test.go) to include both cases for requirePrimaryNetworkSigners,

2. Modifies (expands) TestReceiveWarpMessage:
- first asserting warp messages can be received as expected with the default config (no RequirePrimaryNetworkSigners) for messages from subnets, non-P chain primary and P-Chain primary, this ensures no regression for existing networks,
- disabling then re-enabling the warp configuration with RequirePrimaryNetworkSigners (note disabling is required as it is not allowed allowed to update the precompile config)
- next asserting its behavior again with the RequirePrimaryNetworkSigners (non P-Chain primary network chains should use primary network).

I also added a check the result of the predicate in the block since the previous test asserts behavior about whether the block is valid or not in a different height context, but the block would be considered valid (but failed warp message), then the assertion that does a trace comes later.

It seemed preferable to verify the predicate results are as expected instead of re-ordering the existing code.

## How is this documented
Not yet


precompile/contracts/warp/config.go
```diff
@@ -202,7 +208,7 @@ func (c *Config) VerifyPredicate(predicateContext *precompileconfig.PredicateCon
 		context.Background(),
 		&warpMsg.UnsignedMessage,
 		predicateContext.SnowCtx.NetworkID,
-		warpValidators.NewState(predicateContext.SnowCtx), // Wrap validators.State on the chain snow context to special case the Primary Network
+		warpValidators.NewState(predicateContext.SnowCtx, func() bool { return c.RequirePrimaryNetworkSigners }), // Wrap validators.State on the chain snow context to special case the Primary Network
```
ceyonur, line 211: I think this is fine since we don't call `VerifyPredicate` in bootstrapping, right? Otherwise this might break predicate verification in older blocks. 

hmm or is it fine because in order to activate this config one needs to disable and then re-enable warp with this config?

plugin/evm/vm.go
```diff
@@ -998,7 +999,18 @@ func (vm *VM) CreateHandlers(context.Context) (map[string]http.Handler, error) {
 	}
 
 	if vm.config.WarpAPIEnabled {
-		validatorsState := warpValidators.NewState(vm.ctx)
+		requirePrimaryNetworkSigners := func() bool {
```
ceyonur, line 1002: I wonder if we should specifically enable this at EUpgrade. It does not seem required since one needs to disable/re-enable the precompile with this config (so basically they will perform upgrade), but wanted to get a confirmation.

precompile/contracts/warp/config.go
darioush: this config should always correspond to the active config of the precompile (so seems similar to the 2nd description)

plugin/evm/vm.go
darioush: I don't think it should be, since SoV network is an opt-in change then configuring warp precompile should be also done as part of that opt-in.
I think we should just test both types of activation in some e2e test (opt-in and default).

plugin/evm/vm_warp_test.go
```diff
@@ -408,74 +410,142 @@ func TestReceiveWarpMessage(t *testing.T) {
...
 	}()
 
-	acceptedLogsChan := make(chan []*types.Log, 10)
-	logsSub := vm.eth.APIBackend.SubscribeAcceptedLogsEvent(acceptedLogsChan)
-	defer logsSub.Unsubscribe()
+	// Subnet messages should use subnet signers
+	testReceiveWarpMessage(t, issuer, vm, false, false, upgrade.InitiallyActiveTime)
 
-	payloadData := avagoUtils.RandomBytes(100)
+	/// At genesis, primary network messages should use subnet signers
```
darioush, line 440:
```suggestion
	// At genesis, primary network messages should use subnet signers
```

plugin/evm/vm.go
```diff
@@ -1019,7 +1020,18 @@ func (vm *VM) CreateHandlers(context.Context) (map[string]http.Handler, error) {
...
+			if !ok {
+				return false
+			}
+			warpCfg, ok := warpCfgIntf.(*warpcontract.Config)
+			if !ok {
+				return false
+			}
+			return warpCfg.RequirePrimaryNetworkSigners
+		}
+		validatorsState := warpValidators.NewState(vm.ctx, requirePrimaryNetworkSigners)
```
ceyonur, line 1034: nit: does it make sense to have an interface rather than passing the function (so vm/warpConfig can implement it)?

ceyonur (approved):
overall LGTM, just a nit to convert the function to an interface

plugin/evm/vm_warp_test.go
```diff
@@ -408,74 +410,142 @@ func TestReceiveWarpMessage(t *testing.T) {
...
+	require.True(blockTime.Before(disableTime))
+	testReceiveWarpMessage(t, issuer, vm, true, false, blockTime)
 
+	// After re-enabled, primary network messages should use primary signers
+	testReceiveWarpMessage(t, issuer, vm, true, true, reEnableTime)
+}
+
+func testReceiveWarpMessage(
+	t *testing.T, issuer chan commonEng.Message, vm *VM,
+	fromPrimary, usePrimarySigners bool,
```
ARR4N, line 451: These booleans make it difficult to understand the call sites (made me review out of order because I didn't know what was going on). One fix is to introduce enums:

```go
type warmMsgFrom int

const (
    fromSubnet warpMsgFrom = iota
    fromPrimary
)

type useWarpMsgSigners int

const (
    subnetSigners warpMsgSigners = iota
    primarySigners
)
```

Then the call site is clear:

```go
testReceiveWarpMessage(t, issuer, vm, fromSubnet, subnetSigners, upgrade.InitiallyActiveTime)
```

Also see [Uber style guide](https://github.com/uber-go/guide/blob/master/style.md#avoid-naked-parameters).

plugin/evm/vm_warp_test.go
```diff
@@ -408,74 +410,142 @@ func TestReceiveWarpMessage(t *testing.T) {
...
+	issuer, vm, _, _ := GenesisVM(t, true, string(genesisJSON), "", string(upgradeBytes))
 
 	defer func() {
 		require.NoError(vm.Shutdown(context.Background()))
 	}()
 
-	acceptedLogsChan := make(chan []*types.Log, 10)
-	logsSub := vm.eth.APIBackend.SubscribeAcceptedLogsEvent(acceptedLogsChan)
-	defer logsSub.Unsubscribe()
+	// Subnet messages should use subnet signers
```
ARR4N, line 437: See my comment on `testReceiverWarpMessage()` definition, which would make this comment unnecessary.

plugin/evm/vm_warp_test.go
```diff
@@ -408,74 +410,142 @@ func TestReceiveWarpMessage(t *testing.T) {
...
+	testReceiveWarpMessage(t, issuer, vm, false, false, upgrade.InitiallyActiveTime)
 
-	payloadData := avagoUtils.RandomBytes(100)
+	// At genesis, primary network messages should use subnet signers
+	blockTime := upgrade.InitiallyActiveTime.Add(2 * time.Second) // for fees
```
ARR4N, line 441: `// for fees`

What does this mean?

plugin/evm/vm_warp_test.go
```diff
@@ -408,74 +410,142 @@ func TestReceiveWarpMessage(t *testing.T) {
...
+			weight:    weight,
+		}
+	}
 
-	nodeID2 := ids.GenerateTestNodeID()
-	blsSecretKey2, err := bls.NewSecretKey()
-	require.NoError(err)
-	blsPublicKey2 := bls.PublicFromSecretKey(blsSecretKey2)
-	blsSignature2 := bls.Sign(blsSecretKey2, unsignedMessage.Bytes())
+	var (
```
ARR4N, line 490: Is there a reason that a lot of Ava code uses `var` blocks instead of just `:=`? It jumps out at me as atypical, but "local" style trumps that, especially if it was an active decision.

[Uber is explicit about not doing this](https://github.com/uber-go/guide/blob/master/style.md#local-variable-declarations) while Google only mentions it for nil slices.

plugin/evm/vm_warp_test.go
```diff
@@ -408,74 +410,142 @@ func TestReceiveWarpMessage(t *testing.T) {
...
-					PublicKey: blsPublicKey1,
-					Weight:    50,
-				},
-				nodeID2: {
-					NodeID:    nodeID2,
-					PublicKey: blsPublicKey2,
-					Weight:    50,
-				},
-			}, nil
+			toValidatorOutput := func(signers []signer) map[ids.NodeID]*validators.GetValidatorOutput {
```
ARR4N, line 526: This function is unnecessary. Its contents can just be moved to after the `validators` variable is set.

plugin/evm/vm_warp_test.go
```diff
@@ -519,12 +589,25 @@ func TestReceiveWarpMessage(t *testing.T) {
...
+	headerPredicateResultsBytes, ok := predicate.GetPredicateResultBytes(ethBlock.Extra())
+	require.True(ok)
+	results, err := predicate.ParseResults(headerPredicateResultsBytes)
+	require.NoError(err)
+	txResultsBytes := results.GetResults(
+		getVerifiedWarpMessageTx.Hash(),
+		warp.ContractAddress,
+	)
+	bitset := set.BitsFromBytes(txResultsBytes)
+	require.Zero(bitset.Len()) // Empty bitset indicates success
```
ARR4N, line 609: Why do you need to go via a `bitset`? Am I correct in understanding that you want all bytes to be zero?

For my understanding: why does all-zero mean things worked?

ARR4N:
> ### How this was tested
> CI

I think the intention of this block is an explanation of what tests are being performed. I see that you made changes to the unit tests but I don't have the context to understand why they are sufficient to prove that this works.

precompile/contracts/warp/config.go
michaelkaplan13, line 211: I think we need to make a distinction between the `sourceBlockchainID` of the unsigned message here.

The reason is that validator management related messages from the P-Chain will always need to be signed by only the receiving L1's validator set, because P-Chain validators that do not track the given L1 will not be willing to sign the message for it. Given that all validators still need to track the P-Chain, this should be possible for now.

The precompile configuration flag should only apply for messages where the `sourceChainID` is the C-Chain I believe.

precompile/contracts/warp/config.go
```diff
@@ -59,6 +60,11 @@ func NewConfig(blockTimestamp *uint64, quorumNumerator uint64) *Config {
 	}
 }
 
+func (c *Config) WithRequirePrimaryNetworkSigners(requirePrimaryNetworkSigners bool) *Config {
```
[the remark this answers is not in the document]
michaelkaplan13: Possibly dumb question: is this function meant to only be used by tests? Don't see it being called from anywhere else currently.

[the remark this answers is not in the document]
ceyonur: I think we can just take this a param in the `NewConfig` and modify tests accordingly.

plugin/evm/vm.go
```diff
@@ -1019,7 +1020,18 @@ func (vm *VM) CreateHandlers(context.Context) (map[string]http.Handler, error) {
 	}
 
 	if vm.config.WarpAPIEnabled {
-		validatorsState := warpValidators.NewState(vm.ctx)
+		requirePrimaryNetworkSigners := func() bool {
+			warpCfgIntf, ok := vm.currentRules().ActivePrecompiles[warpcontract.ContractAddress]
```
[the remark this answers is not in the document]
darioush: It's short for interface I'll change it to clarify.

plugin/evm/vm_warp_test.go
darioush: adding a comment to explain it.
for more context, avalanche dynamic fees assumes a 2s block time, while blocks can be issued faster (even at the same time), the fees will exceed what's specified in the UT transaction.

precompile/contracts/warp/config.go
darioush: made a special case for the P-Chain per this comment.
Additionally, there is an alternative implementation we can consider: https://github.com/ava-labs/subnet-evm/pull/1232/files

precompile/contracts/warp/config.go
```diff
@@ -47,22 +47,24 @@ var (
 // adds specific configuration for Warp.
 type Config struct {
 	precompileconfig.Upgrade
-	QuorumNumerator uint64 `json:"quorumNumerator"`
+	QuorumNumerator              uint64 `json:"quorumNumerator"`
+	RequirePrimaryNetworkSigners bool   `json:"requirePrimaryNetworkSigners"`
```
darioush, line 51: We can make this a different type, just need to make sure it serializes as a bool, if readability or passing inline `true`/`false` is a concern.

ceyonur (approved):

ARR4N (approved):
LGTM.

Caveat: I don't know enough about the specific test cases in the table nor the implication of `mySubnetID` in `validators.NewState()` to comment on them, but I think it's all now clear enough that I'm deferring to Cey's existing Approval.

michaelkaplan13 (approved):
Makes sense to me 🙏

PR #1240, opened 2024-07-17
refactor: make `Role` type a `uint64` and use constants

ARR4N:

### DIFF
--- a/precompile/allowlist/allowlist.go
+++ b/precompile/allowlist/allowlist.go
@@ -41,7 +41,11 @@ var (
 func GetAllowListStatus(state contract.StateDB, precompileAddr common.Address, address common.Address) Role {
 	// Generate the state key for [address]
 	addressKey := common.BytesToHash(address.Bytes())
-	return Role(state.GetState(precompileAddr, addressKey))
+	r, err := RoleFromHash(state.GetState(precompileAddr, addressKey))
+	if err != nil {
+		panic(err) // DO NOT MERGE
+	}
+	return r
 }
 
 // SetAllowListRole sets the permissions of [address] to [role] for the precompile
--- a/precompile/allowlist/role.go
+++ b/precompile/allowlist/role.go
@@ -5,56 +5,46 @@ package allowlist
 
 import (
 	"errors"
+	"fmt"
 	"math/big"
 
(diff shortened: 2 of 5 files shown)


precompile/allowlist/role.go
```diff
@@ -5,79 +5,80 @@ package allowlist
...
-
-// IsAdmin returns true if [r] indicates the permission to modify the allow list.
-func (r Role) IsAdmin() bool {
-	switch r {
-	case AdminRole:
-		return true
-	default:
-		return false
-	}
+// IsValid returns whether or not `r` is a valid value.
```
darioush, line 40: nit: while ` is more broadly used avalanche code bases use [] for this type of notation.

precompile/allowlist/role.go
```diff
@@ -107,16 +108,27 @@ func (r Role) String() string {
...
-	default:
-		return Role{}, ErrInvalidRole
+// A Big value may or may not be representable as a uint64. Such types include
+// [big.Int] and [uint256.Int].
+type Big interface {
+	Uint64() uint64
+	IsUint64() bool
+}
+
+// RoleFromBig converts `u.Uint64()` into a [Role] if `u.IsUint64()` is true.
```
darioush, line 122: nit
```suggestion
// RoleFromBig converts `b.Uint64()` into a [Role] if `b.IsUint64()` is true.
```

precompile/allowlist/allowlist.go
```diff
@@ -41,7 +41,11 @@ var (
 func GetAllowListStatus(state contract.StateDB, precompileAddr common.Address, address common.Address) Role {
 	// Generate the state key for [address]
 	addressKey := common.BytesToHash(address.Bytes())
-	return Role(state.GetState(precompileAddr, addressKey))
+	r, err := RoleFromHash(state.GetState(precompileAddr, addressKey))
+	if err != nil {
+		panic(err) // DO NOT MERGE
```
darioush, line 46: what should the actual behavior be in this case?

darioush (commented):
this seems like an improvement so far, could we add the context/motivation to the description when it is r4r?

JonathanOppenheimer:
Conferred with Arran - can close.

Issue #1259, opened 2024-07-29
ACP-118: Implement message handlers 


ceyonur:
We have merged the AvalancheGo counterpart of [ACP-118 ](https://github.com/avalanche-foundation/ACPs/tree/main/ACPs/118-warp-signature-request): https://github.com/ava-labs/avalanchego/pull/3218

We should start converting our existing to comfort ACP-118 message types.

PR #1260, opened 2024-07-30
add p2p handler for acp-118 sig request

darioush:
## Why this should be merged
https://github.com/ava-labs/subnet-evm/issues/1259 

## How this works
Adds support for ACP-118 message types.
Assumes that 32 bytes messages should be interpreted as a block hash
This PR mostly adopts the existing code to the new handlers & protos as defined in ACP-118.

## How this was tested
CI

## How is this documented
N/A

### DIFF
--- a/warp/handlers/signature_request_p2p.go
+++ b/warp/handlers/signature_request_p2p.go
@@ -14,6 +14,7 @@ import (
 	"github.com/ava-labs/avalanchego/snow/engine/common"
 	"github.com/ava-labs/avalanchego/utils/crypto/bls"
 	avalancheWarp "github.com/ava-labs/avalanchego/vms/platformvm/warp"
+	"github.com/ava-labs/avalanchego/vms/platformvm/warp/payload"
 	"github.com/ava-labs/subnet-evm/warp"
 	"google.golang.org/protobuf/proto"
 )
@@ -59,27 +60,45 @@ func (s *SignatureRequestHandlerP2P) AppRequest(
 		}
 	}
 
-	var (
-		sig [bls.SignatureLen]byte
-		err error
-	)
-	if len(req.Message) == ids.IDLen {
-		// Treat requests with 32-byte [message] as a block hash.
-		var blockID ids.ID
-		copy(blockID[:], req.Message)
-		sig, err = s.GetBlockSignature(blockID)
+	unsignedMessage, err := avalancheWarp.ParseUnsignedMessage(req.Message)
+	if err != nil {
+		return nil, &common.AppError{
+			Code:    ErrFailedToParse,
+			Message: "failed to parse unsigned message: " + err.Error(),
+		}
+	}
+	parsed, err := payload.Parse(unsignedMessage.Payload)
+	if err != nil {
+		return nil, &common.AppError{
+			Code:    ErrFailedToParse,
+			Message: "failed to parse payload: " + err.Error(),
+		}
+	}
+
+	var sig [bls.SignatureLen]byte
+	switch p := parsed.(type) {
+	case *payload.AddressedCall:
+		// Note we pass the unsigned message ID to GetMessageSignature since
+		// that is what the backend expects.
+		// However, we verify the types and format of the payload to ensure
+		// the message conforms to the ACP-118 spec.
+		sig, err = s.GetMessageSignature(unsignedMessage.ID())
+		if err != nil {
+			s.stats.IncMessageSignatureMiss()
+		} else {
+			s.stats.IncMessageSignatureHit()
+		}
+	case *payload.Hash:
+		sig, err = s.GetBlockSignature(p.Hash)
 		if err != nil {
 			s.stats.IncBlockSignatureMiss()
 		} else {
 			s.stats.IncBlockSignatureHit()
 		}
-	} else {
-		// Otherwise, treat the request as an unsigned warp message.
-		sig, err = s.GetMessageSignature(req.Message)
-		if err != nil {
-			s.stats.IncMessageSignatureMiss()
-		} else {
-			s.stats.IncMessageSignatureHit()
+	default:
+		return nil, &common.AppError{
+			Code:    ErrFailedToParse,
+			Message: "unknown payload type",
 		}
 	}
 	if err != nil {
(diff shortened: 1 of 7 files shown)


warp/handlers/signature_request_p2p.go
```diff
@@ -0,0 +1,146 @@
...
+// SignatureRequestHandlerP2P serves warp signature requests using the p2p
+// framework from avalanchego. It is a peer.RequestHandler for
+// message.MessageSignatureRequest.
+type SignatureRequestHandlerP2P struct {
+	backend warp.Backend
+	codec   codec.Manager
+	stats   *handlerStats
+}
+
+func NewSignatureRequestHandlerP2P(backend warp.Backend, codec codec.Manager) *SignatureRequestHandlerP2P {
```
darioush, line 38: We can probably remove the existing handler, unless we have some concerns about it being in use.

ceyonur: I think we need to have this implemented in relayer as well before removing from here? cc @michaelkaplan13 

I'd say we should keep it around until E-Upgrade.

warp/handlers/signature_request_p2p.go
```diff
@@ -0,0 +1,146 @@
...
+			s.stats.IncBlockSignatureHit()
+		}
+	} else {
+		// Otherwise, treat the request as an unsigned warp message.
+		sig, err = s.GetMessageSignature(req.Message)
+		if err != nil {
+			s.stats.IncMessageSignatureMiss()
+		} else {
+			s.stats.IncMessageSignatureHit()
+		}
```
ceyonur, line 83: I wonder if we should use our internal codec here to parse `req.Message`. I don't think switching based on length is a reliable way.

On the other side, if we need to parse another data for payload, it will again make these messages specific to per-VM. I think in the long run we should probably define those in AvalancheGo, maybe as protos, so that both relayer and VMs can re-use them.
cc @michaelkaplan13 @StephenButtolph

warp/handlers/signature_request_p2p.go
```diff
@@ -0,0 +1,146 @@
...
+			Code:    ErrFailedToParse,
+			Message: "failed to unmarshal request: " + err.Error(),
+		}
+	}
+
+	var (
+		sig [bls.SignatureLen]byte
+		err error
+	)
+	if len(req.Message) == ids.IDLen {
```
michaelkaplan13, line 66: The `req.Message` to be signed here will be the raw bytes of the unsigned Warp message, so it will need to be parsed as an `UnsignedMessage` [here](https://github.com/ava-labs/avalanchego/blob/master/vms/platformvm/warp/unsigned_message.go#L39).

Then the `Payload` of the message can be determined to either be an `AddressedCall` or `Hash` based on `typeID`s defined [here](https://github.com/ava-labs/avalanchego/tree/master/vms/platformvm/warp/payload).

warp/handlers/signature_request_p2p.go
michaelkaplan13: Oh sorry just saw this comment after adding my own, but yeah the `req.Message` will always be warp `UnsignedMessage` so we can parse that and then switch based on the typeIDs in it's payload. (Added the comment [here](https://github.com/ava-labs/subnet-evm/pull/1260/files#r1696960581) before seeing this).

warp/handlers/signature_request_p2p.go
michaelkaplan13: Yeah the AWM relayer uses the existing handlers today, so I think we should leave them in place until after the E-upgrade goes live and everyone has a chance to switch. They can probably be removed soon after that though 👍

warp/handlers/signature_request_p2p.go
ceyonur: does that mean we have first parse it for `UnsignedMessage` and then do additinal parsing/handling for `AddressedCall` vs `Hash`?

ceyonur: i.e:
```
unsignedWarpMsg := new(UnsignedMessage)
codec.Unmarshal(req.Message, unsignedWarpMsg)

blockhashTtype BlockHashType | AddressedCallType
codec.Unmarshal(blockHashType, payload)
blockHashType.Handle()
```

michaelkaplan13: Yeah I believe so. I think we can use the [`warp.ParseUnsignedMessage`](https://github.com/ava-labs/avalanchego/blob/master/vms/platformvm/warp/unsigned_message.go#L39), [`payload.ParseAddressedCall`](https://github.com/ava-labs/avalanchego/blob/master/vms/platformvm/warp/payload/addressed_call.go#L33), and [`payload.ParseHash`](https://github.com/ava-labs/avalanchego/blob/master/vms/platformvm/warp/payload/hash.go#L29) helpers directly.

warp/handlers/signature_request_p2p.go
darioush: Sounds good. I can add a comment about this.

warp/handlers/signature_request_p2p_test.go
```diff
@@ -38,7 +38,9 @@ func TestMessageSignatureHandlerP2P(t *testing.T) {
 	backend, err := warp.NewBackend(snowCtx.NetworkID, snowCtx.ChainID, warpSigner, warptest.EmptyBlockClient, database, 100, [][]byte{offchainMessage.Bytes()})
 	require.NoError(t, err)
 
-	msg, err := avalancheWarp.NewUnsignedMessage(snowCtx.NetworkID, snowCtx.ChainID, []byte("test"))
+	offchainPayload, err := payload.NewAddressedCall([]byte{0, 0, 0}, []byte("test"))
```
darioush, line 41: Note this "change" effectively adds the requirement that offchain messages be either addressed call or hash payload

warp/handlers/signature_request_p2p.go
```diff
@@ -0,0 +1,159 @@
...
+	if err != nil {
+		return nil, &common.AppError{
+			Code:    ErrFailedToMarshal,
+			Message: "failed to marshal response: " + err.Error(),
+		}
+	}
+	return respBytes, nil
+}
+
+func (s *SignatureRequestHandlerP2P) GetMessageSignature(messageID ids.ID) ([bls.SignatureLen]byte, error) {
```
michaelkaplan13, line 124: This is future looking, but do you want to add a comment here that when the the existing signature request handlers are removed at some point after the E-upgrade, the `backend` can stop storing the `messageID` -> `rawMessage` mapping and instead just store the `messageID` keys? When a message request is received, the check can be if the message ID of the parsed message exists, and assuming it does, the raw message provided in the request can be used to create the signature.

michaelkaplan13 (approved):
This all makes sense to me. 🙏

ceyonur (approved):
LGTM! (just a comment that is not directly related)

PR #1261, opened 2024-07-31
refactor: remove `params` dependency on `modules`

ARR4N:
## Why this should be merged

Importing `core/types` from `precompile/{modules,contract}` results in a circular dependency because `types` -> `params` -> `precompile/modules` -> `precompile/contract` -> `types`.

## How this works

`params` only depends on `modules` for unmarshalling JSON, so all such functionality is abstracted into a new `paramsjson` package. `paramsjson.Unmarshal()` is a drop-in replacement for `json.Unmarshal()` but only accepts problematic `params.<T>` types. Each of those types is also updated so that its `UnmarshalJSON()` method returns an error instructing the user to use `paramsjson` instead.

## How this was tested

All existing CI tests.

## How is this documented

Standard Go comments.


JonathanOppenheimer:
Conferred with Arran - can close.

PR #1263, opened 2024-08-01
Uptime tracking

ceyonur:
This PR contains an unfinished work in AvalancheGo and depends on the PR: https://github.com/ava-labs/avalanchego/pull/3404

## Why this should be merged

* Connects uptime tracking modules with VM
* Starts a new goroutine that fetches current validator set from P-chain to and loads to subnet-evm state
*  `GetCurrentValidator` APIs. 

## How this was tested

* Added UTs
* Local network testing

## How is this documented

* Not yet


plugin/evm/vm.go
```diff
@@ -1345,3 +1398,99 @@ func (vm *VM) createDatabase(dbConfig avalancheNode.DatabaseConfig) (database.Da
...
+	if err := vm.validatorState.WriteState(); err != nil {
+		return fmt.Errorf("failed to write validator state: %w", err)
+	}
+
+	// TODO: add metrics
+	log.Debug("validator update complete", "duration", time.Since(now))
+	return nil
+}
+
+// TODO: cache the last updated height and then load if needed
```
ceyonur, line 1463: not sure if this would bring us any viable advantage, since P-chain height will likely be changing every loadValidatorsFrequency (1 minute).

plugin/evm/vm_validators_state_test.go
```diff
@@ -0,0 +1,334 @@
...
+				require.NoError(err)
+			}
+			// enable mock listener
+			ctrl := gomock.NewController(tt)
+			mockListener := validators.NewMockStateCallbackListener(ctrl)
+			test.registerMockListenerCalls(mockListener)
+
+			validatorState.RegisterListener(mockListener)
+			require.NoError(loadValidators(validatorState, test.newValidators))
+			ctrl.Finish()
```
darioush, line 331: nit: I think calls to ctrl.Finish are not required in newer versions of gomock

plugin/evm/vm_validators_state_test.go
```diff
@@ -0,0 +1,334 @@
...
+	require.NoError(err, "error initializing GenesisVM")
+
+	// Test case 1: state should not be populated until bootstrapped
+	require.Equal(0, vm.validatorState.GetValidationIDs().Len())
+	_, _, err = vm.uptimeManager.CalculateUptime(testNodeIDs[0])
+	require.ErrorIs(database.ErrNotFound, err)
+	require.False(vm.uptimeManager.StartedTracking())
+
+	// Test case 2: state should be populated after bootstrapped
+	require.NoError(vm.SetState(context.Background(), snow.Bootstrapping))
```
darioush, line 86: should we set the state to Bootstrapping before checking that it's not populated, such as to line 80? (i.e., only gets populated in NormalOps)

plugin/evm/vm.go
```diff
@@ -674,38 +695,62 @@ func (vm *VM) initChainState(lastAcceptedBlock *types.Block) error {
...
+	vm.blockChain.InitializeSnapshots()
+
+	return nil
+}
+
+// onNormalOperationsStarted marks this VM as bootstrapped
+func (vm *VM) onNormalOperationsStarted() error {
+	if vm.bootstrapped.Get() {
+		return nil
+	}
```
darioush, line 730: why is this check needed?

plugin/evm/vm.go
darioush: I think we can remove this TODO

plugin/evm/vm.go
```diff
@@ -256,6 +266,9 @@ type VM struct {
 	ethTxPushGossiper  avalancheUtils.Atomic[*gossip.PushGossiper[*GossipEthTx]]
 	ethTxPullGossiper  gossip.Gossiper
 
+	uptimeManager  uptime.PausableManager
+	validatorState validators.State
```
darioush, line 270: could we put the additions in a separate file & ideally a separate struct? (these fields with validatorsDB & the related functions) 
This struct can get called from the VM for initialize, state change, shutdown & for connect/disconnect.
My main concerns are:
- Size of vm.go file (goes from ~1400 -> 1500) with this change,
- Having smaller structs means less tangling of initialization logic and possibly shorter UTs.

plugin/evm/vm_validators_state_test.go
```diff
@@ -0,0 +1,334 @@
...
+		},
+	}
+	vm.ctx.ValidatorState = testState
+	// set VM as bootstrapped
+	require.NoError(vm.SetState(context.Background(), snow.Bootstrapping))
+	require.NoError(vm.SetState(context.Background(), snow.NormalOp))
+
+	// new validator should be added to the state eventually after validatorsLoadFrequency
+	require.EventuallyWithT(func(c *assert.CollectT) {
+		assert.Equal(c, 4, vm.validatorState.GetValidatorIDs().Len())
```
darioush, line 152: nit: can use assert.Len

plugin/evm/vm_validators_state_test.go
```diff
@@ -0,0 +1,334 @@
...
+	// Test case 1: state should not be populated until bootstrapped
+	require.Equal(0, vm.validatorState.GetValidationIDs().Len())
+	_, _, err = vm.uptimeManager.CalculateUptime(testNodeIDs[0])
+	require.ErrorIs(database.ErrNotFound, err)
+	require.False(vm.uptimeManager.StartedTracking())
+
+	// Test case 2: state should be populated after bootstrapped
+	require.NoError(vm.SetState(context.Background(), snow.Bootstrapping))
+	require.NoError(vm.SetState(context.Background(), snow.NormalOp))
+	require.Equal(3, vm.validatorState.GetValidationIDs().Len())
```
darioush, line 88: nit: can be require.Len

plugin/evm/vm_validators_state_test.go
```diff
@@ -0,0 +1,334 @@
...
+		dbManager,
+		genesisBytes,
+		[]byte(""),
+		[]byte(""),
+		issuer,
+		[]*commonEng.Fx{},
+		appSender,
+	)
+	require.NoError(err, "error initializing GenesisVM")
+	require.Equal(3, vm.validatorState.GetValidationIDs().Len())
```
darioush, line 111: nit: can be require.Len

plugin/evm/vm.go
```diff
@@ -239,7 +249,7 @@ type VM struct {
 	// Metrics
 	sdkMetrics *prometheus.Registry
 
-	bootstrapped bool
+	bootstrapped avalancheUtils.Atomic[bool]
```
darioush, line 252: curious what new concurrent access pattern is added for bootstrapped?

plugin/evm/service.go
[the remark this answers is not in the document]
ceyonur: moved changes to https://github.com/ava-labs/subnet-evm/pull/1374

plugin/evm/vm_validators_state_test.go
ceyonur: I think VM is in bootstrapping state by default, but yeah let's make it more explicit.

plugin/evm/vm.go
ceyonur: I think this is in case we would set the state twice.(derived from [P-chain](https://github.com/ava-labs/avalanchego/blob/master/vms/platformvm/vm.go#L341C1-L343C3))

plugin/evm/vm.go
ceyonur: ahh this was for the locked uptime calculator https://github.com/ava-labs/subnet-evm/pull/1367/files#diff-3be5ea9687f8af021d9baaa6584d174a7975b181b604e1a759e4ad2e40136912R515. I moved those to the separate PR but forgot to revert the changes here.

plugin/evm/vm.go
ceyonur: I will try that with a separate PR (including signing changes). Yea I also think VM does too much work of dispatching the update, doing the update itself etc. So they can be moved to a manager-maintainer like struct.

plugin/evm/vm.go
```diff
@@ -129,6 +134,8 @@ const (
 	txGossipThrottlingPeriod             = 10 * time.Second
 	txGossipThrottlingLimit              = 2
 	txGossipPollSize                     = 1
+
+	loadValidatorsFrequency = 1 * time.Minute
```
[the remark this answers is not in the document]
ceyonur: There is none

plugin/evm/vm.go
darioush: I see... makes sense.

darioush:
This LGTM, see this comment/question about the handling of possible duplicate nodeIDs between the staker set and SoVs https://github.com/ava-labs/avalanchego/pull/3404#issuecomment-2458529178

darioush (approved):

darioush (dismissed):

darioush (dismissed):

plugin/evm/validators/state.go
```diff
@@ -98,23 +103,57 @@ func (s *state) GetStartTime(nodeID ids.NodeID) (time.Time, error) {
...
+	if data.IsActive != vdr.IsActive {
+		data.IsActive = vdr.IsActive
+		updated = true
+		for _, listener := range s.listeners {
+			listener.OnValidatorStatusUpdated(data.validationID, data.NodeID, data.IsActive)
+		}
+	}
+
+	if data.Weight != vdr.Weight {
+		data.Weight = vdr.Weight
```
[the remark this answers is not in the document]
ceyonur: No, P-Chain's `GetCurrentValidatorSet` should never have that.

plugin/evm/uptime/pausable_manager.go
```diff
@@ -58,16 +58,10 @@ func (p *pausableManager) Disconnect(nodeID ids.NodeID) error {
...
-// If a node is paused, it will not be tracked
-func (p *pausableManager) StartTracking(nodeIDs []ids.NodeID) error {
-	activeNodeIDs := make([]ids.NodeID, 0, len(nodeIDs))
-	for _, nodeID := range nodeIDs {
-		if !p.IsPaused(nodeID) {
-			activeNodeIDs = append(activeNodeIDs, nodeID)
-		}
-	}
-	return p.Manager.StartTracking(activeNodeIDs)
+// IsConnected returns true if the node with the given ID is connected to the uptime.Manager
```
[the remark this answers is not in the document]
ceyonur: updated

darioush (approved):

plugin/evm/uptime/pausable_manager.go
```diff
@@ -58,16 +58,10 @@ func (p *pausableManager) Disconnect(nodeID ids.NodeID) error {
...
-	activeNodeIDs := make([]ids.NodeID, 0, len(nodeIDs))
-	for _, nodeID := range nodeIDs {
-		if !p.IsPaused(nodeID) {
-			activeNodeIDs = append(activeNodeIDs, nodeID)
-		}
-	}
-	return p.Manager.StartTracking(activeNodeIDs)
+// IsConnected returns true if the node with the given ID is connected to this manager
+// Note: Inner manager may have a different view of the connection status due to pausing
+func (p *pausableManager) IsConnected(nodeID ids.NodeID) bool {
```
michaelkaplan13, line 63: Possibly dumb question, where is `IsConnected` used from?

ceyonur: in the API (in another PR) https://github.com/ava-labs/subnet-evm/pull/1374/files#diff-42f3e9fd1f7cfff775643d1bad698671c0a34b8e3d3e5ab867626fcc0fa6fb8fR50

plugin/evm/validators/state.go
```diff
@@ -293,3 +339,14 @@ func (v *validatorData) getLastUpdated() time.Time {
 func (v *validatorData) getStartTime() time.Time {
 	return time.Unix(int64(v.StartTime), 0)
 }
+
+// constantsAreUnmodified returns true if the constants of this validator have
+// not been modified compared to the other validator.
+func (v *validatorData) constantsAreUnmodified(o interfaces.Validator) bool {
+	if v.validationID != o.ValidationID {
```
michaelkaplan13, line 346: Is there ever a case where we would expect to call this with a different validation ID? Seems likely that would indicate a bug itself and shouldn't be allowed.

plugin/evm/vm.go
```diff
@@ -674,38 +698,62 @@ func (vm *VM) initChainState(lastAcceptedBlock *types.Block) error {
...
 	ctx, cancel := context.WithCancel(context.TODO())
 	vm.cancel = cancel
 
+	// update validators first
+	if err := vm.performValidatorUpdate(ctx); err != nil {
+		return fmt.Errorf("failed to update validators: %w", err)
+	}
+	vdrIDs := vm.validatorState.GetNodeIDs().List()
+	// then start tracking with updated validators
+	if err := vm.uptimeManager.StartTracking(vdrIDs); err != nil {
```
michaelkaplan13, line 745: It looks like this is the only place where we call `StartTracking`. Do we also need to call `StartTracking` for new validators that are added during normal execution?

plugin/evm/vm.go
```diff
@@ -1345,3 +1405,97 @@ func (vm *VM) createDatabase(dbConfig avalancheNode.DatabaseConfig) (database.Da
...
+	if err := loadValidators(vm.validatorState, currentValidatorSet); err != nil {
+		return fmt.Errorf("failed to load current validators: %w", err)
+	}
+
+	// write validators to the database
+	if err := vm.validatorState.WriteState(); err != nil {
+		return fmt.Errorf("failed to write validator state: %w", err)
+	}
+
+	// TODO: add metrics
```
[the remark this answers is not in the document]
ceyonur: no, they will be in a separate PR with refactors.

plugin/evm/vm.go
michaelkaplan13: Gotcha, thanks. The `StartTracking` name is a little confusing in that case, but it was pre-existing from the interface so not worth changing.

Could you add the context that `StartTracking` is used to update the uptimes of existing validators to account for the time we were offline as a comment on this line though?

plugin/evm/validators/state.go
ceyonur: It's not possible to update a validator with a different vID. It should return a not found. That function is more like `newOrCanBeUpdated"` but we only use it in update. so by just looking at that function it looks like we should not allow this. changed it.

michaelkaplan13 (approved):

darioush (approved):

PR #1272, opened 2024-08-08
nits w/ upstream

darioush:
- doesn't seem core package needs embed
- field ordering can match upstream


paramsext/precompile_upgrade.go
ceyonur, line 1: I think this is a leftover from another work.

darioush: indeed

ceyonur (approved):

PR #1289, opened 2024-08-15
nit: resolve TODO in api_test.go

darioush:
## Why this should be merged
resolves the todo and reduces diff w/ upstream


internal/ethapi/api_test.go
```diff
@@ -1487,7 +1487,6 @@ func setupReceiptBackend(t *testing.T, genBlocks int) (*testBackend, []common.Ha
 			b.AddTx(tx)
 			txHashes[i] = tx.Hash()
 		}
-		// b.SetPoS()
```
darioush, line 1490: removing this just removes diff with upstream

ceyonur (approved):

PR #1290, opened 2024-08-15
nit: reduce diff with upstream

darioush:
since london is activated I think this makes no difference.
it looks like this upstream

### DIFF
--- a/internal/ethapi/transaction_args.go
+++ b/internal/ethapi/transaction_args.go
@@ -216,11 +216,11 @@ func (args *TransactionArgs) setSubnetEVMFeeDefault(ctx context.Context, head *t
 		// Set the max fee to be 2 times larger than the previous block's base fee.
 		// The additional slack allows the tx to not become invalidated if the base
 		// fee is rising.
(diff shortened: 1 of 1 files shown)


internal/ethapi/transaction_args.go
```diff
@@ -217,11 +216,11 @@ func (args *TransactionArgs) setSubnetEVMFeeDefault(ctx context.Context, head *t
...
 		// The additional slack allows the tx to not become invalidated if the base
 		// fee is rising.
-		gasFeeCap := new(big.Int).Add(
-			(*big.Int)(args.MaxPriorityFeePerGas),
+		val := new(big.Int).Add(
+			args.MaxPriorityFeePerGas.ToInt(),
 			new(big.Int).Mul(head.BaseFee, big.NewInt(2)),
 		)
-		args.MaxFeePerGas = (*hexutil.Big)(gasFeeCap)
+		args.MaxFeePerGas = (*hexutil.Big)(val)
```
darioush, line 223: this part should be no-op

ceyonur (approved):

PR #1292, opened 2024-08-19
Types and refactoring for ACP-77 ValidatorUptime messages

darioush:
## Why this should be merged
Types and refactoring for ACP-77 ValidatorUptime messages

## How this works
Refactors the warp backend to sign messages that return true for a `MessageHandler` interface.

## How this was tested
CI

## How is this documented
Not yet


warp/backend.go
```diff
@@ -142,15 +157,23 @@ func (b *backend) AddMessage(unsignedMessage *avalancheWarp.UnsignedMessage) err
...
+	messageID := unsignedMessage.ID()
+
 	log.Debug("Getting warp message from backend", "messageID", messageID)
 	if sig, ok := b.messageSignatureCache.Get(messageID); ok {
 		return sig, nil
 	}
 
-	unsignedMessage, err := b.GetMessage(messageID)
+	var err error
+	for _, v := range append(b.messageValidators, b) {
```
ceyonur, line 169: do we expect other interfaces? Rather than adding interfaces via `AddMessageValidator` should message payloads define their own validateMessage function?

warp/backend.go
```diff
@@ -30,14 +30,24 @@ type BlockClient interface {
...
 // Backend tracks signature-eligible warp messages and provides an interface to fetch them.
 // The backend is also used to query for warp message signatures by the signature request handler.
 type Backend interface {
 	// AddMessage signs [unsignedMessage] and adds it to the warp backend database
 	AddMessage(unsignedMessage *avalancheWarp.UnsignedMessage) error
 
-	// GetMessageSignature returns the signature of the requested message hash.
-	GetMessageSignature(messageID ids.ID) ([bls.SignatureLen]byte, error)
+	// AddMessageValidator adds a validator to the backend. The backend will sign
+	// messages that pass any of the validators, in addition to those known in the db.
```
ceyonur, line 46: IMHO it's kind-of odd to expect message to be validated if "any" of these validators pass. Maybe I did not fully understand the case but shouldn't we just need to have validators on payload types rather than adding them to here?

warp/backend.go
```diff
@@ -142,15 +157,23 @@ func (b *backend) AddMessage(unsignedMessage *avalancheWarp.UnsignedMessage) err
 	return nil
 }
 
-func (b *backend) GetMessageSignature(messageID ids.ID) ([bls.SignatureLen]byte, error) {
+func (b *backend) GetMessageSignature(unsignedMessage *avalancheWarp.UnsignedMessage) ([bls.SignatureLen]byte, error) {
```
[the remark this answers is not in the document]
darioush: it is so the backend can sign arbitrary messages on the fly (it will need to inspect their contents, such as for validator uptime message, so knowing the ID aka hash is not enough).
I didn't want to make another struct that has access to the bls signer key.
Do you suggest a different code organization here?

warp/handlers/signature_request.go
```diff
@@ -45,13 +45,20 @@ func (s *SignatureRequestHandler) OnMessageSignatureRequest(ctx context.Context,
...
+	var signature [bls.SignatureLen]byte
+	unsignedMessage, err := s.backend.GetMessage(signatureRequest.MessageID)
 	if err != nil {
-		log.Debug("Unknown warp signature requested", "messageID", signatureRequest.MessageID)
+		log.Debug("Unknown warp message requested", "messageID", signatureRequest.MessageID)
 		s.stats.IncMessageSignatureMiss()
-		signature = [bls.SignatureLen]byte{}
 	} else {
-		s.stats.IncMessageSignatureHit()
+		signature, err = s.backend.GetMessageSignature(unsignedMessage)
```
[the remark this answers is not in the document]
ceyonur: ahh I see we just moved it elsewhere

warp/messages/payload.go
```diff
@@ -0,0 +1,45 @@
...
+	// Bytes returns the binary representation of this payload.
+	Bytes() []byte
+
+	// initialize the payload with the provided binary representation.
+	initialize(b []byte)
+}
+
+// Signable is an optional interface that payloads can implement to allow
+// on-the-fly signing of incoming messages by the warp backend.
+type Signable interface {
```
darioush, line 25: I've made this an optional interface but we can make it required and have non-implementers always return an error if there is a preference for that

ceyonur: how do we make it required?

warp/messages/payload.go
```diff
@@ -0,0 +1,45 @@
...
+	VerifyMesssage(sourceAddress []byte) error
```
ceyonur, line 26: 🚀

warp/backend.go
```diff
@@ -164,6 +166,37 @@ func (b *backend) GetMessageSignature(messageID ids.ID) ([bls.SignatureLen]byte,
...
+	// Known on-chain messages should be signed
+	if _, err := b.GetMessage(unsignedMessage.ID()); err == nil {
+		return nil
+	}
+
+	// Try to parse the payload as an AddressedCall
+	addressedCall, err := payload.ParseAddressedCall(unsignedMessage.Payload)
+	if err != nil {
+		return fmt.Errorf("failed to parse unknown message as AddressedCall: %w", err)
+	}
```
ceyonur, line 179: what happens if that message type is not supposed to comfort any types. i.e a supposed to be a pure addressedcall without any specific type? (like current teleporter messages)

warp/backend.go
```diff
@@ -164,6 +166,37 @@ func (b *backend) GetMessageSignature(messageID ids.ID) ([bls.SignatureLen]byte,
...
+
+	// Check if the message is a known type that can be signed on demand
+	signable, ok := parsed.(messages.Signable)
+	if !ok {
+		return fmt.Errorf("parsed message is not Signable: %T", signable)
+	}
+
+	// Check if the message should be signed according to its type
+	if err := signable.VerifyMesssage(addressedCall.SourceAddress); err != nil {
```
ceyonur, line 194: we don't need to sort it out right not but seems we might need a handler to take different interfaces (VM, providers etc). One example could be the uptime request needs the uptime manager (and/or the state) to verify the message.

ceyonur (approved):

warp/backend.go
ceyonur: here is the PR (based on this one) https://github.com/ava-labs/subnet-evm/pull/1359

PR #1295, opened 2024-08-21
add initial script to do a coreth diff

darioush:
just a script for repeatable work


scripts/apply_diff_and_rename.sh
```diff
@@ -0,0 +1,34 @@
...
+sed_inplace=(sed -i) # Linux
+if [[ $(uname) == "Darwin" ]]; then
+    sed_inplace=(sed -i '')
+fi
+LANG=C find . -type f \! -name 'apply_diff_and_rename.sh' \! -path './.git/*' \! -path './contracts/node_modules/*' -exec "${sed_inplace[@]}" -e "${sed_command}" {} \;
+gofmt -w .
+go mod tidy
+
+# Restore contracts/.gitignore
+git checkout -- contracts/.gitignore 2>/dev/null || true  # Ignore if the file doesn't exist
```
darioush, line 34: Probably should remove these lines for less special casing and merge https://github.com/ava-labs/coreth/pull/638 before getting started instead

PR #1304, opened 2024-08-22
ci: fix NO_RACE + CI re-run on timeouts or fails of known flakes

darioush:
Too much shell quoting makes it not work when you run it with NO_RACE=1
 
For the other PR that got merged with its changes, this is an example showing it works:
https://github.com/ava-labs/subnet-evm/actions/runs/10533951904/job/29190820267?pr=1304


ARR4N (approved):

ceyonur (approved):

darioush:
Actually I want to fix the panic/timeout case as well before merging this.

ceyonur (approved):

PR #1320, opened 2024-08-28
[wip] decouple precompiles from core/vm

darioush:
picking up from https://github.com/ava-labs/subnet-evm/pull/1094
(rebased to master)

Pursuing the goal of the code in `core/vm` only depends on the upstream `param` package,
- since core/vm knows about the precompiles, it needs to pass in the chainConfig as a `precompileconfig.ChainConfig` (https://github.com/ava-labs/subnet-evm/blob/master/core/vm/evm.go#L631-L632)
- so we might as well decouple the precompiles and reduce the diff with geth.

This PR doesn't solve all the issues with the param package, like the one with the accessListGas also depending on which precompiles is active from the avalanche specific rules. But I consider it in the right direction.

Also in the future we can move this NewEVM to its own module, which may be helpful if we want to parameterize `core` more like this.


core/vm/evm.go
```diff
@@ -186,9 +171,19 @@ type EVM struct {
...
 
+type ChainConfig interface {
+	IsEIP158(blockNum *big.Int) bool
+	IsSubnetEVM(timestamp uint64) bool
+	IsCancun(blockNum *big.Int, timestamp uint64) bool
+	IsPrecompileEnabled(addr common.Address, timestamp uint64) bool
+	Rules(blockNum *big.Int, timestamp uint64) params.Rules
+}
+
+var NewEVM = newEVM
```
darioush, line 182: This may be a little jank, but we can consider it a shortcut for this pattern:
Convert functions to methods on the struct and reinstate the top-level function Foo() as an alias to DefaultConfig().Foo() ([slog.Info() example](https://cs.opensource.google/go/go/+/refs/tags/go1.23.0:src/log/slog/logger.go;l=291)).
as suggested by @ARR4N

ceyonur: I think it makes sense. IMHO globals are really annoying.

core/vm/evm.go
```diff
@@ -186,9 +171,19 @@ type EVM struct {
 	callGasTemp uint64
 }
 
+type ChainConfig interface {
```
ARR4N, line 174: Why is this being used instead of an actual `params.ChainConfig`?

core/evm_ext.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/log"
+)
+
+var (
+	newEVM = vm.NewEVM
+)
+
+func init() {
+	vm.NewEVM = NewEVM
```
ARR4N, line 24: This and the above capture of the original of `vm.NewEVM` strike me as a [code smell](https://en.wikipedia.org/wiki/Code_smell). The non-linearity of the logic flow and the "action at a distance" (substituting a global in another package) are anti-patterns that lead to spaghetti code. It will also void all documentation if the implementation can be subbed out in ad hoc ways.

core/evm_ext.go
```diff
@@ -0,0 +1,107 @@
...
+		return true
+	}
+
+	return modules.ReservedAddress(addr)
+}
+
+type EVM struct {
+	*vm.EVM
+
+	chainConfig *params.ChainConfig
```
ARR4N, line 40: `vm.EVM` already carries a `params.ChainConfig`; why is this one needed?

core/evm_ext.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+func NewEVM(
+	blockCtx vm.BlockContext, txCtx vm.TxContext, statedb vm.StateDB,
+	chainConfig vm.ChainConfig, config vm.Config,
+) *vm.EVM {
+	customChainConfig, ok := chainConfig.(*params.ChainConfig)
+	if !ok {
```
ARR4N, line 48: In what circumstances would someone pass another type?

core/evm_ext.go
```diff
@@ -0,0 +1,107 @@
...
+		// If the chainConfig is not a params.ChainConfig, then we can't use the custom
+		// EVM implementation, so we fall back to the default implementation.
+		log.Warn("ChainConfig is not a *params.ChainConfig, falling back to default EVM")
```
ARR4N, line 51: Why does passing a non-geth type result in using the native geth implementation and vice versa? That seems back to front.

core/evm_ext.go
```diff
@@ -0,0 +1,107 @@
+// (c) 2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package core
```
ARR4N, line 4: Why does this live in `core` instead of `vm`? Being here blurs any existing separation of concerns.

core/evm_ext.go
```diff
@@ -0,0 +1,107 @@
...
+
+func (evm *EVM) GetBlockContext() contract.BlockContext {
+	return &evm.EVM.Context
+}
+
+func (evm *EVM) GetStateDB() contract.StateDB {
+	return evm.StateDB
+}
+
+func (evm *EVM) GetChainConfig() precompileconfig.ChainConfig {
```
ARR4N, line 73: Why is this being returned as a `precompileconfig.ChainConfig`?

My immediate thought, before inspecting the interface definition, was that it seems to be a non sequitur. We're in the `core` package, accepting something akin to a `params.ChainConfig`, and then it's converted into something to do with precompiles.

Upon inspection I see that it only has `IsDurango()`; as a rule of thumb, return types should be as broadly useful as possible and accepted interfaces should require the minimum-viable signature to function.

core/vm/eips.go
```diff
@@ -114,7 +114,7 @@ func enable1344(jt *JumpTable) {
 
 // opChainID implements CHAINID opcode
 func opChainID(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext) ([]byte, error) {
-	chainId, _ := uint256.FromBig(interpreter.evm.chainConfig.ChainID)
+	chainId, _ := uint256.FromBig(interpreter.evm.chainRules.ChainID)
```
ARR4N, line 117: Why the change? [Upstream](https://github.com/ethereum/go-ethereum/blob/a9523b6428238a762e1a1e55e46ead47630c3a23/core/vm/eips.go#L110) uses `chainConfig`.

core/vm/evm.go
ARR4N: I'm not sure I follow. The global is introduced so that `core` can substitute it with a different implementation. The `slog` pattern would be more like:

```go
type EVMConfig struct {
//...
}

func (c *EVMConfig) New(...) *EVM {
// ... contents of original constructor, modified to use the `EVMConfig`
}

func DefaultEVMConfig() *EVMConfig {
//...
}

func NewEVM(...) *EVM {
// what used to be here is now in `EVMConfig.New()`
    return DefaultEVMConfig().New()
}
```

core/vm/evm.go
```diff
@@ -305,7 +285,7 @@ func (evm *EVM) Call(caller ContractRef, addr common.Address, input []byte, gas
 	}
 
 	if isPrecompile {
-		ret, gas, err = RunStatefulPrecompiledContract(p, evm, caller.Address(), addr, input, gas, evm.interpreter.readOnly)
+		ret, gas, err = p(caller.Address(), input, gas, evm.interpreter.readOnly)
```
ARR4N, line 288: This moves further from upstream, which uses [`RunPrecompiledContract()`](https://pkg.go.dev/github.com/ethereum/go-ethereum/core/vm#RunPrecompiledContract) to wrap tracing and gas metering.

ARR4N (changes requested):

core/evm_ext.go
darioush: What is your suggested alternative?

We can go back to a style that names the new function (or modifying the imports).
But this is problematic since we cannot control all the callsites, especially new ones that may materialize in upstream, which I also noticed in rebasing this PR. See this commit for the original style: https://github.com/ava-labs/subnet-evm/pull/1320/commits/b1bfabb7feb1e07c64010f2c67df3192b9fc54fe

We already configure the VM globally by importing packages and altering global state, such as the precompile module registry.

I agree that global state is up on the list of bad things, but I think it is preferable to callsite modifications.

core/evm_ext.go
darioush: It is needed to return it as a `precompileconfig.ChainConfig`. The purpose of this type is to keep our modifications of the EVM out of geth code, so we wrap all avalanche specific inputs and configurations and the geth code will try to call a generic callback to call precompiles instead.

core/vm/evm.go
darioush: It is so the core/vm package would be decoupled from params package and can be either geth's original or subnet-evm/params.

darioush: IMO to decouple the geth code from our code, a first important step is to avoid passing avalanche configurations to `core/vm`.

To ensure this, `.../core/vm` should not depend on the modified chain config type (`github.com/ava-labs/subnet-evm/params.ChainConfig`)

Later we can use the libevm/params.ChainConfig` or `go-ethereum/params.ChainConfig` type here, but that would require the `github.com/ava-labs/subnet-evm/params.ChainConfig` to be able to convert itself to one of these types, and I would like to do that refactor as a separate concern.

core/evm_ext.go
darioush: It cannot be in `core/vm` since the purpose is moving to a direction where avalanche packages are not imported in geth code.
So `core` is better than `core/vm`, which is more arguably the inner most geth code. We can move it to another package if desired, originally I picked `core` because it was already imported in most callsites of this code.

core/vm/evm.go
darioush: We are also not using `RunPrecompiledContract`, so I don't think it changes how far this is from upstream.
Also my long term plan is to propose the CustomPrecompiles to upstream as a configuration option, but it may not be accepted, so we see.
But open to suggestions.

core/vm/eips.go
darioush: because this PR makes chainConfig an interface and we cannot reference a field on that. Open to other suggestions

core/evm_ext.go
darioush: It is already being returned as `precompileconfig.ChainConfig` but from `core/vm`, which seems worse.
Also this is the goal of this refactoring, to separate avalanche specifics out of the params package.

Probably you looked at coreth, it also includes `GetFeeConfig` and `AllowedFeeRecipients`.
IMO avalanche specifics should be given access to all the avalanche specific parameters, just it must be transparent to `core/vm` package, and later to the `core` package as well, as you note it is not ideal.

core/evm_ext.go
darioush: Sorry if it was unclear, at the moment I do not consider `github.com/ava-labs/subnet-evm/params` to be a geth implementation (it imports precompiles).

core/evm_ext.go
darioush: In this PR it is not. But in the future it may be possible that geth / libevm code could use that, until the code is fully separated and there is no "bad imports"

core/vm/evm.go
```diff
@@ -31,36 +31,15 @@ import (
...
 	"sync/atomic"
 
-	"github.com/ava-labs/avalanchego/snow"
-	"github.com/ava-labs/subnet-evm/constants"
 	"github.com/ava-labs/subnet-evm/core/types"
 	"github.com/ava-labs/subnet-evm/params"
-	"github.com/ava-labs/subnet-evm/precompile/contract"
-	"github.com/ava-labs/subnet-evm/precompile/contracts/deployerallowlist"
-	"github.com/ava-labs/subnet-evm/precompile/modules"
-	"github.com/ava-labs/subnet-evm/precompile/precompileconfig"
```
darioush, line 41: ^ These are the main goal of this PR

core/vm/evm.go
```diff
@@ -31,36 +31,15 @@ import (
...
 	"github.com/ava-labs/subnet-evm/predicate"
 	"github.com/ava-labs/subnet-evm/vmerrs"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/holiman/uint256"
 )
 
-var (
-	_ contract.AccessibleState = &EVM{}
-	_ contract.BlockContext    = &BlockContext{}
```
darioush, line 51: ^ The new EVM type in `core` is there to implement these interfaces and wrap them as `CustomPrecompiles`, in addition to specifying the exact behavioral modifications.

core/vm/evm.go
darioush: I switched to this pattern. We still somehow need a factory since the arguments provided are used to build the config object.

ARR4N: Requiring a factory is ok; what I'm saying is that it shouldn't be a global variable to allow it to be arbitrarily substituted. The `slog.Info()` pattern unfortunately has a red herring in that you can swap out the default logger; [this is a better example of what I intended](https://github.com/darvaza-proxy/darvaza/blob/e8cb3af76ff47ab319166c60a767c22b7f8b961b/shared/storage/simple/config.go#L17).

darioush:
Closing this PR in favor of the libevm PR https://github.com/ava-labs/coreth/pull/662

PR #1327, opened 2024-08-30
Set http body limit conf

ceyonur:
## Why this should be merged

Adds http-body-limit config for RPC 

## How this works

Adds a new json config and sets for created RPC servers 

## How this was tested

no test in upstream code

## How is this documented

it will be documented as soon as I figure out the new docs repo


darioush (approved):
lgtm, should we add options for SetBatchLimits (itemLimit, maxResponseSize) as well?

PR #1328, opened 2024-08-31
Tail lock revisited

ceyonur:
## Why this should be merged

There is a potential race between txindexer and state sync (ResetToStateSyncedBlock/repairTxIndexTail), where reading can occur before repairTxIndexTail (which can increment the tail) and txindexer can work on an outdate tail (less than actual). 

## How this works

Uses txindexer's `run` function after grabbing lock and then do the actual db read for tail for `run`. 

## How this was tested

Not sure how to test this race

## How is this documented

no need

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -459,7 +459,7 @@ func NewBlockChain(
 	// if txlookup limit is 0 (uindexing disabled), we don't need to repair the tx index tail.
 	if bc.cacheConfig.TransactionHistory != 0 {
 		latestStateSynced := rawdb.GetLatestSyncPerformed(bc.db)
-		bc.setTxIndexTail(latestStateSynced)
+		bc.repairTxIndexTail(latestStateSynced)
 	}
 
 	// Start processing accepted blocks effects in the background
@@ -2147,7 +2147,7 @@ func (bc *BlockChain) ResetToStateSyncedBlock(block *types.Block) error {
 
 	// if txlookup limit is 0 (uindexing disabled), we don't need to repair the tx index tail.
 	if bc.cacheConfig.TransactionHistory != 0 {
-		bc.setTxIndexTail(block.NumberU64())
+		bc.repairTxIndexTail(block.NumberU64())
 	}
 
 	// Update all in-memory chain markers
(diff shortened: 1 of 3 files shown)


core/blockchain.go
```diff
@@ -2183,18 +2183,12 @@ func (bc *BlockChain) CacheConfig() *CacheConfig {
...
 
-	tailP := rawdb.ReadTxIndexTail(bc.db)
-	var tailV uint64
-	if tailP != nil {
-		tailV = *tailP
-	}
-
-	if newTail > tailV {
-		log.Info("Repairing tx index tail", "old", tailV, "new", newTail)
+	if currentTail := rawdb.ReadTxIndexTail(bc.db); currentTail == nil || newTail > *currentTail {
```
ARR4N, line 2190: Question: as far as I can tell this is just a refactor with the same logic; is that your intention?

Comment: I think the variable `currentTail` could just be `curr` (or even `tail`) to make this more succinct and easier to read. The fact that it's a tail is implied by the RHS of the assignment, and it's only used on two lines afterwards so verbosity is unnecessary:

```go
if curr := rawdb.ReadTxIndexTail(bc.db); curr == nil || *curr < newTail {
// or
if tail := rawdb.ReadTxIndexTail(bc.db); tail == nil || *tail < newTail {
```

Afterthought: my using `curr < new` instead of `new > curr` just kinda happened as it felt more natural to frame everything in terms of `curr`. Entirely subjective, and up to you to decide if you think I'm nuts :)

core/txindexer.go
```diff
@@ -99,33 +96,27 @@ func (indexer *txIndexer) run(tail *uint64, head uint64, stop chan struct{}, don
 	}
 
 	// Defensively ensure tail is not nil.
-	if tail == nil {
-		tail = new(uint64)
+	tailValue := uint64(0)
```
ARR4N, line 99: Why the change here? It's effectively the same thing.

core/txindexer.go
```diff
@@ -145,7 +136,12 @@ func (indexer *txIndexer) loop(chain *BlockChain) {
 		done = make(chan struct{})
 		lastHead = head.Number.Uint64()
 		indexer.chain.wg.Add(1)
-		go indexer.run(rawdb.ReadTxIndexTail(indexer.db), head.Number.Uint64(), stop, done)
+		go func() {
```
ARR4N, line 139: Am I correct in understanding that this (and the similar go routine below) is where the fix occurs, because the call to `rawdb.ReadTxIndexTail` is now inside the lock?

Does `ResetToStateSyncedBlock` honour the same lock?

core/txindexer.go
ceyonur: it avoids overriding pointer

core/txindexer.go
ceyonur: yes, `ResetToStateSyncedBlock` uses `repairTxIndexTail` which holds the same lock

core/blockchain.go
ceyonur: yes it's a refactor

I applied your suggestion, thanks.

core/txindexer.go
```diff
@@ -159,15 +155,19 @@ func (indexer *txIndexer) loop(chain *BlockChain) {
 				stop = make(chan struct{})
 				done = make(chan struct{})
 				indexer.chain.wg.Add(1)
-				go indexer.run(rawdb.ReadTxIndexTail(indexer.db), head.Block.NumberU64(), stop, done)
+				go func() {
```
ARR4N, line 158: If this must always be the same as the equivalent go routine then consider abstracting it into shared code, otherwise all LGTM.

ARR4N (approved):

PR #1330, opened 2024-09-02
Bump avago rc

ceyonur:
bumps to avalancheGo RC


README.md
```diff
@@ -18,44 +18,6 @@ The Subnet EVM runs in a separate process from the main AvalancheGo process and
 ### AvalancheGo Compatibility
 
 ```text
-[v0.1.0] AvalancheGo@v1.7.0-v1.7.4 (Protocol Version: 9)
```
StephenButtolph, line 21: Why was this removed? (Or why was it ever here?)

ceyonur: I think this was a request from CLI/AvaCloud team. I don't think we need those old versions. We can even consider removing this altogether, and only keep the current supported versions in readme or in release notes. @ava-labs/platform-evm

ARR4N (approved):

PR #1337, opened 2024-09-10
params: add LondonBlock

darioush:

### DIFF
--- a/params/config.go
+++ b/params/config.go
@@ -75,6 +75,7 @@ var (
 		PetersburgBlock:     big.NewInt(0),
 		IstanbulBlock:       big.NewInt(0),
 		MuirGlacierBlock:    big.NewInt(0),
+		LondonBlock:         big.NewInt(0),
 		CancunTime:          utils.TimeToNewUint64(upgrade.GetConfig(constants.UnitTestID).EtnaTime),
 		NetworkUpgrades:     getDefaultNetworkUpgrades(upgrade.GetConfig(constants.UnitTestID)), // This can be changed to correct network (local, test) via VM.
 		GenesisPrecompiles:  Precompiles{},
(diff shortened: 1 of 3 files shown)


params/config.go
```diff
@@ -317,14 +323,29 @@ func (c *ChainConfig) IsIstanbul(num *big.Int) bool {
...
+func (c *ChainConfig) IsLondon(num *big.Int) bool {
+	return isBlockForked(c.LondonBlock, num)
+}
+
+// IsShanghai returns whether time is either equal to the Shanghai fork time or greater.
+func (c *ChainConfig) IsShanghai(num *big.Int, time uint64) bool {
+	return c.IsLondon(num) && isTimestampForked(c.ShanghaiTime, time)
+}
+
+// IsCancun returns whether num is either equal to the Cancun fork time or greater.
```
darioush, line 341: it's like this upstream.

params/config.go
```diff
@@ -372,6 +393,9 @@ func (c *ChainConfig) CheckConfigForkOrder() error {
 		{name: "petersburgBlock", block: c.PetersburgBlock},
 		{name: "istanbulBlock", block: c.IstanbulBlock},
 		{name: "muirGlacierBlock", block: c.MuirGlacierBlock, optional: true},
+		{name: "berlinBlock", block: c.BerlinBlock},
+		{name: "londonBlock", block: c.LondonBlock},
+		{name: "shanghaiTime", timestamp: c.ShanghaiTime},
```
darioush, line 398: just adding the items we'd need to switch to libevm, not everything needs to be added back at once.

params/config_extra.go
```diff
@@ -260,6 +263,12 @@ func (c *ChainConfig) SetNetworkUpgradeDefaults() {
 	if c.MuirGlacierBlock == nil {
 		c.MuirGlacierBlock = big.NewInt(0)
 	}
+	if c.BerlinBlock == nil {
+		c.BerlinBlock = big.NewInt(0)
+	}
+	if c.LondonBlock == nil {
+		c.LondonBlock = big.NewInt(0)
+	}
```
darioush, line 271: basically, with this we will not support subnet-evm networks running without the subnet-evm fork enabled at genesis.
those networks (if they exist) are welcome to directly specify (by modifying `c`) the blocks at which the forks should be activated, which is also likely the path to go in coreth.

params/config.go
ceyonur, line 1: do we plan to use these forks same places as in upstream? like instructionsSets enabling London features etc?

ceyonur (approved):
left a question that can be solved in a different PR

params/config.go
darioush: yes, this is the plan and we can start making those PRs as well. we can either incrementally change them back or do this change by the package if we have enough confidence (by switching to the libevm or upstream geth packages)

PR #1340, opened 2024-09-12
add pausable uptime manager

ceyonur:
## Why this should be merged

Adds pausable uptime manager interface for uptime tracking

## How this works

Pausable uptime manager extends the avalanchego uptime manager by introducing callback listeners for validators state (under `plugin/evm/validators`). The idea is to pause a node's uptime and then can continue whenever it's resumed. It also modifies `Connect`, `Disconnect,` StartTracking` accordingly. 

## How this was tested

Added tests

## How is this documented

Comments in code

### DIFF
--- a/plugin/evm/uptime/pausable_manager.go
+++ b/plugin/evm/uptime/pausable_manager.go
@@ -13,13 +13,13 @@ import (
 	"github.com/ava-labs/avalanchego/utils/set"
 )
 
-var _ validators.ValidatorsCallbackListener = &pausableManager{}
+var _ validators.StateCallbackListener = &pausableManager{}
 
 var ErrPausedDc = errors.New("paused node cannot be disconnected")
 
 type PausableManager interface {
 	uptime.Manager
-	validators.ValidatorsCallbackListener
+	validators.StateCallbackListener
 	IsPaused(nodeID ids.NodeID) bool
 }
 
--- a/plugin/evm/uptime/pausable_manager.go
+++ b/plugin/evm/uptime/pausable_manager.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 
 	"github.com/ava-labs/subnet-evm/plugin/evm/validators"
+	"github.com/ethereum/go-ethereum/log"
 
 	"github.com/ava-labs/avalanchego/ids"
 	"github.com/ava-labs/avalanchego/snow/uptime"
@@ -15,7 +16,11 @@ import (
 
 var _ validators.StateCallbackListener = &pausableManager{}
 
-var ErrPausedDc = errors.New("paused node cannot be disconnected")
+var (
+	errPausedDisconnect = errors.New("paused node cannot be disconnected")
+	errAlreadyPaused    = errors.New("node is already paused")
+	errNotPaused        = errors.New("node is not paused")
+)
 
 type PausableManager interface {
 	uptime.Manager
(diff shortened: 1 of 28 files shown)


plugin/evm/uptime/pausable_manager.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/ava-labs/subnet-evm/plugin/evm/validators"
+
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/snow/uptime"
+	"github.com/ava-labs/avalanchego/utils/set"
+)
+
+var _ validators.StateCallbackListener = &pausableManager{}
+
+var ErrPausedDc = errors.New("paused node cannot be disconnected")
```
darioush, line 18: let's avoid abbreviation of Disconnected to Dc

plugin/evm/uptime/pausable_manager.go
```diff
@@ -0,0 +1,136 @@
...
+
+// Disconnect disconnects the node with the given ID from the uptime.Manager
+// If the node is paused, it will not be disconnected
+// Invariant: we should never have a connected paused node that is disconnecting
+func (p *pausableManager) Disconnect(nodeID ids.NodeID) error {
+	p.connectedVdrs.Remove(nodeID)
+	if p.Manager.IsConnected(nodeID) {
+		if p.IsPaused(nodeID) {
+			// We should never see this case
+			return ErrPausedDc
```
darioush, line 62: This error should not be triggered by calling exported functions if I understand correctly. (we can unexport it)

plugin/evm/uptime/pausable_manager.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// IsPaused returns true if the node with the given ID is paused.
+func (p *pausableManager) IsPaused(nodeID ids.NodeID) bool {
+	return p.pausedVdrs.Contains(nodeID)
+}
+
+// pause pauses uptime tracking for the node with the given ID
+// pause can disconnect the node from the uptime.Manager if it is connected.
+// Returns an error if the node is already paused.
```
darioush, line 114: seems this is missing?

plugin/evm/uptime/pausable_manager.go
```diff
@@ -0,0 +1,136 @@
...
+		// This should be fine in case tracking has not started yet since
+		// the inner manager should handle disconnects accordingly
+		return p.Manager.Disconnect(nodeID)
+	}
+	return nil
+}
+
+// resume resumes uptime tracking for the node with the given ID
+// resume can connect the node to the uptime.Manager if it was connected.
+// Returns an error if the node is not paused.
```
darioush, line 129: seems this is missing?

plugin/evm/uptime/pausable_manager_test.go
```diff
@@ -0,0 +1,205 @@
...
+		currentTime = addTime(clk, 7*time.Second)
+		// Uptime should not have increased since the node was not connected
+		checkUptime(t, up, nodeID0, 13*time.Second, currentTime)
+
+		// Connect and final uptime check
+		require.NoError(up.Connect(nodeID0))
+		currentTime = addTime(clk, 8*time.Second)
+		// Uptime should have increased by 8 seconds since the node was connected
+		checkUptime(t, up, nodeID0, 21*time.Second, currentTime)
+	}
```
darioush, line 183: I think we could improve this test using a table, may require some moving items around.
One approach could have callback fns like `beforePause`, `afterPause` etc

darioush (approved):

PR #1341, opened 2024-09-13
Validator metastate

ceyonur:
## Why this should be merged

Adds validator metastate that can persist validator information and implements `uptime.State`.

## How this was tested

Added UT

## How is this documented

Comments in code


ceyonur:
closing in favor of #1344

PR #1344, opened 2024-09-16
add validator state

ceyonur:
## Why this should be merged

Adds validator state that can persist validator information and implements `uptime.State`.

## How this was tested

Added UT

## How is this documented

Comments in code


plugin/evm/validators/state.go
```diff
@@ -0,0 +1,313 @@
...
+func (s *state) GetValidatorIDs() set.Set[ids.NodeID] {
+	ids := set.NewSet[ids.NodeID](len(s.index))
+	for nodeID := range s.index {
+		ids.Add(nodeID)
+	}
+	return ids
+}
+
+// RegisterListener registers a listener to the state
+// the listener will be notified of current validators via OnValidatorAdded
```
darioush, line 242:
```suggestion
// OnValidatorAdded is called for all current validators on the provided listener before this function returns
```

plugin/evm/validators/state.go
```diff
@@ -0,0 +1,313 @@
...
+			return fmt.Errorf("failed to parse validator data: %w", err)
+		}
+		if err := s.putData(vID, vdr); err != nil {
+			return err
+		}
+	}
+	return it.Error()
+}
+
+func (s *state) putData(vID ids.ID, data *validatorData) error {
```
darioush, line 287: nit: should we call this add and mention in a comment that it cannot overwrite?

plugin/evm/validators/state.go
```diff
@@ -0,0 +1,313 @@
...
+		return fmt.Errorf("validator data already exists for %s", data.NodeID)
+	}
+
+	s.data[vID] = data
+	s.index[data.NodeID] = vID
+	return nil
+}
+
+// getData returns the data for the validator with the given nodeID
+// returns ErrNotFound if the data does not exist
```
darioush, line 302:
```suggestion
// returns database.ErrNotFound if the data does not exist
```

plugin/evm/validators/state_test.go
```diff
@@ -0,0 +1,288 @@
...
+	_, _, err = state.GetUptime(nodeID)
+	require.ErrorIs(err, database.ErrNotFound)
+
+	// set non-existent uptime
+	startTime := time.Now()
+	err = state.SetUptime(nodeID, 1, startTime)
+	require.ErrorIs(err, database.ErrNotFound)
+
+	// add new validator
+	state.AddNewValidator(vID, nodeID, uint64(startTime.Unix()), true)
```
darioush, line 37: let's check err here

plugin/evm/validators/state_test.go
```diff
@@ -0,0 +1,288 @@
...
+
+	// adding the same validator should fail
+	err = state.AddNewValidator(vID, nodeID, uint64(startTime.Unix()), true)
+	require.Error(err)
```
darioush, line 41: should we add an error like ValidatorAlreadyExists or ValidatorAlreadyAdded and check it here?
also should we test this with a different nodeID (since the repeated vID will cause the failure)

plugin/evm/validators/state_test.go
```diff
@@ -0,0 +1,288 @@
...
+	// adding the same nodeID should fail
+	err = state.AddNewValidator(ids.GenerateTestID(), nodeID, uint64(startTime.Unix()), true)
+	require.Error(err)
+
+	// get uptime
+	upDuration, lastUpdated, err := state.GetUptime(nodeID)
```
darioush, line 47: nit: should we call the local variables uptime, newUptime, etc?

plugin/evm/validators/state_test.go
```diff
@@ -0,0 +1,288 @@
...
+
+	// set status
+	require.NoError(state.SetStatus(vID, false))
+	// get status
+	status, err := state.GetStatus(vID)
+	require.NoError(err)
+	require.False(status)
+
+	// delete uptime
+	state.DeleteValidator(vID)
```
darioush, line 70: let's check err here

plugin/evm/validators/state_test.go
```diff
@@ -0,0 +1,288 @@
...
+	state, err := NewState(db)
+	require.NoError(err)
+	// write empty uptimes
+	require.NoError(state.WriteState())
+
+	// load uptime
+	nodeID := ids.GenerateTestNodeID()
+	vID := ids.GenerateTestID()
+	startTime := time.Now()
+	state.AddNewValidator(vID, nodeID, uint64(startTime.Unix()), true)
```
darioush, line 89: err check

plugin/evm/validators/state_test.go
```diff
@@ -0,0 +1,288 @@
...
+	require.NoError(err)
+
+	// get uptime
+	upDuration, lastUpdated, err := state.GetUptime(nodeID)
+	require.NoError(err)
+	require.Equal(newUpDuration, upDuration)
+	require.Equal(newLastUpdated.Unix(), lastUpdated.Unix())
+
+	// delete
+	state.DeleteValidator(vID)
```
darioush, line 112: err check

plugin/evm/validators/state_test.go
```diff
@@ -0,0 +1,288 @@
...
+				lastUpdated: time.Unix(0, 0),
+				startTime:   time.Unix(0, 0),
+			},
+			expectedErr: nil,
+		},
+		{
+			name: "valid",
+			bytes: []byte{
+				// codec version
+				0x00, 0x00,
```
darioush, line 151: should we create then modify this bytes programmatically? (like cutting it short, or modifying the initial bytes)

plugin/evm/validators/state_test.go
```diff
@@ -0,0 +1,288 @@
...
+	state.RegisterListener(listener)
+
+	// add new validator
+	state.AddNewValidator(expectedvID, expectedNodeID, uint64(expectedStartTime.Unix()), true)
+
+	// set status
+	require.NoError(state.SetStatus(expectedvID, false))
+
+	// remove validator
+	state.DeleteValidator(expectedvID)
```
darioush, line 254: err check

plugin/evm/validators/state_test.go
```diff
@@ -0,0 +1,288 @@
...
+		onStatusUpdate: func(vID ids.ID, nodeID ids.NodeID, isActive bool) {
+			require.Equal(expectedvID, vID)
+			require.Equal(expectedNodeID, nodeID)
+			require.False(isActive)
+		},
+	}
+	state.RegisterListener(listener)
+
+	// add new validator
+	state.AddNewValidator(expectedvID, expectedNodeID, uint64(expectedStartTime.Unix()), true)
```
darioush, line 248: err check

go.mod
```diff
@@ -5,7 +5,7 @@ go 1.21.12
 require (
 	github.com/VictoriaMetrics/fastcache v1.12.1
 	github.com/antithesishq/antithesis-sdk-go v0.3.8
-	github.com/ava-labs/avalanchego v1.11.11
+	github.com/ava-labs/avalanchego v1.11.12-0.20240910093928-8c6a486205ce
```
darioush, line 8: Could you link the modified avalanchego branch in the PR description?

plugin/evm/validators/state.go
```diff
@@ -0,0 +1,313 @@
...
+	data, err := s.getData(nodeID)
+	if err != nil {
+		return time.Time{}, err
+	}
+	return data.startTime, nil
+}
+
+// AddNewValidator adds a new validator to the state
+// the new validator is marked as updated and will be written to the disk when WriteState is called
+func (s *state) AddNewValidator(vID ids.ID, nodeID ids.NodeID, startTimestamp uint64, isActive bool) error {
```
darioush, line 126: should this just be "AddValidator"?

plugin/evm/validators/state.go
```diff
@@ -0,0 +1,313 @@
...
+		}
+	}
+	return it.Error()
+}
+
+func (s *state) putData(vID ids.ID, data *validatorData) error {
+	if _, exists := s.data[vID]; exists {
+		return fmt.Errorf("validator data already exists for %s", vID)
+	}
+	// should never happen
```
darioush, line 291: is there any difference between these 2 cases in that they are always unexpected? (should the comment apply to one/both?)
Since it seems this error could be returned from the exported function `AddNewValidator`

plugin/evm/validators/state_test.go
ceyonur: not sure I understood this, we have cases for `short byte len` and `invalid codec version`

plugin/evm/validators/state.go
```diff
@@ -0,0 +1,313 @@
...
+	validationID ids.ID // database key
+	lastUpdated  time.Time
+	startTime    time.Time
+}
+
+type state struct {
+	data  map[ids.ID]*validatorData // vID -> validatorData
+	index map[ids.NodeID]ids.ID     // nodeID -> vID
+	// updatedData tracks the updates since las WriteValidator was called
+	updatedData map[ids.ID]bool // vID -> true(updated)/false(deleted)
```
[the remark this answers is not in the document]
ceyonur: done

plugin/evm/validators/state.go
```diff
@@ -0,0 +1,313 @@
...
+type validatorData struct {
+	UpDuration  time.Duration `serialize:"true"`
+	LastUpdated uint64        `serialize:"true"`
+	NodeID      ids.NodeID    `serialize:"true"`
+	StartTime   uint64        `serialize:"true"`
+	IsActive    bool          `serialize:"true"`
+
+	validationID ids.ID // database key
+	lastUpdated  time.Time
+	startTime    time.Time
```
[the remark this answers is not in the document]
ceyonur: yup done

plugin/evm/validators/state.go
ceyonur: added

plugin/evm/validators/state_test.go
ceyonur: done

plugin/evm/validators/state_test.go
ceyonur: done

plugin/evm/validators/state.go
ceyonur: the first check (unique `vID`) should always return before second `nodeID` check. the first check is expected in case where `loadFromDisk` is called twice or where the `AddNewValidator` is called multiple times with the same parameters. 

It's actually possible to have the same `nodeID` under a different `vID` but not at the same time. so they're not "always" mutually unique to each other. removed that comment.

ceyonur:
Changed the target branch to `uptime-tracking-base` since these PRs should not be merged to master as they depend on a avalanchego feature branch that not merged yet.

plugin/evm/validators/state_test.go
```diff
@@ -165,15 +165,13 @@ func TestParseValidator(t *testing.T) {
...
-				lastUpdated: time.Unix(900000, 0),
 				NodeID:      testNodeID,
 				StartTime:   6000000,
-				startTime:   time.Unix(6000000, 0),
 				IsActive:    true,
 			},
 		},
 		{
-			name: "invalid codec version",
+			name: "	",
```
darioush, line 174: seems this was modified unintentionally

plugin/evm/validators/state_test.go
```diff
@@ -86,30 +86,30 @@ func TestWriteValidator(t *testing.T) {
...
-	state.AddNewValidator(vID, nodeID, uint64(startTime.Unix()), true)
+	require.NoError(state.AddValidator(vID, nodeID, uint64(startTime.Unix()), true))
 
 	// write state, should reflect to DB
 	require.NoError(state.WriteState())
 	require.True(db.Has(vID[:]))
 
 	// set uptime
-	newUpDuration := 2 * time.Minute
+	newuptime := 2 * time.Minute
```
darioush, line 96: nit: `newUptime`?

plugin/evm/validators/state_test.go
darioush: I mean should we generate the valid bytes using the codec or should we leave it hardcoded as is?

plugin/evm/validators/state_test.go
ceyonur: uhh weird

darioush (dismissed):

PR #1357, opened 2024-10-04
update to avalanchego master

darioush:
## Why this should be merged
For maintaining compatibility with avalanchego

## How this works
Updating avalanchego version and go dependency

## How this was tested
CI

## How is this documented
N/A

### DIFF
--- a/params/config.go
+++ b/params/config.go
@@ -172,7 +172,7 @@ var (
 		NetworkUpgrades: NetworkUpgrades{
 			SubnetEVMTimestamp: utils.NewUint64(0),
 			DurangoTimestamp:   utils.NewUint64(0),
-			EtnaTimestamp:      utils.NewUint64(0),
+			EtnaTimestamp:      utils.TimeToNewUint64(upgrade.InitiallyActiveTime),
 		},
 		GenesisPrecompiles: Precompiles{},
 		UpgradeConfig:      UpgradeConfig{},
--- a/params/network_upgrades.go
+++ b/params/network_upgrades.go
@@ -156,7 +156,9 @@ func verifyWithDefault(configTimestamp *uint64, defaultTimestamp *uint64) error 
 		return errCannotBeNil
 	}
 
-	if *configTimestamp < *defaultTimestamp {
+	// Allow scheduling of UnscheduledActivationTime (or later) to an earlier time.
+	// This is to allow for the network upgrades to be scheduled in tests.
+	if *defaultTimestamp < uint64(upgrade.UnscheduledActivationTime.Unix()) && *configTimestamp < *defaultTimestamp {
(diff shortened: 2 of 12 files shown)


params/network_upgrades.go
```diff
@@ -156,7 +156,9 @@ func verifyWithDefault(configTimestamp *uint64, defaultTimestamp *uint64) error
 		return errCannotBeNil
 	}
 
-	if *configTimestamp < *defaultTimestamp {
+	// Allow scheduling of UnscheduledActivationTime (or later) to an earlier time.
+	// This is to allow for the network upgrades to be scheduled in tests.
+	if *defaultTimestamp < uint64(upgrade.UnscheduledActivationTime.Unix()) && *configTimestamp < *defaultTimestamp {
```
darioush, line 161: in the updated avalanchego, etna is rescheduled on defaults to UnscheduledActivationTime, which means we can't use Etna at genesis unless we use a timestamp past UnscheduledActivationTime, which seems weird.

so it seems more logical that we could override it if it was `UnscheduledActivationTime`, for testing purposes or local testing.

params/network_upgrades_test.go
```diff
@@ -182,8 +182,8 @@ func TestVerifyNetworkUpgrades(t *testing.T) {
 			name: "ValidNetworkUpgrades for custom network",
 			upgrades: &NetworkUpgrades{
 				SubnetEVMTimestamp: utils.NewUint64(0),
-				DurangoTimestamp:   utils.NewUint64(1607144400),
-				EtnaTimestamp:      utils.NewUint64(1607144400),
+				DurangoTimestamp:   utils.TimeToNewUint64(upgrade.InitiallyActiveTime),
+				EtnaTimestamp:      utils.TimeToNewUint64(upgrade.UnscheduledActivationTime),
```
darioush, line 186: this could be undone

params/config.go
```diff
@@ -172,7 +172,7 @@ var (
 		NetworkUpgrades: NetworkUpgrades{
 			SubnetEVMTimestamp: utils.NewUint64(0),
 			DurangoTimestamp:   utils.NewUint64(0),
-			EtnaTimestamp:      utils.NewUint64(0),
+			EtnaTimestamp:      utils.TimeToNewUint64(upgrade.InitiallyActiveTime),
```
darioush, line 175: if we leave this 0 it will get overriden by the default value which is now UnscheduledActivationTime.

ARR4N (approved):

params/network_upgrades.go
ceyonur: I took another route by changing the test snow context here: https://github.com/ava-labs/subnet-evm/pull/1348/files

darioush:
https://github.com/ava-labs/subnet-evm/pull/1348/files was merged instead

PR #1361, opened 2024-10-10
stop RPC servers in shutdown

darioush:
## Why this should be merged
Close the rpc handlers to wait for in-flight rpc requests before calling eth.Stop which closes chaindb, as rawdb package doesn't propagate err correctly this can lead to panics during shutdown.

## How this works
Explicitly calls .Stop for the rpc servers created

## How this was tested
CI

## How is this documented
N/A

### DIFF
--- a/plugin/evm/vm.go
+++ b/plugin/evm/vm.go
@@ -245,6 +245,9 @@ type VM struct {
 	ethTxGossipHandler p2p.Handler
 	ethTxPushGossiper  avalancheUtils.Atomic[*gossip.PushGossiper[*GossipEthTx]]
 	ethTxPullGossiper  gossip.Gossiper
+
+	// RPC handlers (should be stopped before closing chaindb)
+	rpcHandlers []interface{ Stop() }
 }
 
 // Initialize implements the snowman.ChainVM interface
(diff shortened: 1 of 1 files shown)


plugin/evm/vm.go
```diff
@@ -826,6 +829,10 @@ func (vm *VM) Shutdown(context.Context) error {
 		log.Error("error stopping state syncer", "err", err)
 	}
 	close(vm.shutdownChan)
+	// Stop RPC handlers before eth.Stop which will close the database
+	for _, handler := range vm.rpcHandlers {
+		handler.Stop()
```
ceyonur, line 834: In the Stop comment it says ` waits for stopPendingRequestTimeout to allow pending requests to finish`; but I guess that's just a stale comment. It seems they have removed it around v1.4.0 (8 years ago lol).

darioush: I think it still has some impact since it calls close on the codecs [here](https://github.com/ava-labs/subnet-evm/blob/e996852950081016b8a1117b9a4d3d6200fc0072/rpc/server.go#L194)
but I couldn't find that variable either

PR #1363, opened 2024-10-14
remove the noSkip flag

darioush:

### DIFF
--- a/plugin/evm/vm_test.go
+++ b/plugin/evm/vm_test.go
@@ -189,10 +189,8 @@ func setupGenesis(
 	genesisBytes := buildGenesisTest(t, genesisJSON)
 	ctx := NewContext()
 
-	baseDB := memdb.New()
-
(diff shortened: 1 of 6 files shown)


plugin/evm/vm_test.go
```diff
@@ -189,10 +189,8 @@ func setupGenesis(
 	genesisBytes := buildGenesisTest(t, genesisJSON)
 	ctx := NewContext()
 
-	baseDB := memdb.New()
-
 	// initialize the atomic memory
-	atomicMemory := atomic.NewMemory(prefixdb.New([]byte{0}, baseDB))
+	atomicMemory := atomic.NewMemory(prefixdb.New([]byte{0}, memdb.New()))
```
darioush, line 193: I don't think there is any meaningful connection between the atomic memory and the chain memory being connected via some prefixdb.

PR #1366, opened 2024-10-16
use triedb config for reference root

darioush:
## Why this should be merged
This provides better alignment with upstream code, avoiding passing an extra boolean to statedb.Commit and avoids an additional triedb backend method.
Still considering whether we can avoid the concurrent use of the triedb ref. counting entirely

## How this works
Removes UpdateAndReferenceRoot in favor of a triedb backend (hashdb) option that references root when Update is called.

## How this was tested
CI

## How is this documented
N/A

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -198,9 +198,9 @@ func (c *CacheConfig) triedbConfig() *triedb.Config {
 	config := &triedb.Config{Preimages: c.Preimages}
 	if c.StateScheme == rawdb.HashScheme || c.StateScheme == "" {
 		config.HashDB = &hashdb.Config{
-			CleanCacheSize: c.TrieCleanLimit * 1024 * 1024,
-			StatsPrefix:    trieCleanCacheStatsNamespace,
-			ReferenceRoot:  true,
+			CleanCacheSize:                  c.TrieCleanLimit * 1024 * 1024,
+			StatsPrefix:                     trieCleanCacheStatsNamespace,
+			ReferenceRootAtomicallyOnUpdate: true,
 		}
 	}
 	if c.StateScheme == rawdb.PathScheme {
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -196,10 +196,11 @@ type CacheConfig struct {
 // triedbConfig derives the configures for trie database.
 func (c *CacheConfig) triedbConfig() *triedb.Config {
 	config := &triedb.Config{Preimages: c.Preimages}
-	if c.StateScheme == rawdb.HashScheme {
+	if c.StateScheme == rawdb.HashScheme || c.StateScheme == "" {
 		config.HashDB = &hashdb.Config{
 			CleanCacheSize: c.TrieCleanLimit * 1024 * 1024,
 			StatsPrefix:    trieCleanCacheStatsNamespace,
+			ReferenceRoot:  true,
 		}
 	}
 	if c.StateScheme == rawdb.PathScheme {
@@ -1222,9 +1223,9 @@ func (bc *BlockChain) writeBlockWithState(block *types.Block, receipts []*types.
 	// diff layer for the block.
 	var err error
 	if bc.snaps == nil {
-		_, err = state.Commit(block.NumberU64(), bc.chainConfig.IsEIP158(block.Number()), true)
+		_, err = state.Commit(block.NumberU64(), bc.chainConfig.IsEIP158(block.Number()))
 	} else {
-		_, err = state.CommitWithSnap(block.NumberU64(), bc.chainConfig.IsEIP158(block.Number()), bc.snaps, block.Hash(), block.ParentHash(), true)
+		_, err = state.CommitWithSnap(block.NumberU64(), bc.chainConfig.IsEIP158(block.Number()), bc.snaps, block.Hash(), block.ParentHash())
 	}
 	if err != nil {
 		return err
@@ -1756,9 +1757,9 @@ func (bc *BlockChain) reprocessBlock(parent *types.Block, current *types.Block) 
 	// If snapshots are enabled, call CommitWithSnaps to explicitly create a snapshot
 	// diff layer for the block.
 	if bc.snaps == nil {
-		return statedb.Commit(current.NumberU64(), bc.chainConfig.IsEIP158(current.Number()), false)
+		return statedb.Commit(current.NumberU64(), bc.chainConfig.IsEIP158(current.Number()))
 	}
-	return statedb.CommitWithSnap(current.NumberU64(), bc.chainConfig.IsEIP158(current.Number()), bc.snaps, current.Hash(), current.ParentHash(), false)
+	return statedb.CommitWithSnap(current.NumberU64(), bc.chainConfig.IsEIP158(current.Number()), bc.snaps, current.Hash(), current.ParentHash())
 }
 
 // initSnapshot instantiates a Snapshot instance and adds it to [bc]
@@ -1899,8 +1900,7 @@ func (bc *BlockChain) reprocessState(current *types.Block, reexec uint64) error 
 		// Flatten snapshot if initialized, holding a reference to the state root until the next block
 		// is processed.
 		if err := bc.flattenSnapshot(func() error {
-			triedb.Reference(root, common.Hash{})
-			if previousRoot != (common.Hash{}) {
+			if previousRoot != (common.Hash{}) && previousRoot != root {
 				triedb.Dereference(previousRoot)
 			}
 			previousRoot = root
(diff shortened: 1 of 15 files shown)


core/blockchain.go
```diff
@@ -1899,8 +1900,7 @@ func (bc *BlockChain) reprocessState(current *types.Block, reexec uint64) error
 		// Flatten snapshot if initialized, holding a reference to the state root until the next block
 		// is processed.
 		if err := bc.flattenSnapshot(func() error {
-			triedb.Reference(root, common.Hash{})
```
darioush, line 1902: This is removed because the triedb used in reprocess now will reference the root

core/blockchain.go
```diff
@@ -1899,8 +1900,7 @@ func (bc *BlockChain) reprocessState(current *types.Block, reexec uint64) error
...
-			if previousRoot != (common.Hash{}) {
+			if previousRoot != (common.Hash{}) && previousRoot != root {
```
darioush, line 1903: must check that previousRoot != root to avoid an edge case in reprocessing blocks when the parent is the same as the child, as `triedb.Reference` is not called the 2nd time (during block processing through statedb.Commit, because of a check there with root != origin)

eth/state_accessor.go
```diff
@@ -73,13 +74,16 @@ func (eth *Ethereum) hashState(ctx context.Context, block *types.Block, reexec u
 	// The state is both for reading and writing, or it's unavailable in disk,
 	// try to construct/recover the state over an ephemeral trie.Database for
 	// isolating the live one.
+	// Use ReferenceRoot for ephemeral trie.Databases so reference counting
+	// logic is the same as a live trie.Database.
+	tdbConfig := &triedb.Config{HashDB: &hashdb.Config{ReferenceRoot: true}}
```
darioush, line 79: this is where this change gets a little weird, but
- This code doesn't really ever run because we have reexecute = 0 at the top
- I checked all instances that create a triedb (non-test) that could flow here, but it only seems it could be from the chain

darioush: actually this is not needed because the only time statedb is passed into this fn (hashState) is the one that is returned from it (in traceChain api)
So it is safe to not consider this.

eth/state_accessor.go
```diff
@@ -172,8 +172,9 @@ func (eth *Ethereum) hashState(ctx context.Context, block *types.Block, reexec u
 		if err != nil {
 			return nil, nil, fmt.Errorf("state reset after block %d failed: %v", current.NumberU64(), err)
 		}
-		// Note: In subnet-evm, the state reference is held by passing true to [statedb.Commit].
-		// Drop the parent state to prevent accumulating too many nodes in memory.
+		// Hold the state reference and also drop the parent state
+		// to prevent accumulating too many nodes in memory.
+		tdb.Reference(root, common.Hash{})
```
darioush, line 177: I reverted the logic to having a separate call to Reference here instead of using a triedb with ref. counting.
This is because the only triedb that is used in this fn (hashState) is either created here or passed in from return value of this fn.
Basically the live chain db is not used to re-execute blocks, per my understanding

triedb/hashdb/database.go
```diff
@@ -99,6 +99,7 @@ type cache interface {
 type Config struct {
 	CleanCacheSize int    // Maximum memory allowance (in bytes) for caching clean nodes
 	StatsPrefix    string // Prefix for cache stats (disabled if empty)
+	ReferenceRoot  bool   // Whether to reference the root node on update
```
ARR4N, line 102: How about `AtomicReferenceRootOnUpdate` to explain exactly what it does? As a newcomer this was one of the more frustrating modifications because it seemed bigger as I had to drill down a few layers before realising that it was something quite simple.

ARR4N (changes requested):

triedb/hashdb/database.go
darioush: I'm going with ReferenceRootOnUpdateAtomically to avoid confusion with the AtomicTrie

core/blockchain.go
```diff
@@ -196,10 +196,11 @@ type CacheConfig struct {
 // triedbConfig derives the configures for trie database.
 func (c *CacheConfig) triedbConfig() *triedb.Config {
 	config := &triedb.Config{Preimages: c.Preimages}
-	if c.StateScheme == rawdb.HashScheme {
+	if c.StateScheme == rawdb.HashScheme || c.StateScheme == "" {
```
darioush, line 199: note some tests don't set this

ceyonur (commented):
This looks way better 🚀! 

few questions to ensure we're not changing anything in functionality-wise.

core/blockchain.go
```diff
@@ -1899,8 +1900,7 @@ func (bc *BlockChain) reprocessState(current *types.Block, reexec uint64) error
 		// Flatten snapshot if initialized, holding a reference to the state root until the next block
 		// is processed.
 		if err := bc.flattenSnapshot(func() error {
-			triedb.Reference(root, common.Hash{})
-			if previousRoot != (common.Hash{}) {
+			if previousRoot != (common.Hash{}) && previousRoot != root {
 				triedb.Dereference(previousRoot)
 			}
 			previousRoot = root
```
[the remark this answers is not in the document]
darioush: the root is already referenced here because reprocessBlock will call Commit

core/genesis.go
```diff
@@ -328,7 +328,7 @@ func (g *Genesis) toBlock(db ethdb.Database, triedb *triedb.Database) *types.Blo
 		}
 	}
 
-	statedb.Commit(0, false, false)
+	statedb.Commit(0, false)
```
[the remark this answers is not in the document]
darioush: good point, but it ends up not being an issue:
- in most cases in the next line triedb.Commit will be called, so the root will be removed from dirties (the ref-counted map) and put on disk
- if the genesis is empty I think it's fine because there would be no call to Update (because of root != origin check in statedb)

core/blockchain.go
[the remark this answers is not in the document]
darioush: note reprocessBlock is also called by populate missing tries, but right after we commit it, so it's not an issue there as well

eth/state_accessor.go
```diff
@@ -163,7 +163,7 @@ func (eth *Ethereum) hashState(ctx context.Context, block *types.Block, reexec u
 			return nil, nil, fmt.Errorf("processing block %d failed: %v", current.NumberU64(), err)
 		}
 		// Finalize the state so any modifications are written to the trie
-		root, err := statedb.Commit(current.NumberU64(), eth.blockchain.Config().IsEIP158(current.Number()), true)
+		root, err := statedb.Commit(current.NumberU64(), eth.blockchain.Config().IsEIP158(current.Number()))
```
ARR4N, line 166: Where is the config for this to be set to true?

triedb/hashdb/database.go
```diff
@@ -627,6 +631,8 @@ func (db *Database) Initialized(genesisRoot common.Hash) bool {
 
 // Update inserts the dirty nodes in provided nodeset into database and link the
 // account trie with multiple storage tries if necessary.
+// If ReferenceRoot was enabled in the config, it will also add a reference from
```
ARR4N, line 634: `// If ReferenceRootAtomicallyOnUpdate was enabled...`

ARR4N (approved):
Please address the comment in `eth_accessors` re where it's set to `true` before merging. No need for me to re-review though.

eth/state_accessor.go
darioush: Actually I have changed it to semantics using "false", and explicitly re-introduced the call to Reference.

This is more similar to upstream and is not needed to call reference atomically here because the triedb used for retracing is always one created in this function and passed back into it. (the chain triedb is only used in the readOnly case)

See this comment: https://github.com/ava-labs/subnet-evm/pull/1366#discussion_r1802271582

triedb/hashdb/database.go
darioush, line 634:
```suggestion
// If ReferenceRootAtomicallyOnUpdate was enabled in the config, it will also add a reference from
```

triedb/hashdb/database.go
```diff
@@ -137,6 +138,8 @@ type Database struct {
 	childrenSize common.StorageSize // Storage size of the external children tracking
 
 	lock sync.RWMutex
+
+	referenceRoot bool
```
ARR4N, line 142: (optional) Consider making this `referenceRootOnUpdate`. It's a little less important IMO than the exported one in `Config` as it's very close to where it's used, but still out of context. I don't think you need to call it "atomic" here because that's a given as there are locks everywhere in the code.

ARR4N (approved):

ceyonur (approved):

PR #1367, opened 2024-10-17
Sign validator uptime warp msg

ceyonur:
This pull request introduces several new features and improvements to the `warp` package, including the addition of a new example application, a no-op state implementation, and enhanced validator state management. The changes also include updates to the backend and verifier components to support the new functionalities.

* Added support for verifying offchain addressed call messages in the `verifier_backend` (`warp/verifier_backend.go`). [[1]](diffhunk://#diff-42654cfce4c7340bba25db48eff4af122345a6b4c04c01d155d7f723faf5ffd4R41-R42) [[2]](diffhunk://#diff-42654cfce4c7340bba25db48eff4af122345a6b4c04c01d155d7f723faf5ffd4R69-R138)
* Updated imports and added necessary dependencies for the new verification logic (`warp/verifier_backend.go`, `warp/verifier_backend_test.go`). [[1]](diffhunk://#diff-42654cfce4c7340bba25db48eff4af122345a6b4c04c01d155d7f723faf5ffd4R10-R11) [[2]](diffhunk://#diff-42654cfce4c7340bba25db48eff4af122345a6b4c04c01d155d7f723faf5ffd4L56-R60) [[3]](diffhunk://#diff-71ba0931b47ff7554245a4ea01a15871b5dc4312ee4ae9a8bf2316264718c617R17-R25)

* Added a new example application demonstrating how to request a signature for an uptime message from a locally running node (`examples/sign-uptime-message/main.go`).

* Updated the `backend` struct and related functions to include `uptimeCalculator`, `validatorState`, and `stateLock` for improved state management (`warp/backend.go`). [[1]](diffhunk://#diff-be66b853aa52c434887db29553e039e5536b325b159de2d4df44d2e8fb345930R62-R64) [[2]](diffhunk://#diff-be66b853aa52c434887db29553e039e5536b325b159de2d4df44d2e8fb345930R77-R79) [[3]](diffhunk://#diff-be66b853aa52c434887db29553e039e5536b325b159de2d4df44d2e8fb345930R91-R93)
* Modified backend initialization in tests to include the new state management parameters (`warp/backend_test.go`). [[1]](diffhunk://#diff-3a69b001767fa655e727b1d3763e0d128c2145db8b6dca70474c22aa08eb2a9fL47-R50) [[2]](diffhunk://#diff-3a69b001767fa655e727b1d3763e0d128c2145db8b6dca70474c22aa08eb2a9fL70-R73) [[3]](diffhunk://#diff-3a69b001767fa655e727b1d3763e0d128c2145db8b6dca70474c22aa08eb2a9fL89-R92) [[4]](diffhunk://#diff-3a69b001767fa655e727b1d3763e0d128c2145db8b6dca70474c22aa08eb2a9fL116-R119) [[5]](diffhunk://#diff-3a69b001767fa655e727b1d3763e0d128c2145db8b6dca70474c22aa08eb2a9fL170-R173)

### Validator State Management:
* Implemented a no-op state for validators, providing stub methods for state operations (`plugin/evm/validators/interfaces/noop_state.go`).

### Miscellaneous:
* Updated various test files to accommodate the new state management and verification logic (`warp/handlers/signature_request_test.go`). [[1]](diffhunk://#diff-4d0ad5b856387f529805e159cfce54b82b4a1357eed81bb801426e36e4f2d506L36-R38) [[2]](diffhunk://#diff-4d0ad5b856387f529805e159cfce54b82b4a1357eed81bb801426e36e4f2d506R144-R146)
* Removed unused `Signable` interface from the `warp/messages/payload.go` file.
* Added comments and minor adjustments to existing code for clarity and consistency (`warp/messages/validator_uptime.go`).

These changes collectively enhance the functionality and maintainability of the `warp` package, particularly in terms of state management and message verification.

## How this was tested

Added UTs

## How is this documented

Not sure if we have any warp types docs.


plugin/evm/vm.go
```diff
@@ -266,8 +266,9 @@ type VM struct {
 	ethTxPushGossiper  avalancheUtils.Atomic[*gossip.PushGossiper[*GossipEthTx]]
 	ethTxPullGossiper  gossip.Gossiper
 
-	uptimeManager  uptime.PausableManager
-	validatorState validators.State
+	UptimeLockedCalculator avalancheUptime.LockedCalculator
```
darioush, line 269: nit: can this be unexported?

darioush (dismissed):
this PR looks good,

only comment (same as https://github.com/ava-labs/subnet-evm/pull/1263#discussion_r1821338194) is that if validator uptime tracking is moved to its own file/struct, the UptimeLockedCalculator can be moved there too.

warp/verifier_backend.go
```diff
@@ -62,3 +66,70 @@ func (b *backend) verifyBlockMessage(ctx context.Context, blockHashPayload *payl
 
 	return nil
 }
+
+// verifyAddressedCall verifies the addressed call message
+func (b *backend) verifyAddressedCall(addressedCall *payload.AddressedCall) *common.AppError {
```
michaelkaplan13, line 71: Should we rename this `verifyOffChainAddressedCall`? Since all addressed calls emitted on-chain will be verified already above by calling `GetMessage`.

Also, I think we should add the requirement that `len(addressedCall.SourceAddress) == 0` here, so that off-chain messages can always be differentiated from messages emitted on-chain. Otherwise, if a contract wanted to emit a message with a format matching one defined here, a message from that contract could be "forged".

plugin/evm/validators/state.go
```diff
@@ -241,8 +246,8 @@ func (s *state) GetValidationIDs() set.Set[ids.ID] {
 	return ids
 }
 
-// GetValidatorIDs returns the validator IDs in the state
-func (s *state) GetValidatorIDs() set.Set[ids.NodeID] {
+// GetNodeIDs returns the validator IDs in the state
```
michaelkaplan13, line 249:
```suggestion
// GetNodeIDs returns the node IDs of validators in the state
```

warp/verifier_backend.go
```diff
@@ -62,3 +66,70 @@ func (b *backend) verifyBlockMessage(ctx context.Context, blockHashPayload *payl
...
+	if err != nil {
+		return &common.AppError{
+			Code:    VerifyErrCode,
+			Message: fmt.Sprintf("failed to calculate uptime for nodeID %s: %s", nodeID, err.Error()),
+		}
+	}
+
+	currentUptimeSeconds := uint64(currentUptime.Seconds())
+	// verify the current uptime against the total uptime in the message
+	if currentUptimeSeconds < uptimeMsg.TotalUptime {
```
michaelkaplan13, line 127: Nit: Should this be `<=`?

ceyonur: IMHO it should be rewarded if they have same uptime as requested. Also TotalUptime=0 should be rewarding everyone even without any uptime.

michaelkaplan13: Ah, yeah you're totally right I was reading this wrong. Sorry about that

warp/verifier_backend.go
michaelkaplan13: I see now that `verifyUptimeMessage` has the check on `len(sourceAddress)` so the code is correct, though still think it likely would be a bit less error prone to include that check here instead so it applies in all cases for off-chain messages.

plugin/evm/validators/interfaces/noop_state.go
```diff
@@ -0,0 +1,63 @@
+package interfaces
```
michaelkaplan13, line 1: Seems a little odd to put this under the `interfaces` package when it is an implementation and doesn't have any new interface definitions

warp/verifier_backend.go
```diff
@@ -36,6 +38,8 @@ func (b *backend) Verify(ctx context.Context, unsignedMessage *avalancheWarp.Uns
 	}
 
 	switch p := parsed.(type) {
+	case *payload.AddressedCall:
```
michaelkaplan13, line 41: Not new in this PR, but should we check above on line 30 if `err != MessageNotFound`, and return the error in that case?

michaelkaplan13 (dismissed):

warp/verifier_backend.go
ceyonur: yea good idea.

michaelkaplan13 (approved):

PR #1371, opened 2024-10-28
Validator state

ceyonur:
## Why this should be merged

Adds validator state that can persist validator information and implements `uptime.State`.

## How this was tested

Added UT

## How is this documented

Comments in code


plugin/evm/validators/state_test.go
```diff
@@ -0,0 +1,286 @@
...
+	expectedStartTime := time.Now()
+
+	// add listener
+	listener := &testCallbackListener{
+		t: t,
+		onAdd: func(vID ids.ID, nodeID ids.NodeID, startTime uint64, isActive bool) {
+			require.Equal(expectedvID, vID)
+			require.Equal(expectedNodeID, nodeID)
+			require.Equal(uint64(expectedStartTime.Unix()), startTime)
+			require.True(isActive)
```
darioush, line 231: should probably have some verification that these callbacks are actually invoked

darioush (approved):

plugin/evm/validators/state_test.go
ceyonur: dunno what I was thinking, mocks are much easier...

plugin/evm/validators/state_test.go
```diff
@@ -216,71 +217,34 @@ func TestStateListener(t *testing.T) {
 	db := memdb.New()
 	state, err := NewState(db)
 	require.NoError(err)
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
```
darioush, line 221: nit: I think calls to Finish are not required in recent gomock versions.

darioush (approved):

plugin/evm/validators/state_test.go
ceyonur: let's keep it for now

PR #1372, opened 2024-10-28
Pausable uptime manager

ceyonur:
## Why this should be merged

Adds pausable uptime manager interface for uptime tracking

## How this works

Pausable uptime manager extends the avalanchego uptime manager by introducing callback listeners for validators state (under `plugin/evm/validators`). The idea is to pause a node's uptime and then can continue whenever it's resumed. It also modifies `Connect`, `Disconnect,` StartTracking` accordingly. 

## How this was tested

Added tests

## How is this documented

Comments in code

### DIFF
--- a/plugin/evm/uptime/pausable_manager.go
+++ b/plugin/evm/uptime/pausable_manager.go
@@ -13,13 +13,13 @@ import (
 	"github.com/ava-labs/avalanchego/utils/set"
 )
 
-var _ validators.ValidatorsCallbackListener = &pausableManager{}
+var _ validators.StateCallbackListener = &pausableManager{}
 
 var ErrPausedDc = errors.New("paused node cannot be disconnected")
 
 type PausableManager interface {
 	uptime.Manager
-	validators.ValidatorsCallbackListener
+	validators.StateCallbackListener
 	IsPaused(nodeID ids.NodeID) bool
 }
 
--- a/plugin/evm/uptime/pausable_manager.go
+++ b/plugin/evm/uptime/pausable_manager.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 
 	"github.com/ava-labs/subnet-evm/plugin/evm/validators"
+	"github.com/ethereum/go-ethereum/log"
 
 	"github.com/ava-labs/avalanchego/ids"
 	"github.com/ava-labs/avalanchego/snow/uptime"
@@ -15,7 +16,11 @@ import (
 
 var _ validators.StateCallbackListener = &pausableManager{}
 
-var ErrPausedDc = errors.New("paused node cannot be disconnected")
+var (
+	errPausedDisconnect = errors.New("paused node cannot be disconnected")
+	errAlreadyPaused    = errors.New("node is already paused")
+	errNotPaused        = errors.New("node is not paused")
+)
 
 type PausableManager interface {
 	uptime.Manager
@@ -59,7 +64,7 @@ func (p *pausableManager) Disconnect(nodeID ids.NodeID) error {
 	if p.Manager.IsConnected(nodeID) {
 		if p.IsPaused(nodeID) {
 			// We should never see this case
-			return ErrPausedDc
+			return errPausedDisconnect
 		}
 		return p.Manager.Disconnect(nodeID)
 	}
@@ -82,7 +87,10 @@ func (p *pausableManager) StartTracking(nodeIDs []ids.NodeID) error {
 // If the node is inactive, it will be paused.
 func (p *pausableManager) OnValidatorAdded(vID ids.ID, nodeID ids.NodeID, startTime uint64, isActive bool) {
 	if !isActive {
-		p.pause(nodeID)
+		err := p.pause(nodeID)
+		if err != nil {
+			log.Error("failed to pause node %s: %s", nodeID, err)
+		}
 	}
 }
 
@@ -90,17 +98,24 @@ func (p *pausableManager) OnValidatorAdded(vID ids.ID, nodeID ids.NodeID, startT
 // If the node is already paused, it will be resumed.
 func (p *pausableManager) OnValidatorRemoved(vID ids.ID, nodeID ids.NodeID) {
 	if p.IsPaused(nodeID) {
-		p.resume(nodeID)
+		err := p.resume(nodeID)
+		if err != nil {
+			log.Error("failed to resume node %s: %s", nodeID, err)
+		}
 	}
 }
 
 // OnValidatorStatusUpdated is called when the status of a validator is updated.
 // If the node is active, it will be resumed. If the node is inactive, it will be paused.
 func (p *pausableManager) OnValidatorStatusUpdated(vID ids.ID, nodeID ids.NodeID, isActive bool) {
+	var err error
 	if isActive {
-		p.resume(nodeID)
+		err = p.resume(nodeID)
 	} else {
-		p.pause(nodeID)
+		err = p.pause(nodeID)
+	}
+	if err != nil {
+		log.Error("failed to update status for node %s: %s", nodeID, err)
 	}
 }
 
@@ -113,6 +128,9 @@ func (p *pausableManager) IsPaused(nodeID ids.NodeID) bool {
 // pause can disconnect the node from the uptime.Manager if it is connected.
 // Returns an error if the node is already paused.
 func (p *pausableManager) pause(nodeID ids.NodeID) error {
+	if p.IsPaused(nodeID) {
+		return errAlreadyPaused
+	}
 	p.pausedVdrs.Add(nodeID)
 	if p.Manager.IsConnected(nodeID) {
 		// If the node is connected, then we need to disconnect it from
@@ -128,6 +146,9 @@ func (p *pausableManager) pause(nodeID ids.NodeID) error {
 // resume can connect the node to the uptime.Manager if it was connected.
 // Returns an error if the node is not paused.
 func (p *pausableManager) resume(nodeID ids.NodeID) error {
+	if !p.IsPaused(nodeID) {
+		return errNotPaused
+	}
 	p.pausedVdrs.Remove(nodeID)
 	if p.connectedVdrs.Contains(nodeID) {
 		return p.Manager.Connect(nodeID)
--- a/plugin/evm/uptime/pausable_manager.go
+++ b/plugin/evm/uptime/pausable_manager.go
@@ -1,0 +1,136 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package uptime
+
+import (
+	"errors"
+
+	"github.com/ava-labs/subnet-evm/plugin/evm/validators"
+
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/snow/uptime"
+	"github.com/ava-labs/avalanchego/utils/set"
+)
+
+var _ validators.ValidatorsCallbackListener = &pausableManager{}
+
+var ErrPausedDc = errors.New("paused node cannot be disconnected")
+
+type PausableManager interface {
+	uptime.Manager
+	validators.ValidatorsCallbackListener
+	IsPaused(nodeID ids.NodeID) bool
+}
+
+type pausableManager struct {
+	uptime.Manager
+	pausedVdrs set.Set[ids.NodeID]
+	// connectedVdrs is a set of nodes that are connected to the manager.
+	// This is used to keep track of the nodes that are connected to the manager
+	// but are paused.
+	connectedVdrs set.Set[ids.NodeID]
+}
+
+// NewPausableManager takes an uptime.Manager and returns a PausableManager
+func NewPausableManager(manager uptime.Manager) PausableManager {
+	return &pausableManager{
+		pausedVdrs:    make(set.Set[ids.NodeID]),
+		connectedVdrs: make(set.Set[ids.NodeID]),
+		Manager:       manager,
+	}
+}
+
+// Connect connects the node with the given ID to the uptime.Manager
+// If the node is paused, it will not be connected
+func (p *pausableManager) Connect(nodeID ids.NodeID) error {
+	p.connectedVdrs.Add(nodeID)
+	if !p.IsPaused(nodeID) && !p.Manager.IsConnected(nodeID) {
+		return p.Manager.Connect(nodeID)
+	}
+	return nil
+}
+
+// Disconnect disconnects the node with the given ID from the uptime.Manager
+// If the node is paused, it will not be disconnected
+// Invariant: we should never have a connected paused node that is disconnecting
+func (p *pausableManager) Disconnect(nodeID ids.NodeID) error {
+	p.connectedVdrs.Remove(nodeID)
+	if p.Manager.IsConnected(nodeID) {
+		if p.IsPaused(nodeID) {
+			// We should never see this case
+			return ErrPausedDc
+		}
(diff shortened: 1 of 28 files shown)


plugin/evm/uptime/pausable_manager.go
```diff
@@ -0,0 +1,157 @@
...
+// resume resumes uptime tracking for the node with the given ID
+// resume can connect the node to the uptime.Manager if it was connected.
+// Returns an error if the node is not paused.
+func (p *pausableManager) resume(nodeID ids.NodeID) error {
+	if !p.IsPaused(nodeID) {
+		return errNotPaused
+	}
+	p.pausedVdrs.Remove(nodeID)
+	if p.connectedVdrs.Contains(nodeID) {
+		return p.Manager.Connect(nodeID)
```
darioush, line 154: should we check `p.Manager.IsConnected()` here? since line 53 seems to check this?

ceyonur: I think we don't actually need to check them since we should already be disconnected from the manager in both places. but let's keep it safe.

plugin/evm/uptime/pausable_manager_test.go
```diff
@@ -0,0 +1,205 @@
...
+		// Elapsed time
+		currentTime = addTime(clk, 7*time.Second)
+		// Uptime should not have increased since the node was not connected
+		checkUptime(t, up, nodeID0, 13*time.Second, currentTime)
+
+		// Connect and final uptime check
+		require.NoError(up.Connect(nodeID0))
+		currentTime = addTime(clk, 8*time.Second)
+		// Uptime should have increased by 8 seconds since the node was connected
+		checkUptime(t, up, nodeID0, 21*time.Second, currentTime)
```
darioush, line 182: (optional) imo this test would read slightly better if we had a running total for `expectedUptime` then each time calling `checkUptime(t, up, nodeID0, expectedUptime, currentTime)`

plugin/evm/uptime/pausable_manager.go
```diff
@@ -0,0 +1,156 @@
...
+// StartTracking starts tracking uptime for the nodes with the given IDs
+// If a node is paused, it will not be tracked
+func (p *pausableManager) StartTracking(nodeIDs []ids.NodeID) error {
+	var activeNodeIDs []ids.NodeID
+	for _, nodeID := range nodeIDs {
+		if !p.IsPaused(nodeID) {
+			activeNodeIDs = append(activeNodeIDs, nodeID)
+		}
+	}
+	return p.Manager.StartTracking(activeNodeIDs)
```
darioush, line 82: is it needed to wrap StopTracking as well? (I suppose this is checked in Case 4?)

darioush (dismissed):

plugin/evm/uptime/pausable_manager.go
ceyonur: No, since they're guaranteed to be disconnected when `StopTracking` is called; their lastUpdated time updated with the `now` while keeping the uptime when they got paused. So in some sense we only refresh their lastUpdated. This means in `CalculateUptime` we don't double-count the time between `pause()` and `StopTracking`. I added a `checkUptime` right after we call `StopTracking`, it would increase if we wrapped `StopTracking`.

plugin/evm/uptime/pausable_manager_test.go
ceyonur: changed it

plugin/evm/uptime/pausable_manager.go
```diff
@@ -0,0 +1,146 @@
...
+		}
+		return p.Manager.Disconnect(nodeID)
+	}
+	return nil
+}
+
+// StartTracking starts tracking uptime for the nodes with the given IDs
+// If a node is paused, it will not be tracked
+func (p *pausableManager) StartTracking(nodeIDs []ids.NodeID) error {
+	var activeNodeIDs []ids.NodeID
```
yacovm, line 72: perhaps initialize `activeNodeIDs` with a capacity of `nodeIDs` ?

plugin/evm/uptime/pausable_manager.go
```diff
@@ -0,0 +1,146 @@
...
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/snow/uptime"
+	"github.com/ava-labs/avalanchego/utils/set"
+)
+
+var _ validators.StateCallbackListener = &pausableManager{}
+
+var errPausedDisconnect = errors.New("paused node cannot be disconnected")
+
+type PausableManager interface {
```
yacovm, line 21: this is a recurring pattern I see in avalanche codebase - defining an interface where it's implemented, instead of where it's used. 

What is the benefit of doing it this way? 

Anyone that wants to use this interface needs to import this package directly, so you're creating coupling between any consumer package to this package.

plugin/evm/uptime/pausable_manager.go
```diff
@@ -0,0 +1,146 @@
...
+		pausedVdrs:    make(set.Set[ids.NodeID]),
+		connectedVdrs: make(set.Set[ids.NodeID]),
+		Manager:       manager,
+	}
+}
+
+// Connect connects the node with the given ID to the uptime.Manager
+// If the node is paused, it will not be connected
+func (p *pausableManager) Connect(nodeID ids.NodeID) error {
+	p.connectedVdrs.Add(nodeID)
```
yacovm, line 47: what is the meaning of adding it here to the `connectedVdrs` set, and similarly in `Disconnect` - removing it from the set, in case the `if` condition is false?

ceyonur: Paused nodes are disconnected from the inner manager, so inner manager would not know about their connectivity if they resumed again. The set tracks if a node is paused (by deactivation) but still connected in the network layer.

plugin/evm/uptime/pausable_manager.go
```diff
@@ -0,0 +1,146 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
```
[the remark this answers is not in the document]
ceyonur: Yes there is #1263 PR uses this manager.

plugin/evm/uptime/pausable_manager.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+// IsPaused returns true if the node with the given ID is paused.
+func (p *pausableManager) IsPaused(nodeID ids.NodeID) bool {
+	return p.pausedVdrs.Contains(nodeID)
+}
+
+// pause pauses uptime tracking for the node with the given ID
+// pause can disconnect the node from the uptime.Manager if it is connected.
+// Returns an error if the node is already paused.
```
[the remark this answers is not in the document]
ceyonur: that's a stale comment

plugin/evm/uptime/pausable_manager.go
```diff
@@ -0,0 +1,146 @@
...
+		// This should be fine in case tracking has not started yet since
+		// the inner manager should handle disconnects accordingly
+		return p.Manager.Disconnect(nodeID)
+	}
+	return nil
+}
+
+// resume resumes uptime tracking for the node with the given ID
+// resume can connect the node to the uptime.Manager if it was connected.
+// Returns an error if the node is not paused.
```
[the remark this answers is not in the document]
ceyonur: removed

plugin/evm/uptime/pausable_manager.go
ceyonur: Yea I agree this not ideal. I think this is an approach that is valid until it breaks. Ideally we should have a separate subpackage for interfaces.

plugin/evm/uptime/pausable_manager.go
ceyonur: good idea, done

plugin/evm/uptime/pausable_manager.go
ceyonur: We would not know if resumed ones actually connected or not, because we disconnect paused nodes from the inner manager. After that point there is no way to tell if a paused one is actually connected or not. In the `resume` we need to tell the inner manager that the node is connected (so tracking resumes).

plugin/evm/uptime/pausable_manager.go
yacovm: Well I think ideally, the interface is defined where it's consumed, not where it's implemented.

This is not related to this PR, however, it's a general comment.

ceyonur: https://github.com/ava-labs/subnet-evm/pull/1379/files

plugin/evm/uptime/pausable_manager_test.go
```diff
@@ -0,0 +1,234 @@
...
+				expectedUptime += 5 * time.Second
+				checkUptime(t, up, nodeID0, expectedUptime, currentTime)
+
+				// Start tracking and check elapsed time
+				currentTime = addTime(clk, 6*time.Second)
+				require.NoError(up.StartTracking([]ids.NodeID{nodeID0}))
+				// Uptime should have increased by 6 seconds since we started tracking
+				// and node was resumed (we assume the node was online until we started tracking)
+				expectedUptime += 6 * time.Second
+				checkUptime(t, up, nodeID0, expectedUptime, currentTime)
```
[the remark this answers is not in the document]
ceyonur: We assume all nodes are online when we go offline (until we start tracking again). In this case we have first resumed the node, and then started tracking again. It indicates the paused node was resumed while we're offline. So we should increase the uptime of the node. 

In between this and 194 it should not increase the uptime since the nodeID0 is not connected to the manager yet.

plugin/evm/uptime/pausable_manager_test.go
```diff
@@ -19,167 +19,195 @@ func TestPausableManager(t *testing.T) {
...
-		checkUptime(t, up, nodeID0, 0*time.Second, currentTime)
+	tests := []struct {
+		name     string
+		testFunc func(t *testing.T, up PausableManager, clk *mockable.Clock, s uptime.State)
+	}{
+		{
+			name: "Case 1: Connect, pause, start tracking",
+			testFunc: func(t *testing.T, up PausableManager, clk *mockable.Clock, s uptime.State) {
+				// Connect before tracking
+				require.NoError(up.Connect(nodeID0))
```
darioush, line 30: These require will capture the wrong `t` and panic, we can define the require per testFunc or use require.NoError(t, ...) etc.

darioush (approved):

darioush (approved):

michaelkaplan13 (commented):
Could you confirm my understanding based on going through this code in the context of ACP-77?
- Validators can are either connected/disconnected and active/inactive.
    - A connected & inactive validator represents a validators running properly, but out of PAYG on the P-Chain
    - A connected & active validator is both running properly and has non-zero balance on the P-Chain
    - A disconnected & active validator represents an offline node that still has non-zero balance on the P-Chain
    - A disconnected & inactive validator is a node that is both offline and out of PAYG balance.
- We pause nodes when they are marked as inactive, and when we pause them, we disconnect from them if we were previously connected.
- We resume nodes when they are marked as active, and reconnect to them then if necessary.
- A nodes uptime is incremented when: (it is both active and connected) OR (it isn't tracked). The second part is because we assume validators are online while we ourselves are offline.

plugin/evm/uptime/pausable_manager.go
```diff
@@ -0,0 +1,144 @@
...
+	if !isActive {
+		err := p.pause(nodeID)
+		if err != nil {
+			log.Error("failed to handle added validator %s: %s", nodeID, err)
+		}
+	}
+}
+
+// OnValidatorRemoved is called when a validator is removed.
+// If the node is already paused, it will be resumed.
```
michaelkaplan13, line 93: Why is this the case again? Shouldn't we stop tracking the uptime for any nodes that have been removed as validators?

plugin/evm/uptime/pausable_manager.go
```diff
@@ -0,0 +1,144 @@
...
+		if !p.IsPaused(nodeID) {
+			activeNodeIDs = append(activeNodeIDs, nodeID)
+		}
+	}
+	return p.Manager.StartTracking(activeNodeIDs)
+}
+
+// OnValidatorAdded is called when a validator is added.
+// If the node is inactive, it will be paused.
+func (p *pausableManager) OnValidatorAdded(vID ids.ID, nodeID ids.NodeID, startTime uint64, isActive bool) {
```
yacovm, line 83: nit: `vID` and `startTime` are not used.

plugin/evm/uptime/pausable_manager.go
```diff
@@ -0,0 +1,144 @@
...
+		err := p.pause(nodeID)
+		if err != nil {
+			log.Error("failed to handle added validator %s: %s", nodeID, err)
+		}
+	}
+}
+
+// OnValidatorRemoved is called when a validator is removed.
+// If the node is already paused, it will be resumed.
+func (p *pausableManager) OnValidatorRemoved(vID ids.ID, nodeID ids.NodeID) {
```
yacovm, line 94: nit: `vID` is not used

plugin/evm/uptime/pausable_manager.go
```diff
@@ -0,0 +1,144 @@
...
+		err := p.resume(nodeID)
+		if err != nil {
+			log.Error("failed to handle validator removed %s: %s", nodeID, err)
+		}
+	}
+}
+
+// OnValidatorStatusUpdated is called when the status of a validator is updated.
+// If the node is active, it will be resumed. If the node is inactive, it will be paused.
+func (p *pausableManager) OnValidatorStatusUpdated(vID ids.ID, nodeID ids.NodeID, isActive bool) {
```
yacovm, line 105: nit: `vID` is not used.

yacovm (dismissed):

ceyonur:
> Could you confirm my understanding based on going through this code in the context of ACP-77?
> 
>     * Validators can are either connected/disconnected and active/inactive.
Yes, those two states are different though. They can be connected/active disconnected/inactive etc...
>       * A connected & inactive validator represents a validators running properly, but out of PAYG on the P-Chain
True
>       * A connected & active validator is both running properly and has non-zero balance on the P-Chain
True
>       * A disconnected & active validator represents an offline node that still has non-zero balance on the P-Chain
True
>       * A disconnected & inactive validator is a node that is both offline and out of PAYG balance.
True
> 
>     * We pause nodes when they are marked as inactive, and when we pause them, we disconnect from them if we were previously connected.
True, the inner manager (who does the actual uptime tracking) is disconnected.
> 
>     * We resume nodes when they are marked as active, and reconnect to them then if necessary.
> 
True. Note that they're already connected (in p2p), we just start increasing their uptime in manager.
>     * A nodes uptime is incremented when: (it is both active and connected) OR (it isn't tracked). The second part is because we assume validators are online while we ourselves are offline.
true

plugin/evm/uptime/pausable_manager.go
michaelkaplan13: Hmm, do we need to disconnect in the inner manager here then? I don't understand where the inner manager would know to stop tracking the uptime (and delete its previously tracked uptime) when a validator ends currently.

ceyonur: It's the state manages the stored uptime durations. This callback is triggered by the actual validator deletion in the state, which purges the uptime data. https://github.com/ava-labs/subnet-evm/blob/pausable-uptime-manager/plugin/evm/validators/state.go#L156

michaelkaplan13: Got it, thanks for the link 🙏

michaelkaplan13 (approved):

PR #1374, opened 2024-10-29
Uptime tracking api

ceyonur:
## Why this should be merged

Adds Validators API (GetCurrentValidators) 

## How this works

Refactors API (RPC) and Service files, adds new handler for `/validators` endpoint and implement `GetCurrentValidators` API

## How this was tested

Locally

## How is this documented

Need to be documented


plugin/evm/api.go
```diff
@@ -0,0 +1,38 @@
+// (c) 2019-2020, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
```
darioush, line 2: should we leave this in `service.go` and add the new stuff in `service_validator.go`?

or we can rename this to something like `api_snowman.go` and have `api_validator.go`?

plugin/evm/vm.go
```diff
@@ -1082,6 +1083,16 @@ func (vm *VM) CreateHandlers(context.Context) (map[string]http.Handler, error) {
...
+	if vm.config.ValidatorsAPIEnabled {
+		validatorsAPI, err := newHandler("validators", &ValidatorsAPI{vm})
+		if err != nil {
+			return nil, fmt.Errorf("failed to register service for admin API due to %w", err)
+		}
+		apis[validatorsEndpoint] = validatorsAPI
+		enabledAPIs = append(enabledAPIs, "validators")
+	}
+
+	// RPC APIs
```
darioush, line 1095: nit: not sure which RPC this comment is referring to?

plugin/evm/vm.go
```diff
@@ -1082,6 +1083,16 @@ func (vm *VM) CreateHandlers(context.Context) (map[string]http.Handler, error) {
 		enabledAPIs = append(enabledAPIs, "subnet-evm-admin")
 	}
 
+	if vm.config.ValidatorsAPIEnabled {
+		validatorsAPI, err := newHandler("validators", &ValidatorsAPI{vm})
+		if err != nil {
+			return nil, fmt.Errorf("failed to register service for admin API due to %w", err)
```
darioush, line 1089:
```suggestion
			return nil, fmt.Errorf("failed to register service for validators API due to %w", err)
```

darioush (dismissed):

plugin/evm/vm.go
ceyonur: `/rpc` endpoint

plugin/evm/api.go
ceyonur: My intention was to keep avalancheRPC server handlers separate from `/rpc` handler, and follow avalanche endpoints -> service, /rpc endpoint -> api convention.

michaelkaplan13 (approved):
LGTM, only question is if there is an existing E2E test suite that we could add a new test case to in order to confirm the new `/validators` endpoints is working as expected in CI?

michaelkaplan13 (approved):

PR #1378, opened 2024-11-04
Uptime validation nits

ceyonur:
## Why this should be merged

* Adds a example for uptime warp msg signing
* Adds `IsSoV` and `weight` fields to validators 

## How this was tested

Added e2e test

## How is this documented

should document whole `GetCurrentValidators` API

### DIFF
--- a/examples/sign-uptime-message/main.go
+++ b/examples/sign-uptime-message/main.go
@@ -1,0 +1,118 @@
+// Copyright (C) 2019-2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package main
+
+import (
+	"context"
+	"log"
+	"net/netip"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"google.golang.org/protobuf/proto"
+
+	"github.com/ava-labs/avalanchego/api/info"
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/network/p2p"
+	"github.com/ava-labs/avalanchego/network/peer"
+	"github.com/ava-labs/avalanchego/proto/pb/sdk"
+	"github.com/ava-labs/avalanchego/snow/networking/router"
+	"github.com/ava-labs/avalanchego/utils/compression"
+	"github.com/ava-labs/avalanchego/utils/logging"
+	"github.com/ava-labs/avalanchego/vms/platformvm/warp"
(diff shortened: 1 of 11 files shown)


examples/sign-uptime-message/main.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/ava-labs/avalanchego/utils/logging"
+	"github.com/ava-labs/avalanchego/vms/platformvm/warp"
+	"github.com/ava-labs/avalanchego/vms/platformvm/warp/payload"
+	"github.com/ava-labs/avalanchego/wallet/subnet/primary"
+	"github.com/ava-labs/subnet-evm/warp/messages"
+
+	p2pmessage "github.com/ava-labs/avalanchego/message"
+)
+
+func main() {
```
darioush, line 31: optional: should we format this as Example so it would be runnable by the test framework?

examples/sign-uptime-message/main.go
```diff
@@ -0,0 +1,118 @@
...
+	uri := primary.LocalAPIURI
+	validationID := ids.FromStringOrPanic("p3NUAY4PbcAnyCyvUTjGVjezNEQCdnVdfAbJcZScvKpxP5tJr")
```
darioush, line 33: could we add a comment about where these IDs are coming from or use a generated ID if it's arbitrary?

plugin/evm/validators/noop_state.go
```diff
@@ -45,7 +45,7 @@ func (n *noOpState) GetStartTime(
 	return time.Time{}, nil
 }
 
-func (n *noOpState) AddValidator(vID ids.ID, nodeID ids.NodeID, startTimestamp uint64, isActive bool) error {
+func (n *noOpState) AddValidator(vID ids.ID, nodeID ids.NodeID, weight uint64, startTimestamp uint64, isActive bool, isSoV bool) error {
```
darioush, line 48: nit: should we just use the Validator struct here?

darioush (approved):

examples/sign-uptime-message/main.go
ceyonur: I think we can add a separate, this is to provide an example how to send sign request.

plugin/evm/validators/noop_state.go
ceyonur: done

examples/sign-uptime-message/main.go
ceyonur: added

darioush (approved):

PR #1383, opened 2024-11-13
remove shared memory from precompiles and accept

ceyonur:
## Why this should be merged

Currently shared memory writes version DB to the shared database (avaDB). This conflicts and corrupts the standaloneDB and leads to a duality in data sources: acceptedBlockDB and metaDB were being written to the shared db (avaDB) and chaindb was being written to the standalone DB. This PR removes the unused shared memory components from subnet-evm.
 
## How this works

This pull request includes significant refactoring and cleanup of the `plugin/evm` package, specifically focusing on the removal of shared memory operations and simplifying database interactions. The most important changes include the removal of the `sharedMemoryWriter`, modifications to the `Accept` method, and updates to the test cases to reflect these changes.

* [`plugin/evm/block.go`](diffhunk://#diff-9a870b1572ea65697aea0a8f959e0ee20efa4ff3370c3e39dcc2f3ec436b4fc3L64-R66): Removed the use of `sharedMemoryWriter` and related logic from the `Accept` method and `handlePrecompileAccept` function. [[1]](diffhunk://#diff-9a870b1572ea65697aea0a8f959e0ee20efa4ff3370c3e39dcc2f3ec436b4fc3L64-R66) [[2]](diffhunk://#diff-9a870b1572ea65697aea0a8f959e0ee20efa4ff3370c3e39dcc2f3ec436b4fc3L80-R82) [[3]](diffhunk://#diff-9a870b1572ea65697aea0a8f959e0ee20efa4ff3370c3e39dcc2f3ec436b4fc3L112)
* [`plugin/evm/shared_memory_writer.go`](diffhunk://#diff-ff39fc9a42e325fe6e37d26f214866669a4bf9c2127e201da62f3373b93a9dffL1-L37): Deleted the entire `shared_memory_writer.go` file, which included the `sharedMemoryWriter` implementation.

* [`plugin/evm/vm.go`](diffhunk://#diff-3be5ea9687f8af021d9baaa6584d174a7975b181b604e1a759e4ad2e40136912L1258-R1258): Modified the `initializeDBs` method to use `avaDB` directly instead of creating a versioned database.
* [`plugin/evm/vm_test.go`](diffhunk://#diff-7cc8115837d92067c0e48165d5b49fcd63797639f8680ceee734caa4983e9b32L209-R210): Updated test setup to use `baseDB` directly and added a new test `TestStandaloneDB` to ensure the standalone database is used correctly. [[1]](diffhunk://#diff-7cc8115837d92067c0e48165d5b49fcd63797639f8680ceee734caa4983e9b32L209-R210) [[2]](diffhunk://#diff-7cc8115837d92067c0e48165d5b49fcd63797639f8680ceee734caa4983e9b32R3154-R3207)

## How this was tested

Added regression UT & locally tested

## Need to be documented?

No

No


ARR4N (commented):
I don't know enough about the problem to give a safe approval, but overall LGTM.

plugin/evm/vm_test.go
```diff
@@ -206,8 +207,7 @@ func setupGenesis(
 	ctx.Keystore = userKeystore.NewBlockchainKeyStore(ctx.ChainID)
 
 	issuer := make(chan commonEng.Message, 1)
-	prefixedDB := prefixdb.New([]byte{1}, baseDB)
-	return ctx, prefixedDB, genesisBytes, issuer, atomicMemory
+	return ctx, baseDB, genesisBytes, issuer, atomicMemory
```
darioush, line 210: IMO this is not entirely correct since it's returning the baseDB which has a prefixDB defined on top of it, let's either return a new memdb or retain the prefixDB so the atomicMemory is separate. 
This could be an issue in iteration if the atomic memory is used

plugin/evm/vm_test.go
```diff
@@ -3151,3 +3151,57 @@ func TestParentBeaconRootBlock(t *testing.T) {
 		})
 	}
 }
+
+func TestStandaloneDB(t *testing.T) {
+	vm := &VM{}
+	ctx, dbManager, genesisBytes, issuer, _ := setupGenesis(t, genesisJSONLatest)
+	// alter network ID to use standalone database
```
darioush, line 3158:
```suggestion
	// alter network ID to use standalone database, as by default unit test network does not use it.
```

plugin/evm/vm_test.go
ceyonur: In the regression test we should check the `baseDB` rather than `prefixedDB` where the `baseDB` is used in `sharedMemory` and we test out if `sharedMemory` writes to the `baseDB`.

darioush (approved):

darioush (approved):

PR #1388, opened 2024-11-14
Refactor uptime tracking

ceyonur:
## Why this should be merged

Current VM package was operating state + uptime manager together and had an increased `Initialialize` complexity. This PR is to move that to another component. 

## How this works

This pull request introduces significant changes to the validator management system in the `subnet-evm` plugin. The main focus is on refactoring the code to use a new `validatorsManager` for managing validators instead of the previous `validatorState` and `uptimeManager`. Additionally, it includes the introduction of new interfaces and restructuring of existing code to improve modularity and maintainability.

Key changes include:

### Refactoring and Code Improvements:
* Refactored the `GetCurrentValidators` function in `plugin/evm/service.go` to use `validatorsManager` instead of `validatorState` for retrieving validator information.
* Introduced new interfaces in `plugin/evm/validators/interfaces/interfaces.go` to define `ValidatorReader` and `Manager` for better abstraction and modularity.
* Added a new `manager` implementation in `plugin/evm/validators/manager.go` to handle validator synchronization and state management.

### Testing Enhancements:
* Added comprehensive tests in `plugin/evm/validators/manager_test.go` to validate the new validator management functionality, including scenarios for adding, updating, and removing validators.

### File Restructuring:
* Renamed and moved several files for better organization, including renaming `plugin/evm/validators/codec.go` to `plugin/evm/validators/state/codec.go` and `plugin/evm/validators/interfaces/interface.go` to `plugin/evm/validators/state/interfaces/state.go`. [[1]](diffhunk://#diff-f9e7efc5ecf92d59d942e09adc8c6aed99b3529442338fe0ec91e2497800d7d8L4-R4) [[2]](diffhunk://#diff-190cee8f203955ebb957f4da27c648c58fdb9363e70a4e7329303ae5dc58dd39L7-L31)

### Code Cleanup:
* Removed the `noop_state.go` file from `plugin/evm/validators/validatorstest` as it is no longer needed.

These changes collectively aim to enhance the maintainability and functionality of the validator management system in the `subnet-evm` plugin.

## How this was tested

Locally and UTs

## Need to be documented?

No

No


plugin/evm/vm.go
```diff
@@ -740,28 +728,28 @@ func (vm *VM) onNormalOperationsStarted() error {
 	ctx, cancel := context.WithCancel(context.TODO())
 	vm.cancel = cancel
 
-	// update validators first
-	if err := vm.performValidatorUpdate(ctx); err != nil {
+	// sync validators first
+	if err := vm.validatorsManager.Sync(ctx); err != nil {
```
darioush, line 732: Should we move the initialization logic (from here to line 747) to the validatorsManager?
such as:
```suggestion
	if err := vm.validatorsManager.Init(ctx); err != nil {
```

One small complication seems the shutdownWg, which we could share or add a .Stop or .Shutdown method to the manager with its own wg

plugin/evm/vm_test.go
```diff
@@ -3173,3 +3176,142 @@ func TestStandaloneDB(t *testing.T) {
 	assert.False(t, isDBEmpty(vm.db))
 	assert.False(t, isDBEmpty(vm.acceptedBlockDB))
 }
+
+func TestValidatorState(t *testing.T) {
```
darioush, line 3180: Can we put this in a separate file eg, `vm_validator_state_test.go` as this file is already very long?

warp/backend.go
```diff
@@ -88,9 +85,8 @@ func NewBackend(
 		warpSigner:                warpSigner,
 		blockClient:               blockClient,
 		signatureCache:            signatureCache,
-		uptimeCalculator:          uptimeCalculator,
-		validatorState:            validatorsState,
-		stateLock:                 stateLock,
+		validatorReader:           validatorReader,
+		validatorLock:             validatorLock,
```
darioush, line 89: Seem a bit odd that we pass this interface but also a lock for it separately?
maybe some kind of locker wrapper for `GetValidator` / `CalculateUptime` is appropriate here? does the lock need to be held atomically for both operations?
This can be out of scope for this PR.

plugin/evm/vm.go
ceyonur: I think in general VM might want to manage processes directly, at least this is the current approach.

plugin/evm/vm_test.go
ceyonur: It was in that file but intentionally moved to vm_test as this is the only test left in that package. But I agree vm_test became too long. Moving it back again.

warp/backend.go
ceyonur: yes they should be locked together. I added another method to interface that locks the chainLock.

plugin/evm/vm.go
darioush: IMO we should steer the VM to being more modular with a bunch of composable/optional components.

However this PR is already a big improvement so I am going to approve it and we can deal with this later.

darioush (dismissed):

plugin/evm/validators/manager.go
```diff
@@ -0,0 +1,161 @@
...
+// DispatchSync starts the sync process
+// DispatchSync holds the chain context lock while performing the sync.
+func (m *manager) DispatchSync(ctx context.Context) {
+	ticker := time.NewTicker(SyncFrequency)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			m.chainCtx.Lock.Lock()
```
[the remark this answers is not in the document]
ceyonur: no, `Sync` is also being called in `SetBootstrapped` which already holds chain lock.

PR #1390, opened 2024-11-18
use nodeID filter and uptimePerc

ceyonur:
## Why this should be merged

`getCurrentValidators` in Platform VM API returns uptime percentage. This PR changes the uptime seconds to uptime percentage in `GetCurrentValidators` API to keep the consistency between APIs. It also adds NodeID filtering.

## How this works

* [`plugin/evm/client.go`](diffhunk://#diff-f2ee9b6048cfd95ed277b400ea49d3ab7eaf30e8d033cdb0b6c7cfca9cae336cL27-R28): Modified the `GetCurrentValidators` method to accept `nodeIDs` as a parameter, allowing for filtered queries. [[1]](diffhunk://#diff-f2ee9b6048cfd95ed277b400ea49d3ab7eaf30e8d033cdb0b6c7cfca9cae336cL27-R28) [[2]](diffhunk://#diff-f2ee9b6048cfd95ed277b400ea49d3ab7eaf30e8d033cdb0b6c7cfca9cae336cL80-R85)
* [`plugin/evm/service.go`](diffhunk://#diff-42f3e9fd1f7cfff775643d1bad698671c0a34b8e3d3e5ab867626fcc0fa6fb8fR7-R21): Introduced `GetCurrentValidatorsRequest` to handle node ID filtering and updated the `GetCurrentValidators` method to use this request. Additionally, changed the `Uptime` field to `UptimePercentage` for consistency with the PlatformVM API. [[1]](diffhunk://#diff-42f3e9fd1f7cfff775643d1bad698671c0a34b8e3d3e5ab867626fcc0fa6fb8fR7-R21) [[2]](diffhunk://#diff-42f3e9fd1f7cfff775643d1bad698671c0a34b8e3d3e5ab867626fcc0fa6fb8fL29-R73) [[3]](diffhunk://#diff-42f3e9fd1f7cfff775643d1bad698671c0a34b8e3d3e5ab867626fcc0fa6fb8fL61-R82)

Support for new method in state interfaces:

* [`plugin/evm/validators/state/interfaces/state.go`](diffhunk://#diff-190cee8f203955ebb957f4da27c648c58fdb9363e70a4e7329303ae5dc58dd39R19-R20): Added `GetValidationID` method to the `StateReader` interface to retrieve validation IDs based on node IDs.
* [`plugin/evm/validators/state/state.go`](diffhunk://#diff-e46380917993bdd51b392434fa7a53079d569379714560baa561445319eacd1eR242-R250): Implemented the `GetValidationID` method in the `state` struct to support the new functionality.

## How this was tested

Locally

## Need to be documented?

Yes

Not since we have not captured the API.

### DIFF
--- a/plugin/evm/service.go
+++ b/plugin/evm/service.go
@@ -4,14 +4,19 @@
 package evm
 
 import (
+	"fmt"
 	"net/http"
-	"time"
 
 	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/utils/set"
 )
 
 type ValidatorsAPI struct {
 	vm *VM
+}
+
+type GetCurrentValidatorsRequest struct {
+	NodeIDs []ids.NodeID `json:"nodeIDs"`
 }
 
 type GetCurrentValidatorsResponse struct {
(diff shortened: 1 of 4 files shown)


plugin/evm/service.go
```diff
@@ -4,61 +4,82 @@
...
-	IsConnected    bool          `json:"isConnected"`
-	Uptime         time.Duration `json:"uptime"`
+	ValidationID     ids.ID     `json:"validationID"`
+	NodeID           ids.NodeID `json:"nodeID"`
+	Weight           uint64     `json:"weight"`
+	StartTimestamp   uint64     `json:"startTimestamp"`
+	IsActive         bool       `json:"isActive"`
+	IsL1Validator    bool       `json:"isL1Validator"`
+	IsConnected      bool       `json:"isConnected"`
+	UptimePercentage float32    `json:"uptimePercentage"`
```
michaelkaplan13, line 34: I think including both the raw uptime seconds and the percentage would be beneficial because validators will need to know the value in seconds for constructing an uptime message that peers will be willing to sign.

ceyonur: It seemed  a bit redundant info as it can be calculable from the other. Only thing that might be changing is the lastUpdated/currentTime that's being used for percentage calculation. I added it nevertheless.

PR #1396, opened 2024-11-22
refactor: block on `StopPrefetchers()` instead of `cleanup`

ARR4N:
If you use a shared `BoundedWorkers` across all tries[^1] then it's sufficient[^2][^3] to block on `StateDB.StopPrefetcher()` or on `triePrefetcher.close()` when you don't have a full DB.

[^1]:  I think this is what we should be doing as it's the point of `BoundedWorkers`. It's also what your original implementation had too.

[^2]:  There was a bug in my ordering of the `defer` statements at the top of `subfetcher.loop()`. I fixed it here; on my libevm branch I also added a test to prove that `StopPrefetcher()` blocks until `WorkerPool.Wait()` returns.

[^3]: I (wrongly) assumed that `BoundedWorkers` would allow extra work while `Wait()`ing.

### DIFF
--- a/core/state/trie_prefetcher.go
+++ b/core/state/trie_prefetcher.go
@@ -79,7 +79,6 @@ func newTriePrefetcher(db Database, root common.Hash, namespace string, opts ...
 // close iterates over all the subfetchers, aborts any that were left spinning
 // and reports the stats to the metrics subsystem.
 func (p *triePrefetcher) close() {
-	p.abortFetchersAndReleaseWorkerPools()
 	for _, fetcher := range p.fetchers {
 		fetcher.abort() // safe to do multiple times
 
@@ -105,6 +104,7 @@ func (p *triePrefetcher) close() {
 			}
 		}
 	}
+	p.releaseWorkerPools()
 	// Clear out all fetchers (will crash on a second call, deliberate)
 	p.fetchers = nil
 }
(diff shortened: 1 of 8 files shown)


core/state/trie_prefetcher.go
```diff
@@ -303,9 +304,11 @@ func (sf *subfetcher) abort() {
 // loop waits for new tasks to be scheduled and keeps loading them until it runs
 // out of tasks or its underlying trie is retrieved for committing.
 func (sf *subfetcher) loop() {
-	defer sf.wait()
 	// No matter how the loop stops, signal anyone waiting that it's terminated
-	defer close(sf.term)
+	defer func() {
+		sf.pool.wait()
+		close(sf.term)
```
darioush, line 310: 👍

core/state/trie_prefetcher.libevm.go
```diff
@@ -68,43 +71,66 @@ func (c *prefetcherConfig) applyTo(sf *subfetcher) {
...
+	// Calling abort() sequentially may result in later fetchers accepting new
+	// work in the interim.
+	var wg sync.WaitGroup
+	for _, f := range p.fetchers {
+		wg.Add(1)
+		go func(f *subfetcher) {
+			f.abort()
+			wg.Done()
+		}(f)
+	}
```
darioush, line 84: I don't think this is needed, the user of triePrefetcher should not call `prefetch` after or concurrently with `abort`.

Also this doesn't guarantee the cancellation.

IMO we can remove this part, rename the fn to releaseWorkerPools() and call it after the sequential abort loop before line 109

ARR4N: Agreed, it was YAGNI thinking on my part.

PR #1410, opened 2024-12-17
bump avalanchego to master (updates x/crypto)

darioush:
While we don't use any impacted code, better to not reference older versions with known issues.
Context https://github.com/ava-labs/avalanchego/pull/3608

Additionally changes `bls.NewSecretKey() -> bls.NewSigner()` to match avalanchego master


go.mod
```diff
@@ -5,7 +5,7 @@ go 1.22.8
 require (
 	github.com/VictoriaMetrics/fastcache v1.12.1
 	github.com/antithesishq/antithesis-sdk-go v0.3.8
-	github.com/ava-labs/avalanchego v1.12.0-config-pebble-sync
+	github.com/ava-labs/avalanchego v1.12.1-0.20241217204724-0ae6f7cc90e4
```
qdm12, line 8: shouldn't we use v1.12.1 (probably a simple answer I'm not aware of! 😉)

qdm12 (dismissed):

go.mod
ceyonur: I think we had not released 1.12.1 at the time this PR was created, but seems we can use it now.

ARR4N (approved):

PR #1413, opened 2024-12-27
chore(scripts): simplify mock.gen.sh to use go generate commands

qdm12:
## Why this should be merged

💁  [**Sibling PR for coreth**](https://github.com/ava-labs/coreth/pull/721) so both repositories would look similar in this aspect

- Add missing license headers
- Localized mockgen commands in the package where they are needed
- Less shell scripting needed
- Remove unneeded `source` commands for constants.sh and versions.sh

## How this works

- New `mocks_generate_test.go` file per package where mocks need to be generated, containing only `//go:generate mockgen` commands. Each command is relative to the current package directory
- Define license header in `license_header`
- Use `-copyright_file=license_header` mockgen flag to add the license header to mocks
- Remove ineffective and unneeded scripting from `scripts/mocks.gen.sh` and remove now unneeded `scripts/mocks.mockgen.txt`
- Restrict to generate using mockgen only commands containing `license_header` such that we don't re-generate eventual geth mocks (it shouldn't hurt to do it as well though)

## How this was tested

`./scripts/mocks.gen.sh`

## Need to be documented?

No

Not really


license_header
ceyonur, line 1: I agree with @darioush about not needing the header.

scripts/mock.gen.sh
```diff
@@ -2,40 +2,4 @@
...
-  echo "go-license not found, installing..."
-  # https://github.com/palantir/go-license
-  go install -v github.com/palantir/go-license@v1.25.0
-fi
-
-# Load the constants
-source "$SUBNET_EVM_PATH"/scripts/constants.sh
-
-# tuples of (source interface import path, comma-separated interface names, output file path)
-input="scripts/mocks.mockgen.txt"
```
ceyonur, line 29: should we remove this file as well?

scripts/mock.gen.sh
ceyonur, line 1: do we have any sibling PR for avalanchego? This is a direct copy from avalanchego. I'd prefer to keep workflows in repos (coreth/subnet-evm/avalanchego) as similar as possible.

qdm12: Done in https://github.com/ava-labs/avalanchego/pull/3628

I'm happy to wait for that PR to be discussed first and then merge the coreth and subnet-evm ones after.

qdm12: https://github.com/ava-labs/avalanchego/pull/3628 merged, and I pushed a few commits here to make it as similar as possible 😉

darioush (approved):

ceyonur (approved):
LGTM

PR #1421, opened 2025-01-09
libevm phase 2: Use libevm statedb from upstream

darioush:
## Why this should be merged
Supersedes https://github.com/ava-labs/subnet-evm/pull/1401

## How this works
Continues libevm development by referencing a version of avalanchego without cyclical imports to avoid initializing libevm incorrectly from wallet & e2e tests.

In this phase of the refactoring, subnet-evm uses the upstream statedb & trie_prefetching mechanism.

Additional avalanchego changes: https://github.com/ava-labs/avalanchego/pull/3645

New changes are commit [28942af](https://github.com/ava-labs/subnet-evm/pull/1421/commits/28942af7d885a615052de2e21d70c9bc5ebb99ce) and after

## How this was tested
CI

## Need to be documented?
No

No

### DIFF
--- a/scripts/build_antithesis_images.sh
+++ b/scripts/build_antithesis_images.sh
@@ -47,7 +47,7 @@ source "${AVALANCHEGO_CLONE_PATH}"/scripts/lib_build_antithesis_images.sh
 build_antithesis_builder_image "${GO_VERSION}" "antithesis-subnet-evm-builder:${IMAGE_TAG}" "${AVALANCHEGO_CLONE_PATH}" "${SUBNET_EVM_PATH}"
 
 # Ensure avalanchego and subnet-evm binaries are available to create an initial db state that includes subnets.
-"${AVALANCHEGO_CLONE_PATH}"/scripts/build.sh
+pushd "${AVALANCHEGO_CLONE_PATH}" && ./scripts/build.sh && popd
 "${SUBNET_EVM_PATH}"/scripts/build.sh
 
 echo "Generating compose configuration"


scripts/build_antithesis_images.sh
```diff
@@ -47,7 +47,7 @@ source "${AVALANCHEGO_CLONE_PATH}"/scripts/lib_build_antithesis_images.sh
 build_antithesis_builder_image "${GO_VERSION}" "antithesis-subnet-evm-builder:${IMAGE_TAG}" "${AVALANCHEGO_CLONE_PATH}" "${SUBNET_EVM_PATH}"
 
 # Ensure avalanchego and subnet-evm binaries are available to create an initial db state that includes subnets.
-"${AVALANCHEGO_CLONE_PATH}"/scripts/build.sh
+pushd "${AVALANCHEGO_CLONE_PATH}" && ./scripts/build.sh && popd
```
qdm12, line 50: What does this do? Why is this change needed (asking out of curiosity really)

qdm12 (approved):

scripts/build_antithesis_images.sh
darioush: Yeah... this was another 20 minute adventure.
See this commit: https://github.com/ava-labs/subnet-evm/pull/1401/commits/33bd19399ea863babec7b7e4b6e4a57e599afb20

For some reason, (perhaps since we dropped the circular import and import less stuff), go.sum for subnet-evm doesn't contain "github.com/jackpal/gateway" anymore.
Therefore if we trigger the avalanchego build script from subnet-evm's repo root (it gets checked out to avalanchego/) it doesn't work (I think these scripts assume they are run from the repo root).

ethclient/ethclient.go
```diff
@@ -479,14 +477,6 @@ func (ec *client) BalanceAt(ctx context.Context, account common.Address, blockNu
 	return (*big.Int)(&result), err
 }
 
-// AssetBalanceAt returns the [assetID] balance of the given account
-// The block number can be nil, in which case the balance is taken from the latest known block.
-func (ec *client) AssetBalanceAt(ctx context.Context, account common.Address, assetID ids.ID, blockNumber *big.Int) (*big.Int, error) {
```
ceyonur, line 484: ughhh 😞

scripts/build_antithesis_images.sh
qdm12: What on earth! I would prefer funny shell scripting than having a blank import for `github.com/jackpal/gateway` really, if that's possible. Weird adventure yet again!

ceyonur (approved):
lgtm

PR #1422, opened 2025-01-13
chore(metrics): use geth metrics package and delete local metrics

qdm12:
## Why this should be merged

Contribution to the libevm effort.
Note there are the same changes for both coreth and subnet-evm, but they should not be part of libevm since these are extra features specific to subnet-evm/coreth, even if they're the same for both.

## How this works

Comparing the following:

- subnet-evm `master` vs coreth `master`: subnet-evm is missing the extra file https://github.com/ava-labs/coreth/blob/b6b4dfbc4bfc7f17e4cfe5f3c0fb44944176c884/metrics/init_test.go (which is present in geth as well)
- subnet-evm `master` vs geth `v1.13.14`: subnet-evm has many deletions, very few minor unneeded changes and the new non-conflicting file [`metrics/prometheus/prometheus.go`](https://github.com/ava-labs/subnet-evm/blob/387bbc1cab40b40cd06d87d38db71b425df2d01e/metrics/prometheus/prometheus.go)

Therefore:

- file kept and refactored `metrics/prometheus/prometheus.go` with the `Gatherer` implementation we use
- file kept and refactored `metrics/prometheus/prometheus_test.go`
- new file `metrics/prometheus/interfaces.go` added for refactoring
- geth global variable `metrics.Enabled` is set to true in `plugin/evm.VM.initializeMetrics`

Note for the two files refactored (with fixes as well), this was done in some intermediary step in https://github.com/ava-labs/libevm/pull/103, so I decided to bring this over here so it doesn't get trashed.

## How this was tested

CI passing

## Need to be documented?

No

Not really, since it *should* not change anything 🙏


plugin/evm/vm.go
```diff
@@ -32,8 +33,7 @@ import (
 	"github.com/ava-labs/subnet-evm/core/types"
 	"github.com/ava-labs/subnet-evm/eth"
 	"github.com/ava-labs/subnet-evm/eth/ethconfig"
-	"github.com/ava-labs/subnet-evm/metrics"
-	subnetEVMPrometheus "github.com/ava-labs/subnet-evm/metrics/prometheus"
+	subnetevmprometheus "github.com/ava-labs/subnet-evm/metrics/prometheus"
```
qdm12, line 36: nit I took the liberty to have this all lowercase, I believe this is Go convention for package name aliases (correct me if I'm wrong!)

plugin/evm/vm.go
```diff
@@ -555,7 +555,7 @@ func (vm *VM) initializeMetrics() error {
 		return nil
 	}
 
-	gatherer := subnetEVMPrometheus.Gatherer(metrics.DefaultRegistry)
+	gatherer := subnetevmprometheus.NewGatherer(metrics.DefaultRegistry)
```
qdm12, line 558: nit I took the liberty (again) to rename `Gatherer` to `NewGatherer` in order to export `gatherer` as `Gatherer` and return a `*Gatherer` from `NewGatherer` to satisfy the "return concrete types" concept

warp/handlers/signature_request_test.go
```diff
@@ -21,6 +22,14 @@ import (
 )
 
 func TestMessageSignatureHandler(t *testing.T) {
+	metricsEnabled := metrics.Enabled
+	if !metricsEnabled {
+		metrics.Enabled = true
+		t.Cleanup(func() {
+			metrics.Enabled = false
+		})
+	}
```
qdm12, line 31: This might not be needed given the `init()` block setting the metrics to Enabled, but I preferred to keep it here if we one day revert the default Metrics Enabled to false (as it is in geth)

metrics/prometheus/prometheus.go
```diff
@@ -1,196 +1,193 @@
...
+				Summary: &dto.Summary{
+					SampleCount: ptrTo(uint64(snapshot.Count())), //nolint:gosec
+					// TODO: do we need to specify SampleSum here? and if so
+					// what should that be?
+					Quantile: dtoQuantiles,
+				},
+			}},
+		}, nil
+	default:
+		return nil, fmt.Errorf("metric type is not supported: %T", metric)
```
qdm12, line 195: I took the liberty to change this to return an error (instead of silently ignoring it) on an unsupported type to prevent in the future to define new types and forget to add the implementation in the switch here.

metrics/prometheus/prometheus.go
```diff
@@ -1,196 +1,193 @@
...
 
-			ps := snapshot.Percentiles(pvShortPercent)
-			qs := make([]*dto.Quantile, len(pv))
-			for i := range pvShort {
-				v := pv[i]
-				s := ps[i]
-				qs[i] = &dto.Quantile{
-					Quantile: &v,
-					Value:    &s,
-				}
```
qdm12, line 173: Note this would trigger a slice index error, it's fixed in the production code, and the test is refactored to cover those lines which were not previously covered.

metrics/prometheus/prometheus.go
```diff
@@ -1,196 +1,197 @@
-// (c) 2021, Ava Labs, Inc. All rights reserved.
+// (c) 2025 Ava Labs, Inc. All rights reserved.
```
qdm12, line 1:
```suggestion
// (c) 2021-2025 Ava Labs, Inc. All rights reserved.
```

metrics/prometheus/prometheus_test.go
```diff
@@ -1,4 +1,4 @@
-// (c) 2021, Ava Labs, Inc. All rights reserved.
+// (c) 2025 Ava Labs, Inc. All rights reserved.
```
qdm12, line 1:
```suggestion
// (c) 2021-2025 Ava Labs, Inc. All rights reserved.
```

metrics/prometheus/prometheus.go
```diff
@@ -1,196 +1,197 @@
...
 )
 
-var (
-	pv             = []float64{.5, .75, .95, .99, .999, .9999}
-	pvShortPercent = []float64{50, 95, 99}
-	pvShort        = []float64{.50, .95, .99}
-)
+func init() {
+	metrics.Enabled = true
+}
```
qdm12, line 21: Needed to keep the same behavior, since we modified geth to have `Enabled = true` at global scope instead of `false`.

go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/VictoriaMetrics/fastcache v1.12.1
 	github.com/antithesishq/antithesis-sdk-go v0.3.8
 	github.com/ava-labs/avalanchego v1.12.1
+	github.com/ava-labs/libevm v1.13.14-0.1.0-rc.1
```
darioush, line 9: We should keep the dependency as go-ethereum until we are able to switch to libevm (importing both go-ethereum and libevm sometimes creates linker issues)

warp/handlers/signature_request_test.go
darioush: Could we put this code in a different location such as `plugin/evm/metricstest`
then have `metricstest.WithMetrics(t)` for example?

I'm assuming this code was added only to tests that were failing when metrics was disabled?

metrics/prometheus/prometheus.go
darioush: Could we add this to a different package's init, such as `plugin/evm` or make a new package like `plugin/evm/metrics`?

More ideal would be to set it in VM initialize but I'm not sure if that breaks too many tests.

qdm12: I suppose we could, tests affected are already changed to set enabled to true for the test duration. Also I hate global state, but oh well that's geth, no way around it (yet... 😈 )  🤷

warp/handlers/signature_request_test.go
qdm12: Done in `plugin/evm/testutils/metrics.go` with a function `WithMetrics(t)`. I went a bit crazy on the use-testutils-only-for-tests-outside-plugin/evm (inspired by your import test + arran's check for test files), since I think it would be good to only have a `testutils` to reduce changes in geth code, not for our own code (prefer copy pasting and defining local code for readability + reduce package dependencies)

metrics/prometheus/prometheus.go
qdm12: Done in initializeMetrics 😉

darioush (dismissed):

qdm12:
TIL rebasing with github does not sign commits 😢

qdm12:
> A random PR doesn't show any metrics on the subnets dashboard for the same job:

Ah, let me check if I can revert it to no-metric, and open another PR (based off this one I suppose) to enable metrics.
Since this PR is not meant to change anything and should be pure maintenance / code movement.

qdm12:
Working as expected, this is good to go (same as https://github.com/ava-labs/coreth/pull/745), just need reviews 😉

scripts/geth-allowed-packages.txt
```diff
@@ -17,4 +17,6 @@
 "github.com/ethereum/go-ethereum/ethdb/pebble"
 "github.com/ethereum/go-ethereum/event"
 "github.com/ethereum/go-ethereum/log"
+"github.com/ethereum/go-ethereum/metrics"
+"github.com/ethereum/go-ethereum/metrics.(*meterArbiter).tick"
```
darioush, line 21: this seems a tad bit hacky, perhaps we could exclude files with `main_test.go`?
seems slightly less hacky to build the string as `"github.com/ethereum/go-ethereum/metrics" + ".(*meterArbiter).tick"` in the go file where there can be a comment explaining what's going on

plugin/evm/testutils/testutils_test.go
```diff
@@ -0,0 +1,81 @@
...
+	for _, packagePath := range evmPackagePaths {
+		packageName := "github.com/ava-labs/subnet-evm/plugin/evm/" + strings.TrimPrefix(packagePath, "../")
+		imports, err := getPackageImports(packagePath)
+		require.NoError(t, err)
+		_, ok := imports["github.com/ava-labs/subnet-evm/plugin/evm/testutils"]
+		assert.Falsef(t, ok, "package %s imports testutils: testutils should only be used outside plugin/evm.", packageName)
+	}
+}
+
+func getPackageImports(packagePath string) (imports map[string]struct{}, err error) {
```
darioush, line 39: There's an existing test in coreth that does something similar in case you wanted to adopt that style,
this is not exactly the same as here you walk plugin/evm to get the subdirs.

https://github.com/ava-labs/coreth/blob/master/plugin/evm/imports_test.go#L45

darioush (approved):
some optional suggestions

plugin/evm/testutils/metrics.go
```diff
@@ -0,0 +1,21 @@
...
+	"testing"
+
+	"github.com/ethereum/go-ethereum/metrics"
+)
+
+// WithMetrics enables go-ethereum metrics globally for the test.
+// If the [metrics.Enabled] is already true, nothing is done.
+// Otherwise, it is set to true and is reverted to false when the test finishes.
+func WithMetrics(t *testing.T) {
+	panicIfCallsFromNonTest()
```
ceyonur, line 13: I wonder if we should rather do that with a CI script/linter. Not sure if this is maintainable for all functions that we want to avoid using in prod code.

plugin/evm/testutils/testutils_test.go
```diff
@@ -0,0 +1,81 @@
...
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestMustNotImport fails if a package in plugin/evm imports the testutils package.
+func TestMustNotImport(t *testing.T) {
```
ceyonur, line 18: I think either a UT like that or a CI script should be sufficient, rather than having a function panicking in prod code.

plugin/evm/vm.go
```diff
@@ -550,13 +550,9 @@ func (vm *VM) Initialize(
 }
 
 func (vm *VM) initializeMetrics() error {
+	metrics.Enabled = true
```
ceyonur, line 553: can we add few comments why we explicitly enable it here?

plugin/evm/vm.go
```diff
@@ -550,13 +550,9 @@ func (vm *VM) Initialize(
...
 	vm.sdkMetrics = prometheus.NewRegistry()
-	// If metrics are enabled, register the default metrics registry
-	if !metrics.Enabled {
-		return nil
-	}
```
ceyonur, line 557: isn't this something we want to keep?

scripts/geth-allowed-packages.txt
ceyonur: what's the requirement here and what does this do? I thought `github.com/ethereum/go-ethereum/metrics` should be enough to allow the whole package, no?

darioush: the regex picks up "github.com/ethereum/go-ethereum/metrics.(*meterArbiter).tick" as a false positive. 
always regex's providing the entertainment factor in life.

qdm12: Done in e389c7f1adb615c132e21dccbadee5e37acb1455

plugin/evm/testutils/testutils_test.go
qdm12: Ha I know, I shamelessly copied and modified your code 😄 Thanks 💯

plugin/evm/vm.go
qdm12: Added

```go
	// [metrics.Enabled] is a global variable imported from go-ethereum/metrics
	// and must be set to true to enable metrics collection.
```

Does that suffice? 🤔

plugin/evm/vm.go
qdm12: From the previous logic, we always have metrics enabled, see

https://github.com/ava-labs/subnet-evm/blob/b049381bf3f50eb8687ca131c5237abdc0d2aa29/metrics/metrics.go#L13

If at some point we want a configuration option to disable metrics, we can refactor this code, but for now there is not even an option to disable/enable metrics.

plugin/evm/testutils/testutils.go
```diff
@@ -0,0 +1,5 @@
+// Package testutils contains test utilities ONLY to be used outside plugin/evm.
+// The aim is to reduce changes in geth tests by using the utilities defined here.
+// This package MUST NOT be imported by non-test packages.
```
qdm12, line 3:
```suggestion
// This package MUST NOT be imported by non-test files.
```

plugin/evm/testutils/testutils_test.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestMustNotImport fails if:
+// - a package in plugin/evm imports the testutils package
+// - a non-test file imports the testutils package
+func TestMustNotImport(t *testing.T) {
+	prodOnly := false
+	checkNotImporting(t, "../", prodOnly, "testutils should only be used outside plugin/evm")
```
ceyonur, line 23: why do we have such limitation? Does that mean we can't use this pkg in plugin/evm test files?

qdm12: See [this comment](https://github.com/ava-labs/subnet-evm/pull/1422#discussion_r1916368549). TLDR:

- this testutils package is only to limit diffs in geth code
- (subjective) we should prefer to copy paste test helping utilities, to reduce package dependencies and possible cycle import issues

metrics/prometheus/prometheus.go
```diff
@@ -1,196 +1,202 @@
...
+func metricFamily(registry Registry, name string) (mf *dto.MetricFamily, err error) {
+	metric := registry.Get(name)
+	name = strings.ReplaceAll(name, "/", "_")
+
+	switch m := metric.(type) {
+	case metrics.NilCounter, metrics.NilCounterFloat64, metrics.NilEWMA,
+		metrics.NilGauge, metrics.NilGaugeFloat64, metrics.NilGaugeInfo,
+		metrics.NilHealthcheck, metrics.NilHistogram, metrics.NilMeter,
+		metrics.NilResettingTimer, metrics.NilSample, metrics.NilTimer:
+		return nil, fmt.Errorf("%w: %q metric is nil", errMetricSkip, name)
```
ceyonur, line 74: if we're going to skip this, can't we just return nil err?

metrics/prometheus/prometheus.go
```diff
@@ -1,196 +1,202 @@
...
+			Type: dto.MetricType_SUMMARY.Enum(),
+			Metric: []*dto.Metric{{
+				Summary: &dto.Summary{
+					SampleCount: ptrTo(uint64(snapshot.Count())), //nolint:gosec
+					Quantile:    dtoQuantiles,
+				},
+			}},
+		}, nil
+	default:
+		return nil, fmt.Errorf("metric %q: type is not supported: %T", name, metric)
```
ceyonur, line 200: this won't cause any unknown geth metric type to break this, right? (e.g a new upstream type in the future)

plugin/evm/testutils/testutils_test.go
ceyonur: I don't think we should limit this pkg like that. if we nee such a pkg it should be in internal if possible and under a different pkg than plugin/evm. I think it should be sufficient to check this is not imported in any prod code, no?

I'm fine with copy pasting `WithMetrics` if we feel like it's best not to import `metrics` here.

ceyonur: thinking again, having a `testutils` package actually indicates this should be not called from prod code. So not sure if there is any point to strict that down too much.

metrics/prometheus/prometheus.go
qdm12: no we need to signal somehow to the caller to not collect that metric and to skip it

metrics/prometheus/prometheus.go
qdm12: Shouldn't it break it? This would show in the CI right, and we should add the type to our switch case and either handle it or explicitly ignore it, I think.

plugin/evm/testutils/testutils_test.go
qdm12: Moved to internal/testutils, without import restrictions

ceyonur (approved):
changes lgtm, just a small concern/question with potential unknown metrics in upstream.

metrics/prometheus/prometheus.go
qdm12: The only place that gatherer is registered is:

https://github.com/ava-labs/subnet-evm/blob/4f2ea6381cb6ae0dd911cf46f5d295b3291ee7a4/plugin/evm/vm.go#L558

Which registers it to [avalanchego's multi gatherer](https://github.com/ava-labs/avalanchego/blob/9c9b0042c76f2075f99faa478489bab9fed7e011/snow/context.go#L54) and all errors are checked/none-are-discarded.

Maybe it's not covered by tests, but it should definitely show up in e2e tests (and, if not, we would have an error handling problem in avalanchego).

metrics/prometheus/prometheus_test.go
```diff
@@ -8,80 +8,121 @@ import (
...
+	"github.com/ethereum/go-ethereum/metrics"
 )
 
-func TestGatherer(t *testing.T) {
+func TestGatherer_Gather(t *testing.T) {
+	enabledOriginal := metrics.Enabled
+	t.Cleanup(func() {
+		metrics.Enabled = enabledOriginal
+	})
+	metrics.Enabled = true
```
ARR4N, line 21: Isn't this what `testutils.WithMetrics()` is for?

metrics/prometheus/prometheus_test.go
```diff
@@ -8,80 +8,121 @@ import (
...
+
 	registry := metrics.NewRegistry()
+	register := func(t *testing.T, name string, collector any) {
+		t.Helper()
+		err := registry.Register(name, collector)
+		require.NoError(t, err)
```
ARR4N, line 27:
```suggestion
		require.NoErrorf(t, err, "registering collector %q", name)
```

metrics/prometheus/prometheus_test.go
```diff
@@ -8,80 +8,121 @@ import (
...
-
-	_, err = g.Gather()
-	assert.NoError(t, err)
+	register(t, "test/empty_resetting_timer", emptyResettingTimer)
+
+	emptyResettingTimer.Update(time.Second) // no effect because of snapshot below
+	register(t, "test/empty_resetting_timer_snapshot", emptyResettingTimer.Snapshot())
+
+	// Nil metrics
+	metrics.Enabled = false
```
ARR4N, line 76: Why does this need to be done with metrics disabled?

ARR4N, line 76: (optional, readability) I'd abstract this into a separate function for a couple of reasons:

1. It's really just boilerplate noise; necessary but not something that the reader needs to grok and remember. The important thing to internalise is "we're registering some nil metrics", which a function call will signal.
2. The re-enabling of metrics would be clearer with a `defer`.

metrics/prometheus/prometheus.go
```diff
@@ -1,196 +1,202 @@
...
 	"sort"
 	"strings"
 
-	"github.com/ava-labs/subnet-evm/metrics"
-
 	"github.com/prometheus/client_golang/prometheus"
 
+	"github.com/ethereum/go-ethereum/metrics"
+
 	dto "github.com/prometheus/client_model/go"
```
ARR4N, line 16: Any idea why `dto` is the standard way to alias this package?

metrics/prometheus/prometheus_test.go
```diff
@@ -8,80 +8,121 @@ import (
...
+	metrics.Enabled = true
+
+	gatherer := NewGatherer(registry)
+
+	families, err := gatherer.Gather()
+	require.NoError(t, err)
+	familyStrings := make([]string, len(families))
+	for i := range families {
+		familyStrings[i] = families[i].String()
+	}
```
ARR4N, line 111: (fyi) This is where the `cmp` package shines. In this case you'd probably want a [Comparer](https://pkg.go.dev/github.com/google/go-cmp/cmp#Comparer). Although a `Transformer` would do the same thing, it's a means (converting to a string) to an end (checking equality).

metrics/prometheus/prometheus_test.go
```diff
@@ -8,80 +8,121 @@ import (
...
+		`name:"test_histogram" type:SUMMARY metric:<summary:<sample_count:0 sample_sum:0 quantile:<quantile:0.5 value:0 > quantile:<quantile:0.75 value:0 > quantile:<quantile:0.95 value:0 > quantile:<quantile:0.99 value:0 > quantile:<quantile:0.999 value:0 > quantile:<quantile:0.9999 value:0 > > > `,
+		`name:"test_meter" type:GAUGE metric:<gauge:<value:9.999999e+06 > > `,
+		`name:"test_resetting_timer" type:SUMMARY metric:<summary:<sample_count:1 quantile:<quantile:50 value:1e+09 > quantile:<quantile:95 value:1e+09 > quantile:<quantile:99 value:1e+09 > > > `,
+		`name:"test_timer" type:SUMMARY metric:<summary:<sample_count:6 sample_sum:2.3e+08 quantile:<quantile:0.5 value:2.25e+07 > quantile:<quantile:0.75 value:4.8e+07 > quantile:<quantile:0.95 value:1.2e+08 > quantile:<quantile:0.99 value:1.2e+08 > quantile:<quantile:0.999 value:1.2e+08 > quantile:<quantile:0.9999 value:1.2e+08 > > > `,
+	}
+	assert.Equal(t, want, familyStrings)
+
+	register(t, "unsupported", metrics.NewGaugeInfo())
+	families, err = gatherer.Gather()
+	assert.EqualError(t, err, "metric \"unsupported\": type is not supported: *metrics.StandardGaugeInfo")
```
ARR4N, line 126: Use a common error and `errors.Is()` as this is a change-detector test for the error message.

metrics/prometheus/prometheus.go
qdm12: It's for data transfer object... Package is `io_prometheus_client` in a `go` directory, so not too sure how it should be named 😄

internal/testutils/metrics.go
```diff
@@ -0,0 +1,20 @@
...
+import (
+	"testing"
+
+	"github.com/ethereum/go-ethereum/metrics"
+)
+
+// WithMetrics enables go-ethereum metrics globally for the test.
+// If the [metrics.Enabled] is already true, nothing is done.
+// Otherwise, it is set to true and is reverted to false when the test finishes.
+func WithMetrics(t *testing.T) {
```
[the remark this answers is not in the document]
qdm12: > unlocks in the cleanup and the early return.

In the test cleanup in all cases; in the early return it should not be unlocked I think, just in case ™️

metrics/prometheus/interfaces.go
```diff
@@ -0,0 +1,12 @@
+// (c) 2025 Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+package prometheus
+
```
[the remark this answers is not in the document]
qdm12: It doesn't because

> Registry is a narrower interface of [prometheus.Registry]

It's the other way around, but that can't be done afaik. This however showed me the comment used `[prometheus.Gatherer]` which isn't resolvable without the prometheus import. God the world needs a linter/gopls to show non resolvable godoc links!

metrics/prometheus/prometheus.go
```diff
@@ -1,196 +1,202 @@
...
+)
+
+func ptrTo[T any](x T) *T { return &x }
+
+func metricFamily(registry Registry, name string) (mf *dto.MetricFamily, err error) {
+	metric := registry.Get(name)
+	name = strings.ReplaceAll(name, "/", "_")
+
+	switch m := metric.(type) {
+	case metrics.NilCounter, metrics.NilCounterFloat64, metrics.NilEWMA,
```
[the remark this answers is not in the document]
qdm12: Thinking about it, it's kind of strange we can do that given yes `m` is then of `any` type. In this case we don't use `m` so I suppose it's fine 🤷

metrics/prometheus/prometheus.go
```diff
@@ -1,196 +1,202 @@
...
+		metrics.NilGauge, metrics.NilGaugeFloat64, metrics.NilGaugeInfo,
+		metrics.NilHealthcheck, metrics.NilHistogram, metrics.NilMeter,
+		metrics.NilResettingTimer, metrics.NilSample, metrics.NilTimer:
+		return nil, fmt.Errorf("%w: %q metric is nil", errMetricSkip, name)
+	case metrics.Counter:
```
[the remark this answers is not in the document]
qdm12: The only block changed is for metrics.ResettingTimer which was just wrong, so now it's fixed

metrics/prometheus/prometheus_test.go
qdm12: Indeed; however my (maybe strange) original idea for internal/testutils was to be used only in code coming from geth, to reduce diffs. Old package comment:

```go
// Package testutils contains test utilities ONLY to be used outside plugin/evm.
// The aim is to reduce changes in geth tests by using the utilities defined here.
// This package MUST NOT be imported by non-test files.
```

I'm not a fan of generic packages like testutils which can later grow out of proportion and create dependency cycle hell, so I originally wanted to keep it restricted to geth-code.
I even went [a bit too far enforcing that testutils package would not be imported i.e. by plugin/evm](https://github.com/ava-labs/subnet-evm/pull/1422/commits/c3befa454f7bc88cb0713628bd76995680c59536).
I'm essentially in the fan club of "prefer a little copying over adding a dependency", even within the same module.

metrics/prometheus/prometheus_test.go
qdm12: 1. Comparer would need both want and got to be of the same type right? Which is not the case here, and, if it would be, we could just use `assert.Equal`. I have my `want` as strings since it's easier to read/more compact than having it as `*dto.MetricFamily` structs.
2. I tried using `cmp.Transformer`, using:

    ```go
    transformer := cmp.Transformer("family_string", func(f *dto.MetricFamily) string {
        return f.String()
    })
    assert.Empty(t, cmp.Diff(want, families, transformer))
    ```

    but it doesn't seem to convert the `families` to a string slice? 🤔

metrics/prometheus/prometheus_test.go
qdm12: Because nil metrics are only built when Enabled is false.

For example `NilCounter` is returned by `NewCounter()` only if `Enabled` is `false`.

qdm12: Done in `registerNilMetrics`

metrics/prometheus/interfaces.go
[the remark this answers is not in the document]
ARR4N: In that case the comment is wrong; either it's not a narrower set of methods or that should be `[metrics.Registry]`, not `[prometheus.Registry]`. This compiles for me and that's the concrete type being used in the test:

```go
package prometheus

import "github.com/ethereum/go-ethereum/metrics"

var _ Registry = metrics.Registry(nil)
```

internal/testutils/metrics.go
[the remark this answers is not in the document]
ARR4N: Yeah, I think you're right because it could be `true` because another holder of the lock is about to flip it back to `false`. Terrible package API.

metrics/prometheus/prometheus_test.go
ARR4N: I learnt that later in the day. Not a fan of the API.

metrics/prometheus/prometheus.go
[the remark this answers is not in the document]
ARR4N: I suspect that it actually just keeps its original type, which just so happens to be `any` in this case.

metrics/prometheus/prometheus_test.go
ARR4N: You may need to define the transformer as `func([]*dto.MetricFamily) []string {`. If you want to go ahead with using it then DM me and I'll help figure it out otherwise let's just leave it.

ARR4N (approved):
Pending resolution of the comment re the comment

metrics/prometheus/interfaces.go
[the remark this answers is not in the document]
qdm12: Yep thanks! Comment was completely wrong, removed it and added your suggestion!

qdm12:
We might need to change the base of this to libevm given the global Enabled boolean is set to false in geth, and that causes metrics to initialize as Nil* (aka no-op) metrics, without a proper workaround.

qdm12:
Force push rebasing this on current master branch (no changes or merge conflicts resolved otherwise)

qdm12:
Force push rebasing the branch on `qdm12/libevm/master-merge-1` (libevm up to date with master), history is kept more or less, but everything will have to be re-reviewed.

---

this PR got merged and I have no idea how. I may have pressed keyboard shortcuts merging it, but I really have no idea, I re-opened another draft PR #1508 + hard reset my branch `qdm12/libevm/master-merge-1` to discard that out-of-nowhere-merge.

PR #1424, opened 2025-01-16
Add Docker Usage Instructions to README.md

containerman17:
## Why this should be merged
The README.md currently lacks instructions on running subnet-evm using Docker. Adding a Docker section will help users set up and configure subnet-evm easily in containerized environments.

## How this works
Added a new Run in Docker section to the README.md with brief instructions on configuration, data persistence, updating, networking, and example setups for both a Fuji subnet validator and RPC server.

## How this was tested
I tested the examples manually.

## Need to be documented?
No additional documentation is needed; all information is included in the updated README.md.

No update to RELEASES.md is necessary since this is a documentation addition.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,3 +98,80 @@ There are two options when using the Avalanche-CLI:
 
 1. Use an official Subnet-EVM release: https://docs.avax.network/subnets/build-first-subnet
 2. Build and deploy a locally built (and optionally modified) version of Subnet-EVM: https://docs.avax.network/subnets/create-custom-subnet
+
+## Run in Docker
+
+### Configuration
+
+You can configure the `subnet-evm` Docker container using both environment variables and standard AvalancheGo config files.
+
+- **Environment Variables**: Use uppercase variables prefixed with `AVAGO_`. For example, `AVAGO_NETWORK_ID` corresponds to the `--network-id` flag in AvalancheGo.
+- **Config Files**: Configure as you would with the regular AvalancheGo binary using config files. Ensure `HOME` is set to `/home/avalanche` and mount the config directory with `-v ~/.avalanchego:/home/avalanche/.avalanchego`.
+
+### Data Persistence
+
+To persist data across container restarts, you need to mount the `.avalanchego` directory.
+
+- **Create Directory**: Run `mkdir -p ~/.avalanchego/staking` beforehand to create the necessary directories and avoid file system access issues.
+- **Set Permissions**: Run the container with `--user $(id -u):$(id -g)` to ensure proper access rights.
+- **Set Home and Mount**: Set `HOME` to `/home/avalanche` and mount the data directory with `-v ~/.avalanchego:/home/avalanche/.avalanchego`.
+
+### Updating
+
+Run `docker stop avago; docker rm avago;` then start a new container with the latest version tag in your `docker run` command.
+
+### Networking
+
+Using `--network host` is recommended to avoid any issues.
+
+If you know what you are doing, you will need port `AVAGO_STAKING_PORT` (default `9651`) open for the validator to connect to the subnet. For the RPC server, open `AVAGO_HTTP_PORT` (default `9650`). Do not attempt to remap `AVAGO_STAKING_PORT` using the Docker `-p` flag (e.g., `-p 9651:1234`); it will not work. Instead, set `AVAGO_STAKING_PORT=1234` and then use `-p 1234:1234`.
+
+### Example Configs
+
+#### Fuji Subnet Validator
+
+```bash
+mkdir -p ~/.avalanchego/staking; docker run -it -d \
+  --name avago \
+  --network host \
+  -v ~/.avalanchego:/home/avalanche/.avalanchego \
+  -e AVAGO_NETWORK_ID=fuji \
+  -e AVAGO_PARTIAL_SYNC_PRIMARY_NETWORK=true \
+  -e AVAGO_TRACK_SUBNETS=REPLACE_THIS_WITH_YOUR_SUBNET_ID \
+  -e AVAGO_PUBLIC_IP_RESOLUTION_SERVICE=ifconfigme \
+  -e AVAGO_PLUGIN_DIR=/avalanchego/build/plugins/ \
+  -e HOME=/home/avalanche \
+  --user $(id -u):$(id -g) \
+  avaplatform/subnet-evm:v0.7.1
+```
+
+- `AVAGO_PARTIAL_SYNC_PRIMARY_NETWORK`: Ensures you don't sync the X and C-Chains.
+- `AVAGO_PLUGIN_DIR`: Sets the directory for the `subnet-evm` plugin.
+- `AVAGO_TRACK_SUBNETS`: Sets the subnet ID to track. It will track all chains in the subnet.
+- `AVAGO_NETWORK_ID=fuji`: Sets the network ID to Fuji. Remove to sync Mainnet.
+- `AVAGO_PUBLIC_IP_RESOLUTION_SERVICE=ifconfigme`: Required for AWS EC2 instances to be accessed from outside AWS.
+
+#### Fuji Subnet RPC
+
+```bash
+const dockerCommand = (subnetID: string, chainID: string) => `mkdir -p ~/.avalanchego_rpc/staking; docker run -it -d \\
+  --name rpc \\
+  --network host \\
+  -v ~/.avalanchego_rpc/:/home/avalanche/.avalanchego \\
+  -e AVAGO_NETWORK_ID=fuji \\
+  -e AVAGO_PARTIAL_SYNC_PRIMARY_NETWORK=true \\
+  -e AVAGO_TRACK_SUBNETS=REPLACE_THIS_WITH_YOUR_SUBNET_ID \\
+  -e AVAGO_HTTP_PORT=8080 \\
+  -e AVAGO_STAKING_PORT=9653 \\
+  -e AVAGO_HTTP_ALLOWED_HOSTS="*" \\
+  -e AVAGO_HTTP_HOST=0.0.0.0 \\
+  -e AVAGO_PUBLIC_IP_RESOLUTION_SERVICE=ifconfigme \\
+  -e AVAGO_PLUGIN_DIR=/avalanchego/build/plugins/ \\
+  -e HOME=/home/avalanche \\
+  --user $(id -u):$(id -g) \\
+  avaplatform/subnet-evm:v0.7.1`;
+```
+
+- `AVAGO_STAKING_PORT` is set to `9653` in case you want to run this on the same machine as the validator.
+- `AVAGO_HTTP_PORT` is set to `8080` instead of `9650` to avoid conflicts with the validator.
+- `AVAGO_HTTP_ALLOWED_HOSTS` and `AVAGO_HTTP_HOST` are required to allow the RPC server to be accessed from outside. You'll need to secure it with HTTPS; Caddy is recommended.


martineckardt:
LGTM

README.md
```diff
@@ -98,3 +98,84 @@ There are two options when using the Avalanche-CLI:
...
 1. Use an official Subnet-EVM release: https://docs.avax.network/subnets/build-first-subnet
 2. Build and deploy a locally built (and optionally modified) version of Subnet-EVM: https://docs.avax.network/subnets/create-custom-subnet
+
+## Run in Docker
+
+The `subnet-evm` Docker image comes with AvalancheGo pre-installed, making it easy to run a node. You can find the latest image tags on [Docker Hub](https://hub.docker.com/r/avaplatform/subnet-evm/tags).
+
+### Configuration
+
+You can configure the `subnet-evm` Docker container using both environment variables and standard AvalancheGo config files.
```
qdm12, line 108: Technically you could still pass flags to the docker run command at the end to pass them to the binary I suppose?

```suggestion
You can configure the `subnet-evm` Docker container using environment variables, flags or standard AvalancheGo config files.
```

README.md
```diff
@@ -98,3 +98,84 @@ There are two options when using the Avalanche-CLI:
...
+
+- **Environment Variables**: Use uppercase variables prefixed with `AVAGO_`. For example, `AVAGO_NETWORK_ID` corresponds to the `--network-id` flag in AvalancheGo.
+- **Config Files**: Configure as you would with the regular AvalancheGo binary using config files. Ensure `HOME` is set to `/home/avalanche` and mount the config directory with `-v ~/.avalanchego:/home/avalanche/.avalanchego`.
+
+### Data Persistence
+
+To persist data across container restarts, you need to mount the `.avalanchego` directory.
```
qdm12, line 115: to avoid any ambiguity

```suggestion
To persist data across container restarts, you need to mount the container `/home/avalanche/.avalanchego` directory.
```

README.md
```diff
@@ -98,3 +98,84 @@ There are two options when using the Avalanche-CLI:
...
+
+- **Create Directory**: Run `mkdir -p ~/.avalanchego/staking` beforehand to create the necessary directories and avoid file system access issues.
+- **Set Permissions**: Run the container with `--user $(id -u):$(id -g)` to ensure proper access rights.
```
qdm12, line 118: It looks like the container runs as root as default (it should not ideally but that's another topic), so there is no need to change permissions on the directory or run as a specific user really.

I'm also wondering, why can't it create the staking directory? That seems odd, and it shouldn't be the user doing it really.

README.md
```diff
@@ -98,3 +98,84 @@ There are two options when using the Avalanche-CLI:
...
+- **Set Home and Mount**: Set `HOME` to `/home/avalanche` and mount the data directory with `-v ~/.avalanchego:/home/avalanche/.avalanchego`.
```
qdm12, line 119: If you leave `HOME` unset, what's its default then? 🤔

qdm12 (commented):
Nice! 💯 

Just a few questions/comments to try to make this as simple as possible!

README.md
```diff
@@ -98,3 +98,84 @@ There are two options when using the Avalanche-CLI:
...
+## Run in Docker
+
+The `subnet-evm` Docker image comes with AvalancheGo pre-installed, making it easy to run a node. You can find the latest image tags on [Docker Hub](https://hub.docker.com/r/avaplatform/subnet-evm/tags).
+
+### Configuration
+
+You can configure the `subnet-evm` Docker container using both environment variables and standard AvalancheGo config files.
+
+- **Environment Variables**: Use uppercase variables prefixed with `AVAGO_`. For example, `AVAGO_NETWORK_ID` corresponds to the `--network-id` flag in AvalancheGo.
+- **Config Files**: Configure as you would with the regular AvalancheGo binary using config files. Ensure `HOME` is set to `/home/avalanche` and mount the config directory with `-v ~/.avalanchego:/home/avalanche/.avalanchego`.
```
ceyonur, line 111: I dont think the container has `/home/avalanche`. Why do we need to set the $HOME?

README.md
```diff
@@ -98,3 +98,84 @@ There are two options when using the Avalanche-CLI:
...
+
+### Data Persistence
+
+To persist data across container restarts, you need to mount the `.avalanchego` directory.
+
+- **Create Directory**: Run `mkdir -p ~/.avalanchego/staking` beforehand to create the necessary directories and avoid file system access issues.
```
ceyonur, line 117: Is there any issue in the default run or was it an issue with specific use-case? I think avalanchego creates the folder itself. 

I kinda think we should keep these steps simple and clear.

README.md
ceyonur, line 119: This can be dangerous if people accidently set this from their host machine. We should either make it ultra clear that this is to be run from the container, or really don't mention this at all if this is not a blocker to run the container. If there is something wrong with the docker's own container, we should fix it in the dockerfile.

qdm12: Or perhaps have it as `/yourpath:/home/avalanche/.avalanchego` to highlight the host path is of your choosing.

README.md
qdm12: Related to this: where does it output files if HOME is not set?

It also feels this should be written in the Dockerfile as an `ENV HOME=somehomepath` + have a `WORKDIR /somehomepath` instead of requiring the user to set it.

README.md
darioush: is there actually a path for (assuming command line) flags to propagate to the Docker image?

README.md
```diff
@@ -98,3 +98,84 @@ There are two options when using the Avalanche-CLI:
...
+- **Set Permissions**: Run the container with `--user $(id -u):$(id -g)` to ensure proper access rights.
+- **Set Home and Mount**: Set `HOME` to `/home/avalanche` and mount the data directory with `-v ~/.avalanchego:/home/avalanche/.avalanchego`.
+
+### Updating
+
+Run `docker stop avago; docker rm avago;` then start a new container with the latest version tag in your `docker run` command.
```
darioush, line 123: nit (or maybe `&&` is more appropriate)
```suggestion
Run `docker stop avago; docker rm avago` then start a new container with the latest version tag in your `docker run` command.
```

README.md
```diff
@@ -98,3 +98,84 @@ There are two options when using the Avalanche-CLI:
...
+
+### Networking
+
+Using `--network host` is recommended to avoid any issues.
+
+If you know what you are doing, you will need port `AVAGO_STAKING_PORT` (default `9651`) open for the validator to connect to the subnet. For the RPC server, open `AVAGO_HTTP_PORT` (default `9650`). Do not attempt to remap `AVAGO_STAKING_PORT` using the Docker `-p` flag (e.g., `-p 9651:1234`); it will not work. Instead, set `AVAGO_STAKING_PORT=1234` and then use `-p 1234:1234`.
```
darioush, line 129: maybe can add a short explanation of why the port mapping doesn't work?

README.md
```diff
@@ -98,3 +98,84 @@ There are two options when using the Avalanche-CLI:
...
+  -e HOME=/home/avalanche \
+  --user $(id -u):$(id -g) \
+  avaplatform/subnet-evm:v0.7.1
+```
+
+- `AVAGO_STAKING_PORT` is set to `9653` in case you want to run this on the same machine as the validator.
+- `AVAGO_HTTP_PORT` is set to `8080` instead of `9650` to avoid conflicts with the validator.
+- `AVAGO_HTTP_ALLOWED_HOSTS` and `AVAGO_HTTP_HOST` are required to allow the RPC server to be accessed from outside. You'll need to secure it with HTTPS; Caddy is recommended.
+
+RPC example uses another folder `~/.avalanchego_rpc` to avoid conflicts with the validator if you want to run both on the same machine.
```
darioush, line 181: curious is it common / recommended to run both on the same machine?

README.md
containerman17: > is there actually a path for (assuming command line) flags to propagate to the Docker image?
Yes, but I think we don’t need to add it to the README—too much information.

README.md
containerman17: Thanks!

README.md
containerman17: If you add `--user $(id -u):$(id -g)`, it sets HOME to an empty value.

README.md
containerman17: Reverted everything to match how AvalancheGo works. Those issues need to be fixed in AvalancheGo, not Subnet-EVM.

README.md
containerman17: Reverted to the default behavior. Thanks for the input! It was resetting HOME when I enforced a non-root user.

README.md
containerman17: Not an issue anymore—I reverted to root. Root doesn’t have any access rights problems.

README.md
containerman17: I would keep `;` for idempotency. It doesn’t make an actual difference since `docker stop` always returns zero, but `;` makes a clear statement: "all commands will execute no matter what."

README.md
containerman17: Added thanks!

README.md
containerman17: I don’t see why not. It’s a trusted binary, so Docker-level isolation should be enough, as long as you have sufficient resources. Subnet-EVM without X and C chains doesn’t require much. Would I run it on the same machine for DFK? Probably not. But for 95% of other networks, especially on Fuji, I don’t see an issue.

Actually, that raises another question—why is it not recommended to run the same node as both a validator and a public RPC node?

darioush: It's not recommended as the validator could be targeted with many API requests when it is its turn to propose a block, or otherwise targeted so it can miss participating in necessary consensus queries.

README.md
```diff
@@ -98,3 +98,75 @@ There are two options when using the Avalanche-CLI:
...
+#### Fuji Subnet RPC
+
+```bash
+docker run -it -d \
+  --name rpc \
+  --network host \
+  -v ~/.avalanchego_rpc/:/home/avalanche/.avalanchego \
+  -e AVAGO_NETWORK_ID=fuji \
+  -e AVAGO_PARTIAL_SYNC_PRIMARY_NETWORK=true \
+  -e AVAGO_TRACK_SUBNETS=hk755meusfKqBb9C9RfzzCxZFkdSXhaFHTcbtycMUSQ11o2cd \
```
darioush, line 159: should this be?
```suggestion
  -e AVAGO_TRACK_SUBNETS=REPLACE_THIS_WITH_YOUR_SUBNET_ID \
```

README.md
qdm12: "it will not work" is not super clear 😉  I suppose this port is shared with other nodes and so it needs to be accessible on publicip:AVAGO_STAKING_PORT right? If so we should also specify any port forwarding on the router should map the same port right?

README.md
```diff
@@ -98,3 +98,75 @@ There are two options when using the Avalanche-CLI:
...
+To persist data across container restarts, you need to mount the `/root/.avalanchego` directory. For example, `-v ~/.avalanchego:/root/.avalanchego`. The container runs as root by default.
+
+### Updating
+
+Run `docker stop avago; docker rm avago;` then start a new container with the latest version tag in your `docker run` command.
+
+### Networking
+
+Using `--network host` is recommended to avoid any issues.
+If you know what you are doing, you will need port `AVAGO_STAKING_PORT` (default `9651`) open for the validator to connect to the subnet. For the RPC server, open `AVAGO_HTTP_PORT` (default `9650`). Do not attempt to remap `AVAGO_STAKING_PORT` using the Docker `-p` flag (e.g., `-p 9651:1234`); it will not work. Instead, set `AVAGO_STAKING_PORT=1234` and then use `-p 1234:1234`.
```
qdm12, line 125:
```suggestion
If you know what you are doing, you will need port `AVAGO_STAKING_PORT` (default `9651`) open for the validator to connect to the subnet. For the RPC server, open `AVAGO_HTTP_PORT` (default `9650`). Only port map the `AVAGO_STAKING_PORT` to the same port number, on both the container (e.g. `-p 1234:1234`) and your router port forwarding, otherwise it won't be reachable, which is explained below.
```

README.md
```diff
@@ -98,3 +98,75 @@ There are two options when using the Avalanche-CLI:
...
+  -e AVAGO_PARTIAL_SYNC_PRIMARY_NETWORK=true \
+  -e AVAGO_TRACK_SUBNETS=REPLACE_THIS_WITH_YOUR_SUBNET_ID \
+  -e AVAGO_PUBLIC_IP_RESOLUTION_SERVICE=ifconfigme \
+  avaplatform/subnet-evm:v0.7.1-rc.0
+```
+
+- `AVAGO_PARTIAL_SYNC_PRIMARY_NETWORK`: Ensures you don't sync the X and C-Chains.
+- `AVAGO_TRACK_SUBNETS`: Sets the subnet ID to track. It will track all chains in the subnet.
+- `AVAGO_NETWORK_ID=fuji`: Sets the network ID to Fuji. Remove to sync Mainnet.
+- `AVAGO_PUBLIC_IP_RESOLUTION_SERVICE=ifconfigme`: Required for AWS EC2 instances to be accessed from outside AWS.
```
qdm12, line 148: nit A markdown table with env variable, default value and description would be nice

README.md
```diff
@@ -98,3 +98,75 @@ There are two options when using the Avalanche-CLI:
...
+  -e AVAGO_HTTP_PORT=8080 \
+  -e AVAGO_STAKING_PORT=9653 \
+  -e AVAGO_HTTP_ALLOWED_HOSTS="*" \
+  -e AVAGO_HTTP_HOST=0.0.0.0 \
+  -e AVAGO_PUBLIC_IP_RESOLUTION_SERVICE=ifconfigme \
+  avaplatform/subnet-evm:v0.7.1-rc.0
+```
+
+- `AVAGO_STAKING_PORT` is set to `9653` in case you want to run this on the same machine as the validator. Remove this to set to the default `9651`.
+- `AVAGO_HTTP_PORT` is set to `8080` instead of `9650` to avoid conflicts with the validator.
```
qdm12, line 169: I would leave `AVAGO_HTTP_PORT=9650` in the docker run command, and specify that the port can be changed in case of a conflict.

README.md
```diff
@@ -98,3 +98,75 @@ There are two options when using the Avalanche-CLI:
...
+  -e AVAGO_TRACK_SUBNETS=hk755meusfKqBb9C9RfzzCxZFkdSXhaFHTcbtycMUSQ11o2cd \
+  -e AVAGO_HTTP_PORT=8080 \
+  -e AVAGO_STAKING_PORT=9653 \
+  -e AVAGO_HTTP_ALLOWED_HOSTS="*" \
+  -e AVAGO_HTTP_HOST=0.0.0.0 \
+  -e AVAGO_PUBLIC_IP_RESOLUTION_SERVICE=ifconfigme \
+  avaplatform/subnet-evm:v0.7.1-rc.0
+```
+
+- `AVAGO_STAKING_PORT` is set to `9653` in case you want to run this on the same machine as the validator. Remove this to set to the default `9651`.
```
qdm12, line 168: I would leave `AVAGO_STAKING_PORT=9651` in the docker run command, and specify that the port can be changed in case of a conflict.

README.md
```diff
@@ -98,3 +98,75 @@ There are two options when using the Avalanche-CLI:
...
+- `AVAGO_PARTIAL_SYNC_PRIMARY_NETWORK`: Ensures you don't sync the X and C-Chains.
+- `AVAGO_TRACK_SUBNETS`: Sets the subnet ID to track. It will track all chains in the subnet.
+- `AVAGO_NETWORK_ID=fuji`: Sets the network ID to Fuji. Remove to sync Mainnet.
+- `AVAGO_PUBLIC_IP_RESOLUTION_SERVICE=ifconfigme`: Required for AWS EC2 instances to be accessed from outside AWS.
+
+#### Fuji Subnet RPC
+
+```bash
+docker run -it -d \
+  --name rpc \
```
qdm12, line 154: A more explicit name would be nice, for example

```suggestion
  --name avago-fuji-rpc \
```

README.md
```diff
@@ -98,3 +98,75 @@ There are two options when using the Avalanche-CLI:
...
+
+This is because the staking port is used for peer-to-peer communication between validators, and the port number is part of the node's identity in the network. If you remap it using Docker's port mapping, other nodes will still try to connect to the original port number advertised by your node, leading to connection failures. The HTTP port can be remapped freely since it's only used for API calls.
+
+### Example Configs
+
+#### Fuji Subnet Validator
+
+```bash
+docker run -it -d \
+  --name avago \
```
qdm12, line 135:
```suggestion
  --name avago-fuji-validator \
```

README.md
```diff
@@ -98,3 +98,75 @@ There are two options when using the Avalanche-CLI:
...
+  --network host \
+  -v ~/.avalanchego:/home/avalanche/.avalanchego \
```
qdm12, line 137: I would rather have it as

```suggestion
  -v /your/validator/path:/home/avalanche/.avalanchego \
```

to make it explicit it can be chosen. Plus `~` doesn't work on Windows, if that matters.

README.md
```diff
@@ -98,3 +98,75 @@ There are two options when using the Avalanche-CLI:
...
+- `AVAGO_NETWORK_ID=fuji`: Sets the network ID to Fuji. Remove to sync Mainnet.
+- `AVAGO_PUBLIC_IP_RESOLUTION_SERVICE=ifconfigme`: Required for AWS EC2 instances to be accessed from outside AWS.
+
+#### Fuji Subnet RPC
+
+```bash
+docker run -it -d \
+  --name rpc \
+  --network host \
+  -v ~/.avalanchego_rpc/:/home/avalanche/.avalanchego \
```
qdm12, line 156: It would be nicer/clearer to just make it explicit the user should pick a path of their choosing

```suggestion
  -v /your/rpc/path/:/home/avalanche/.avalanchego \
```

README.md
qdm12: I agree to remove the trailing `;` I'm not sure why it's there.
For the middle `;`, anyway `docker rm avago` will fail if the stop failed.

README.md
```diff
@@ -98,3 +98,75 @@ There are two options when using the Avalanche-CLI:
...
+
+You can configure the `subnet-evm` Docker container using environment variables, flags or standard AvalancheGo config files.
+
+- **Environment Variables**: Use uppercase variables prefixed with `AVAGO_`. For example, `AVAGO_NETWORK_ID` corresponds to the `--network-id` [flag in AvalancheGo](https://docs.avax.network/nodes/configure/configs-flags#--network-id-string).
+- **Flags**: Please refer to the [AvalancheGo documentation](https://docs.avax.network/nodes/configure/configs-flags) for a list of available flags.
+- **Config Files**: Configure as you would with the regular AvalancheGo binary using config files. Mount the config directory with `-v ~/.avalanchego:/root/.avalanchego`.
+
+### Data Persistence
+
+To persist data across container restarts, you need to mount the `/root/.avalanchego` directory. For example, `-v ~/.avalanchego:/root/.avalanchego`. The container runs as root by default.
```
qdm12, line 116: I'm confused, the commands below bind mount `/home/avalanche/.avalanchego` although they don't have a user setting. Does it run as root or not by default??

qdm12 (changes requested):

PR #1427, opened 2025-01-22
refactor: align accounts/abi/bind with coreth+upstream

darioush:
Includes changes to the `accounts/abi/bind` package, focusing on renaming types and functions for consistency with upstream and coreth by moving some functionality to a new file.

The most important changes include renaming `TmplContract`, `TmplMethod`, and `TmplStruct` to `tmplContract`, `tmplMethod`, and `tmplStruct`, respectively, and moving the `BindHook` function and related types to a new file `bind_extra.go`.

This makes our modifications much clearer.

Some additional cosmetic changes are also reduced.

### DIFF
--- a/cmd/abigen/namefilter.go
+++ b/cmd/abigen/namefilter.go
@@ -7,7 +7,6 @@
 // original code from which it is derived.
 //
 // Much love to the original authors for their work.
-// **********
 package main
 
 import (
--- a/cmd/abigen/namefilter_test.go
+++ b/cmd/abigen/namefilter_test.go
@@ -7,7 +7,6 @@
 // original code from which it is derived.
 //
 // Much love to the original authors for their work.
-// **********
 package main
 
 import (
--- a/triedb/pathdb/database_test.go
+++ b/triedb/pathdb/database_test.go
@@ -319,7 +319,7 @@ func (t *tester) generate(parent common.Hash) (common.Hash, *trienode.MergedNode
 	return root, ctx.nodes, triestate.New(ctx.accountOrigin, ctx.storageOrigin, nil)
 }
 
-// lastRoot returns the latest root hash, or empty if nothing is cached.
+// lastHash returns the latest root hash, or empty if nothing is cached.
 func (t *tester) lastHash() common.Hash {
 	if len(t.roots) == 0 {
 		return common.Hash{}
--- a/accounts/abi/bind/bind.go
+++ b/accounts/abi/bind/bind.go
@@ -43,9 +43,6 @@ import (
 	"github.com/ethereum/go-ethereum/log"
 )
 
-// BindHook is a callback function that can be used to customize the binding.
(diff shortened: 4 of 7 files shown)


ceyonur (approved):

cmd/abigen/namefilter.go
```diff
@@ -7,7 +7,6 @@
 // original code from which it is derived.
 //
 // Much love to the original authors for their work.
-// **********
```
qdm12, line 10: Why the removal? The geth files seem to have no copyright at all, so I'm not sure why we would modify this 🤔

cmd/abigen/namefilter_test.go
```diff
@@ -7,7 +7,6 @@
 // original code from which it is derived.
 //
 // Much love to the original authors for their work.
-// **********
```
qdm12, line 10: Why the removal? The geth files seem to have no copyright at all, so I'm not sure why we would modify this 🤔

triedb/pathdb/database_test.go
```diff
@@ -319,7 +319,7 @@ func (t *tester) generate(parent common.Hash) (common.Hash, *trienode.MergedNode
 	return root, ctx.nodes, triestate.New(ctx.accountOrigin, ctx.storageOrigin, nil)
 }
 
-// lastRoot returns the latest root hash, or empty if nothing is cached.
+// lastHash returns the latest root hash, or empty if nothing is cached.
```
qdm12, line 322: this is a typo in geth, I propose we revert this, however, if you want, I can be the typo comment man and accumulate these in a branch on the libevm repository, that we can PR to geth when it gets large enough?

qdm12 (dismissed):
👍  Just a few minor comments

triedb/pathdb/database_test.go
qdm12: Coreth already contains this typo fix, so I suppose why not then. Ideally we should revert the fix in coreth but that seems like a lot of trouble for just this.

Anyway, I opened https://github.com/ava-labs/libevm/pull/108 feel free to commit to it when you find a geth code typo.

cmd/abigen/namefilter_test.go
qdm12: Never mind, this aligns with coreth 👍

cmd/abigen/namefilter.go
darioush: to match coreth's libevm branch.

triedb/pathdb/database_test.go
darioush: Same as above

cmd/abigen/namefilter.go
qdm12: Don't we want to match geth more than coreth/subnet-evm ultimately? 🤔

qdm12 (approved):

ceyonur (approved):

PR #1429, opened 2025-01-22
chore(core/types): remove `Blocks`'s `Timestamp()` method

qdm12:
## Why this should be merged

In `core/types`, the `Block` method `Timestamp` does the same as the method `Time` and **is not present upstream**, this causes small difficulties migrating to libevm core/types.Block which doesn't have a Timestamp method.

## How this works

- `core/types.Block` method `Timestamp()` removed
- Use `core.NewBlockContext` to convert `types.Block` and `types.Header` for calls to `core.ApplyUpgrades`

## How this was tested

Existing CI passing

## Need to be documented?

No

No

### DIFF
--- a/core/chain_makers.go
+++ b/core/chain_makers.go
@@ -187,8 +187,8 @@ func (b *BlockGen) Number() *big.Int {
 	return new(big.Int).Set(b.header.Number)
 }
 
-// Time returns the time of the block being generated.
-func (b *BlockGen) Time() uint64 {
+// Timestamp returns the timestamp of the block being generated.
+func (b *BlockGen) Timestamp() uint64 {
 	return b.header.Time
 }
 
--- a/core/chain_makers.go
+++ b/core/chain_makers.go
@@ -187,8 +187,8 @@ func (b *BlockGen) Number() *big.Int {
 	return new(big.Int).Set(b.header.Number)
 }
 
-// Timestamp returns the timestamp of the block being generated.
-func (b *BlockGen) Timestamp() uint64 {
+// Time returns the time of the block being generated.
+func (b *BlockGen) Time() uint64 {
 	return b.header.Time
 }
 
--- a/core/chain_makers.go
+++ b/core/chain_makers.go
@@ -284,7 +284,8 @@ func GenerateChain(config *params.ChainConfig, parent *types.Block, engine conse
 		b := &BlockGen{i: i, cm: cm, parent: parent, statedb: statedb, engine: engine}
 		b.header = cm.makeHeader(parent, gap, statedb, b.engine)
 
-		err := ApplyUpgrades(config, &parent.Header().Time, b, statedb)
+		blockContext := NewBlockContext(b.header.Number, b.header.Time)
+		err := ApplyUpgrades(config, &parent.Header().Time, blockContext, statedb)
 		if err != nil {
 			return nil, nil, fmt.Errorf("failed to configure precompiles %w", err)
 		}
--- a/core/state_processor.go
+++ b/core/state_processor.go
@@ -27,7 +27,6 @@
 package core
 
 import (
-	"encoding/json"
 	"fmt"
 	"math/big"
 
@@ -36,13 +35,9 @@ import (
 	"github.com/ava-labs/libevm/crypto"
 	"github.com/ava-labs/libevm/log"
 	"github.com/ava-labs/subnet-evm/consensus"
-	"github.com/ava-labs/subnet-evm/core/extstate"
 	"github.com/ava-labs/subnet-evm/core/state"
 	"github.com/ava-labs/subnet-evm/core/types"
 	"github.com/ava-labs/subnet-evm/params"
-	"github.com/ava-labs/subnet-evm/precompile/contract"
-	"github.com/ava-labs/subnet-evm/precompile/modules"
-	"github.com/ava-labs/subnet-evm/stateupgrade"
 )
 
 // StateProcessor is a basic Processor, which takes care of transitioning
@@ -98,6 +93,7 @@ func (p *StateProcessor) Process(block *types.Block, parent *types.Header, state
 	if beaconRoot := block.BeaconRoot(); beaconRoot != nil {
 		ProcessBeaconBlockRoot(*beaconRoot, vmenv, statedb)
 	}
+
 	// Iterate over and process the individual transactions
 	for i, tx := range block.Transactions() {
 		msg, err := TransactionToMessage(tx, signer, header.BaseFee)
@@ -130,7 +126,6 @@ func applyTransaction(msg *Message, config *params.ChainConfig, gp *GasPool, sta
 	if err != nil {
 		return nil, err
 	}
-
 	// Update the state with pending changes.
 	var root []byte
 	if config.IsByzantium(blockNumber) {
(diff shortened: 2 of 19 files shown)


precompile/contracts/deployerallowlist/module.go
```diff
@@ -43,10 +44,10 @@ func (*configurator) MakeConfig() precompileconfig.Config {
 
 // Configure configures [state] with the given [cfg] precompileconfig.
 // This function is called by the EVM once per precompile contract activation.
-func (c *configurator) Configure(chainConfig precompileconfig.ChainConfig, cfg precompileconfig.Config, state contract.StateDB, blockContext contract.ConfigurationBlockContext) error {
+func (c *configurator) Configure(chainConfig precompileconfig.ChainConfig, cfg precompileconfig.Config, state contract.StateDB, blockNumber *big.Int) error {
```
darioush, line 47: Could we keep both the block number and time here even if the time is unused? As we previously exposed this interface in `precompile-evm` and it seems it could be used in the future?

precompile/allowlist/config.go
```diff
@@ -21,8 +21,8 @@ type AllowListConfig struct {
 }
 
 // Configure initializes the address space of [precompileAddr] by initializing the role of each of
-// the addresses in [AllowListAdmins].
-func (c *AllowListConfig) Configure(chainConfig precompileconfig.ChainConfig, precompileAddr common.Address, state contract.StateDB, blockContext contract.ConfigurationBlockContext) error {
+// the addresses in [AllowListConfig].
```
darioush, line 24: Could we keep the arguments for this method for symmetry with the other methods (even though unused)?

precompile/contracts/deployerallowlist/module.go
qdm12: Sure, one of the last commits removes the time argument, I wanted to see your opinion on this 😉

precompile/allowlist/config.go
qdm12: I'm not sure I understand what symmetry you are thinking about? With the `Configurator.Configure`??

darioush: That was my thought, even though the AllowList is not a precompile it follows through the phases of Configure along with the precompile.

qdm12: Related question: what would be the cost of changing the signature of AllowListConfig.Configure if needed in the future? 🤔

precompile/contracts/deployerallowlist/module.go
qdm12: Actually, similar question: what would be the cost to change the function signature in the future?

darioush: It would require changes to all precompiles, both those that we have developed and any external code bases that may have developed precompiles for use in subnet-evm or precompile-evm.
So I think we should avoid changing it without good reason, but we can still change it to deliver improvements, features, and a better interface.

It seems natural the code that sets up the precompile would have access to both the timestamp and the number of the block where it's activated.

precompile/allowlist/config.go
darioush: Less than the other method, as the AllowList is not part of the precompile interface but more a partial implementation that's used by some of the precompiles.

As I mentioned I have a subjective preference for methods that all take the same arguments but this is not a strong preference.

accounts/abi/bind/precompilebind/precompile_module_template.go
```diff
@@ -59,7 +60,7 @@ func (*configurator) MakeConfig() precompileconfig.Config {
 // This function is called by the EVM once per precompile contract activation.
 // You can use this function to set up your precompile contract's initial state,
 // by using the [cfg] config and [state] stateDB.
-func (*configurator) Configure(chainConfig precompileconfig.ChainConfig, cfg precompileconfig.Config, state contract.StateDB, blockContext contract.ConfigurationBlockContext) error {
+func (*configurator) Configure(chainConfig precompileconfig.ChainConfig, cfg precompileconfig.Config, state contract.StateDB, blockNumber *big.Int, blockTime uint64) error {
```
ceyonur, line 63: this is a breaking change to all precompiles (external and internal precompiles). While having a simple params is better, we should also evaluate this with the future extension possibilities. IMO having a blockContext interface is less cumbersome to extend.

ceyonur (commented):
changes in the subnet-evm context are lgtm, but we should consider that this will break all precompiles (including precompiles in precompile-evm + subnet-evm forks)

precompile/contracts/deployerallowlist/module.go
qdm12: Interesting, should we then not change the signature, such that the block context interface (now removed) can be later extended?

accounts/abi/bind/precompilebind/precompile_module_template.go
qdm12: @darioush thoughts on this? 🤔

qdm12: No longer breaking this.

precompile/contracts/deployerallowlist/module.go
qdm12: No longer changing this

precompile/contract/interfaces.go
```diff
@@ -62,7 +62,7 @@ type AccessibleState interface {
 // ConfigurationBlockContext defines the interface required to configure a precompile.
 type ConfigurationBlockContext interface {
 	Number() *big.Int
-	Timestamp() uint64
+	Time() uint64
```
ceyonur, line 65: I think this still introduces a breaking change (but slightly less). It still breaks precompiles that use Block.timestamp in their configuration. But I believe this is a far less problem and if it's required for libevm (not targeted for master), it should be fine.

ceyonur (dismissed):

precompile/contract/interfaces.go
qdm12: Look at the last commit: it uses core.NewBlockContext to go around breaking changes, such that this interface still has a Timestamp method and the Timestamp method is removed from the types.Block. It's pretty good I think?

core/state_processor.go
```diff
@@ -284,16 +284,16 @@ func ApplyUpgrades(c *params.ChainConfig, parentTimestamp *uint64, blockContext
 }
 
 type blockContext struct {
-	number    *big.Int
-	timestamp uint64
+	number *big.Int
+	time   uint64
```
qdm12, line 289: note: changing this to time since it should usually be the block.Time() set here. We can revert this field, but the argument to `NewBlockContext` should be `time` imo

ceyonur (approved):

core/chain_makers.go
```diff
@@ -187,8 +187,8 @@ func (b *BlockGen) Number() *big.Int {
 	return new(big.Int).Set(b.header.Number)
 }
 
-// Timestamp returns the timestamp of the block being generated.
-func (b *BlockGen) Timestamp() uint64 {
+// Time returns the time of the block being generated.
+func (b *BlockGen) Time() uint64 {
```
darioush, line 191: I think it's actually Timestamp() in upstream, let's leave it as upstream

qdm12: Totally, plus that's an unneeded change anyway 👍

darioush (approved):

ceyonur (approved):

PR #1430, opened 2025-01-22
refactor: params/extras (coreth a7f7061)

darioush:
Applies the same changes as were applied to coreth here: https://github.com/ava-labs/coreth/commit/a7f7061ce0662671eea01e973676c52879a4461c

I realized we could import most of these types back to params if we wanted, but explicit seems clearer + it's already like this in coreth.

### DIFF
--- a/params/config_libevm.go
+++ b/params/config_libevm.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/ava-labs/libevm/common"
 	ethparams "github.com/ava-labs/libevm/params"
+	"github.com/ava-labs/subnet-evm/params/extras"
 	"github.com/ava-labs/subnet-evm/precompile/modules"
 	"github.com/ava-labs/subnet-evm/precompile/precompileconfig"
 )
@@ -15,19 +16,19 @@ import (
 // libevmInit would ideally be a regular init() function, but it MUST be run
 // before any calls to [ChainConfig.Rules]. See `config.go` for its call site.
 func libevmInit() any {
-	extras = ethparams.RegisterExtras(ethparams.Extras[*ChainConfigExtra, RulesExtra]{
+	payloads = ethparams.RegisterExtras(ethparams.Extras[*extras.ChainConfig, RulesExtra]{
 		ReuseJSONRoot: true, // Reuse the root JSON input when unmarshalling the extra payload.
 		NewRules:      constructRulesExtra,
 	})
 	return nil
 }
 
-var extras ethparams.ExtraPayloads[*ChainConfigExtra, RulesExtra]
+var payloads ethparams.ExtraPayloads[*extras.ChainConfig, RulesExtra]
 
 // constructRulesExtra acts as an adjunct to the [params.ChainConfig.Rules]
 // method. Its primary purpose is to construct the extra payload for the
 // [params.Rules] but it MAY also modify the [params.Rules].
-func constructRulesExtra(c *ethparams.ChainConfig, r *ethparams.Rules, cEx *ChainConfigExtra, blockNum *big.Int, isMerge bool, timestamp uint64) RulesExtra {
+func constructRulesExtra(c *ethparams.ChainConfig, r *ethparams.Rules, cEx *extras.ChainConfig, blockNum *big.Int, isMerge bool, timestamp uint64) RulesExtra {
 	var rules RulesExtra
 	if cEx == nil {
 		return rules
@@ -39,7 +40,7 @@ func constructRulesExtra(c *ethparams.ChainConfig, r *ethparams.Rules, cEx *Chai
 	rules.Predicaters = make(map[common.Address]precompileconfig.Predicater)
 	rules.AccepterPrecompiles = make(map[common.Address]precompileconfig.Accepter)
 	for _, module := range modules.RegisteredModules() {
-		if config := cEx.getActivePrecompileConfig(module.Address, timestamp); config != nil && !config.IsDisabled() {
+		if config := cEx.GetActivePrecompileConfig(module.Address, timestamp); config != nil && !config.IsDisabled() {
 			rules.Precompiles[module.Address] = config
 			if predicater, ok := config.(precompileconfig.Predicater); ok {
 				rules.Predicaters[module.Address] = predicater
--- a/params/extras/config.go
+++ b/params/extras/config.go
@@ -1,0 +1,352 @@
+// (c) 2024 Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package extras
+
+import (
+	"encoding/json"
+	"fmt"
+	"math/big"
+
+	"github.com/ava-labs/avalanchego/snow"
+	"github.com/ava-labs/avalanchego/upgrade"
+	"github.com/ava-labs/avalanchego/utils/constants"
+	"github.com/ava-labs/libevm/common"
+	ethparams "github.com/ava-labs/libevm/params"
+	"github.com/ava-labs/subnet-evm/commontype"
+	"github.com/ava-labs/subnet-evm/utils"
+)
+
+var (
+	DefaultFeeConfig = commontype.FeeConfig{
+		GasLimit:        big.NewInt(8_000_000),
+		TargetBlockRate: 2, // in seconds
+
+		MinBaseFee:               big.NewInt(25_000_000_000),
+		TargetGas:                big.NewInt(15_000_000),
+		BaseFeeChangeDenominator: big.NewInt(36),
+
+		MinBlockGasCost:  big.NewInt(0),
+		MaxBlockGasCost:  big.NewInt(1_000_000),
+		BlockGasCostStep: big.NewInt(200_000),
+	}
+
+	TestChainConfig = &ChainConfig{
+		AvalancheContext:   AvalancheContext{utils.TestSnowContext()},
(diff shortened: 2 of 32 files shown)


params/extras/precompile_upgrade_test.go
```diff
@@ -15,8 +16,7 @@ import (
 
 func TestVerifyUpgradeConfig(t *testing.T) {
 	admins := []common.Address{{1}}
-	chainConfigCpy := Copy(TestChainConfig)
-	chainConfig := GetExtra(&chainConfigCpy)
+	chainConfig := &ChainConfig{}
```
darioush, line 19: note I tried to use the empty ChainConfig (former ChainConfigExtra) where possible
We can switch back to TestChainConfig if others prefer.
Less test setup seems preferred to me, but we could keep it to avoid any risk here.

params/extras/config.go
```diff
@@ -0,0 +1,352 @@
...
+	"github.com/ava-labs/avalanchego/upgrade"
+	"github.com/ava-labs/avalanchego/utils/constants"
+	"github.com/ava-labs/libevm/common"
+	ethparams "github.com/ava-labs/libevm/params"
+	"github.com/ava-labs/subnet-evm/commontype"
+	"github.com/ava-labs/subnet-evm/utils"
+)
+
+var (
+	DefaultFeeConfig = commontype.FeeConfig{
```
darioush, line 21: I moved this so it can be used in tests, but also put it back in it's original location to avoid having to change all params.DefaultFeeConfig's in the code.

params/extras/config.go
```diff
@@ -0,0 +1,352 @@
...
+		MinBlockGasCost:  big.NewInt(0),
+		MaxBlockGasCost:  big.NewInt(1_000_000),
+		BlockGasCostStep: big.NewInt(200_000),
+	}
+
+	TestChainConfig = &ChainConfig{
+		AvalancheContext:   AvalancheContext{utils.TestSnowContext()},
+		FeeConfig:          DefaultFeeConfig,
+		AllowFeeRecipients: false,
+		NetworkUpgrades:    GetDefaultNetworkUpgrades(upgrade.GetConfig(constants.UnitTestID)), // This can be changed to correct network (local, test) via VM.
```
darioush, line 38: note we could unexport this GetDefaultNetworkUpgrades if we had something like GetDefaultExtras(networkID), but seems mostly the same to me.

params/config_libevm.go
```diff
@@ -8,26 +8,27 @@ import (
...
+	"github.com/ava-labs/subnet-evm/params/extras"
 	"github.com/ava-labs/subnet-evm/precompile/modules"
 	"github.com/ava-labs/subnet-evm/precompile/precompileconfig"
 )
 
 // libevmInit would ideally be a regular init() function, but it MUST be run
 // before any calls to [ChainConfig.Rules]. See `config.go` for its call site.
 func libevmInit() any {
-	extras = ethparams.RegisterExtras(ethparams.Extras[*ChainConfigExtra, RulesExtra]{
+	payloads = ethparams.RegisterExtras(ethparams.Extras[*extras.ChainConfig, RulesExtra]{
```
qdm12, line 19: nit why the rename `extras` -> `payload`? I don't really mind at all, just curious

params/extras/config.go
```diff
@@ -0,0 +1,352 @@
+// (c) 2024 Ava Labs, Inc. All rights reserved.
```
qdm12, line 1:
```suggestion
// (c) 2025 Ava Labs, Inc. All rights reserved.
```

params/extras/config.go
qdm12: Maybe that's out of scope, but can't we unexport and put these configurations `DefaultFeeConfig` and `TestChainConfig` in a `_test.go` file, since these are only used in tests in this package?

params/extras/precompile_upgrade_test.go
```diff
@@ -277,16 +277,16 @@ type upgradeCompatibilityTest struct {
 func (tt *upgradeCompatibilityTest) run(t *testing.T, chainConfig ChainConfig) {
 	// apply all the upgrade bytes specified in order
 	for i, upgrade := range tt.configs {
-		newCfg := Copy(&chainConfig)
-		GetExtra(&newCfg).UpgradeConfig = *upgrade
+		newCfg := chainConfig
+		newCfg.UpgradeConfig = *upgrade
 
-		err := chainConfig.CheckCompatible(&newCfg, 0, tt.startTimestamps[i])
+		err := chainConfig.checkConfigCompatible(&newCfg, new(big.Int), tt.startTimestamps[i])
```
qdm12, line 283: why the change from `CheckCompatible` to `checkConfigCompatible`? Can't we keep it to `checkCompatible` (unexported)?

params/extras/config.go
```diff
@@ -0,0 +1,352 @@
...
+// Implements precompile.ChainConfig interface.
+func (c *ChainConfig) AllowedFeeRecipients() bool {
+	return c.AllowFeeRecipients
+}
+
+// IsForkTransition returns true if [fork] activates during the transition from
+// [parent] to [current].
+// Taking [parent] as a pointer allows for us to pass nil when checking forks
+// that activate during genesis.
+// Note: [parent] and [current] can be either both timestamp values, or both
```
qdm12, line 335: nit - maybe out of scope - but brackets only work for type definitions or global variables, they won't work for arguments 🤷  Maybe worth changing to use backticks instead

```suggestion
// IsForkTransition returns true if `fork` activates during the transition from
// `parent` to `current`.
// Taking `parent` as a pointer allows for us to pass nil when checking forks
// that activate during genesis.
// Note: `parent` and `current` can be either both timestamp values, or both
```

params/extras/config.go
```diff
@@ -0,0 +1,352 @@
...
+
+	// Verify the network upgrades are internally consistent given the existing chainConfig.
+	if err := c.verifyNetworkUpgrades(c.SnowCtx.NetworkUpgrades); err != nil {
+		return fmt.Errorf("invalid network upgrades: %w", err)
+	}
+
+	return nil
+}
+
+// IsPrecompileEnabled returns whether precompile with [address] is enabled at [timestamp].
```
qdm12, line 313: nit

```suggestion
// IsPrecompileEnabled returns whether precompile with `address` is enabled at `timestamp`.
```

params/extras/config.go
```diff
@@ -0,0 +1,352 @@
...
+	// Avalanche upgrades are enabled in order.
+	// Note: we do not add the precompile configs here because they are optional
+	// and independent, i.e. the order in which they are enabled does not impact
+	// the correctness of the chain config.
+	return checkForks(c.forkOrder(), false)
+}
+
+// checkForks checks that forks are enabled in order and returns an error if not.
+// [blockFork] is true if the fork is a block number fork, false if it is a timestamp fork
+// TODO: This code was adapted from CheckConfigForkOrder, consider refactoring to avoid duplication.
```
qdm12, line 253: New TODO added, just in case it was forgotten - otherwise discard 😉

params/extras/config.go
```diff
@@ -0,0 +1,352 @@
...
+	// it would not be a meaningful comparison. Instead, we only check that the
+	// Avalanche upgrades are enabled in order.
+	// Note: we do not add the precompile configs here because they are optional
+	// and independent, i.e. the order in which they are enabled does not impact
+	// the correctness of the chain config.
+	return checkForks(c.forkOrder(), false)
+}
+
+// checkForks checks that forks are enabled in order and returns an error if not.
+// [blockFork] is true if the fork is a block number fork, false if it is a timestamp fork
```
qdm12, line 252: nit brackets don't work for function arguments

```suggestion
// `blockFork` is true if the fork is a block number fork, false if it is a timestamp fork
```

qdm12 (approved):

params/extras/config.go
darioush: I appreciate the comments on details that help us have a more consistent codebase, but since we are moving code between repositories, it is more important that they are textually in sync so the same diffs apply without merge conflicts.
This is a losing game to begin with, which is why we are pursuing the repository merger of coreth & subnet-evm.
Once we a single repos, making small improvements to code quality while moving it seems more reasonable.

params/extras/config.go
darioush: Using the brackets seems very specific to the avalanche code bases, I would appreciate if you were able to figure out the best practices here and communicate them with the team, especially for code on a go-forward basis.

params/extras/config.go
darioush: It is re-exported (and then referred to in many places) as `params.DefaultFeeConfig`

params/config_libevm.go
darioush: `extras` is the name of an imported module here.

params/extras/precompile_upgrade_test.go
darioush: `CheckConfigCompatible` is referred to in libvem, and `CheckConfigCompatible` converts types and calls the unexported `checkConfigCompatible`.
Since this is how it is in coreth & libevm, I'm not sure what we gain by changing it.

params/extras/config.go
qdm12: My apologies, totally right 💯

ceyonur (approved):

Issue #1428, opened 2025-01-22
Remove deprecated legacy gossip handler


ceyonur:
We have deprecated legacy gossip handler in favor of p2p sdk gossip handler (push/pull gossips). We have removed it from coreth with the PR https://github.com/ava-labs/coreth/pull/706 

Coreth PR is a big one but these changes should be same to remove the deprecated handler from subnet-evm as well:

* Network file: https://github.com/ava-labs/coreth/pull/706/files#diff-9cd76a0999351a7520dce33ca95e771d39a4560eb9f589abd38d392506fdfc39

* Network test file: https://github.com/ava-labs/coreth/pull/706/files#diff-305c10713f73ecc40109c234bebfcdcac4bfc4a32a0e4a894e04bc8a0d3b1635

* Handler file removal: https://github.com/ava-labs/coreth/pull/706/files#diff-99c1d8c878c89edcd40d329c34104fd12df71d1bdeb3dfae51ad7f5758d42424

* Gossip stats removal: https://github.com/ava-labs/coreth/pull/706/files#diff-956126bb535d3b8f9fe540954850254433316fabe9e908b55a48de6a1cb153c1

* Skip registration for `EthTxsGossip` type in codec: https://github.com/ava-labs/coreth/pull/706/files#diff-295ec2bc5b901db96bc911c27f3ef2df3eaf51e592081cab92d1dd51cbe372bc

* no-op handler changes: https://github.com/ava-labs/coreth/pull/706/files#diff-f89e2ab248338043520f9f69119ec289998cdca066fa80e0d86160cfe00717dd

* message, message_test, handler_test removal: https://github.com/ava-labs/coreth/pull/706/files#diff-a447f410a435eacd25a234db3635e85c4664c11c7feb7326556b62662fc5405e

JonathanOppenheimer:
Closed in #1435

PR #1435, opened 2025-01-28
Remove legacy gossip

alarso16:
## Why this should be merged
This removes the deprecated legacy gossip manager.

## How this works
All revisions exactly follow the recommendations laid out in #1428, following the example set in the [coreth PR](https://github.com/ava-labs/subnet-evm/issues/1428). Additionally, a change to `plugin/evm/vm.go` to completely remove the implementation of the deprecated handlers.

## How this was tested
Unit tests

## Need to be documented?
No

Yes - updated.


plugin/evm/message/codec.go
```diff
@@ -15,19 +15,14 @@ const (
...
+var Codec codec.Manager
 
 func init() {
 	Codec = codec.NewManager(maxMessageSize)
 	c := linearcodec.NewDefault()
 
 	errs := wrappers.Errs{}
 	errs.Add(
-		// Gossip types
-		c.RegisterType(EthTxsGossip{}),
```
darioush, line 29: We should replace this with:
`c.SkipRegistrations(1)`
this will keep the numerical types registered below unchanged.

alarso16: Added above err wrapper.

peer/network_test.go
```diff
@@ -503,15 +505,15 @@ func TestHandleInvalidMessages(t *testing.T) {
...
 	clientNetwork.SetRequestHandler(&testRequestHandler{})
 
 	assert.NoError(t, clientNetwork.Connected(context.Background(), nodeID, defaultPeerVersion))
 
 	defer clientNetwork.Shutdown()
 
 	// Ensure a valid gossip message sent as any App specific message type does not trigger a fatal error
-	gossipMsg, err := buildGossip(codecManager, HelloGossip{Msg: "hello there!"})
+	marshaller := helloGossipMarshaller{codec: codecManager}
+	gossipMsg, err := marshaller.MarshalGossip(&HelloGossip{Msg: "hello sthere!"})
```
qdm12, line 516: nit there is an extra s! 😨 

```suggestion
	gossipMsg, err := marshaller.MarshalGossip(&HelloGossip{Msg: "hello there!"})
```

darioush (dismissed):

RELEASES.md
```diff
@@ -3,6 +3,7 @@
 ## Pending Release
 
 * Refactored trie_prefetcher.go to be structurally similar to upstream.
+* Remove legacy gossip handler and metrics (deprecated since v0.10.0)
```
ceyonur, line 6: coreth and subnet-evm versions are different. Can we just write "removed deprecated legacy gossip handler and metrics"?

peer/network_test.go
```diff
@@ -719,28 +716,22 @@ type HelloGossip struct {
...
 }
 
-func (t *testGossipHandler) HandleEthTxs(nodeID ids.NodeID, msg message.EthTxsGossip) error {
-	t.received = true
-	t.nodeID = nodeID
-	return nil
+func (g helloGossipMarshaller) UnmarshalGossip(bytes []byte) (*HelloGossip, error) {
+	h := &HelloGossip{}
+	_, error := g.codec.Unmarshal(bytes, h)
+	return h, error
```
qdm12, line 734: nit convention in Go (🤷) is to have error variables as `err`

```suggestion
	_, err := g.codec.Unmarshal(bytes, h)
	return h, err
```

peer/network_test.go
```diff
@@ -30,6 +30,10 @@ import (
 	"github.com/ava-labs/avalanchego/version"
 )
 
+const (
+	codecVersion uint16 = 0
+)
```
qdm12, line 35: nit: since this only used for tests, perhaps let's rename it to `testCodecVersion` to avoid any confusion?

Also would it be possible to use another version, like `1`? Personal opinion, it's nice to test with non-zero values to verify nothing was forgotten (since default = zero value)

RELEASES.md
alarso16: Fixed

ceyonur (approved):

plugin/evm/message/codec.go
```diff
@@ -15,19 +15,17 @@ const (
 	maxMessageSize = 2*units.MiB - 64*units.KiB // Subtract 64 KiB from p2p network cap to leave room for encoding overhead from AvalancheGo
 )
 
-var (
-	Codec codec.Manager
-)
+var Codec codec.Manager
```
qdm12, line 18: nit unneeded diff, let's revert it

```suggestion
var (
	Codec codec.Manager
)
```

plugin/evm/message/handler.go
```diff
@@ -6,27 +6,10 @@ package message
...
-	HandleEthTxs(nodeID ids.NodeID, msg EthTxsGossip) error
-}
-
-type NoopMempoolGossipHandler struct{}
-
-func (NoopMempoolGossipHandler) HandleEthTxs(nodeID ids.NodeID, msg EthTxsGossip) error {
-	log.Debug("dropping unexpected EthTxsGossip message", "peerID", nodeID)
-	return nil
-}
+var _ RequestHandler = NoopRequestHandler{}
```
qdm12, line 12: nit to reduce git diffs

```suggestion
var (
	_ RequestHandler = NoopRequestHandler{}
)
```

qdm12 (approved):
I like 🔴  deletions 👍 
Just a few nitpicky comments here and there, but looks good 💯

ceyonur (approved):

darioush (approved):

PR #1448, opened 2025-02-26
Refactor blockgascost

ceyonur:
## Why this should be merged

Cherry picks following PR from Coreth and syncs coreth master up to [b9890a7](https://github.com/ava-labs/coreth/commit/b9890a7d6be65e2e54364762858a8b3c2471d439)

https://github.com/ava-labs/coreth/pull/792

## How this was tested

Added relevant UTs for subnet-evm fee configs.

## Need to be documented?

No

No

### DIFF
--- a/consensus/dummy/consensus.go
+++ b/consensus/dummy/consensus.go
@@ -17,6 +17,7 @@ import (
 	"github.com/ava-labs/subnet-evm/core/types"
 	"github.com/ava-labs/subnet-evm/params"
 	"github.com/ava-labs/subnet-evm/trie"
+	"github.com/ava-labs/subnet-evm/utils"
 	"github.com/ava-labs/subnet-evm/vmerrs"
 	"github.com/ethereum/go-ethereum/common"
 
@@ -26,11 +27,10 @@ import (
 var (
 	allowedFutureBlockTime = 10 * time.Second // Max time from current time allowed for blocks, before they're considered future blocks
 
-	errInvalidBlockTime     = errors.New("timestamp less than parent's")
-	errUnclesUnsupported    = errors.New("uncles unsupported")
-	errBlockGasCostNil      = errors.New("block gas cost is nil")
-	errBlockGasCostTooLarge = errors.New("block gas cost is not uint64")
-	errBaseFeeNil           = errors.New("base fee is nil")
+	errInvalidBlockTime  = errors.New("timestamp less than parent's")
+	errUnclesUnsupported = errors.New("uncles unsupported")
+	errBlockGasCostNil   = errors.New("block gas cost is nil")
+	errBaseFeeNil        = errors.New("base fee is nil")
 )
 
 type Mode struct {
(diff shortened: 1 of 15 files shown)


plugin/evm/blockgascost/cost.go
```diff
@@ -0,0 +1,57 @@
...
+	safemath "github.com/ava-labs/avalanchego/utils/math"
+	"github.com/ava-labs/subnet-evm/commontype"
+)
+
+// BlockGasCost calculates the required block gas cost.
+//
+// cost = parentCost + step * (TargetBlockRate - timeElapsed)
+//
+// The returned cost is clamped to [MinBlockGasCost, MaxBlockGasCost].
+func BlockGasCost(
```
darioush, line 19: seems this method has a different type than coreth, but the code is very similar to ap4 package with additional parameters.

probably later we can consider merging these code paths?

plugin/evm/header/block_gas_cost.go
```diff
@@ -0,0 +1,61 @@
...
+	"math/big"
+
+	"github.com/ava-labs/subnet-evm/commontype"
+	"github.com/ava-labs/subnet-evm/core/types"
+	"github.com/ava-labs/subnet-evm/plugin/evm/blockgascost"
+)
+
+// BlockGasCost calculates the required block gas cost based on the parent
+// header and the timestamp of the new block.
+func BlockGasCost(
```
darioush, line 16: I notice this has a different type than coreth. Is the intention to have 2 packages for custom header? (sort of makes sense)

darioush (approved):

plugin/evm/blockgascost/cost.go
ceyonur: yes, the main thing is we need to support dynamic configurations here so we need to pass fee configs.

plugin/evm/header/block_gas_cost.go
ceyonur: I think we can merge it by passing coreth fee configs as a commontype.FeeConfig. These will be refactored in several times (with upcoming cherry picked commits), as long as I see in the final form we have very similar fee configs in `ap3, ap4` packages. So we can just define them as `ap4Configs commontype.FeeConfig` etc. (wishful thinking)

alarso16 (approved):

consensus/dummy/consensus.go
```diff
@@ -358,32 +349,31 @@ func (eng *DummyEngine) verifyBlockFee(
 }
 
 func (eng *DummyEngine) Finalize(chain consensus.ChainHeaderReader, block *types.Block, parent *types.Header, state *state.StateDB, receipts []*types.Receipt) error {
-	if chain.Config().IsSubnetEVM(block.Time()) {
+	config := chain.Config()
+	timestamp := block.Time()
+	if config.IsSubnetEVM(block.Time()) {
```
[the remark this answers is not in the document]
ceyonur: ah good catch

PR #1451, opened 2025-02-28
Refactor AP3 `Window` and Params (#808)

ceyonur:
Cherry picks following PR from Coreth

https://github.com/ava-labs/coreth/pull/808


alarso16 (approved):
I left some comments on potential diffs from coreth (and there were more with the `:=` replacements in this PR). If this isn't an actual issue, feel free to ignore the comments.

plugin/evm/header/block_gas_cost.go
```diff
@@ -59,3 +66,31 @@ func BlockGasCostWithStep(
...
+	if header.BlockGasCost == nil {
+		return nil, errBlockGasCostNil
+	}
+
+	// minTip = requiredBlockFee/blockGasUsage
+	requiredBlockFee := new(big.Int).Mul(
+		header.BlockGasCost,
+		header.BaseFee,
+	)
+	return new(big.Int).Div(requiredBlockFee, new(big.Int).SetUint64(header.GasUsed)), nil
```
qdm12, line 95: nit it would be nicer split, this is a bit long to follow

```suggestion
	blockGasUsage := new(big.Int).SetUint64(header.GasUsed)
	return new(big.Int).Div(requiredBlockFee, blockGasUsage), nil
```

ceyonur: This is same as in https://github.com/ava-labs/coreth/pull/808, I think we should keep it that way.

PR #1462, opened 2025-03-04
Acp 176 config

ceyonur:


plugin/evm/upgrade/acp176/acp176.go
```diff
@@ -18,24 +18,24 @@ import (
...
 
 const (
-	MinTargetPerSecond  = 1_000_000                                 // P
+	MaxTargetExcessDiff = 1 << 15                                   // Q
 	TargetConversion    = MaxTargetChangeRate * MaxTargetExcessDiff // D
-	MaxTargetExcessDiff = 2 ^ 15                                    // Q
 
-	TimeToFillCapacity            = 10   // in seconds
-	TargetToMax                   = 2    // multiplier to convert from target per second to max per second
-	TargetToPriceUpdateConversion = 87   // 87s ~= 60s * ln(2) which makes the price double at most every ~60 seconds
```
ceyonur, line 31: I wonder if the calculation here made by using `TargetToMax` if so we need to factor out 60s as a separate parameter.

plugin/evm/upgrade/acp176/acp176.go
```diff
@@ -18,24 +18,24 @@ import (
...
-	MaxTargetChangeRate           = 1024 // Controls the rate that the target can change per block.
```
ceyonur, line 32: this is something we cannot parameterize because EVM also imposes the same limit.

plugin/evm/upgrade/acp176/acp176.go
```diff
@@ -18,24 +18,24 @@ import (
...
 }
 
 const (
-	MinTargetPerSecond  = 1_000_000                                 // P
+	MaxTargetExcessDiff = 1 << 15                                   // Q
 	TargetConversion    = MaxTargetChangeRate * MaxTargetExcessDiff // D
-	MaxTargetExcessDiff = 2 ^ 15                                    // Q
 
-	TimeToFillCapacity            = 10   // in seconds
-	TargetToMax                   = 2    // multiplier to convert from target per second to max per second
```
ceyonur, line 30: Does `2` here control doubling the price? I.e does 3 here means triple it?

plugin/evm/upgrade/acp176/acp176.go
StephenButtolph: This is something we can parameterize

plugin/evm/upgrade/acp176/acp176.go
alarso16: Relating to my comment above, this equation (as a function of this ratio and `TargetToMax`, would look like this:
`DoublingTime = TargetToPriceUpdateConversion * ln(2) / (TargetToMax - 1)`

plugin/evm/upgrade/acp176/acp176.go
ceyonur: isn't this the same value for https://github.com/ava-labs/subnet-evm/blob/master/params/protocol_params.go#L36?

PR #1465, opened 2025-03-04
use ctx in validatorsManager

ceyonur:
## Why this should be merged

`snow.Context.Lock` provided by AvalancheGo(rpcchainvm) is not reliable for engine calls. Validator manager should not depend on that, and rather hold the ctx itself.

## How this works

Moves ctx lock management to validators manager, hold ctx lock in `Connected`, `Disconnected`

## How this was tested

UT, locally

## Need to be documented?

No

updated

### DIFF
--- a/plugin/evm/vm.go
+++ b/plugin/evm/vm.go
@@ -720,17 +720,11 @@ func (vm *VM) onNormalOperationsStarted() error {
 	ctx, cancel := context.WithCancel(context.TODO())
 	vm.cancel = cancel
 
-	// sync validators first
-	if err := vm.validatorsManager.Sync(ctx); err != nil {
-		return fmt.Errorf("failed to update validators: %w", err)
+	// Start the validators manager
+	if err := vm.validatorsManager.Initialize(ctx); err != nil {
+		return fmt.Errorf("failed to initialize validators manager: %w", err)
 	}
-	vdrIDs := vm.validatorsManager.GetNodeIDs().List()
-	// Then start tracking with updated validators
-	// StartTracking initializes the uptime tracking with the known validators
-	// and update their uptime to account for the time we were being offline.
-	if err := vm.validatorsManager.StartTracking(vdrIDs); err != nil {
-		return fmt.Errorf("failed to start tracking uptime: %w", err)
-	}
+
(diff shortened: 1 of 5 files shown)


plugin/evm/vm_test.go
```diff
@@ -153,10 +153,6 @@ func setupGenesis(
 	atomicMemory := atomic.NewMemory(prefixdb.New([]byte{0}, baseDB))
 	ctx.SharedMemory = atomicMemory.NewSharedMemory(ctx.ChainID)
 
-	// NB: this lock is intentionally left locked when this function returns.
-	// The caller of this function is responsible for unlocking.
-	ctx.Lock.Lock()
```
ceyonur, line 158: uh this was a weird assumption. This might work for coreth (as it's expected a lock to be held in engine before calling VM functions) but this does not work here.

plugin/evm/vm.go
```diff
@@ -720,17 +720,11 @@ func (vm *VM) onNormalOperationsStarted() error {
...
-		return fmt.Errorf("failed to update validators: %w", err)
-	}
-	vdrIDs := vm.validatorsManager.GetNodeIDs().List()
-	// Then start tracking with updated validators
-	// StartTracking initializes the uptime tracking with the known validators
-	// and update their uptime to account for the time we were being offline.
-	if err := vm.validatorsManager.StartTracking(vdrIDs); err != nil {
-		return fmt.Errorf("failed to start tracking uptime: %w", err)
+	// Start the validators manager
+	if err := vm.validatorsManager.Initialize(ctx); err != nil {
```
ceyonur, line 724: this will hold the lock in `SetState` path (when bootstrapped).

darioush (approved):

StephenButtolph (approved):
Looks correct afaict. Although mixing methods that assume the lock is held with methods that will grab the lock internally seems suboptimal in the long run.

ceyonur:
> Looks correct afaict. Although mixing methods that assume the lock is held with methods that will grab the lock internally seems suboptimal in the long run.

I will work on a bigger refactor to cleanup interfaces and lock management.

PR #1469, opened 2025-03-05
Simplify avalanchego version maintenance

maru-ava:
## Why this should be merged

Previously it was necessary to manually maintain the avalanchego version in versions.sh. Now the version is automatically detected via go modules.

## How this works

 - updates versions.sh to automatically determine the version/SHA from go module

## How this was tested

- CI
- Manually checked the version used matched the go module version when invoking `bash -x scripts/build.sh`

## Need to be documented?

N/A

N/A

### DIFF
--- a/scripts/versions.sh
+++ b/scripts/versions.sh
@@ -4,5 +4,25 @@
 # shellcheck disable=SC2034
 
 # Don't export them as they're used in the context of other calls
-AVALANCHE_VERSION=${AVALANCHE_VERSION:-'f1ec9a13'}
+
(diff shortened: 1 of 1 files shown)


darioush (approved):
seems fine, can we apply this to coreth as well?

darioush:
I think we will want to be able to pass an optional AVALANCHE_VERSION in the docker build job if we do this.
Just for the other incident I was able to set AVALANCHE_VERSION and build the docker image with the fixed avalanchego but it wouldn't be possible if we merged this PR. (cc @maru-ava)

ceyonur (approved):
huge thanks 💯

PR #1473, opened 2025-03-05
eliminate an impossible error case from state.WriteState

tsachiherman:
## Why this should be merged
This PR removes an impossible error case from state.WriteState.

## How this works
The switch case was covering three states:
```golang
switch variable ( of type bool ) {
case false:
   ...
case true:
   ...
default:
   <generate error>
```
Given that a boolean can be either true or false, and nothing else, the default path is not possible.

## How this was tested
By the virtue of the error, this code change is not testable.

## Need to be documented?
no.

no.

### DIFF
--- a/plugin/evm/validators/state/state.go
+++ b/plugin/evm/validators/state/state.go
@@ -203,7 +203,7 @@ func (s *state) WriteState() error {
 		return err
 	}
 	// we've successfully flushed the updates, clear the updated marker.
-	s.updatedData = make(map[ids.ID]dbUpdateStatus)
+	clear(s.updatedData)
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


plugin/evm/validators/state/state.go
```diff
@@ -197,13 +197,14 @@ func (s *state) WriteState() error {
...
 		}
-		// we're done, remove the updated marker
-		delete(s.updatedData, vID)
 	}
-	return batch.Write()
+	if err := batch.Write(); err != nil {
+		return err
+	}
+	// we've successfully flushed the updates, clear the updated marker.
+	s.updatedData = make(map[ids.ID]dbUpdateStatus)
```
darioush, line 206: possible alternative:

```suggestion
	clear(s.updatedData)
```

darioush (approved):

PR #1479, opened 2025-03-08
refactor: use t.TempDir() instead of os.MkdirTemp

yuhangcangqian:
## Why this should be merged

 TempDir returns a temporary directory for the test to use.The directory is automatically removed when the test and
all its subtests complete. More info  https://pkg.go.dev/testing#B.TempDir


qdm12:
Thanks for the PR! 👍 
Can you instead please fix it in [go-ethereum in this file](https://github.com/ethereum/go-ethereum/blob/master/core/txpool/blobpool/blobpool_test.go) (exact same lines/issues), and this should be ported downstream to subnet-evm later 😉 
I'll close this PR, thank you for your understanding!

PR #1481, opened 2025-03-09
refactor: replace golang.org/x/exp with stdlib

wangjingcun:
## Why this should be merged

Since Go 1.21, the functions in x/exp used here can already be replaced by functions from the standard library

### DIFF
--- a/accounts/keystore/account_cache.go
+++ b/accounts/keystore/account_cache.go
@@ -32,6 +32,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"slices"
 	"sort"
 	"strings"
 	"sync"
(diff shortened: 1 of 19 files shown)


qdm12:
Thanks for the pull request! 🎖️ 
We currently have the code in subnet-evm based off [geth v1.13.14](https://github.com/ethereum/go-ethereum/releases/tag/v1.13.14) which contains these imports using the exp packages.
However, newer releases of geth contain the same changes as the proposed changes (only use  non-exp imports) so these will be picked up automatically when we upgrade subnet-evm with the last geth release, and we want to keep differences with geth v1.13.14 to a minimum to have a smoother upgrade. I'll thus close this, but we appreciate the effort! 👍

PR #1493, opened 2025-03-12
TargetBlockRate as milliseconds

qdm12:
## Why this should be merged

Do not merge.

Just to try having sub-second targetblockrate.


qdm12:
Closing this since it's largely untested and a very rough draft.

PR #1498, opened 2025-03-13
add fortuna default as nil

ceyonur:
## Why this should be merged

Fortuna is currently activated on Fuji. We have no corresponding Subnet-EVM version as Fortuna have no effect on subnet-evm networks. Therefore we should keep the default as nil so that it won't break when nodes update to any newer Subnet-EVM version. 

Fortuna is still kept as a defined upgrade in Subnet-EVM to reduce diff between Coreth. 

## How this works

* Adds a nil default for Fortuna
* returns nil error if default ts is nil in `verifyWithDefault`.

## How this was tested

Added UT

## Need to be documented?

Not probably

Updated

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -5,6 +5,7 @@
 * - IMPORTANT: `eth_getProof` calls for historical state will be rejected by default.
   - On archive nodes (`"pruning-enabled": false`): queries for historical proofs for state older than approximately 24 hours preceding the last accepted block will be rejected by default. This can be adjusted with the new option `historical-proof-query-window` which defines the number of blocks before the last accepted block which should be accepted for state proof queries, or set to `0` to accept any block number state query (previous behavior).
   - On `pruning` nodes: queries for proofs past the tip buffer (32 blocks) will be rejected. This is in support of moving to a path based storage scheme, which does not support historical state proofs.
+* Added `Fortuna` upgrade as optional and defaulting to a `nil` timestamp.
 
 ## [v0.7.2](https://github.com/ava-labs/subnet-evm/releases/tag/v0.7.2)
 
--- a/params/network_upgrades.go
+++ b/params/network_upgrades.go
@@ -171,7 +171,7 @@ func getDefaultNetworkUpgrades(agoUpgrade upgrade.Config) NetworkUpgrades {
 		SubnetEVMTimestamp: utils.NewUint64(0),
 		DurangoTimestamp:   utils.TimeToNewUint64(agoUpgrade.DurangoTime),
 		EtnaTimestamp:      utils.TimeToNewUint64(agoUpgrade.EtnaTime),
-		FortunaTimestamp:   nil, // Fortuna is optional and have no effect on Subnet-EVM
+		FortunaTimestamp:   nil, // Fortuna is optional and has no effect on Subnet-EVM
 	}
 }
 
--- a/params/network_upgrades.go
+++ b/params/network_upgrades.go
@@ -15,6 +15,7 @@ var errCannotBeNil = fmt.Errorf("timestamp cannot be nil")
 
 // NetworkUpgrades contains timestamps that enable network upgrades.
 // Avalanche specific network upgrades are also included here.
+// (nil = no fork, 0 = already activated)
 type NetworkUpgrades struct {
 	// SubnetEVMTimestamp is a placeholder that activates Avalanche Upgrades prior to ApricotPhase6
 	SubnetEVMTimestamp *uint64 `json:"subnetEVMTimestamp,omitempty"`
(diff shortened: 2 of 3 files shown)


darioush (approved):
please fix the fuji thing before merge.

(completely optional) change the test names to include "_" instead of ` ` while we're modifying this

params/network_upgrades_test.go
```diff
@@ -187,78 +215,89 @@ func TestVerifyNetworkUpgrades(t *testing.T) {
...
 		},
 		{
 			name: "Valid Fuji Durango reconfigured to Mainnet",
 			upgrades: &NetworkUpgrades{
 				SubnetEVMTimestamp: utils.NewUint64(0),
 				DurangoTimestamp:   utils.TimeToNewUint64(upgrade.GetConfig(constants.MainnetID).DurangoTime),
 			},
-			avagoUpgrades: upgrade.GetConfig(constants.FujiID),
-			expected:      false,
+			avagoUpgrades: upgrade.Mainnet,
```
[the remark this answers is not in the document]
ceyonur: fat fingered that, fixed it now.

RELEASES.md
```diff
@@ -5,6 +5,7 @@
 * - IMPORTANT: `eth_getProof` calls for historical state will be rejected by default.
   - On archive nodes (`"pruning-enabled": false`): queries for historical proofs for state older than approximately 24 hours preceding the last accepted block will be rejected by default. This can be adjusted with the new option `historical-proof-query-window` which defines the number of blocks before the last accepted block which should be accepted for state proof queries, or set to `0` to accept any block number state query (previous behavior).
   - On `pruning` nodes: queries for proofs past the tip buffer (32 blocks) will be rejected. This is in support of moving to a path based storage scheme, which does not support historical state proofs.
+* Added `Fortuna` upgrade as optional and defaulting to a `nil` timestamp.
```
qdm12, line 8: Should we mention this anyway, given

> Fortuna have no effect on subnet-evm networks

This does feel like maintenance/code-similarity-for-coreth and I don't think users would care about it really?

params/network_upgrades.go
```diff
@@ -156,18 +165,22 @@ func (n *NetworkUpgrades) GetAvalancheRules(time uint64) AvalancheRules {
...
 // getDefaultNetworkUpgrades returns the network upgrades for the specified avalanchego upgrades.
-// These should not return nil values.
+// Nil values are used to indicate optional upgrades.
 func getDefaultNetworkUpgrades(agoUpgrade upgrade.Config) NetworkUpgrades {
 	return NetworkUpgrades{
 		SubnetEVMTimestamp: utils.NewUint64(0),
 		DurangoTimestamp:   utils.TimeToNewUint64(agoUpgrade.DurangoTime),
 		EtnaTimestamp:      utils.TimeToNewUint64(agoUpgrade.EtnaTime),
-		FortunaTimestamp:   utils.TimeToNewUint64(agoUpgrade.FortunaTime),
+		FortunaTimestamp:   nil, // Fortuna is optional and have no effect on Subnet-EVM
```
qdm12, line 174:
```suggestion
		FortunaTimestamp:   nil, // Fortuna is optional and has no effect on Subnet-EVM
```

params/network_upgrades.go
```diff
@@ -156,18 +165,22 @@ func (n *NetworkUpgrades) GetAvalancheRules(time uint64) AvalancheRules {
...
 	}
 }
 
 // verifyWithDefault checks that the provided timestamp is greater than or equal to the default timestamp.
 func verifyWithDefault(configTimestamp *uint64, defaultTimestamp *uint64) error {
+	if defaultTimestamp == nil {
+		return nil
+	}
+
```
qdm12, line 183: supernit maybe move this below the configTimestamp nil check so it matches arguments order

params/network_upgrades_test.go
```diff
@@ -177,88 +205,99 @@ func TestVerifyNetworkUpgrades(t *testing.T) {
...
 				DurangoTimestamp:   utils.TimeToNewUint64(upgrade.GetConfig(constants.MainnetID).DurangoTime),
 			},
-			avagoUpgrades: upgrade.GetConfig(constants.FujiID),
-			expected:      false,
+			avagoUpgrades: upgrade.Fuji,
+			valid:         false,
 		},
 		{
-			name: "Invalid Etna nil",
+			name: "Invalid_Etna_nil",
```
qdm12, line 266: nit (no action to take) That's a nice addition to use underscores, although a bit out of scope; we can keep it now that it's done, but I think we should address this and similar issues at repo-wide scale after the repo-merge so it's done everywhere, since this likely applies to a lot of test files.

qdm12 (approved):

qdm12:
Make sure to squash and merge this PR into merge-coreth-master-clean 😉

RELEASES.md
ceyonur: I think we still want to mention this as they will see it startup messages

params/network_upgrades.go
ceyonur: Ordering matters here since both values are nil.

RELEASES.md
ceyonur: I will put up a more descriptive note once we release a new version after avago mainnet release

PR #1502, opened 2025-03-18
update libevm branch with coreth's libevm branch commits

qdm12:

PR #1504, opened 2025-03-18
fix extra state to noop rlp

darioush:


qdm12:
Incorporated in #1502 commit "chore(core/types): header JSON and RLP serialization hooks" (191be9a597 at the time of this comment, but will get overwritten later)

PR #1507, opened 2025-03-21
[ci] Use SHAs instead of tags for 3rd-party github actions

maru-ava:
## Why this should be merged

We should be able to trust that built-in actions (actions/*) have stable tags, but for everything else a SHA should be preferred to a tag to avoid the possibility of a supply-chain attack.

## How this works

Updates 3rd party github action references to use SHAs.

## How this was tested

CI

## Need to be documented?

N/A

## Need to be documented in RELEASES.md?

N/A

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -44,7 +44,7 @@ jobs:
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@89c54ab98078857ad0c420448dcbd03df3e1d014 #v3
+      uses: github/codeql-action/init@1b549b9259bda1cb5ddde3b41741a82a2d15a841 #v3.28.13
       with:
         languages: ${{ matrix.language }}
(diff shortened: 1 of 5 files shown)


qdm12 (approved):
A problem with this is we won't get bugfixes automatically.
Is this also needed versus pinning to bugfix versions i.e. `v2.1.2`? Doesn't Github have some security in place to prevent an action from being re-tagged? 🤔

ARR4N (approved):

ceyonur (approved):

.github/dependabot.yml
```diff
@@ -9,3 +9,7 @@ updates:
     directory: "/" # Location of package manifests
     schedule:
       interval: "daily"
+  - package-ecosystem: github-actions
+    directory: "/"
+    schedule:
+      interval: weekly
```
qdm12, line 15: Will this do anything?
Given we pin versions using hashes, either it won't do anything or it will spam with PR updates using vx.y.z versioning I suppose? Or does it support bumping up using hashes as well??

PR #1509, opened 2025-03-21
don't assume rpc format on the plugin evm client

felipemadero:
## Why this should be merged
Currently, NewClient on plugin/evm/client/client.go assumes the RPC is of the form: %s/ext/bc/%s, requiring the user
to provide an uri and a blockchain ID. Instead, it should just ask for plain rpc.


plugin/evm/client/client.go
```diff
@@ -48,14 +47,13 @@ type client struct {
 }
 
 // NewClient returns a Client for interacting with EVM [chain]
-func NewClient(uri, chain string) Client {
-	requestUri := fmt.Sprintf("%s/ext/bc/%s", uri, chain)
+func NewClient(url string) Client {
 	return &client{
 		adminRequester: rpc.NewEndpointRequester(
-			requestUri + "/admin",
+			url + "/admin",
```
ceyonur, line 53: can we keep the old `NewClient` and have this as a new method to ensure backwards-compatibility? I.e:
`NewClientWithURL(url string)` and then the `NewClient` can call the `NewClientWithURL` with 
```
requestUri := fmt.Sprintf("%s/ext/bc/%s", uri, chain)
return NewClientWithURL(requestUri)
```

ceyonur (approved):

PR #1510, opened 2025-03-23
Switch from `go install` to `go run` to avoid cluttering shared bin path

maru-ava:
## Why this should be merged

Using `go run` invokes the chosen command without the side-effect of installing tooling dependencies to the global $GOPATH/bin path. This attempts to ensure a degree of tooling isolation between projects.

## How this was tested

CI

## Need to be documented?

N/A

N/A

### DIFF
--- a/bin/ginkgo
+++ b/bin/ginkgo
@@ -1,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+# Ensure the go command is run from the root of the repository so that its go.mod file is used
+REPO_ROOT=$(cd "$( dirname "${BASH_SOURCE[0]}" )"; cd .. && pwd )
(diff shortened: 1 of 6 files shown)


bin/ginkgo
ceyonur, line 1: does it make sense to rename this to `run_ginkgo.sh` or such and keep it under scripts folder? This might be mistaken as if we're putting ginkgo binary directly here.

ceyonur (commented):
just a slight folder/script name concern/suggest other than LGTM.

bin/ginkgo
qdm12: Why not scripts/bin perhaps then? 🤔 As in, finding shell scripts in a bin/ directory is a bit out of the ordinary I would say

qdm12 (approved):

PR #1527, opened 2025-03-27
chore(github): dependabot ignore some actions and Go modules

qdm12:
## Why this should be merged

Dependabot currently spams us with Go modules upgrades, which we need to handle manually/through a geth sync anyway, as well as with incompatible Github actions updates.

## How this works

- Ignore specific versions of golangci-lint-action versions only supporting golangci-lint v2.x.x which we can't support
- Ignore avalanchego Github action updates (do this manually)
- Ignore avalanchego and libevm Go module updates (do this manually instead)
- Ignore untouched geth Go modules to stay aligned with geth code


.github/dependabot.yml
```diff
@@ -5,7 +5,7 @@
 
 version: 2
 updates:
-  - package-ecosystem: "gomod" # See documentation for possible values
+  - package-ecosystem: "github-actions" # See documentation for possible values
```
ceyonur, line 8: I think we now have 2 github actions. I think we still want to know about critical updates/vulns in go packages, even if we cannot immediately update with dependabot.

We also have a NPM package in this repo, which we should probably remove before merging repos.

.github/dependabot.yml
```diff
@@ -9,7 +9,27 @@ updates:
...
+      - dependency-name: "github.com/tyler-smith/go-bip39" # geth dependency
+      - dependency-name: "github.com/urfave/cli/v2" # geth dependency
+      - dependency-name: "golang.org/x/time" # geth dependency
+      - dependency-name: "gopkg.in/natefinch/lumberjack.v2" # geth dependency
   - package-ecosystem: github-actions
     directory: "/"
     schedule:
       interval: weekly
+    ignore:
+      - dependency-name: "ava-labs/avalanchego" # manually updated
```
michaelkaplan13, line 32: Just out of curiousity, what Github actions are sourced from AvalancheGo that dependabot picks up on?

michaelkaplan13 (approved):

.github/dependabot.yml
qdm12: For now it's only `ava-labs/avalanchego/.github/actions/run-monitored-tmpnet-cmd` which is enforced by CI to match the Go.mod avalanchego version. However, it is true we could add more actions that don't need to match the go.mod avalanchego version further down the line, so I changed this to be an exact match `ava-labs/avalanchego/.github/actions/run-monitored-tmpnet-cmd`

PR #1528, opened 2025-03-27
chore: fix problematic comment

evenevent:
## Why this should be merged

fix problematic comment

### DIFF
--- a/accounts/abi/argument.go
+++ b/accounts/abi/argument.go
@@ -137,7 +137,7 @@ func (arguments Arguments) Copy(v interface{}, values []interface{}) error {
 	return arguments.copyAtomic(v, values[0])
 }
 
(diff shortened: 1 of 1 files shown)


evenevent:
> Thank you for the PR! Unfortunately, this is a comment error coming from geth v1.13.14 ([that line](https://github.com/ethereum/go-ethereum/blob/2bd6bd01d2e8561dd7fc21b631f4a34ac16627a1/accounts/abi/argument.go#L130)) and is [already fixed on their master branch](https://github.com/ethereum/go-ethereum/blob/6143c350ae1ecf3330678be02b4c2745bb6b8134/accounts/abi/argument.go#L130), which we will eventually sync with automatically.
> 
> Closing this, but we appreciate the effort! 👍 💯 🎖️

Thank you for your patient explanation!

PR #1530, opened 2025-03-27
New Docker image name and tag

martineckardt:
## Why this should be merged

The docker image contains the binary for AvalancheGo and Subnet-EVM. To make this more clear to the developer the image name will be changed to `avalanchego-subnet-evm`. The tag will now include both versions/commits.

## How this works

Change image name in github workflow and create combined version tag in `builder_docker_image.sh`.

## How this was tested

TBD

## Need to be documented?

Will be updated in the Avalanche Builder tooling

No


.github/workflows/publish_docker.yml
```diff
@@ -31,7 +31,7 @@ jobs:
         env:
           DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
           DOCKER_PASS: ${{ secrets.DOCKER_PASS }}
-          IMAGE_NAME: "avaplatform/subnet-evm"
+          IMAGE_NAME: "avaplatform/avalanchego-subnet-evm"
```
darioush, line 34: pretty sure we rely on this in much of the infra, better check with infra first.

ceyonur: we have checked this with infra, and seems there is no blocker as they manage their own images.

michaelkaplan13:
Closing in favor of https://github.com/ava-labs/subnet-evm/pull/1566

PR #1551, opened 2025-04-03
fix(all): critical fixes to the libevm branch

qdm12:
## Why this should be merged

Critical fixes **only** from review on #1501 

Minor changes can be applied in #1552

### DIFF
--- a/core/genesis.go
+++ b/core/genesis.go
@@ -184,6 +184,12 @@ func SetupGenesisBlock(
 		customrawdb.WriteChainConfig(db, stored, newcfg)
 		return newcfg, stored, nil
 	}
+
+	// Notes on the following line:
+	// - this is needed in coreth to handle the case where existing nodes do not
+	//   have the Berlin or London forks initialized by block number on disk.
+	//   See https://github.com/ava-labs/coreth/pull/667/files
+	// - this is not needed in subnet-evm but it does not impact it either
 	params.SetEthUpgrades(storedcfg, params.GetExtra(storedcfg).NetworkUpgrades)
 	// Check config compatibility and write the config. Compatibility errors
 	// are returned to the caller unless we're already at block zero.
--- a/core/genesis.go
+++ b/core/genesis.go
@@ -184,6 +184,7 @@ func SetupGenesisBlock(
 		customrawdb.WriteChainConfig(db, stored, newcfg)
 		return newcfg, stored, nil
 	}
+	params.SetEthUpgrades(storedcfg, params.GetExtra(storedcfg).NetworkUpgrades)
 	// Check config compatibility and write the config. Compatibility errors
 	// are returned to the caller unless we're already at block zero.
 	// we use last accepted block for cfg compatibility check. Note this allows
--- a/params/extras/config.go
+++ b/params/extras/config.go
@@ -176,6 +176,13 @@ func (c *ChainConfig) Description() string {
 	}
 	banner += fmt.Sprintf("Upgrade Config: %s", string(upgradeConfigBytes))
 	banner += "\n"
+
+	feeBytes, err := json.Marshal(c.FeeConfig)
+	if err != nil {
+		feeBytes = []byte("cannot marshal FeeConfig")
+	}
+	banner += fmt.Sprintf("Fee Config: %s\n", string(feeBytes))
+
 	return banner
 }
 
--- a/params/extras/config.go
+++ b/params/extras/config.go
@@ -318,6 +318,10 @@ func checkForks(forks []fork, blockFork bool) error {
 
 // Verify verifies chain config.
 func (c *ChainConfig) Verify() error {
+	if err := c.FeeConfig.Verify(); err != nil {
+		return fmt.Errorf("invalid fee config: %w", err)
+	}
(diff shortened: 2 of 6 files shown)


ARR4N (changes requested):
If these are critical then they need tests, specifically ones that:

1. Confirm that the ETH upgrades have been applied; and
2. Verify a ChainConfig with invalid FeeConfig.

core/genesis.go
```diff
@@ -184,6 +184,7 @@ func SetupGenesisBlock(
 		customrawdb.WriteChainConfig(db, stored, newcfg)
 		return newcfg, stored, nil
 	}
+	params.SetEthUpgrades(storedcfg, params.GetExtra(storedcfg).NetworkUpgrades)
```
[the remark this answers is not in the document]
ceyonur: @darioush can we confirm that. It looks like either we accidentally deleted it in subnet-evm or we forgot to delete in coreth.

params/extras/config.go
```diff
@@ -176,6 +176,15 @@ func (c *ChainConfig) Description() string {
...
 	banner += fmt.Sprintf("Upgrade Config: %s", string(upgradeConfigBytes))
 	banner += "\n"
+
+	feeBytes, err := json.Marshal(c.FeeConfig)
+	if err != nil {
+		feeBytes = []byte("cannot marshal FeeConfig")
+	}
+	banner += fmt.Sprintf("Fee Config: %s\n", string(feeBytes))
+
+	banner += fmt.Sprintf("Allow Fee Recipients: %v\n", c.AllowFeeRecipients)
```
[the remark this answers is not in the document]
ceyonur: yes

core/genesis.go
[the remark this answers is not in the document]
darioush: This was added to coreth in this context: https://github.com/ava-labs/coreth/pull/667/files
Specifically to handle the case with existing nodes that do not have the Berlin/London forks initialized by number on disk.
It is not needed in subnet-evm afaik but it should also do no harm (eg, to keep diffs minimal).
Probably worth adding a comment about this context to coreth side at least.

qdm12:
Tests added:

- ChainConfig Description() test `TestChainConfigDescription`
- ChainConfig Verify test `TestChainConfigVerify` (Verify coverage, does not test in-depth unlike other existing tests such as precompile ones)
- Genesis params.SetEthUpgrades test stolen from coreth `TestGenesisEthUpgrades` and adapted with small changes, with the aim of having small diffs between the 2 repositories.

alarso16 (approved):

params/extras/config_test.go
```diff
@@ -0,0 +1,216 @@
...
+	"github.com/ava-labs/libevm/common"
+	"github.com/ava-labs/subnet-evm/commontype"
+	"github.com/ava-labs/subnet-evm/precompile/contracts/txallowlist"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func pointer[T any](v T) *T { return &v }
+
+func TestChainConfigDescription(t *testing.T) {
```
ARR4N, line 22: This is just a change-detector test that IMO not only adds no value, results in maintenance burden because any tiny change to the hook just has to be mirrored here. If it's absolutely necessary to check that the hook is working (I'm skeptical but open to convincing) then just test for the presence of a substring like "SubnetEVM".

params/extras/config_test.go
```diff
@@ -0,0 +1,216 @@
...
+					GasLimit: nil,
+				},
+			},
+			errRegex: "^invalid fee config: ",
+		},
+		"invalid_precompile_upgrades": {
+			// Also see precompile_config_test.go TestVerifyWithChainConfig* tests
+			config: ChainConfig{
+				FeeConfig: commontype.FeeConfig{
+					GasLimit:                 big.NewInt(1),
```
ARR4N, line 118: What is the significance of all these incrementing values? It suggests that they're deliberate but it's not obvious why. I also see that you're using the same thing in every test case—if it's just because they're valid then abstract a `validFeeConfig` variable and reuse it in all. A reader needs to be able to pinpoint the key information in a test without any distraction.

ARR4N (approved):
Approving with the caveat of the comments I've made. This will still need Cey's review as a code owner.

params/extras/config_test.go
qdm12: We're not really testing FeeConfig in particular, and there is no hook here.
However, the part checking the network upgrades will have to be changed everytime a new upgrade timestamp is set in the `NetworkUpgrades` struct, so that's a change detector issue: test was changed to use a regex instead with N timestamp lines, so that would adapt to future timestamps.

The point of the test is to detect a missing piece in the description method (such as the feeconfig); so we need to check *most* of the string, not just that it contains 1 element, such that this won't happen for other fields used in the description.

params/extras/config_test.go
qdm12: Good point, changed them all to 1 in the feeconfig, and set a `validFeeConfig` variable as suggested.

ceyonur (approved):

PR #1552, opened 2025-04-03
chore(all): minor changes post-libevm merge

qdm12:
## Why this should be merged

Minor changes suggested from the review of #1501


core/extstate/statedb.go
```diff
@@ -62,7 +62,7 @@ func (s *StateDB) GetLogData() (topics [][]common.Hash, data [][]byte) {
 // GetPredicateStorageSlots(AddrA, 0) -> Predicate1
 // GetPredicateStorageSlots(AddrB, 0) -> Predicate2
 // GetPredicateStorageSlots(AddrA, 1) -> Predicate3
-func (s *StateDB) GetPredicateStorageSlots(address common.Address, index int) ([]byte, bool) {
+func (s *StateDB) GetPredicateStorageSlots(address common.Address, index int) (predicate []byte, exists bool) {
```
alarso16, line 65: I don't like these named returns - what's the point if we don't return them?

qdm12: Documentation; it's not really clear what the `bool` would be from `GetPredicateStorageSlots` from a consumer of that function (without looking at its implementation).

If the function is `IsLast() bool` it's obvious what the return value is so no need for a named return, but in some cases it makes sense when it's not that clear imo.

alarso16 (approved):
If you decide to resolve use libevm for the other params in a different PR, looks good. But seems easier to do here

params/extras/protocol_params.go
```diff
@@ -1,11 +0,0 @@
...
-// See the file LICENSE for licensing terms.
-
-package extras
-
-const (
-	GasLimitBoundDivisor uint64 = 1024               // The bound divisor of the gas limit, used in update calculations.
-	MaxGasLimit          uint64 = 0x7fffffffffffffff // Maximum the gas limit (2^63-1).
-	MaximumExtraDataSize uint64 = 64                 // Maximum size extra data may be after Genesis.
-	MinGasLimit          uint64 = 5000               // Minimum the gas limit may ever be.
-)
```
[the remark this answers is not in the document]
qdm12: Done in #1563 for all the protocol params, given it is significant changes still.

accounts/abi/bind/util_test.go
```diff
@@ -125,10 +125,7 @@ func TestWaitDeployedCornerCases(t *testing.T) {
 	// Create a transaction to an account.
 	code := "6060604052600a8060106000396000f360606040526008565b00"
 	tx := types.NewTransaction(0, common.HexToAddress("0x01"), big.NewInt(0), 3000000, gasPrice, common.FromHex(code))
-	tx, err := types.SignTx(tx, types.LatestSignerForChainID(big.NewInt(1337)), testKey)
-	if err != nil {
-		t.Fatalf("Failed to sign transaction: %s", err)
-	}
```
ceyonur, line 131: is this to reduce diffs? I'm pretty sure I've seen one of these unhandled errors caused trouble in tests. So I'd be willing to keep this

core/extstate/statedb.go
ceyonur: necro-posting this: can we rather keep this in comments and not use named returns?

core/state/statedb.go
```diff
@@ -118,19 +118,3 @@ func (s *StateDB) Copy() *StateDB {
...
 	}
 }
-
-// NormalizeCoinID ORs the 0th bit of the first byte in
-// `coinID`, which ensures this bit will be 1 and all other
-// bits are left the same.
-// This partitions multicoin storage from normal state storage.
-func NormalizeCoinID(coinID *common.Hash) {
-	coinID[0] |= 0x01
-}
```
ceyonur, line 128: 💣

accounts/abi/bind/util_test.go
qdm12: It's geth code so we should align with it.
In the libevm branch, we did change this to be an error check, I suggested to revert it since it was out of scope / no obvious reason, to which we agreed let's check the error it doesn't hurt - but I did not realise it was geth code. Anyway, I don't think this would fail really 🤷

core/block_validator_test.go
```diff
@@ -63,3 +66,8 @@ func TestCalcGasLimit(t *testing.T) {
 		}
 	}
 }
+
+func TestUpstreamParamsValues(t *testing.T) {
+	assert.Equal(t, uint64(1024), ethparams.GasLimitBoundDivisor, "gas limit bound divisor")
+	assert.Equal(t, uint64(5000), ethparams.MinGasLimit, "min gas limit")
+}
```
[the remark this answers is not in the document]
qdm12: This was meant to be removed, good catch! Removed!

precompile/contract/interfaces.go
```diff
@@ -40,7 +40,7 @@ type StateDB interface {
 
 	AddLog(*ethtypes.Log)
 	GetLogData() (topics [][]common.Hash, data [][]byte)
-	GetPredicateStorageSlots(address common.Address, index int) ([]byte, bool)
+	GetPredicateStorageSlots(address common.Address, index int) (predicate []byte, exists bool)
```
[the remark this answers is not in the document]
qdm12: Oh I misread your comment. Let's agree to agree then!

ceyonur (approved):

PR #1566, opened 2025-04-17
Docker image combined tag

ceyonur:
## Why this should be merged

Replaces #1530 

### Consolidation of versioning logic:
* Removed `scripts/versions.sh` and migrated its functionality to `scripts/constants.sh`, including logic for determining `AVALANCHE_VERSION` and `DOCKERHUB_TAG`.

### Docker image naming and tagging improvements:
* Updated `IMAGE_NAME` to `avaplatform/subnet-evm_avalanchego` 
* Adjusted Docker image tags in test scripts to align with the updated naming convention.

### Workflow updates:
* Introduced a `tag_latest` parameter to `workflow_dispatch` trigger to conditionally tag Docker images as `latest`, which by default should be `true` in `push` events. 

## How this was tested

locally created images  + existing UT should cover

image

## Need to be documented?

Yes

Yes


scripts/build_docker_image.sh
```diff
@@ -70,9 +66,6 @@ else
 fi
 
 VM_ID=${VM_ID:-"${DEFAULT_VM_ID}"}
-if [[ "${VM_ID}" != "${DEFAULT_VM_ID}" ]]; then
-  DOCKERHUB_TAG="${VM_ID}-${DOCKERHUB_TAG}"
```
ceyonur, line 74: not sure if this was needed at all.

scripts/constants.sh
```diff
@@ -14,7 +14,7 @@ DEFAULT_VM_ID="srEXiWaHuhNyGwPUi444Tu47ZEDwxTWrbQiuD7FmgSAQ6X7Dy"
 # Avalabs docker hub
 # avaplatform/avalanchego - defaults to local as to avoid unintentional pushes
 # You should probably set it - export IMAGE_NAME='avaplatform/subnet-evm'
-IMAGE_NAME=${IMAGE_NAME:-"subnet-evm"}
+IMAGE_NAME=${IMAGE_NAME:-"avalanchego_subnet-evm"}
```
ceyonur, line 17: `_` seems better than `-` as we already have it in the name `subnet-evm` i

scripts/constants.sh
```diff
@@ -14,7 +14,7 @@ DEFAULT_VM_ID="srEXiWaHuhNyGwPUi444Tu47ZEDwxTWrbQiuD7FmgSAQ6X7Dy"
 # Avalabs docker hub
 # avaplatform/avalanchego - defaults to local as to avoid unintentional pushes
 # You should probably set it - export IMAGE_NAME='avaplatform/subnet-evm'
-IMAGE_NAME=${IMAGE_NAME:-"subnet-evm"}
+IMAGE_NAME=${IMAGE_NAME:-"subnet-evm_avalanchego"}
```
ceyonur, line 17: well I think subnet_evm should come before avalanchego in case we publish same subnet-evm version under different avalanchego versions. Just felt it could be better to reorder/display in docker hub. We already too many components starting with avalanchego. cc @martineckardt

martineckardt: Works for me 👍🏻

scripts/tests.build_docker_image.sh
```diff
@@ -61,7 +58,7 @@ build_and_test() {
 VM_ID="${VM_ID:-${DEFAULT_VM_ID}}"
 
 echo "checking build of single-arch image"
-build_and_test "subnet-evm" "${VM_ID}" false "avalanchego"
+build_and_test "avalanchego-subnet-evm" "${VM_ID}" false "avalanchego"
```
martineckardt, line 61: @ceyonur do we need to change the order here to?

scripts/tests.build_docker_image.sh
```diff
@@ -81,4 +78,4 @@ function cleanup {
 trap cleanup EXIT
 
 echo "checking build of multi-arch images"
-build_and_test "localhost:${REGISTRY_PORT}/subnet-evm" "${VM_ID}" true "localhost:${REGISTRY_PORT}/avalanchego"
+build_and_test "localhost:${REGISTRY_PORT}/avalanchego-subnet-evm" "${VM_ID}" true "localhost:${REGISTRY_PORT}/avalanchego"
```
martineckardt, line 81: same here

scripts/tests.build_docker_image.sh
ceyonur: actually tests pass as they're but yea let's keep it same for simplicity. thanks.

RELEASES.md
```diff
@@ -2,10 +2,44 @@
...
+
+- plugin/evm: add local API docs (#1513)
+- 2023 Least Authority Audit (#1499)
+- precompile/contracts/warp:
+  - changes for 9000 upgrade (#1494)
+  - differentiate between L1 and Subnet self-signing optimization cases (#1529)
+- simulator: fix wrong flag name `staking-enabled` -> `--sybil-protection-enabled=false` (#1476)
+- typo fixes (#1491, #1506)
+
+**Full Changelog**: <https://github.com/ava-labs/subnet-evm/compare/v0.7.2...v0.7.3>
```
michaelkaplan13, line 42: A bit confused why these notes are in this PR..should they have been included when we released v0.7.3 a few weeks ago?

RELEASES.md
```diff
@@ -2,10 +2,44 @@
 
 ## Pending Release
 
-* * IMPORTANT: `eth_getProof` calls for historical state will be rejected by default.
-  * On archive nodes (`"pruning-enabled": false`): queries for historical proofs for state older than approximately 24 hours preceding the last accepted block will be rejected by default. This can be adjusted with the new option `historical-proof-query-window` which defines the number of blocks before the last accepted block which should be accepted for state proof queries, or set to `0` to accept any block number state query (previous behavior).
-  * On `pruning` nodes: queries for proofs past the tip buffer (32 blocks) will be rejected. This is in support of moving to a path based storage scheme, which does not support historical state proofs.
-* Added `Fortuna` upgrade as optional and defaulting to a `nil` timestamp.
+- Updated dockerhub image name to `avaplatform/avalanchego_subnet-evm` and tags to accommodate the new versioning scheme: {avalanchego version}_{subnet-evm version}
```
michaelkaplan13, line 5: Looks like this line needs updating to match the naming convention

.github/workflows/publish_docker.yml
```diff
@@ -4,15 +4,20 @@ on:
...
         type: string
       avalanche_version:
-        description: 'The SHA or tag of avalanchego to use for the base image (must be compatible with the version in go.mod)'
-        default: ''
+        description: "The SHA or tag of avalanchego to use for the base image (must be compatible with the version in go.mod)"
+        default: ""
         required: false
         type: string
+      tag_latest:
+        description: "Whether or not to tag the image with the latest"
```
michaelkaplan13, line 17:
```suggestion
        description: "Whether or not to tag the image created as latest"
```

scripts/update_avalanchego_version.sh
```diff
@@ -7,6 +7,11 @@ if ! [[ "$0" =~ scripts/update_avalanchego_version.sh ]]; then
   exit 255
 fi
 
+SUBNET_EVM_PATH=$(
```
michaelkaplan13, line 10: Looks like this is unused currently, did you mean to use it to run `scripts/constants.sh`?

scripts/build_docker_image.sh
```diff
@@ -112,7 +105,7 @@ ${DOCKER_CMD} -t "$IMAGE_NAME:$BUILD_IMAGE_ID" -t "$IMAGE_NAME:${DOCKERHUB_TAG}"
   --build-arg CURRENT_BRANCH="$CURRENT_BRANCH" \
   --build-arg VM_ID="$VM_ID"
 
-if [[ -n "${PUBLISH}" && $CURRENT_BRANCH == "master" ]]; then
+if [[ -n "${PUBLISH}" && $CURRENT_BRANCH == "master" && $TAG_LATEST == true ]]; then
```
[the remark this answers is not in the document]
michaelkaplan13: I'm having difficultly following in exactly which cases an image will or won't be tagged as latest.

If the `publish_docker.yml` workflow is run, `TAG_LATEST` is `true` if it was either triggered by a new tag or push to `master`, or if it was manually set to true in a workflow dispatch. But then here, even if `TAG_LATEST` is true, we won't tag as latest unless the script is being run from a git repo on the `master` branch.

Is there a way we can simplify this so the conditions are defined in a single place, and we just check against `TAG_LATEST` here?

scripts/update_avalanchego_version.sh
ceyonur: yes, updated

RELEASES.md
ceyonur: they have been included, we just did not update them here. Maybe we should draft the release notes in this file before we actually cut and release the version. But the intention if this file is actually to keep pending changes for the next release, not to keep previous notes.

scripts/build_docker_image.sh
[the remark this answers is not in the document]
ceyonur: We can create images from the same commit in subnet-evm but with a different avalanchego versions, without this flag it can tag that with latest. in this case a "latest" can have an old avalanchego version which could be weird. so this one prevents accidentally tagging with latest. I can reword it, it should be `ALLOW_TAG_LATEST`, or really try to refactor the whole thing.

RELEASES.md
ceyonur: updated it

scripts/build_docker_image.sh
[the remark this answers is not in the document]
michaelkaplan13: Gotcha, so `latest` is true if:
- The image is built from the `master` branch AND
- `ALLOW_TAG_LATEST` is set to true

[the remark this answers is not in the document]
ceyonur: yes

.github/workflows/publish_docker.yml
```diff
@@ -4,15 +4,20 @@ on:
   workflow_dispatch:
     inputs:
       vm_id:
-        description: 'The ID of the VM (binary dst in Docker image)'
-        default: ''
+        description: "The ID of the VM (binary dst in Docker image)"
+        default: ""
```
qdm12, line 8: Side note: we should add a yml linter to have consistent yml formatting (I suppose this change is auto-formatting)

RELEASES.md
qdm12: > we have seen that some context (like a critical change to API/flags) was missing in the release notes

That's why we could have a draft release notes where everyone can update it before a release. A bit more decoupled from the pull request though.

> because it was not possibly find the relevant context during the release process

IMO that's a commit message/description issue then

> I don't think a draft release would be ideal to be used as a documenting place

That's perhaps subjective, from my point of view, I would say the RELEASES.md file would be more overlooked (like in that case here) than the github releases 😄

.github/workflows/publish_docker.yml
```diff
@@ -31,9 +36,9 @@ jobs:
         env:
           DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
           DOCKER_PASS: ${{ secrets.DOCKER_PASS }}
-          IMAGE_NAME: "avaplatform/subnet-evm"
           VM_ID: ${{ inputs.vm_id }}
           PUBLISH: 1
           PLATFORMS: "linux/amd64,linux/arm64"
           AVALANCHE_VERSION: ${{ inputs.avalanche_version }}
+          ALLOW_TAG_LATEST: ${{ github.event_name != 'workflow_dispatch' || inputs.allow_tag_latest }}
```
qdm12, line 43: nit Perhaps we could have this as `TAG_LATEST` with the expression:

```
${{ github.ref == format('refs/heads/{0}', github.event.repository.default_branch) && (github.event_name != 'workflow_dispatch' || inputs.allow_tag_latest) }}
```

And remove the master branch check from the script, since it would be done here?

scripts/build_docker_image.sh
qdm12: It looks like the vm id can be set by a dispatch event, maybe we should keep this?

scripts/constants.sh
```diff
@@ -34,17 +34,40 @@ else
...
+  # Extract the version part
+  AVALANCHE_VERSION="$(echo "${MODULE_DETAILS}" | awk '{print $2}')"
+
+  # Check if the version matches the pattern where the last part is the module hash
+  # v*YYYYMMDDHHMMSS-abcdef123456
+  #
+  # If not, the value is assumed to represent a tag
+  if [[ "${AVALANCHE_VERSION}" =~ ^v.*[0-9]{14}-[0-9a-f]{12}$ ]]; then
+    # Extract module hash from version
+    MODULE_HASH="$(echo "${AVALANCHE_VERSION}" | grep -Eo '[0-9a-f]{12}$')"
```
qdm12, line 51: nit unneeded comment

```suggestion
    MODULE_HASH="$(echo "${AVALANCHE_VERSION}" | grep -Eo '[0-9a-f]{12}$')"
```

scripts/constants.sh
```diff
@@ -34,17 +34,40 @@ else
     SUBNET_EVM_COMMIT="$(git --git-dir="$SUBNET_EVM_PATH/.git" rev-parse HEAD || :)"
 fi
 
+# Don't export them as they're used in the context of other calls
+if [[ -z ${AVALANCHE_VERSION:-} ]]; then
+  # Get module details from go.mod
+  MODULE_DETAILS="$(go list -m "github.com/ava-labs/avalanchego" 2>/dev/null)"
+
+  # Extract the version part
+  AVALANCHE_VERSION="$(echo "${MODULE_DETAILS}" | awk '{print $2}')"
```
qdm12, line 43: nit unneeded comment

```suggestion
  AVALANCHE_VERSION="$(echo "${MODULE_DETAILS}" | awk '{print $2}')"
```

scripts/constants.sh
```diff
@@ -34,17 +34,40 @@ else
     SUBNET_EVM_COMMIT="$(git --git-dir="$SUBNET_EVM_PATH/.git" rev-parse HEAD || :)"
 fi
 
+# Don't export them as they're used in the context of other calls
```
qdm12, line 37: I'm not too sure what this comment refers to, maybe mention what variables should not be exported? i.e. `AVALANCHE_VERSION`?

scripts/constants.sh
```diff
@@ -34,17 +34,40 @@ else
...
+
+  # Check if the version matches the pattern where the last part is the module hash
+  # v*YYYYMMDDHHMMSS-abcdef123456
+  #
+  # If not, the value is assumed to represent a tag
+  if [[ "${AVALANCHE_VERSION}" =~ ^v.*[0-9]{14}-[0-9a-f]{12}$ ]]; then
+    # Extract module hash from version
+    MODULE_HASH="$(echo "${AVALANCHE_VERSION}" | grep -Eo '[0-9a-f]{12}$')"
+
+    # The first 8 chars of the hash is used as the tag of avalanchego images
```
qdm12, line 53: I think that's what was meant, given the final tag is subnetevm8_avalanchego8?

```suggestion
    # The first 8 chars of the hash is used as the avalanchego part of the image tag
```

scripts/constants.sh
```diff
@@ -34,17 +34,40 @@ else
...
 
 # Static compilation
 STATIC_LD_FLAGS=''
 if [ "${STATIC_COMPILATION:-}" = 1 ]; then
-    export CC=musl-gcc
-    command -v $CC || (echo $CC must be available for static compilation && exit 1)
-    STATIC_LD_FLAGS=' -extldflags "-static" -linkmode external '
+  export CC=musl-gcc
+  command -v $CC || (echo $CC must be available for static compilation && exit 1)
+  STATIC_LD_FLAGS=' -extldflags "-static" -linkmode external '
```
qdm12, line 70: Question for future reference, what shell script formatting tool are you using? We could also use that to have consistent shell formatting and not have out of scope diffs changing the formatting of scripts (like here)

scripts/update_avalanchego_version.sh
```diff
@@ -17,7 +22,7 @@ if [[ -n "${VERSION}" ]]; then
 fi
 
 # Discover AVALANCHE_VERSION
-. scripts/versions.sh
+. "$SUBNET_EVM_PATH"/scripts/constants.sh
```
qdm12, line 25: (Apologies for my lack of shell knowledge!)

Couldn't this work as before with `. scripts/constants.sh`?

scripts/run_simulator.sh
```diff
@@ -11,34 +11,31 @@ if ! [[ "$0" =~ scripts/run_simulator.sh ]]; then
...
+  echo "running simulator from $PWD"
+  ./cmd/simulator/simulator \
+    --endpoints="$RPC_ENDPOINTS" \
+    --key-dir=./cmd/simulator/.simulator/keys \
+    --timeout=300s \
+    --workers=1 \
+    --txs-per-worker=50000 \
+    --batch-size=50000 \
+    --max-fee-cap=1000000 \
+    --max-tip-cap=10000
```
qdm12, line 38: nit ideally revert this, these are auto-formatting out of scope changes

.github/workflows/publish_docker.yml
ceyonur: sounds good

RELEASES.md
ceyonur: The problem here is not that we have forgotten to update the `RELEASES.md` in this PR, but we did not need to update it after the release and before this PR so the previous release was left under "Draft Release". I do think this `RELEASES.md` saves a ton of trouble in the release process and the one doing the release. Having a file make the changes reviewable/enforcable. We just need to improve our release process so we update this file + the actual version in `version.go` after cutting a new release.

scripts/constants.sh
ceyonur: I think it also refers to the pulled avalanchego image tag.

scripts/constants.sh
ceyonur: I frankly don't know why my editor keep changing these. I'm using VS code and dont have any special formatter (at least not for script files). This one looks like a tab vs spaces issue.

scripts/run_simulator.sh
ceyonur: I think this actually fixes the formatting (we used 2 spaces) but reverted it, but I agree we should probably find a linter plugin for these.

scripts/update_avalanchego_version.sh
ceyonur: I think that would work, but better to use the full path.

PR #1585, opened 2025-05-22
refactor: use the built-in max/min to simplify the code

knowmost:
## Why this should be merged

In Go 1.21, the standard library includes built-in [max/min](https://pkg.go.dev/builtin@go1.21.0#max) function, which can greatly simplify the code.


JonathanOppenheimer (approved):
Hey @knowmost - good catch! We try to avoid touching the original go-ethereum code unless absolutely necessary, but we are behind from them.

JonathanOppenheimer:
Hi @knowmost can you please sign your commit?

You can usually do it with `git commit -S --amend --no-edit && git push -f`, provided your git has a signing setup, see https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

Thanks!

Issue #977, opened 2023-11-01
Admin API duplicated register


ceyonur:
From: https://github.com/ava-labs/avalanchego/issues/2243

A developer is trying to run AvalancheGo 1.10.14 with SubnetEVM 0.5.8 and running into the following error:

```
ERROR server/server.go:185 failed to create handlers {"chainName": "guvaqv679hxAjtD3Xq9QPVCAtqHERdazVz7eb8dPJYnj36kaw", "error": "rpc error: code = Unknown desc = failed to register service for admin API due to rpc: service already defined: \"admin\""}
```

PR #1593, opened 2025-06-04
remove request parse check in network forwarding (#977)

ceyonur:
Cherry picks https://github.com/ava-labs/coreth/pull/977


peer/network_test.go
ARR4N, line 1: Given that 0 vs 1 has special meaning, I think it's worth defining constants with comments.

ARR4N (approved):

alarso16 (approved):

PR #1597, opened 2025-06-18
Update avalanchego version to v1.13.2-rc.0

StephenButtolph:
## Why this should be merged

Fixes the antithesis tests.

## How this works

Uses the new version of avalanchego which correctly builds the antithesis docker compose.

## How this was tested

N/A

## Need to be documented?

No


alarso16 (approved):
Seems good to me once you fix the interface issues

PR #1601, opened 2025-06-24
chore(plugin/evm/customtypes): use upstream RLPHash function (#918)

ceyonur:


ceyonur:
> Looks good to me (the only other changes I see between the hashes that aren't included in this PR is:
> 
>     * libevm update (already done)
> 
>     * coreth specific changes

coreth specific changes mostly consist of acp-176 or atomic operations related changes

PR #1607, opened 2025-06-24
Parse genesis

ceyonur:


params/config_extra.go
```diff
@@ -31,19 +35,51 @@ var (
 // TODO: Prior to Cancun, Avalanche upgrades are referenced inline in the
```
alarso16, line 35: This TODO seems like it's starting to be addressed, right?

alarso16 (approved):
fwict it looks good, but I think another review would be good

params/config_extra.go
ceyonur: yes

PR #1610, opened 2025-06-26
Move CI tasks to Taskfile.yml

JonathanOppenheimer:
## Why this should be merged

See issue [1023](https://github.com/ava-labs/coreth/issues/1023) in the coreth repository. @ceyonur requested that issue be addressed here too.

- _Enables both discoverability of available tasks (./scripts/run_task.sh without arguments outputs all tasks) and maximizes the potential for reproducing what runs in CI - it can be as simple as task [ci job]._
- Makes testing patterns consistent with the AvalancheGo repository and should reduce the complexity of the workflow. 
- Simplifies the `test-e2e-precompile` by creating individual test tasks and then creating a parent task that runs everything. 

## How this works
Reassigns how scripts are called instead of directly calling them in workflow files. 

## How this was tested
As this changed how the unit tests themselves are executed, I will let the automatic unit tests run 

## Need to be documented?
No 

No


Taskfile.yml
```diff
@@ -0,0 +1,89 @@
...
+# If in the nix dev shell, `task` is available.
+
+version: '3'
+
+tasks:
+  default: ./scripts/run_task.sh --list
+
+  actionlint:
+    desc: Run the actionlint.sh
+    cmd: bash ./scripts/actionlint.sh # tests.yml
```
JonathanOppenheimer, line 12: AvalancheGo doesn't do any tests on Windows but I missed that - to ensure that tests continue to run correctly on Windows, the scripts have to ran in bash directly as Windows doesn't know how to execute the .sh e.g. `fork/exec D:\a\subnet-evm\subnet-evm\scripts\build.sh: %1 is not a valid Win32 application` 

I tried and failed a couple other more elegant solutions than just appending bash to no avail.

Taskfile.yml
```diff
@@ -0,0 +1,89 @@
...
+
+  coverage:
+    desc: Run the coverage.sh
+    cmd: bash ./scripts/coverage.sh # tests.yml
+
+  install-avalanchego-release:
+    desc: Install AvalancheGo release for testing
+    env:
+      BASEDIR: /tmp/e2e-test
+      AVALANCHEGO_BUILD_PATH: /tmp/e2e-test/avalanchego
```
ceyonur, line 42: we should probably let CI to pass that arg

Taskfile.yml
ceyonur: I don't think we need to run actionlint with Windows, actually we can even start removing CI tests for Windows as it's not supported in avalanchego anymore.

Taskfile.yml
```diff
@@ -0,0 +1,89 @@
...
+    desc: Run E2E precompile tests with CI setup
+    cmds:
+      - task: install-avalanchego-release
+      - task: build
+      - task: test-e2e-precompile
+
+  test-e2e-warp:
+    desc: Run E2E warp tests
+    env:
+      AVALANCHEGO_BUILD_PATH: /tmp/e2e-test/avalanchego
```
ceyonur, line 84: again path should be passed from CI

.github/workflows/tests.yml
```diff
@@ -216,13 +210,13 @@ jobs:
           go-version-file: "go.mod"
       - name: Install AvalancheGo Release
         shell: bash
-        run: BASEDIR=/tmp/e2e-test AVALANCHEGO_BUILD_PATH=/tmp/e2e-test/avalanchego ./scripts/install_avalanchego_release.sh
+        run: ./scripts/run_task.sh install-avalanchego-release
```
ceyonur, line 213: should we also move this to task?

.github/workflows/tests.yml
```diff
@@ -180,14 +174,14 @@ jobs:
...
```
ceyonur, line 177: should we also move this to task?

.github/workflows/tests.yml
```diff
@@ -152,14 +146,14 @@ jobs:
         working-directory: ./contracts
       - name: Install AvalancheGo Release
         shell: bash
-        run: BASEDIR=/tmp/e2e-test AVALANCHEGO_BUILD_PATH=/tmp/e2e-test/avalanchego ./scripts/install_avalanchego_release.sh
+        run: ./scripts/run_task.sh install-avalanchego-release
```
ceyonur, line 149: should we also move this to task? 
(`test-e2e-warp`)

ceyonur (commented):
this overall lgtm, left few comments. we should get @maru-ava's approval.

Taskfile.yml
JonathanOppenheimer: Done!

.github/workflows/tests.yml
JonathanOppenheimer: I created a new task to install the avalanchego release to prevent duplication

Taskfile.yml
ceyonur: It should definetly be possible to pass env vars from CI with `run_task` so that moving the `env` from here to `ci.yml` should work. 

I'm just not sure if we need to, but this looks to me that if we want to run these scripts in local we have no way to pass a custom path bc the task.yml just overrides it with `/tmp/e2e-test` (unless I'm mistaken and task.yml just provides a default value here).

JonathanOppenheimer: From reading the [order precedence](https://taskfile.dev/experiments/env-precedence/), the order in which environment variables can be determined by the user (e.g. a variable set in .bashrc vs. via the env taskfile). I am not a fan of modifying `run_task.sh` to pass environment variables because it feels very clunky, and isn't how it's configured in other repositories.

Instead, I've clarified that these are default values, and can be overridden by the user locally if they wish to run a task locally - for example `task test-e2e-precompile AVALANCHEGO_BUILD_PATH=/my/local/path DATA_DIR=/my/local/data` (the same thing `run_task.sh` does). [See this](https://taskfile.dev/usage/#variables), and of course you can always run the scripts directly.

.github/workflows/tests.yml
```diff
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - run: ./scripts/lint_allowed_eth_imports.sh
+      - run: ./scripts/run_task.sh lint_allowed_eth_imports
```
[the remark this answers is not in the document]
JonathanOppenheimer: Perhaps more aggressive than you asked for but I've just consolidated all linting, as is done in AvalancheGo.

.github/workflows/tests.yml
```diff
@@ -50,16 +50,12 @@ jobs:
...
       - uses: actions/setup-go@v5
         with:
           go-version-file: "go.mod"
-      - name: Set timeout on Windows # Windows UT run slower and need a longer timeout
-        shell: bash
-        if: matrix.os == 'windows-latest'
-        run: echo "TIMEOUT=1200s" >> "$GITHUB_ENV"
       - run: go mod download
         shell: bash
       - name: fjl/gencodec generated files are up to date
```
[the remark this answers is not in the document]
JonathanOppenheimer: I've done so - I've mirror Stephen's file choice placement, but would think it'd be better in `/scripts`

Taskfile.yml
```diff
@@ -0,0 +1,107 @@
...
+    cmd: ./scripts/run_ginkgo_warp.sh # tests.yml
+
+  test-e2e-load-ci: # consolidated test-e2e-load
+    desc: Run E2E load tests with CI setup
+    cmds:
+      - task: install-avalanchego-release
+      - task: build
+      - task: test-e2e-load
+
+  test-e2e-warp-ci: # consolidated test-e2e-warp
```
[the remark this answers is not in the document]
JonathanOppenheimer: I think I'm missing larger picture here - I see node install steps in `tests.yml` under `e2e_warp` and `e2e_precompile.` I've moved that into it's own job to prevent the duplication, but I have a feeling that's not exactly what you were angling for..?

.github/workflows/tests.yml
```diff
@@ -11,37 +11,35 @@ on:
     types: [checks_requested]
 
 jobs:
+  setup-node:
```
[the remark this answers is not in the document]
JonathanOppenheimer: Yeah I'm definitely missing something then! I thought from [here](https://docs.github.com/en/actions/how-tos/writing-workflows/choosing-what-your-workflow-does/using-jobs-in-a-workflow#defining-prerequisite-jobs) that it required on the node setup successfully running first. What's the 'correct' way to share the setup step?

Taskfile.yml
```diff
@@ -0,0 +1,168 @@
...
+  generate-mocks:
+    desc: Generates testing mocks
+    cmds:
+      - cmd: grep -lr -E '^// Code generated by MockGen\. DO NOT EDIT\.$' . | xargs -r rm
+      - cmd: go generate -run "go.uber.org/mock/mockgen" ./...
+
+  install-avalanchego-release:
+    desc: Install AvalancheGo release for testing
+    env:
+      BASEDIR: '{{.BASEDIR | default "/tmp/e2e-test"}}'
```
JonathanOppenheimer, line 75: This is actually required here to tell the antithesis tests where to find the binary - moving the env the the task specifically (see line 112) makes sense but still results in the path being unable to be found, as tasks [do not inherit environment variables from the tasks that call them](https://github.com/go-task/task/issues/1277). I am okay leaving it as it is, but curious as to your opinion.

JonathanOppenheimer: Ah I'll set it globally!

.github/workflows/check-clean-branch.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
JonathanOppenheimer: It doesn't; I copied it, and the structure from [AvalancheGo](https://github.com/ava-labs/avalanchego/blob/master/.github/workflows/check-clean-branch.sh). I personally did not like it being a shell script in `/workflows` and will change it to a task.

PR #957, opened 2023-10-16
rename stuttering files

ceyonur:
Renames stustuttering files.

PR #1614, opened 2025-07-03
Sync coreth PR #957: move log to separate package

JonathanOppenheimer:
Mirrors https://github.com/ava-labs/coreth/pull/957

It's been awhile since a sync - I will introduce these PRs one at a time to bring in coreth changes until we reach parity, unless anyone has objections (@ceyonur suggested doing these one at a time).


StephenButtolph (approved):
The diff lgtm.

Do we know why `TestTrimPrefixes` is in coreth but not subnet-evm? It seems like `trimPrefixes` is slightly different between the two repos and the test is also missing in subnet-evm... Pretty odd.

JonathanOppenheimer:
> The diff lgtm.
> 
> Do we know why `TestTrimPrefixes` is in coreth but not subnet-evm? It seems like `trimPrefixes` is slightly different between the two repos and the test is also missing in subnet-evm... Pretty odd.

Doesn't answer your question but I can align them and bring in the test if you'd like 🤷‍♂️

ceyonur:
> `TestTrimPrefixes`

Probably another missing commit/changes in subnet-evm. 

@JonathanOppenheimer let's add it in another PR.

PR #1623, opened 2025-07-09
Sync coreth test cleanup v2

JonathanOppenheimer:
Largely mirrors https://github.com/ava-labs/coreth/pull/950, but this time I don't make mistakes 😎

Do note that this PR goes beyond a simple drop in and replace in from https://github.com/ava-labs/coreth/pull/950 as the testing framework upgrade had implications for tests and functionality not included in coreth (which is why I messed up on first revision) 

No breaking changes

Does not require release notes


params/extras/config.go
```diff
@@ -38,7 +39,13 @@ var (
 	}
 
 	TestChainConfig = &ChainConfig{
-		AvalancheContext:   AvalancheContext{SnowCtx: utils.TestSnowContext()},
+		AvalancheContext: AvalancheContext{
```
JonathanOppenheimer, line 42: If someone with additional context could explain why this is necessary here, but was able to be removed in https://github.com/ava-labs/coreth/pull/950/files#r2080300565, that would be appreciated. I think I can also modify so they're set during VM init

plugin/evm/vm_test.go
```diff
@@ -3272,10 +3271,13 @@ func TestFeeManagerRegressionMempoolMinFeeAfterRestart(t *testing.T) {
...
 	// Contract is initialized but no preconfig is given, reader should return genesis fee config
-	feeConfig, lastChangedAt, err := restartedVM.blockChain.GetFeeConfigAt(restartedVM.blockChain.Genesis().Header())
+	// We must query the current block header here (not genesis) because the FeeManager precompile
+	// is only activated at precompileActivationTime, not at genesis. Querying the genesis header would
+	// return the chain config fee config and lastChangedAt as zero, which is not correct after activation.
+	feeConfig, lastChangedAt, err := restartedVM.blockChain.GetFeeConfigAt(restartedVM.blockChain.CurrentBlock())
 	require.NoError(t, err)
 	require.EqualValues(t, feeConfig, testHighFeeConfig)
-	require.Zero(t, restartedVM.blockChain.CurrentBlock().Number.Cmp(lastChangedAt))
+	require.EqualValues(t, restartedVM.blockChain.CurrentBlock().Number, lastChangedAt)
```
JonathanOppenheimer, line 3280: I do not know why this changed or why it worked before. Again would appreciate another set of eyes and brain!

params/extras/config.go
```diff
@@ -351,7 +358,7 @@ func (c *ChainConfig) Verify() error {
 	}
 
 	// Verify the network upgrades are internally consistent given the existing chainConfig.
-	if err := c.verifyNetworkUpgrades(c.SnowCtx.NetworkUpgrades); err != nil {
+	if err := c.NetworkUpgrades.verifyNetworkUpgrades(c.SnowCtx.NetworkUpgrades); err != nil {
 		return fmt.Errorf("invalid network upgrades: %w", err)
```
ceyonur, line 362: can we revert this?

params/extras/config.go
ceyonur: I think the only difference is in Verify() where we pass c.SnowCtx.NetworkUpgrades. I think we should explicitly handle that in tests.

plugin/evm/vm_test.go
```diff
@@ -92,97 +101,158 @@ var (
...
+	// These will be initialized after init() runs
+	// forkToChainConfig maps a fork to a chain config
+	forkToChainConfig = map[upgradetest.Fork]*params.ChainConfig{
+		upgradetest.Durango: params.TestDurangoChainConfig,
+		upgradetest.Etna:    params.TestEtnaChainConfig,
+		// upgradetest.Fortuna: params.TestFortunaChainConfig,
+		upgradetest.Granite: params.TestGraniteChainConfig,
+	}
+
+	genesisJSONPreSubnetEVM string
```
ceyonur, line 113: can we initialize them here as `genesisJSONPreSubnetEVM = toGenesisJSON(params.TestPreSubnetEVMChainConfig)`

utils/snow.go
```diff
@@ -7,59 +7,13 @@ import (
...
-	aliaser := ctx.BCLookup.(ids.Aliaser)
-	_ = aliaser.Alias(testCChainID, "C")
-	_ = aliaser.Alias(testCChainID, testCChainID.String())
-	_ = aliaser.Alias(testXChainID, "X")
-	_ = aliaser.Alias(testXChainID, testXChainID.String())
-
-	return ctx
-}
-
 func NewTestValidatorState() *validatorstest.State {
```
ceyonur, line 17: we should remove this entire file

utils/snow_test.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"testing"
+
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/snow/snowtest"
+	"github.com/ava-labs/subnet-evm/utils/utilstest"
+	"github.com/stretchr/testify/require"
+)
+
+func TestNewTestSnowContext(t *testing.T) {
```
ceyonur, line 15: this test should be in utilstest

plugin/evm/vm_test.go
```diff
@@ -92,97 +101,158 @@ var (
...
+		atomicMemory: atomicMemory,
+		appSender:    appSender,
+	}
+}
+
+// setupGenesis sets up the genesis
+func setupGenesis(
+	t *testing.T,
+	fork upgradetest.Fork,
+	fallbackGenesisJSON string,
```
ceyonur, line 214: I feel like we should not need that and just use genesisJson option in newVM

plugin/evm/vm_validators_test.go
```diff
@@ -99,7 +101,7 @@ func TestValidatorState(t *testing.T) {
 	vm = &VM{}
 	err = vm.Initialize(
 		context.Background(),
-		utils.TestSnowContext(), // this context does not have validators state, making VM to source it from the database
+		utilstest.NewTestSnowContext(t, snowtest.CChainID), // this context does not have validators state, making VM to source it from the database
```
ceyonur, line 104: We should not use CChainID here and generate a new ID. If we're only using a single chain ID (CChainID) then I think we can just skip taking a ChainID in `NewTestSnowContext` and just define a subnet-evm chain ID and use that.

params/extras/config.go
JonathanOppenheimer: Yup!

plugin/evm/vm_test.go
JonathanOppenheimer: Even if this would diverge further from coreth? Unless you're asking about the argument

params/extras/network_upgrades.go
```diff
@@ -93,19 +93,19 @@ func (n *NetworkUpgrades) SetDefaults(agoUpgrades upgrade.Config) {
 func (n *NetworkUpgrades) verifyNetworkUpgrades(agoUpgrades upgrade.Config) error {
 	defaults := GetNetworkUpgrades(agoUpgrades)
 	if err := verifyWithDefault(n.SubnetEVMTimestamp, defaults.SubnetEVMTimestamp); err != nil {
-		return fmt.Errorf("SubnetEVM fork block timestamp is invalid: %w", err)
+		return fmt.Errorf("subnetEVM fork block timestamp is invalid: %w", err)
```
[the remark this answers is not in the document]
ceyonur: ah I was not aware that, weird why does it errors now?

plugin/evm/vm_test.go
```diff
@@ -57,20 +58,21 @@ import (
...
 	avagoconstants "github.com/ava-labs/avalanchego/utils/constants"
 )
 
 var (
 	testNetworkID uint32 = avagoconstants.UnitTestID
 
-	testMinGasPrice int64 = 225_000_000_000
-	testKeys        []*ecdsa.PrivateKey
+	testMinGasPrice int64            = 225_000_000_000
+	testKeys                         = secp256k1.TestKeys()[:2]
```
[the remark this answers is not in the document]
ceyonur: let's keep it same just in case

params/extras/network_upgrades.go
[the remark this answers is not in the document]
JonathanOppenheimer: I can revert and add an ignore statement thing I think

params/extras/config.go
JonathanOppenheimer: Can you expand upon what you mean by this - I see now why we do need the context as the `verifyNetworkUpgrades` method needs to compare configured timestamps against the expected defaults from the Avalanche config.

plugin/evm/vm_test.go
JonathanOppenheimer: Ah I see - new Go fact learned!

plugin/evm/vm_test.go
```diff
@@ -93,84 +101,151 @@ var (
...
+		testEthAddrs = append(testEthAddrs, key.EthAddress())
+	}
+
+	// Initialize forkToChainConfig after testEthAddrs is populated
+	forkToChainConfig = map[upgradetest.Fork]*params.ChainConfig{
+		upgradetest.Durango: params.TestDurangoChainConfig,
+		upgradetest.Etna:    params.TestEtnaChainConfig,
+		// upgradetest.Fortuna: params.TestFortunaChainConfig,
+		upgradetest.Granite: params.TestGraniteChainConfig,
+	}
```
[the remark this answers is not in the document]
JonathanOppenheimer: Yes!

params/extras/precompile_config_test.go
```diff
@@ -16,13 +16,17 @@ import (
...
 	"github.com/stretchr/testify/require"
 )
 
 func TestVerifyWithChainConfig(t *testing.T) {
 	admins := []common.Address{{1}}
 	copy := *TestChainConfig
 	config := &copy
+	config.AvalancheContext = AvalancheContext{
+		SnowCtx: utilstest.NewTestSnowContext(t),
+	}
```
ceyonur, line 29: config.SnowCtx = utilstest.NewTestSnowContext(t) 

and similary in all other occurrences

ceyonur (approved):

PR #1629, opened 2025-07-15
sync: coreth PR #968-969: add linters

JonathanOppenheimer:


precompile/allowlist/allowlist_test.go
```diff
@@ -1,4 +1,4 @@
-// (c) 2019-2023, Ava Labs, Inc. All rights reserved.
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE for licensing terms.
 
 package allowlist_test
```
ceyonur, line 4: I think we can now move this to allowlisttest pkg

precompile/allowlist/config_test.go
```diff
@@ -1,26 +1,27 @@
-// (c) 2019-2023, Ava Labs, Inc. All rights reserved.
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE for licensing terms.
 
 package allowlist_test
```
ceyonur, line 4: we can now move this to allowlisttest pkg

rpc/client_opt_test.go
```diff
@@ -8,6 +9,8 @@
 //
 // Much love to the original authors for their work.
 // **********
+// TODO: FIND THE ORIGINAL LICENSE BECAUSE IT IS NOT HERE! >:O
```
ceyonur, line 12: revert this, if there is nothing we can leave it as it's

scripts/generated_file_list.txt
```diff
@@ -0,0 +1,10 @@
+accounts/abi/bind/precompilebind/precompile_config_template.go
+accounts/abi/bind/precompilebind/precompile_config_test_template.go
+accounts/abi/bind/precompilebind/precompile_contract_template.go
+accounts/abi/bind/precompilebind/precompile_contract_test_template.go
+accounts/abi/bind/precompilebind/precompile_event_template.go
+accounts/abi/bind/precompilebind/precompile_module_template.go
```
ceyonur, line 6: why did we special case this? I think these can be excluded in `scripts/upstream_files.txt` and then we can use avalanche header right?

scripts/generated_file_list.txt
```diff
@@ -0,0 +1,10 @@
...
+precompile/allowlist/event.go
+precompile/contracts/feemanager/event.go
+precompile/contracts/nativeminter/event.go
+precompile/contracts/rewardmanager/event.go
```
ceyonur, line 10: again we can just use avalanche header in these files right? (these are generated but we can ignore that header)

scripts/generated_file_list.txt
ceyonur, line 1: I feel like we can remove this and special handling of generated files

cmd/abigen/namefilter_test.go
```diff
@@ -7,6 +8,23 @@
 // original code from which it is derived.
 //
 // Much love to the original authors for their work.
+// **********
+// Copyright 2016 The go-ethereum Authors
```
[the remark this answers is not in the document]
JonathanOppenheimer: I think it's a bit strange to have a header referencing a license below, but then there is no license, but yeah I can remove it.

precompile/allowlist/config_test.go
JonathanOppenheimer: this and config_test which also used to be `package allowlist_test`

rpc/client_opt_test.go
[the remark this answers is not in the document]
JonathanOppenheimer: What do you mean more changes?

alarso16 (approved):
A lot of these file moves seem out of scope for this PR. If Cey said to include these, i think it's ready to merge

rpc/client_opt_test.go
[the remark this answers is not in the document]
alarso16: There's more license below, right? It only has the first half of the license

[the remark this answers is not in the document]
JonathanOppenheimer: Oh yeah - I pointed this out to Cey on a couple other files (https://github.com/ava-labs/subnet-evm/pull/1629#discussion_r2216726697) - there's upstream files we took which don't have licensure. I agree it's weird.

JonathanOppenheimer:
> A lot of these file moves seem out of scope for this PR. If Cey said to include these, i think it's ready to merge

All of the file moves are a result of the test package linting - for example "testing" as an import. The only file moves here that aren't in the `coreth` PR are the `allowlist` ones as those files don't exist in `coreth.`

thank you Austin :)

PR #1630, opened 2025-07-16
Update releasing README.md

ceyonur:
Updates releasing md to use the correct version of avalanchego in local testing

### DIFF
--- a/docs/releasing/README.md
+++ b/docs/releasing/README.md
@@ -170,11 +170,11 @@ Once the tag is created, you need to test it on the Fuji testnet both locally an
     git clone git@github.com:ava-labs/avalanchego.git
     ```
 
-1. Build AvalancheGo using those VM ids:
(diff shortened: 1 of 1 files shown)


JonathanOppenheimer (approved):

docs/releasing/README.md
```diff
@@ -170,11 +170,11 @@ Once the tag is created, you need to test it on the Fuji testnet both locally an
     git clone git@github.com:ava-labs/avalanchego.git
     ```
 
-1. Build AvalancheGo using those VM ids:
+1. Checkout correct AvalancheGo version, the version should match the one used in Subnet-EVM `go.mod` file
 
     ```bash
     cd avalanchego
-    ./scripts/build.sh
+    git checkout v1.13.0
```
rkuris, line 177: Seems like they should just check out the branch you need with:
```
git clone git@github.com:ava-labs/avalanchego.git --branch v1.13.0
```

Don't they still need to run `./scripts/build.sh`? Even if it's only to verify that it builds fine?

PR #1631, opened 2025-07-16
feat: Add Firewood Trie implementation

alarso16:
Cherry-picks ava-labs/coreth#1029

### DIFF
--- a/triedb/firewood/database.go
+++ b/triedb/firewood/database.go
@@ -1,0 +1,612 @@
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package firewood
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"sync"
+	"time"
+
+	"github.com/ava-labs/libevm/common"
+	"github.com/ava-labs/libevm/core/rawdb"
+	"github.com/ava-labs/libevm/core/types"
+	"github.com/ava-labs/libevm/ethdb"
+	"github.com/ava-labs/libevm/libevm/stateconf"
+	"github.com/ava-labs/libevm/log"
+	"github.com/ava-labs/libevm/metrics"
+	"github.com/ava-labs/libevm/trie/trienode"
+	"github.com/ava-labs/libevm/trie/triestate"
+	"github.com/ava-labs/libevm/triedb"
+	"github.com/ava-labs/libevm/triedb/database"
+
+	ffi "github.com/ava-labs/firewood-go-ethhash/ffi"
+)
+
+var (
+	_ proposable = (*ffi.Database)(nil)
+	_ proposable = (*ffi.Proposal)(nil)
+
+	// FFI triedb operation metrics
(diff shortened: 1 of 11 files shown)


JonathanOppenheimer (dismissed):
Could you add a bullet in the commit .txt file that this PR is cherry picking coreth (as the sync will be out of order)

JonathanOppenheimer (commented):
Thank you for the docs update!

triedb/firewood/database.go
```diff
@@ -0,0 +1,612 @@
...
+	"github.com/ava-labs/libevm/ethdb"
+	"github.com/ava-labs/libevm/libevm/stateconf"
+	"github.com/ava-labs/libevm/log"
+	"github.com/ava-labs/libevm/metrics"
+	"github.com/ava-labs/libevm/trie/trienode"
+	"github.com/ava-labs/libevm/trie/triestate"
+	"github.com/ava-labs/libevm/triedb"
+	"github.com/ava-labs/libevm/triedb/database"
+
+	ffi "github.com/ava-labs/firewood-go-ethhash/ffi"
```
JonathanOppenheimer, line 25: Random question but doesn't this alias not do anything? the package is already ffi. What's the point of it? I've seen aliases throughout the subnet-evm seemingly at random and am curious as to when to use them / not use them

triedb/firewood/database.go
```diff
@@ -0,0 +1,612 @@
...
+// Size returns the storage size of diff layer nodes above the persistent disk
+// layer and the dirty nodes buffered within the disk layer
+// Only used for metrics and Commit intervals in APIs.
+// This will be implemented in the firewood database eventually.
+// Currently, Firewood stores all revisions in disk and proposals in memory.
+func (db *Database) Size() (common.StorageSize, common.StorageSize) {
+	return 0, 0
+}
+
+// This isn't called anywhere in subnet-evm.
```
JonathanOppenheimer, line 375: minimizes diff with coreth? Seen other uncalled functions as well and get linting warnings

triedb/firewood/database.go
```diff
@@ -0,0 +1,612 @@
...
+}
+
+// reader is a state reader of Database which implements the Reader interface.
+type reader struct {
+	db   *Database
+	root common.Hash // The root of the state this reader is reading.
+}
+
+// Node retrieves the trie node with the given node hash. No error will be
+// returned if the node is not found.
```
JonathanOppenheimer, line 522: Humor me but! Why wouldn't an error want to be returned - what is returned instead?

JonathanOppenheimer (approved):

triedb/firewood/database.go
alarso16: Nope, not called anywhere there either :). This is just to fulfill the interface to be a `triedb`

triedb/firewood/database.go
alarso16: `nil` - if you're looking for a value in a database and it isn't found, you wouldn't expect to get an error.

triedb/firewood/database.go
alarso16: I'm going to push this to a different PR so syncing is better, and this must also be fixed in coreth

Issue #987, opened 2023-11-04
A question about EstimateBaseFee logic


shubhamgoyal42:
In a attempt to test the network resilience of our subnet against spam, I'm testing base fee increase on local network. In [EstimateBaseFee](https://github.com/ava-labs/subnet-evm/blob/master/eth/gasprice/gasprice.go#L244), I see that we use the min of 2 base fees. One is calculated from the last few blocks’ base fees, and the other from fee config’s BaseFeeChangeDenominator and TargetGas. Why do we use the minimum of these 2 values? In our subnet on local and fuji network, I observed that base fee never increased even though `nextBaseFee` in the above function increases when a burst of transactions is made. Even when making ~10 transactions per second, I was not able to increase the baseFee.

From what I understand from the code, [baseFee](https://github.com/ava-labs/subnet-evm/blob/master/eth/gasprice/gasprice.go#L226) will never increase if we keep using a min of suggestDynamicFees and estimateNextBaseFee. For example, if the minBaseFee is 60 gwei, and baseFee of the last few blocks is also 60 gwei,  suggestDynamicFees will always return 60 gwei.

Here's my fee config on local network:

```json
{
    "gasLimit": 500000000,
    "targetBlockRate": 1,
    "minBaseFee": 60000000000,
    "targetGas": 1000000,
    "baseFeeChangeDenominator": 5,
    "minBlockGasCost": 0,
    "maxBlockGasCost": 5000000,
    "blockGasCostStep": 10000
}
```

ceyonur:
`suggestDynamicFees` does a lookback analysis and returns a percentile of sorted historic fees, so if all previous blocks had 60 gwei fee, then it will return 60 gwei.

`nextBaseFee` is calculated as if the block is being build at that time when you call the API. However when it's actually being build, it can show a lower price since time will pass. 

In your case ~10 tps is a bit too low to see any changes. On my tests I had to push to ~150 TPS to see it changes. 

You should either decrease min fee, or decrease target gas to see any changes. or just simply increase TPS. BTW 1.000.000 target gas is a bit low. Keep in mind that it is the target for the whole window (10 sec). 

You can take a look at fee config field explanations in the code [here](https://github.com/ava-labs/subnet-evm/blob/master/commontype/fee_config.go).

PR #1639, opened 2025-07-21
sync: coreth PR #987: add granite configs

JonathanOppenheimer:


JonathanOppenheimer:
This is actually meaningless - see: https://github.com/ava-labs/subnet-evm/pull/1498

PR #1651, opened 2025-07-24
ci: Add GH token

alarso16:
## Why this should be merged

Ports the related changes from ava-labs/coreth#1003, as well as ava-labs/coreth#1064 and ava-labs/coreth#1066

## How this works

Adds debug printing back, moves github token in

## How this was tested

In coreth

## Need to be documented?

No

No

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -173,4 +173,4 @@ tasks:
 
   update-avalanchego-version:
     desc: Update AvalancheGo version in go.mod and sync GitHub Actions workflow custom action version
-    cmd: ./scripts/update_avalanchego_version.sh # tests.yml
+    cmd: bash -x ./scripts/update_avalanchego_version.sh # tests.yml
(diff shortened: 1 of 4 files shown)


Taskfile.yml
```diff
@@ -173,4 +173,4 @@ tasks:
 
   update-avalanchego-version:
     desc: Update AvalancheGo version in go.mod and sync GitHub Actions workflow custom action version
-    cmd: ./scripts/update_avalanchego_version.sh # tests.yml
+    cmd: bash -x ./scripts/update_avalanchego_version.sh # tests.yml
```
JonathanOppenheimer, line 176: Is the bash -x still necessary?

JonathanOppenheimer (approved):

Taskfile.yml
alarso16: Stephen wanted it. Also, if we have issues with this script in the future (like this token isn't sufficient), we can address them easier.

PR #1679, opened 2025-08-07
sync: coreth PR #963,981,1009: sync package extension and related items

JonathanOppenheimer:
This PR syncs:
- https://github.com/ava-labs/coreth/pull/963
- https://github.com/ava-labs/coreth/pull/981
- https://github.com/ava-labs/coreth/pull/1009

Tthis PR brings in a lot of changes that aren't particular useful for subnet-evm, like the extender. This is intended, as I am trying to minimize the difference between coreth's version of these files, and subnet-evms. The eventual goal is the complete merger of all of these files, which is why I am making these changes now. 

This PR introduces practically 0 "originally written by me code" - this is a culmination of the code of many other people (and in effect, has already been reviewed)  

Also syncs this PR: https://github.com/ava-labs/coreth/pull/1338 as surfaced by @alarso16

PR #1716, opened 2025-09-04
sync: coreth PR #1141: update tx gossip rate limit dynamically

JonathanOppenheimer:
## Why this should be merged
This PR syncs https://github.com/ava-labs/coreth/pull/1141 and additionally brings in the gossip related refactor work that was done in https://github.com/ava-labs/coreth/pull/1018, which for some reason was never synced to subnet-evm. 

These changes are necessary because in order to release a new version of subnet-evm, we need to bump AvalancheGo, but v1.13.5 of AvalancheGo introduces breaking changes, which this PR addresses. Thus, these changes are currently blocking a subnet-evm release. 

## How this was tested
CI

## Need to be documented?
No

No


JonathanOppenheimer:
> There are some things left out of the atomic refactor. I know they don't pertain to this specifically, but it would be good to address eventually

Yeah I only included the minimum set of changes. All of 1018 can't (or really should not) be brought in until

https://github.com/ava-labs/subnet-evm/pull/1679/files

is merged.

PR #1724, opened 2025-09-09
Delete latest commit file

JonathanOppenheimer:
This is annoying to maintain and leads to merge conflicts and isn't even accurate.


StephenButtolph (approved):
> isn't even accurate

all I needed to hear

PR #1728, opened 2025-09-15
sync: coreth PR #1168: move block fee verification to customheader

JonathanOppenheimer:
Syncs https://github.com/ava-labs/coreth/pull/1168


plugin/evm/header/block_gas_cost.go
```diff
@@ -117,3 +120,70 @@ func EstimateRequiredTip(
...
+		totalBlockFee        = new(big.Int)
+	)
+
+	// Add in the external contribution
+	if extraStateChangeContribution != nil {
+		if extraStateChangeContribution.Cmp(common.Big0) < 0 {
+			return fmt.Errorf("invalid extra state change contribution: %d", extraStateChangeContribution)
+		}
+		totalBlockFee.Add(totalBlockFee, extraStateChangeContribution)
+	}
```
ceyonur, line 150: for the context: this is unnecessary but we can keep it to minimize the diffs

consensus/dummy/consensus.go
```diff
@@ -42,7 +41,31 @@ type Mode struct {
...
+		parent *types.Header,
+		statedb *state.StateDB,
+	) (
+		blockFeeContribution *big.Int,
+		err error,
+	)
+	ConsensusCallbacks struct {
+		OnFinalizeAndAssemble OnFinalizeAndAssembleCallbackType
+		OnExtraStateChange    OnExtraStateChangeType
+	}
```
ceyonur, line 66: these are unnecessary and not being used, but we can keep them to minimize diffs

plugin/evm/header/block_gas_cost.go
```diff
@@ -117,3 +120,70 @@ func EstimateRequiredTip(
...
+	txs []*types.Transaction,
+	receipts []*types.Receipt,
+	extraStateChangeContribution *big.Int,
+) error {
+	if baseFee == nil || baseFee.Sign() <= 0 {
+		return fmt.Errorf("invalid base fee (%d) in apricot phase 4", baseFee)
+	}
+	if requiredBlockGasCost == nil || !requiredBlockGasCost.IsUint64() {
+		return fmt.Errorf("invalid block gas cost (%d) in apricot phase 4", requiredBlockGasCost)
+	}
```
ceyonur, line 136: it's not in scope of this PR but we should probably skip verifying block fee if required block gas cost is 0, like here https://github.com/ava-labs/coreth/pull/1215

JonathanOppenheimer: This will be synced when the sync work catches up - right now I am waiting for the existing sync PRs open to be reviewed and merged, as they need to be done in chronological order.

consensus/dummy/consensus.go
```diff
@@ -41,7 +40,31 @@ type Mode struct {
 }
 
 type (
+	OnFinalizeAndAssembleCallbackType = func(
```
michaelkaplan13, line 43: It looks like these changes are related to the original PR we're syncing here. Is there a reason they need to be included in this PR?

consensus/dummy/consensus.go
```diff
@@ -384,13 +362,16 @@ func (eng *DummyEngine) FinalizeAndAssemble(chain consensus.ChainHeaderReader, h
...
-			receipts,
-		); err != nil {
-			return nil, err
+		if !eng.consensusMode.ModeSkipBlockFee {
+			if err := customheader.VerifyBlockFee(
+				header.BaseFee,
+				headerExtra.BlockGasCost,
+				txs,
+				receipts,
+				nil,
```
michaelkaplan13, line 371: Above in `Finalize` it looks like we declare `contribution`, but that it is always left as `nil`. Here, we just pass `nil` directly. I assume the benefit of declaring the unassigned var is to reduce the diff with `coreth`, but think we should be consistent with either declaring or not (I'd personally lean not declaring just to reduce the diff, because it's not clear in the code itself why the var is there if it's never assigned).

consensus/dummy/consensus.go
JonathanOppenheimer: Yeah totally understand that - if syncing a coreth PR requires bringing in other code (that wasn't brought in the past, I've done it in the same PR (as otherwise would sacrifice speed). But yes generally understand and agree with what you're saying. If you'd like me to separate these out into two PRs I certainly can.

consensus/dummy/consensus.go
JonathanOppenheimer: Yeah I think this is generally a more philosophical question. I think I'd be inclined to agree if the merger wasn't something that hopefully is happening in the near term. Muddying the subnet-evm code is absolutely valid, but it but it hopefully won't be muddied for long (as it will cease to exist), and I'd further hope that it wouldn't matter that much for people looking at it, as I'd hope no one would be looking at it - are we going to be making consensus specific subnet-evm changes prior to the repository merger? Surely the answer is no. 

@ceyonur do you have thoughts?

consensus/dummy/consensus.go
```diff
@@ -28,10 +28,9 @@ import (
 var (
 	allowedFutureBlockTime = 10 * time.Second // Max time from current time allowed for blocks, before they're considered future blocks
 
-	ErrInsufficientBlockGas = errors.New("insufficient gas to cover the block cost")
-
-	errInvalidBlockTime  = errors.New("timestamp less than parent's")
-	errUnclesUnsupported = errors.New("uncles unsupported")
+	errInvalidBlockTime    = errors.New("timestamp less than parent's")
+	errUnclesUnsupported   = errors.New("uncles unsupported")
+	errInvalidBlockGasCost = errors.New("invalid blockGasCost")
```
michaelkaplan13, line 33: I think my final nit: Most of the changes to this file don't seem necessary or directly related to this PR. Can we limit it to only match the changes made to `consensus.go` in the [coreth PR](https://github.com/ava-labs/coreth/pull/1168/files#diff-532a2c6a5365d863807de5b435d8d6475552904679fd611b1b4b10d3bf4f5010) and just pass `nil` for `contribution`?

I think that would make the diff of this PR as minimal as possible

michaelkaplan13 (approved):

plugin/evm/customheader/block_gas_cost.go
```diff
@@ -117,3 +120,70 @@ func EstimateRequiredTip(
 	estimatedTip := totalRequiredTips.Div(totalRequiredTips, totalGasUsed)
 	return estimatedTip, nil
 }
+
+func VerifyBlockFee(
+	baseFee *big.Int,
+	requiredBlockGasCost *big.Int,
+	txs []*types.Transaction,
+	receipts []*types.Receipt,
+	extraStateChangeContribution *big.Int,
```
StephenButtolph, line 129: Should we remove `extraStateChangeContribution` as well? We don't currently have it in subnet-evm.

I'm fine leaving it if its to reduce the diff with coreth, but I want to be intentional about it.

PR #1738, opened 2025-09-19
Resolve npm security vulnerabilities

JonathanOppenheimer:
## Why this should be merged
As discussed, while we should be working towards removing `npm` dependancies in general we have a large number of concerning security alerts due to the packages we rely on. This PR resolves all `npm` security alerts, by updating 

## How this works

@avalabs/avalanchejs: 4.0.5 → 5.0.0 ⬆️
@openzeppelin/contracts: 4.9.6 → 5.4.0 ⬆️
@ethersproject/signing-key: Not installed → 5.8.0 ➕
@sentry/node: Not installed → 10.12.0 ➕
solc: Not installed → 0.8.30 ➕

signing-key, sentry, and solc were installed explicitly to resolve security vulnerabilities created by an old version of `hard-hat` and it's transitive dependencies. Explicitly updating `hard-hat` to `v3.x` would introduce too many breaking changes, AFAICT. 

## How this was tested
CI, .sol contracts compile

## Need to be documented?
No

No

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -175,7 +175,7 @@ const (
 	// clean cache's underlying fastcache.
 	trieCleanCacheStatsNamespace = "hashdb/memcache/clean/fastcache"
 
-	firewoodFileName = "firewood_state"
+	firewoodFileName = "firewood.db"
 )
 
 // cacheableFeeConfig encapsulates fee configuration itself and the block number that it has changed at,
--- a/core/extstate/firewood_database.go
+++ b/core/extstate/firewood_database.go
@@ -29,12 +29,14 @@ func (db *firewoodAccessorDb) OpenTrie(root common.Hash) (state.Trie, error) {
 }
 
 // OpenStorageTrie opens a wrapped version of the account trie.
-func (*firewoodAccessorDb) OpenStorageTrie(_ common.Hash, _ common.Address, accountRoot common.Hash, self state.Trie) (state.Trie, error) {
+//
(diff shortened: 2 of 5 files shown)


contracts/tasks.ts
```diff
@@ -5,7 +5,7 @@ const CONTRACT_ALLOW_LIST_ADDRESS = "0x0200000000000000000000000000000000000000"
 const MINT_ADDRESS = "0x0200000000000000000000000000000000000001"
 const TX_ALLOW_LIST_ADDRESS = "0x0200000000000000000000000000000000000002"
 const FEE_MANAGER_ADDRESS = "0x0200000000000000000000000000000000000003"
-const REWARD_MANAGER_ADDDRESS = "0x0200000000000000000000000000000000000004"
+const REWARD_MANAGER_ADDRESS = "0x0200000000000000000000000000000000000004"
```
michaelkaplan13, line 8: Not to be nit picky or inflate the number of PRs, but if it's a small standalone fix that's separate from the purpose of the PR, it's generally best to do it in its own trivial PR rather than batching it together in a bigger one.

JonathanOppenheimer: I've broken out these changes into a separate PR: https://github.com/ava-labs/subnet-evm/pull/1740

michaelkaplan13 (approved):
I confirmed the `package-lock.json` updates by running `npm install` and checking that there was no local diff from the version that is checked in.

ceyonur (approved):

PR #1761, opened 2025-09-30
sync: coreth PR #1149: add precompile reversion logic for Granite activation

JonathanOppenheimer:
Syncs https://github.com/ava-labs/coreth/pull/1149. 

- Requesting x for review as he was the original author
- Requesting x for review as he was the merging approver.


plugin/evm/vm_test.go
```diff
@@ -3926,3 +3931,176 @@ func TestCreateHandlers(t *testing.T) {
 		require.ErrorIs(t, elem.Error, rpc.ErrMissingBatchResponse)
 	}
 }
+
+// newSignedLegacyTx builds a legacy transaction and signs it using the
+// LatestSigner derived from the provided chain config.
+func newSignedLegacyTx(
```
JonathanOppenheimer, line 3937: this function was introduced in the original delegate call PR, and touched a bunch of other tests. due to how subnet-evm is not synced/up to date with coreth, making those other changes is very difficult / impossible without further work. We include it here for use in the test introduced, but leave the other changes to a future sync.

params/hooks_libevm.go
```diff
@@ -100,6 +100,7 @@ func makePrecompile(contract contract.StatefulPrecompiledContract) libevm.Precom
 			panic(err) // Should never happen
 		}
 		var predicateResults predicate.BlockResults
+		rules := GetRulesExtra(env.Rules()).AvalancheRules
```
michaelkaplan13, line 103: Nit: Typically it'd be cleanest to put the var definition as close as possible to its use below. In this case, it's a bit of a gray area because putting it here aligns it most closely with coreth (which needs it for `PredicateBytesFromExtra`).

Nonetheless, I generally think that best practices shouldn't be bypassed for contextual reasons not apparent in this code repo itself, so would lean towards putting it as close as possible to its use for now.

plugin/evm/vm.go
```diff
@@ -531,6 +532,13 @@ func parseGenesis(ctx *snow.Context, genesisBytes []byte, upgradeBytes []byte, a
 	// Set network upgrade defaults
 	configExtra.SetDefaults(ctx.NetworkUpgrades)
 
+	// If Durango is scheduled, schedule the Warp Precompile at the same time.
```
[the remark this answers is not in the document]
michaelkaplan13: IIUC, the tests just depend on the warp precompile being active. Can we configure that in the test set up code itself rather than changing production code here?

plugin/evm/vm.go
```diff
@@ -531,6 +532,29 @@ func parseGenesis(ctx *snow.Context, genesisBytes []byte, upgradeBytes []byte, a
 	// Set network upgrade defaults
 	configExtra.SetDefaults(ctx.NetworkUpgrades)
 
+	// If Durango is scheduled, schedule the Warp Precompile at the same time
+	// (unless it's already been configured manually).
+	if configExtra.DurangoTimestamp != nil {
```
ceyonur, line 537: Why we have added these? I don't think they're related with the PR or correct in general.

ceyonur (changes requested):

plugin/evm/vm.go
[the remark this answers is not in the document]
ceyonur: Because warp was never intended to be a required/default precompile for subnet-evm. We cannot really retroactively activate this with Durango even if we wanted it to be auto-enabled.

plugin/evm/vm.go
JonathanOppenheimer: I can remove it and you can see everything that fails without it - gotcha regarding though! 

> Because warp was never intended to be a required/default precompile for subnet-evm. We cannot really retroactively activate this with Durango even if we wanted it to be auto-enabled.

plugin/evm/vm_test.go
```diff
@@ -3972,3 +3977,176 @@ func TestBlockGasValidation(t *testing.T) {
...
+			wantReceiptStatus:     types.ReceiptStatusSuccessful,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			vm := newVM(t, testVMConfig{
+				genesisJSON: genesisJSONSubnetEVM,
+				fork:        &tt.fork,
+			}).vm
```
ceyonur, line 4099: we need to add warp here if these tests rely on warp. Could you take a look at how it's done in `vm_warp_test` for an example?

ceyonur: yup!

ceyonur (approved):

PR #1770, opened 2025-10-02
chore: remove deprecated genesis types

JonathanOppenheimer:
## Why this should be merged
These types are explicitly deprecated and just serve as a type alias. 

## How this was tested
Existing UTs

## Need to be documented?
No

No


core/genesis.go
ceyonur: should we also get rid of these?

ceyonur (changes requested):

PR #1217, opened 2024-06-20
fix build: change protobuf version

darioush:
I'm seeing:  `undefined: descriptorpb.Default_FileOptions_PhpGenericServices` when running `./scripts/build_test.sh`
See: https://github.com/protocolbuffers/protobuf/issues/16163

PR #1772, opened 2025-10-02
sync: coreth PR #1217: have `checkPredicates` account for maximum available gas capacity

JonathanOppenheimer:
Sync ava-labs/coreth#1217

- Requesting Cey for review as he was the first approver
- Requesting Stephen for review as he was the merging approver.

### DIFF
--- a/plugin/evm/vm_test.go
+++ b/plugin/evm/vm_test.go
@@ -37,8 +37,6 @@ import (
 	"github.com/ava-labs/avalanchego/vms/evm/predicate"
 	avalancheWarp "github.com/ava-labs/avalanchego/vms/platformvm/warp"
 	"github.com/ava-labs/avalanchego/vms/platformvm/warp/payload"
-	"github.com/ava-labs/coreth-internal/plugin/evm/vmtest"
-	"github.com/ava-labs/coreth/plugin/evm/upgrade/ap0"
 	"github.com/ava-labs/libevm/common"
 	"github.com/ava-labs/libevm/common/math"
 	"github.com/ava-labs/libevm/core/rawdb"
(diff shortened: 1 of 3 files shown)


plugin/evm/vm_test.go
```diff
@@ -3812,3 +3821,162 @@ func TestCreateHandlers(t *testing.T) {
...
+	newBlock := func(
+		t *testing.T,
+		vm *VM,
+		claimedGasUsed uint64,
+	) *types.Block {
+		require := require.New(t)
+
+		chainExtra := params.GetExtra(vm.chainConfig)
+		parent := vm.eth.APIBackend.CurrentBlock()
+		const timeDelta = acp176.TimeToFillCapacity
```
ceyonur, line 3838: we're not using acp176 here so we should probably stay avoid importing it.

plugin/evm/vm_test.go
```diff
@@ -3926,3 +3934,169 @@ func TestCreateHandlers(t *testing.T) {
...
+	newBlock := func(
+		t *testing.T,
+		vm *VM,
+		claimedGasUsed uint64,
+	) *types.Block {
+		require := require.New(t)
+
+		chainExtra := params.GetExtra(vm.chainConfig)
+		parent := vm.eth.APIBackend.CurrentBlock()
+		const timeDelta = 5 // seconds to fill gas capacity
```
michaelkaplan13, line 3951: Nit: Subnet-EVM doesn't have ACP-176 yet, so there is no notion of "gas capacity". I think this is really just a mocked block delay in this case

ceyonur (approved):
changes LGTM, just a comment for a follow up PR.

PR #1313, opened 2024-08-27
build_docker_image.sh: use AVALANCHEGO_NODE_IMAGE

darioush:
Failing job on master

PR #1777, opened 2025-10-04
sync: coreth PR #1313: remove secondary validator override

ceyonur:
## Why this should be merged

Coreth https://github.com/ava-labs/coreth/pull/1313 sync PR

### DIFF
--- a/plugin/evm/vm_upgrade_bytes_test.go
+++ b/plugin/evm/vm_upgrade_bytes_test.go
@@ -189,9 +189,8 @@ func TestNetworkUpgradesOverridden(t *testing.T) {
 
 	// verify initial state
 	require.True(t, tvm.vm.chainConfigExtra().IsGranite(uint64(upgrade.InitiallyActiveTime.Unix())))
-	rules := currentRules(tvm.vm)
-	require.True(t, rules.IsSubnetEVM)
-	require.False(t, rules.IsGranite)
+	require.True(t, tvm.vm.currentRules().IsSubnetEVM)
+	require.False(t, tvm.vm.currentRules().IsGranite)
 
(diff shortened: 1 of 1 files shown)


plugin/evm/vm_upgrade_bytes_test.go
```diff
@@ -210,7 +211,7 @@ func TestNetworkUpgradesOverridden(t *testing.T) {
 	// verify upgrade overrides
 	require.False(t, restartedVM.chainConfigExtra().IsGranite(uint64(upgrade.InitiallyActiveTime.Unix())))
 	require.True(t, restartedVM.chainConfigExtra().IsGranite(graniteTimestamp))
-	require.False(t, restartedVM.currentRules().IsGranite)
+	require.False(t, currentRules(tvm.vm).IsGranite)
```
StephenButtolph, line 214: This seems to have changed which vm we are getting the rules from.
```suggestion
	require.False(t, currentRules(restartedVM).IsGranite)
```

plugin/evm/vm_upgrade_bytes_test.go
```diff
@@ -411,3 +412,9 @@ func TestVMEtnaActivatesCancun(t *testing.T) {
 		})
 	}
 }
+
+// currentRules returns the chain rules for the current block.
+func currentRules(vm *VM) extras.Rules {
```
StephenButtolph, line 417: We don't necessarily need to do this, but we could avoid any other changes in the tests if we just did:
```suggestion
func (vm *VM) currentRules() extras.Rules {
```
which aligns with the prior implementation.

StephenButtolph (approved):

PR #1823, opened 2025-10-23
Remove `validators` package and use AvalancheGo `uptimetracker`

JonathanOppenheimer:
## Why this should be merged
With the [successful uplift ](https://github.com/ava-labs/avalanchego/pull/4166)of the `validators` package to AvalancheGo, we can now remove the original package from subnet-evm. I called attention to specific changes in the comments below. 

Note that as AvalancheGo is on Go `v1.24.9`, this PR is blocked by https://github.com/ava-labs/subnet-evm/pull/1821, which should be merged before this PR. 

## How this was tested
Exiting CI

## Need to be documented?
No

No

### DIFF
--- a/plugin/evm/vm_uptime_test.go
+++ b/plugin/evm/vm_uptime_test.go
@@ -42,59 +42,36 @@ func TestUptimeTracker(t *testing.T) {
 		}
 	}
 
-	t.Run("uptime not tracked before NormalOp", func(t *testing.T) {
-		require := require.New(t)
-		ctx, dbManager, genesisBytes := setupGenesis(t, upgradetest.Latest)
-		ctx.ValidatorState = makeValidatorState()
-		appSender := &enginetest.Sender{T: t}
+	require := require.New(t)
+	ctx, dbManager, genesisBytes := setupGenesis(t, upgradetest.Latest)
+	ctx.ValidatorState = makeValidatorState()
+	appSender := &enginetest.Sender{T: t}
 
-		vm := &VM{}
-		require.NoError(vm.Initialize(
-			context.Background(),
-			ctx,
-			dbManager,
-			genesisBytes,
-			[]byte(""),
-			[]byte(""),
-			[]*commonEng.Fx{},
-			appSender,
-		))
-		defer vm.Shutdown(context.Background())
+	vm := &VM{}
+	require.NoError(vm.Initialize(
+		context.Background(),
+		ctx,
+		dbManager,
+		genesisBytes,
+		[]byte(""),
+		[]byte(""),
+		[]*commonEng.Fx{},
+		appSender,
+	))
+	defer vm.Shutdown(context.Background())
 
-		require.NoError(vm.SetState(context.Background(), snow.Bootstrapping))
+	require.NoError(vm.SetState(context.Background(), snow.Bootstrapping))
 
-		// After bootstrapping but before NormalOp, uptimeTracker hasn't started syncing yet
-		_, _, found, err := vm.uptimeTracker.GetUptime(testValidationID)
-		require.NoError(err)
-		require.False(found, "uptime should not be tracked yet")
-	})
+	// After bootstrapping but before NormalOp, uptimeTracker hasn't started syncing yet
+	_, _, found, err := vm.uptimeTracker.GetUptime(testValidationID)
+	require.NoError(err)
+	require.False(found, "uptime should not be tracked yet")
 
-	t.Run("uptime tracked after NormalOp", func(t *testing.T) {
-		require := require.New(t)
-		ctx, dbManager, genesisBytes := setupGenesis(t, upgradetest.Latest)
-		ctx.ValidatorState = makeValidatorState()
-		appSender := &enginetest.Sender{T: t}
+	require.NoError(vm.SetState(context.Background(), snow.NormalOp))
 
-		vm := &VM{}
-		require.NoError(vm.Initialize(
-			context.Background(),
-			ctx,
-			dbManager,
-			genesisBytes,
-			[]byte(""),
-			[]byte(""),
-			[]*commonEng.Fx{},
-			appSender,
-		))
-		defer vm.Shutdown(context.Background())
-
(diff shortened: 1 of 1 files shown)


plugin/evm/vm_validators_test.go
```diff
@@ -78,22 +76,21 @@ func TestValidatorState(t *testing.T) {
...
+	_, _, found, err := vm.uptimeTracker.GetUptime(testValidationIDs[0])
+	require.NoError(err)
+	require.False(found, "uptime should not be tracked yet")
 
 	// Test case 2: state should be populated after bootstrapped
 	require.NoError(vm.SetState(context.Background(), snow.NormalOp))
-	require.Len(vm.validatorsManager.GetValidationIDs(), 3)
-	_, _, err = vm.validatorsManager.CalculateUptime(testNodeIDs[0])
+	// Give the sync goroutine time to run
+	time.Sleep(2 * time.Second)
```
JonathanOppenheimer: is this okay?

plugin/evm/vm_validators_test.go
```diff
@@ -76,84 +78,50 @@ func TestValidatorState(t *testing.T) {
 	)
 	require.NoError(err, "error initializing GenesisVM")
 
-	// Test case 1: state should not be populated until bootstrapped
```
JonathanOppenheimer: this test isn't really super relevant anymore as `state` is no longer externally exposed -- this changes a lot of what can be tested (and a lot of tested in AvalancheGo now), so this test is kind of awkward now. thoughts?

plugin/evm/vm.go
```diff
@@ -776,16 +775,23 @@ func (vm *VM) onNormalOperationsStarted() error {
...
-	}
-
-	// dispatch validator set update
+	// Start syncing validator set updates
 	vm.shutdownWg.Add(1)
 	go func() {
-		vm.validatorsManager.DispatchSync(ctx, &vm.vmLock)
-		vm.shutdownWg.Done()
+		defer vm.shutdownWg.Done()
+		ticker := time.NewTicker(time.Second)
```
alarso16: So you wait one second and then call `Sync`? Why?

plugin/evm/vm_validators_test.go
```diff
@@ -76,84 +78,50 @@ func TestValidatorState(t *testing.T) {
...
+	// Test case 1: uptime should not be tracked until NormalOp
```
alarso16: Are these tests independent? Because it seems like it, so it seems more intuitive to me that test1 failing shouldn't make test3 never run

plugin/evm/vm_validators_test.go
```diff
@@ -42,118 +30,181 @@ func TestValidatorState(t *testing.T) {
...
+
+		// Manually call Sync to ensure state is updated after Connect
+		require.NoError(vm.uptimeTracker.Sync(context.Background()))
+
+		_, _, found, err := vm.uptimeTracker.GetUptime(testValidationIDs[0])
+		require.NoError(err)
+		require.True(found, "uptime should be tracked after validators are connected")
+	})
+
+	t.Run("uptime persisted across restarts", func(t *testing.T) {
```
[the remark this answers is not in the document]
JonathanOppenheimer: I will remove this

warp/verifier_backend_test.go
```diff
@@ -326,41 +326,5 @@ func TestUptimeSignatures(t *testing.T) {
 		protoBytes, _ = getUptimeMessageBytes([]byte{}, vID)
 		_, appErr = handler.AppRequest(context.Background(), ids.GenerateTestNodeID(), time.Time{}, protoBytes)
 		require.ErrorIs(t, appErr, &common.AppError{Code: VerifyErrCode})
-		require.Contains(t, appErr.Error(), "failed to get validator")
```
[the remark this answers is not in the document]
JonathanOppenheimer: I restored these 3 tests - my apologies, didn't mean to delete them.

plugin/evm/service.go
```diff
@@ -18,61 +18,69 @@ type ValidatorsAPI struct {
...
+		}
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to get uptime for validation ID %s: %w", validationID, err)
 		}
+
 		var uptimeFloat float64
-		startTime := time.Unix(int64(validator.StartTimestamp), 0)
+		var uptimePercentage float32
+		var uptimeSeconds uint64
```
StephenButtolph: We don't need to define these as `var`s anymore.

plugin/evm/vm_uptime_test.go
```diff
@@ -0,0 +1,141 @@
...
+		))
+		defer vm.Shutdown(context.Background())
+
+		require.NoError(vm.SetState(context.Background(), snow.Bootstrapping))
+		require.NoError(vm.SetState(context.Background(), snow.NormalOp))
+
+		// Connect the validators to start tracking their uptime
+		for _, nodeID := range testNodeIDs {
+			require.NoError(vm.uptimeTracker.Connect(nodeID))
+		}
```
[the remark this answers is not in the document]
StephenButtolph: I don't think we are testing `Connect` in this test (outside of testing that it doesn't panic). If we were testing `Connect` then removing it from the test should necessarily cause the test to fail

warp/verifier_backend_test.go
```diff
@@ -295,19 +300,49 @@ func TestUptimeSignatures(t *testing.T) {
...
+		validatorState := &validatorstest.State{
+			GetMinimumHeightF: func(context.Context) (uint64, error) {
+				return 0, nil
+			},
+			GetCurrentHeightF: func(context.Context) (uint64, error) {
+				return 0, nil
+			},
+			GetSubnetIDF: func(context.Context, ids.ID) (ids.ID, error) {
+				return snowCtx.SubnetID, nil
+			},
```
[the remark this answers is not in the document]
JonathanOppenheimer: I won't be offended if you just phrase your comments as observations rather than questions 😂. removed.

PR #1827, opened 2025-10-29
feat: Add solc-based contract compilation and Go binding generation (Phase 1)

JonathanOppenheimer:
## Why this should be merged
This is the first phase of the `npm` to Go testing migration. It adds native solc compilation and Go binding generation while
maintaining existing Hardhat-based TypeScript tests.

This PR 
- Add OpenZeppelin contracts as git submodule
- Add contracts/compile.go with solc compilation directives
- Add contracts/bindings/bindings.go with abigen directives  
- Update CI to install solc via setup-solc@v0.2.0
- Update Taskfile.yml to run both Hardhat and solc compilation
- Update .gitignore to exclude generated artifacts and bindings
- Update contracts/README.md with new compilation instructions

## How this was tested
This just needs to run successfully in CI, it doesn't directly impact tests yet

## Need to be documented?
Yes, done 

No

### DIFF
--- a/scripts/lint.sh
+++ b/scripts/lint.sh
@@ -38,6 +38,7 @@ function read_dirs {
     ! -name 'mock.go'
     ! -name 'gen_*.go'
     ! -path './**/*mock/*.go'
+    ! -path './contracts/bindings/*.go'
   )
 
   # Combined loop: build both upstream licensed find and exclude args


.github/workflows/ci.yml
```diff
@@ -59,10 +59,15 @@ jobs:
...
           fetch-depth: 0
+          submodules: recursive
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
           go-version-file: "go.mod"
+      - name: Set up solc
+        uses: ARR4N/setup-solc@v0.2.0
+        with:
+          version: '0.8.30'
```
ARR4N:
```suggestion
          versions: '0.8.30'
```

The singular will install it as `solc`, which doesn't lock in a version at the point of use. The plural will force use of a specific version as `solc-v0.8.30`, which is important when checking in generated code.

.github/workflows/ci.yml
```diff
@@ -87,6 +92,11 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
+          submodules: recursive
+      - name: Set up solc
+        uses: ARR4N/setup-solc@v0.2.0
+        with:
+          version: '0.8.30'
```
ARR4N:
```suggestion
          versions: '0.8.30'
```

contracts/bindings/bindings.go
```diff
@@ -0,0 +1,26 @@
...
+//go:generate go run github.com/ava-labs/libevm/cmd/abigen --pkg bindings --type Example --abi ../artifacts/Example.abi --bin ../artifacts/Example.bin --out example.go
+//go:generate go run github.com/ava-labs/libevm/cmd/abigen --pkg bindings --type ExampleFeeManager --abi ../artifacts/ExampleFeeManager.abi --bin ../artifacts/ExampleFeeManager.bin --out examplefeemanager.go
+//go:generate go run github.com/ava-labs/libevm/cmd/abigen --pkg bindings --type ExampleRewardManager --abi ../artifacts/ExampleRewardManager.abi --bin ../artifacts/ExampleRewardManager.bin --out examplerewardmanager.go
+//go:generate go run github.com/ava-labs/libevm/cmd/abigen --pkg bindings --type ExampleTxAllowList --abi ../artifacts/ExampleTxAllowList.abi --bin ../artifacts/ExampleTxAllowList.bin --out exampletxallowlist.go
+//go:generate go run github.com/ava-labs/libevm/cmd/abigen --pkg bindings --type ExampleWarp --abi ../artifacts/ExampleWarp.abi --bin ../artifacts/ExampleWarp.bin --out examplewarp.go
+
+// Generate bindings for interface contracts
```
ARR4N:
```suggestion
```

contracts/bindings/bindings.go
```diff
@@ -0,0 +1,26 @@
...
+//go:generate go run github.com/ava-labs/libevm/cmd/abigen --pkg bindings --type IAllowList --abi ../artifacts/IAllowList.abi --bin ../artifacts/IAllowList.bin --out iallowlist.go
```
ARR4N: If the contracts implement the interfaces then there's no need to generate bindings for them. Please confirm that they do and, if so, remove all of these.

contracts/submodules/openzeppelin-contracts
ARR4N: The idiomatic name for this directory is `lib`, not `submodules`.

contracts/compile.go
```diff
@@ -0,0 +1,26 @@
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+// Package contracts provides contract compilation directives.
+// This file contains go:generate directives to compile Solidity contracts using solc.
+package contracts
+
+// Compile main contracts
```
ARR4N:
```suggestion
```

contracts/compile.go
```diff
@@ -0,0 +1,26 @@
...
+//go:generate solc -o ./artifacts --overwrite --abi --bin --base-path . contracts/ERC20NativeMinter.sol @openzeppelin/contracts/=submodules/openzeppelin-contracts/contracts/
+//go:generate solc -o ./artifacts --overwrite --abi --bin --base-path . contracts/ExampleDeployerList.sol @openzeppelin/contracts/=submodules/openzeppelin-contracts/contracts/
+//go:generate solc -o ./artifacts --overwrite --abi --bin --base-path . contracts/ExampleFeeManager.sol @openzeppelin/contracts/=submodules/openzeppelin-contracts/contracts/
+//go:generate solc -o ./artifacts --overwrite --abi --bin --base-path . contracts/ExampleRewardManager.sol @openzeppelin/contracts/=submodules/openzeppelin-contracts/contracts/
+//go:generate solc -o ./artifacts --overwrite --abi --bin --base-path . contracts/ExampleTxAllowList.sol @openzeppelin/contracts/=submodules/openzeppelin-contracts/contracts/
+//go:generate solc -o ./artifacts --overwrite --abi --bin --base-path . contracts/ExampleWarp.sol @openzeppelin/contracts/=submodules/openzeppelin-contracts/contracts/
```
ARR4N:
```suggestion
//go:generate solc -o ./artifacts --overwrite --abi --bin --base-path . @openzeppelin/contracts/=submodules/openzeppelin-contracts/contracts/ contracts/AllowList.sol contracts/ERC20NativeMinter.sol contracts/ExampleDeployerList.sol contracts/ExampleFeeManager.sol contracts/ExampleRewardManager.sol contracts/ExampleTxAllowList.sol contracts/ExampleWarp.sol
```

Taskfile.yml
```diff
@@ -123,12 +123,16 @@ tasks:
...
+    desc: Set up contracts by compiling Solidity contracts and generating Go bindings
     dir: ./contracts
     cmds:
+      # Keep npm/Hardhat compilation for existing TypeScript tests (Phase 2-3 migration)
       - cmd: npm ci
       - cmd: npx hardhat clean
       - cmd: npx hardhat compile
+      # New: Compile contracts with solc and generate Go bindings
+      - cmd: go generate ./compile.go
+      - cmd: go generate ./bindings/bindings.go
```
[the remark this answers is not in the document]
ARR4N: > what's the rationale for this change?

Why force explicit listing of every file that has a `//go:generate` in it when we know that we always want all of them?

Taskfile.yml
```diff
@@ -123,12 +123,16 @@ tasks:
...
+    desc: Set up contracts by compiling Solidity contracts and generating Go bindings
     dir: ./contracts
     cmds:
+      # Keep npm/Hardhat compilation for existing TypeScript tests (Phase 2-3 migration)
       - cmd: npm ci
       - cmd: npx hardhat clean
       - cmd: npx hardhat compile
+      # New: Compile contracts with solc and generate Go bindings
+      - cmd: go generate ./contracts/compile.go
+      - cmd: go generate ./bindings/bindings.go
```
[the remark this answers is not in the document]
ARR4N: No need to block this further so let's merge, but requiring specific ordering is a [smell](https://en.wikipedia.org/wiki/Code_smell) suggesting something should change (e.g. putting all `go:generate` in the same file, or checking in generated files).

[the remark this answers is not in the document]
JonathanOppenheimer: Given how tightly coupled these are, I think a single file makes more sense. Then we can still use `./...`. Thanks for the explanation!

PR #1830, opened 2025-11-03
feat: Add go test utilities for precompiles and migrate one test (Phase 2)

JonathanOppenheimer:
## Why this should be merged
This is phase 2 of the conversion from a `npm`/typescript based testing framework to a go based framework. 

To start, let me provide some background:  there are two basic approaches we could take here in transitioning the framework: 

**Option A**: keep existing tmpnet infrastructure
- Use existing `tests/utils/subnet.go` infrastructure
- Deploy real subnet with precompile-enabled genesis
- RPC 
- _This is what TypeScript tests currently do_

**Option B**: configure precompiles in `libevm`'s `simulated.Backend` directly and test 
- More performant 
- No RPC/ginkgo

Upon investigation, I determined that Option B was not possible without modifications to `libevm`:

The `libevm` simulated backend always boots with the stock dev genesis ([params.AllDevChainProtocolChanges](https://github.com/ava-labs/libevm/blob/1bccf4f2ddb28ff440f9e24f46506a2e98ad2a0d/core/genesis.go#L541)) and never passes through any Avalanche-specific extras. You can see the hard-coded setup in [libevm/ethclient/simulated/backend.go](https://github.com/ava-labs/libevm/blob/main/ethclient/simulated/backend.go) (NewBackend), where the [default Genesis config](https://github.com/ava-labs/libevm/blob/1bccf4f2ddb28ff440f9e24f46506a2e98ad2a0d/ethclient/simulated/backend.go#L83) is created with plain libevm params -- there’s no call into subnet-evm’s extras or precompile registration.

Even if we changed that to inject a `subnet-evm` `ChainConfig`, `libevm`’s core genesis logic has no idea what "GenesisPrecompiles" are. The go file in `libevm` simply initializes accounts and code; there’s no hook that walks the config and installs stateful precompiles. The `subnet-evm` tree explicitly does that work (it calls [ApplyPrecompileActivations](https://github.com/ava-labs/subnet-evm/blob/65f4d47b3929ce5028668ea73c26893a02a0d235/core/state_processor_ext.go#L26) during genesis creation), but that function doesn’t exist in the libevm copy. Without it, the precompile contracts never get written to state, and as a result, every `0x0200…` address still looks empty and the bindings revert with “no contract code”.  I saw this first hand when trying to test it out.

In the absence of a better way to test these with libevm's `simulated.Backend,` I moved forward with Option A.  

## How this works
- `tests/precompile/contracttest/helpers.go` - Test backend setup, utilities
- `tests/precompile/contracttest/roles.go` - AllowList role management  (needs tmpnet for precompiles)
- `tests/precompile/contracttest/tmpnet.go` - Tmpnet/RPC backend helpers. 
- `tests/precompile/contracttest/example_test.go` - Examples (simulated backend)

I also migrated a single test `contract_deployer_allow_list_test,` as a proof of concept for this approach. I was largely able to match the format of the typescript test. Here is the mapping of tests to show no coverage was lost:  

| TypeScript Test Name | Solidity Method | Go Test Name |
|---------------------|----------------|--------------|
| `test("precompile should see owner address has admin role")` | `step_verifySenderIsAdmin` | `ginkgo.It("should verify deployer list shows admin has admin role via precompile")` |
| `test("precompile should see test address has no role")` | `step_newAddressHasNoRole` | `ginkgo.It("should verify new address has no role")` |
| `test("contract should report test address has no admin role")` | `step_noRoleIsNotAdmin` | `ginkgo.It("should verify contract correctly reports admin status")` (partial) |
| `test("contract should report owner address has admin role")` | `step_ownerIsAdmin` | `ginkgo.It("should verify contract correctly reports admin status")` (combined) |
| `test("should not let test address deploy")` | `step_noRoleCannotDeploy` | `ginkgo.It("should not let address with no role deploy contracts")` |
| `test("should allow admin to add contract as admin")` | `step_adminAddContractAsAdmin` | `ginkgo.It("should allow admin to add contract as admin via precompile")` |
| `test("should allow admin to add deployer address as deployer through contract")` | `step_addDeployerThroughContract` | `ginkgo.It("should allow admin to add deployer via contract")` |
| `test("should let deployer address to deploy")` | `step_deployerCanDeploy` | `ginkgo.It("should allow enabled address to deploy contracts")` |
| `test("should let admin revoke deployer")` | `step_adminCanRevokeDeployer` | `ginkgo.It("should allow admin to revoke deployer role")` |

I have also called out specific things of note via comments below. The next steps are to migrate each of the additional 6 tests, one at a time. 

## How this was tested
CI.  We deliberately kept the Hardhat bridge in place so the TypeScript suites can run beside the new Go tests during migration. The tmpnet-backed Go suites (labeled `Go`) now execute via Ginkgo in CI, but `RunHardhatTests()` and the async TypeScript descriptors remain until Phase 3 completes the rewrites, and in Phase 4, we remove all the npm code in one go. 

## Need to be documented?
No

No.

### DIFF
--- a/tests/precompile/contracttest/tmpnet.go
+++ b/tests/precompile/contracttest/tmpnet.go
@@ -88,9 +88,9 @@ func NewTmpnetBackend(t testing.TB, rpcURL string) *TmpnetBackend {
 }
 
 // Close closes the client connection
-func (tb *TmpnetBackend) Close() {
-	if tb.Client != nil {
-		tb.Client.Close()
+func (be *TmpnetBackend) Close() {
+	if be.Client != nil {
+		be.Client.Close()
 	}
 }
 


contracts/contracts/compile.go
```diff
@@ -4,13 +4,21 @@
 package contracts
 
 // Step 1: Compile Solidity contracts to generate ABI and bin files
-//go:generate solc-v0.8.30 -o ../artifacts --overwrite --abi --bin --base-path . @openzeppelin/contracts/=../node_modules/@openzeppelin/contracts/ AllowList.sol ERC20NativeMinter.sol ExampleDeployerList.sol ExampleFeeManager.sol ExampleRewardManager.sol ExampleTxAllowList.sol ExampleWarp.sol
+//go:generate solc-v0.8.30 -o ../artifacts --overwrite --abi --bin --base-path . --evm-version paris @openzeppelin/contracts/=../node_modules/@openzeppelin/contracts/ AllowList.sol ERC20NativeMinter.sol ExampleDeployerList.sol ExampleFeeManager.sol ExampleRewardManager.sol ExampleTxAllowList.sol ExampleWarp.sol interfaces/IAllowList.sol interfaces/IFeeManager.sol interfaces/INativeMinter.sol interfaces/IRewardManager.sol interfaces/IWarpMessenger.sol
```
JonathanOppenheimer: We are compiling to `paris`, rather than `shanghai` because the `PUSH0` opcode introduced in Shanghai causes "invalid opcode" errors. This seems fine for now, but as the tests are about behavior, not gas optimization, but this should be investigated further. I think that perhaps libevm's backend stuff is out of date? 

This is why all the tests are slightly larger in bytecode is because we do not have this optimization anymore.

tests/precompile/contract_deployer_allow_list_test.go
```diff
@@ -0,0 +1,328 @@
...
+)
+
+// Register Go-based contract deployer allow list tests
+// Tests within this suite run serially to avoid nonce conflicts on the same subnet
+func init() {
+	ginkgo.Describe("[Go] Contract Deployer Allow List", ginkgo.Serial, ginkgo.Label("Precompile", "ContractDeployerAllowList", "Go"), func() {
+		var (
+			require        *requirePkg.Assertions
+			backend        *contracttest.TmpnetBackend
+			allowList      *bindings.IAllowList
```
JonathanOppenheimer: Here is where the interface binding is needed.

scripts/run_ginkgo_precompile.sh
```diff
@@ -17,6 +17,10 @@ TEST_SOURCE_ROOT=$(pwd)
...
-TEST_SOURCE_ROOT="$TEST_SOURCE_ROOT" "${SUBNET_EVM_PATH}"/bin/ginkgo run -procs=5 tests/precompile \
-  --ginkgo.vv \
-  --ginkgo.label-filter="${GINKGO_LABEL_FILTER:-""}"
+# Use GINKGO_LABEL_FILTER environment variable to filter by label (e.g., GINKGO_LABEL_FILTER="Go")
+LABEL_FILTER_FLAG=""
+if [ -n "${GINKGO_LABEL_FILTER:-}" ]; then
+  LABEL_FILTER_FLAG="--label-filter=${GINKGO_LABEL_FILTER}"
+fi
+
+TEST_SOURCE_ROOT="$TEST_SOURCE_ROOT" "${SUBNET_EVM_PATH}"/bin/ginkgo run -procs=5 --vv ${LABEL_FILTER_FLAG} tests/precompile
```
JonathanOppenheimer: This is for local testing, and will be removed once the transition is complete.

tests/precompile/contracttest/example_test.go
```diff
@@ -0,0 +1,56 @@
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
```
JonathanOppenheimer: This file is an example of how testing can be done directly on libevm's `simulated.Backend,` without doing a tmpnet, RPC, etc. 

It's cleaner and faster to execute. It's also currently unworkable.

tests/precompile/contracttest/tmpnet.go
```diff
@@ -0,0 +1,228 @@
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
```
JonathanOppenheimer: A lot of these functions are unused, but will be as we convert more tests to go. Here's an overview:

```
func NewTmpnetBackendSimple(rpcURL string) *TmpnetBackend { … }
…
func NewTmpnetBackend(t testing.TB, rpcURL string) *TmpnetBackend { … }
…
func WaitForTxReceipt(t testing.TB, client *ethclient.Client, tx *types.Transaction) *types.Receipt { … }
…
func SetupAllowListRoleOnTmpnet( … ) { … }
…
func GetAllowListRoleOnTmpnet( … ) uint8 { … }
``` 

- `NewTmpnetBackend(t testing.TB, …)` mirrors the simulated-backend helper (`NewTestBackend`) so the remaining TypeScript suites can move over without inventing new patterns.
- `WaitForTxReceipt`, `SetupAllowListRoleOnTmpnet`, and `GetAllowListRoleOnTmpnet `port the DS-Test helpers the TS suite uses ([`test/utils.ts`](https://github.com/ava-labs/subnet-evm/blob/master/contracts/test/utils.ts)) into Go. They aren’t consumed by the first migrated suite yet, but they will be once we migrate the scenarios that in TS call `setAllowListRole`, `requireRole`, etc. 

We'll need these functions in the remaining precompile test suites.

tests/precompile/contracttest/example_test.go
```diff
@@ -0,0 +1,57 @@
...
+	SetupAllowListRole(t, backend, ContractDeployerAllowListAddress,
+		backend.Admin.Address, RoleAdmin, backend.Admin)
+
+	// Verify the role was set
+	RequireRole(t, backend, ContractDeployerAllowListAddress,
+		backend.Admin.Address, RoleAdmin)
+
+	// Verify other address has no role
+	RequireRole(t, backend, ContractDeployerAllowListAddress,
+		backend.OtherAddr.Address, RoleNone)
```
ARR4N: As above re levels.

tests/precompile/contracttest/helpers.go
```diff
@@ -0,0 +1,142 @@
...
+
+var _ bind.ContractBackend = (simulated.Client)(nil)
+
+// TestBackend wraps a simulated backend with common test utilities
+type TestBackend struct {
+	*simulated.Backend
+	Accounts
+}
+
+// Common test accounts matching the TypeScript test setup
```
ARR4N: There won't be a TypeScript setup when someone reads this in a year.

tests/precompile/contracttest/helpers.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+// TestAccount represents a funded test account
+type TestAccount struct {
+	Key     *ecdsa.PrivateKey
+	Address common.Address
+	Auth    *bind.TransactOpts
+}
+
+// Common addresses from TypeScript tests
```
ARR4N: As above re TypeScript.

tests/precompile/contracttest/helpers.go
```diff
@@ -0,0 +1,142 @@
...
+		switch addr {
+		case AdminAddress:
+			fundedAccounts.Admin = account
+		case OtherAddress:
+			fundedAccounts.OtherAddr = account
+		}
+	}
+
+	// Create simulated backend from libevm (uses AllDevChainProtocolChanges by default)
+	backend := simulated.NewBackend(alloc)
```
ARR4N: There's no need for a variable. Just add this straight to the struct literal.

tests/precompile/contracttest/helpers.go
```diff
@@ -0,0 +1,142 @@
...
+		if account.Address == addr {
+			return account
+		}
+	}
+	return nil
+}
+
+// WaitForReceipt waits for a transaction receipt and commits a block if needed
+func WaitForReceipt(t testing.TB, tb *TestBackend, tx *types.Transaction) *types.Receipt {
+	require := require.New(t)
```
ARR4N:
```suggestion
```

tests/precompile/contracttest/example_test.go
```diff
@@ -0,0 +1,56 @@
...
+	require.NotNil(contract, "contract should not be nil")
+
+	// Example: Check owner (should be admin account)
+	owner, err := contract.Owner(nil)
+	require.NoError(err)
+	require.Equal(backend.Admin.Address, owner, "owner should be admin account")
+}
+
+// TestAllowListRoles demonstrates role management
+func TestAllowListRoles(t *testing.T) {
```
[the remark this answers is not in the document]
JonathanOppenheimer: See this comment to Austin which hopefully provides more context to the original post: 

https://github.com/ava-labs/subnet-evm/pull/1830#issuecomment-3486912737

tests/precompile/contracttest/tmpnet.go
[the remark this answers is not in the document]
JonathanOppenheimer: See my original PR, post, [this comment thread ](https://github.com/ava-labs/subnet-evm/pull/1830#discussion_r2487828813), and my [reply to Austin asking the same thing](https://github.com/ava-labs/subnet-evm/pull/1830#issuecomment-3486912737).

In short: personal skill issue. I don't think it works.

contracts/contracts/compile.go
JonathanOppenheimer: Sounds good! Once that gets merged, I'll change the target to Shanghai which will match the current tests.

tests/precompile/contracttest/helpers.go
```diff
@@ -0,0 +1,142 @@
...
+	for _, account := range tb.FundedKeys {
+		if account.Address == addr {
+			return account
+		}
+	}
+	return nil
+}
+
+// WaitForReceipt waits for a transaction receipt and commits a block if needed
+func WaitForReceipt(t testing.TB, tb *TestBackend, tx *types.Transaction) *types.Receipt {
```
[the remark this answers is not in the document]
JonathanOppenheimer: @joshua-kim argued against this in general for reasons I can't quite remember. I'll make a habit of including it.

tests/precompile/contracttest/example_test.go
```diff
@@ -0,0 +1,57 @@
...
+	require := require.New(t)
+
+	// Create test backend with funded accounts
+	backend := NewTestBackend(t)
+	defer backend.Close()
+
+	// Deploy ExampleDeployerList contract using generated binding
+	addr, tx, contract, err := bindings.DeployExampleDeployerList(backend.Admin.Auth, backend.Client())
+	require.NoError(err, "failed to deploy contract")
+	require.NotEqual(addr.Hex(), "0x0000000000000000000000000000000000000000", "contract address should not be zero")
```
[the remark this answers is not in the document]
JonathanOppenheimer: We can!

PR #1843, opened 2025-11-13
chore: Use go 1.24's 'go tool' support for CLI tooling

alarso16:
## Why this should be merged

Cherry-picks ava-labs/coreth#1317. A pre-req for repo merger

## How this works

Same as coreth

## How this was tested

CI

## Need to be documented?

No

No


JonathanOppenheimer (approved):
The line differences occur only in our ci.yml, the taskfile, and the contributing guide. This is otherwise a cherry-pick.

PR #1853, opened 2025-11-18
sync: coreth PR #1391: ci: Add lint-fix to taskfile

alarso16:
## Why this should be merged

Allows easier linter integration. Cherry-picks ava-labs/coreth#1391, but not that simple. There were some differences between the lint scripts in coreth and subnet-evm, which uncovered some improperly named generated files. 

## How this works

Cherry-picks from coreth.

## How this was tested

Run `./scripts/lint_fix.sh`

## Need to be documented?

No

No


scripts/lint_setup.sh
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/env bash
```
JonathanOppenheimer: cherry pick

PR #1865, opened 2025-11-19
Add geth license to `blockchain_test.go` 

JonathanOppenheimer:
## Why this should be merged
This is not our code - we should not pass it off as ours.


alarso16 (changes requested):
Remove it from `scripts/upstream_files.txt`

JonathanOppenheimer:
> Remove it from `scripts/upstream_files.txt`

done

PR #1271, opened 2024-08-08
refactor: allow `types` to be imported by `precompile/contract`

ARR4N:
## Why this should be merged

Importing `core/types` from `precompile/{modules,contract}` results in a circular dependency because `types` -> `params` -> `precompile/modules` -> `precompile/contract` -> `types`.

## How this works

`precompile/modules` only depended on `precompile/contract` for interfaces; by moving the interfaces into the new `precompile/interfaces` package, there is no longer a transitive dependency `params -> precompile/modules -> precompile/contract`.

Note that this only allows importing `core/types` from `precompile/contract` but not from `precompile/modules`. This is OK as the latter is merely a registry that has implementations _injected_ into it.

Visualising the dependency graph (raw Graphviz below) with:

```shell
$ go install github.com/loov/goda@latest
$ goda graph -cluster -short ./params ./core/types ./precompile/{contract,modules,precompileconfig} | dot -Tpng > depgraph.png
```

### Before

image

### After (as at `12febd1` PoC)

image

### Dependency graph at `caf34ea` master:

```dot
digraph G {
    node [penwidth=2 fontsize=10 shape=rectangle target="_graphviz"];
    edge [tailport=e penwidth=2];
    compound=true;
    rankdir=LR;
    newrank=true;
    ranksep="1.5";
    quantum="0.5";
subgraph "cluster_./precompile/interfaces" {
    label="./precompile/interfaces"
    tooltip="./precompile/interfaces"
    href="https://pkg.go.dev/./precompile/interfaces"
    "./precompile/interfaces" [label="" tooltip="./precompile/interfaces" shape=point color="#07918cb2" rank=0];
}
subgraph "cluster_github.com/ava-labs/subnet-evm" {
    label="github.com/ava-labs/subnet-evm (local)"
    tooltip="github.com/ava-labs/subnet-evm (local)"
    href="https://pkg.go.dev/github.com/ava-labs/subnet-evm@"
    "github.com/ava-labs/subnet-evm/core/types" [label="core/types\l3822 / 130.6KB\l" tooltip="github.com/ava-labs/subnet-evm/core/types" href="https://pkg.go.dev/github.com/ava-labs/subnet-evm/core/types" color="#910750b2"];
    "github.com/ava-labs/subnet-evm/params" [label="params\l1669 / 78.0KB\l" tooltip="github.com/ava-labs/subnet-evm/params" href="https://pkg.go.dev/github.com/ava-labs/subnet-evm/params" color="#915107b2"];
[...]

PR #1886, opened 2025-12-02
sync: coreth PR #1271: style: enable depguard linter

JonathanOppenheimer:
Syncs https://github.com/ava-labs/coreth/pull/1271

- Requesting @alarso16 for review as he was the first approver 
- Requesting @StephenButtolph for review as he was the merging approver.


sync/handlers/stats/statstest/test_stats.go
```diff
@@ -1,17 +1,19 @@
 // Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
```
JonathanOppenheimer: This is part of the alignment with coreth

JonathanOppenheimer:
> Compared against the corresponding coreth PR, this all LGTM.
> 
> Question though: Why/how is it that `github.com/stretchr/testify/assert` is still imported within the `accounts/abi` and `eth` packages?

`accounts/abi` and `eth` are upstream, old geth packages, and we don't enforce the AvalancheGo linting configuration on upstream files (so assert is okay). Only the more limited `.golangci.yml` is enforced, which does not contain `depguard`.

alarso16 (approved):

ceyonur (approved):

PR #1905, opened 2025-12-04
chore: fix shellcheck directives

JonathanOppenheimer:
## Why this should be merged

These failures broke `./scripts/shellcheck.sh`, which was not caught in this repository's CI for some reason. This PR adds the correct directives. 

## How this was tested
Locally (CI doesn't catch)

## Need to be documented?
No

No

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -24,4 +24,5 @@ fi
 
 # Build Subnet EVM, which is run as a subprocess
 echo "Building Subnet EVM @ GitCommit: $SUBNET_EVM_COMMIT at $BINARY_PATH"
+# shellcheck disable=SC2153 # STATIC_LD_FLAGS is defined in constants.sh
 go build -ldflags "-X github.com/ava-labs/subnet-evm/plugin/evm.GitCommit=$SUBNET_EVM_COMMIT $STATIC_LD_FLAGS" -o "$BINARY_PATH" "plugin/"*.go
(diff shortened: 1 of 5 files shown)


scripts/build.sh
```diff
@@ -24,4 +24,5 @@ fi
 
 # Build Subnet EVM, which is run as a subprocess
 echo "Building Subnet EVM @ GitCommit: $SUBNET_EVM_COMMIT at $BINARY_PATH"
+# shellcheck disable=SC2153 # STATIC_LD_FLAGS is defined in constants.sh
```
ceyonur: are we sure that these are needed? I don't see anything on the avalanchego/coreth side for these changes? Could it be your local using a different shellcheck version?

JonathanOppenheimer: I think these are needed! Check CI in the AvalancheGo graft PR

JonathanOppenheimer: I could also make these changes in the AvalancheGo PR.

alarso16: IMO you should not change the shellcheck script. You can just exclude the `graft/subnet-evm` directory in avalanchego's linter, and you can use subnet-evm's shellcheck for subnet-evm

alarso16: After migration, we can cleanup anything necessary

JonathanOppenheimer: sounds good!

PR #1907, opened 2025-12-05
test: convert warp npm test to go test

JonathanOppenheimer:
## Why this should be merged

This converts the testing present in `contracts/test/warp.ts` to go based testing, and removes the hard hat test. The typescript test really wasn't doing much -- just verifying the message and blockchain ID.  

You can inspect that test to ensure coverage was not lost here: 

https://github.com/ava-labs/subnet-evm/blob/master/contracts/test/warp.ts

## How this was tested

CI

## Need to be documented?
No

No


tests/warp/warp_test.go
```diff
@@ -171,8 +168,8 @@ var _ = ginkgo.Describe("[Warp]", func() {
 		log.Info("Delivering block hash payload to receiving subnet")
 		w.deliverBlockHashPayload()
 
-		log.Info("Executing HardHat test")
-		w.executeHardHatTest()
```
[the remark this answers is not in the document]
JonathanOppenheimer: I thought you wanted that in a separate PR?

ceyonur (dismissed):
We should add bindings tests here.

tests/warp/warp_test.go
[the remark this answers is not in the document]
JonathanOppenheimer: that's where the existing filtering test is done.

contracts/test/utils.ts
```diff
@@ -1,7 +1,3 @@
-import { ethers } from "hardhat"
```
JonathanOppenheimer: This file will not compile unless you delete most of it: 
Screenshot 2025-12-05 at 2 00 52 PM

contracts/index.ts
```diff
@@ -1 +1 @@
-export { test } from './test/utils';
```
JonathanOppenheimer: Just trying to get this to pass with a minimal difference in code.

ceyonur (dismissed):

tests/warp/warp_test.go
```diff
@@ -549,33 +589,92 @@ func (w *warpTest) deliverBlockHashPayload() {
...
+	require.NoError(err)
+	require.Equal(types.ReceiptStatusSuccessful, sendReceipt.Status)
+
+	log.Info("Filtering SendWarpMessage events using binding")
+	warpFilterer, err := warpbindings.NewIWarpMessengerFilterer(warp.Module.Address, client)
+	require.NoError(err)
+
+	// The event sender is the proxy contract , since the proxy calls the precompile
+	endBlock := sendReceipt.BlockNumber.Uint64()
+	iter, err := warpFilterer.FilterSendWarpMessage(
```
alarso16: Isn't all this code repeated with the above function `verifyAndExtractWarpMessage`?

JonathanOppenheimer: yup, I meant to reuse the helper, and THEN check the payload matches, but I did it in the wrong order. thanks for keeping me in check

tests/warp/warp_test.go
```diff
@@ -315,54 +346,66 @@ func (w *warpTest) sendMessageFromSendingSubnet() {
...
+	)
 	require.NoError(err)
+	defer iter.Close()
 
-	// Set local variables for the duration of the test
-	w.addressedCallUnsignedMessage = unsignedMsg
-	log.Info("Parsed unsignedWarpMsg", "unsignedWarpMessageID", w.addressedCallUnsignedMessage.ID(), "unsignedWarpMessage", w.addressedCallUnsignedMessage)
+	// Verify we got exactly one event with the correct data
+	require.True(iter.Next(), "expected at least one SendWarpMessage event")
+	event := iter.Event // event is *IWarpMessengerSendWarpMessage
```
alarso16: I don't think this comment is necessary since your IDE says this
```suggestion
	event := iter.Event
```

tests/warp/warp_test.go
```diff
@@ -315,54 +346,66 @@ func (w *warpTest) sendMessageFromSendingSubnet() {
...
+		[]common.Address{sender},
+		nil, // messageID filter: any
+	)
 	require.NoError(err)
+	defer iter.Close()
 
-	// Set local variables for the duration of the test
-	w.addressedCallUnsignedMessage = unsignedMsg
-	log.Info("Parsed unsignedWarpMsg", "unsignedWarpMessageID", w.addressedCallUnsignedMessage.ID(), "unsignedWarpMessage", w.addressedCallUnsignedMessage)
+	// Verify we got exactly one event with the correct data
```
alarso16:
```suggestion
```

I think that the `require` statement at the end of this function explains this sufficiently, but you could also document what this function does a little better

tests/warp/warp_test.go
```diff
@@ -315,54 +346,66 @@ func (w *warpTest) sendMessageFromSendingSubnet() {
...
+	require.True(iter.Next(), "expected at least one SendWarpMessage event")
```
alarso16:
```suggestion
	require.True(iter.Next(), "expected a SendWarpMessage event")
```
`at least` makes it seems like more than one is acceptable

plugin/evm/ExampleWarp.sol
```diff
@@ -2,7 +2,7 @@
 pragma solidity ^0.8.24;
```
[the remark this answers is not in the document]
JonathanOppenheimer: Yeah in the npm PR, I completely delete the `contracts/` folder, so it'd seem weird if we left it hanging around with a single example contract and the abigen files. I could make a subfolder for the three files?

alarso16 (approved):
I believe that we did choose the abigen approach. In this case, I think this test looks sufficient to check all the necessary behavior

tests/warp/warp_test.go
```diff
@@ -549,33 +589,92 @@ func (w *warpTest) deliverBlockHashPayload() {
...
+
+	log.Info("Filtering SendWarpMessage events using binding")
+	warpFilterer, err := warpbindings.NewIWarpMessengerFilterer(warp.Module.Address, client)
+	require.NoError(err)
+
+	// The event sender is the proxy contract , since the proxy calls the precompile
+	endBlock := sendReceipt.BlockNumber.Uint64()
+	iter, err := warpFilterer.FilterSendWarpMessage(
+		&bind.FilterOpts{
+			Start:   startBlock,
```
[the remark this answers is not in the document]
alarso16: This was resolved by the refactor

contracts/bindings/gen_examplewarp.go
michaelkaplan13: Am I correct that this was previously unused already?

JonathanOppenheimer: Yup!

plugin/evm/ExampleWarp.sol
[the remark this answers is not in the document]
JonathanOppenheimer: Yeah let me know what you'd like - my thought was a `plugin/evm/contracts` or something of that nature.

precompile/contracts/warp/warptest/bindings/WarpTest.sol
michaelkaplan13: Why do we need this file? Can't we just call the precompile directly from tests using the precompile ABI bindings?

JonathanOppenheimer: We could call the precompile directly, but I thought we had settled on using the contracts as a testing strategy (it's how we did all of the other precompile tests thus far). 

It exists so we can test contract to precompile calls, rather than an EOA calling it directly: 

e.g.
```
require.Equal(testPayload, addressedCall.Payload, "payload mismatch in warp message")
require.Equal(proxyAddr.Bytes(), addressedCall.SourceAddress, "source address should be proxy contract")
```

If this is no longer important to test, we can remove the contract (and we'd thus also have to remove some testing).

tests/warp/warp_test.go
```diff
@@ -315,54 +346,67 @@ func (w *warpTest) sendMessageFromSendingSubnet() {
...
-	logs, err := client.FilterLogs(ctx, ethereum.FilterQuery{
-		BlockHash: &blockHash,
-		Addresses: []common.Address{warp.Module.Address},
-	})
+// verifyAndExtractWarpMessage queries the given block for SendWarpMessage events
+// emitted by the specified sender using the IWarpMessengerFilterer binding. It
+// asserts that exactly one such event exists and then parses the unsigned warp
+// message from the event. The unsigned warp message is stored in
+// w.addressedCallUnsignedMessage for use in subsequent test steps.
+func (w *warpTest) verifyAndExtractWarpMessage(
```
michaelkaplan13: I think it would be cleaner and more useful have this a static helper rather than as a method of `warpTest`. It should return `avalancheWarp.UnsignedMessage`, which the caller can use however they'd like.

precompile/contracts/warp/warptest/bindings/WarpTest.sol
michaelkaplan13: After looking a bit more, I think we should get rid of this `WarpTest` contract entirely and just call the precompile directly using the ABI bindings for `IWarpMessenger.sol`. That tests the precompile interface, bindings, and implementation, which is exactly what we want. We do not need the proxy contract

michaelkaplan13: The reasoning stands though, I do not think we are trying to test contract to contract call logic itself here. My understanding is that is the equivalent of testing that solidity correctly ABI packs arguments to make the calls, which is out of this scope.

tests/warp/warp_test.go
alarso16: I agree

JonathanOppenheimer: makes sense - done!

precompile/contracts/warp/warptest/bindings/WarpTest.sol
ARR4N: > The reasoning stands though, I do not think we are trying to test contract to contract call logic itself here. My understanding is that is the equivalent of testing that solidity correctly ABI packs arguments to make the calls, which is out of this scope.

In recommending that we have an intermediary contract, my intention was that Solidity acts as the ground truth for a proper ABI encoding, only necessary if we expose a (Solidity) interface with which we expect someone to call the precompile. Assuming there is some `IPrecompile.sol` and a respective ABI, various scenarios for a Go `TestPrecompile`:

1. `TestPrecompile` -> `abigen` bindings -> precompile.
2. `TestPrecompile` -> intermediary contract[^1] -> `IPrecompile` -> precompile.

Since we expect an end user to consume `IPrecompile` in their own smart contract, we MUST be including it as part of the SUT (i.e. `IPrecompile` + precompile). Achieving this with (2) is self-evident. To achieve this with (1) we MUST have the ABI+bindings generated from `IPrecompile` and the results confirmed by CI[^2].

It is not sufficient to say that the precompile was generated from some ABI of unknown provenance and that it's the source of truth because then the tests exclude a key element on which our users depend.

[^1]: This can be called any way, e.g. with its own `abigen` bindings, but for tests that don't require dynamic arguments it can be no more than a constructor with `revert()`s, and then successful deployment is the test.

[^2]: Long-term we have to have a reliable Solidity build system. For now, it is sufficient to use the `arr4n/setup-solc` action, but we'll need to move to Maru's nix setup in the end.

JonathanOppenheimer: I have never seen footnotes before in a GitHub comment - the formatting is crazy! 

I will copy Arran's comment largely wholesale into the codebase as I think it is extremely well written as an explanation. Let me know if you object or want it in a different place.

plugin/evm/ExampleWarp.sol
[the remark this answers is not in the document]
michaelkaplan13: Not for this PR, but is it possible to achieve the same test effect coverage by using the `WarpTest.sol` (and corresponding ABI/bindings) in `vm_warp_test.go`? If so, we could remove the `ExampleWarp` files all together.

I _think_ this should be possible and pretty straightforward. The only difference is that `require` statements in `validateWarpMessage` below will need to made by the test itself, but I think that is a good change to make because otherwise the test is actually testing the example contract implementation rather than the VM behavior itself.

If you agree, could you create an issue for this? Definitely not a priority right now but don't want to lose track of it.

tests/warp/warp_test.go
```diff
@@ -171,8 +172,8 @@ var _ = ginkgo.Describe("[Warp]", func() {
 		log.Info("Delivering block hash payload to receiving subnet")
 		w.deliverBlockHashPayload()
 
-		log.Info("Executing HardHat test")
-		w.executeHardHatTest()
+		log.Info("bindings test: verifying warp message and blockchain ID")
+		w.warpBindingsTest()
```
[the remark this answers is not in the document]
JonathanOppenheimer: This is significantly cleaner and definitely possible. 

I create an issue for this so we don't forget about this! https://github.com/ava-labs/subnet-evm/issues/1919

michaelkaplan13 (approved):

precompile/contracts/warp/warptest/bindings/WarpTest.sol
michaelkaplan13: Either way, we want to keep the `*Test.sol` contracts currently for the reason that Arran described. Thanks for adding the comment 👍

PR #1908, opened 2025-12-05
feat: remove npm directory and npm test framework

JonathanOppenheimer:
## Why this should be merged
We have now completely migrated all node based tests to go tests. This makes all of the existing node infrastructure useless, so we can delete it all 🎉🎉🎉. 

## How this works
This just deletes a *lot* of dead code

## How this was tested
CI

## Need to be documented?
No

No


JonathanOppenheimer:
@ceyonur this will do more than just delete the directory, as I will delete the E2E tests in here as well now.

PR #1910, opened 2025-12-05
style: remaining NPM conversion cleanup items

JonathanOppenheimer:
## Why this should be merged
There were a couple clean up items I made in the past series of PRs that should be broken out into a separate PR (this one).

This PR 
- removes comments that are simply a "what" not a "why"
- aligns the formatting on all solidity files
- remove compile naming convention of all `.abi` files to `contract.go`
  - This is quite unclear, and having the source and output have the same name (default) would be much clearer. 
- Move `genesis.json`  outside of genesis folder -- after refactors, each of these folders only had a single item, which is silly. 

## How this was tested
There are no functional changes in this PR 

## Need to be documented?
No

No


alarso16 (approved):
I don't see the point in unifying style whitespace if it won't be enforced by anything. It really doesn't matter, so I'll approve

PR #1912, opened 2025-12-08
test: use generated ABI

JonathanOppenheimer:
## Why this should be merged

Currently, the committed ABI files are 2 years old, and there is no clear path of how they were generated. This PR deletes those files, and commits the files generated by `compile.go` instead.

You can verify that each folders `contract.abi` files matches the newly commited ABI file through the following:

1. pretty format the new ABI through https://jsonformatter.org/ 
2. diff check 

You can also compare them directly using json, e.g. `<(git show HEAD:precompile/allowlist/allowlist.abi | jq -S .) <(cat precompile/allowlist/IAllowList.abi | jq -S .) && echo "Files are identical!"` 

There are two exceptions which are commented below: 

## How this was tested
CI

## Need to be documented?
No

No


alarso16 (dismissed):
Since it can all be generated in CI (right?), I think we should verify that there are no differences between what is committed vs. what is generated, just like all other generated files. Can you modify the Taskfile.yml to include that?

ceyonur (commented):
I don't understand the intention of this PR. Because of:
> Currently, the committed ABI files are 2 years old,
1- yes and the reason is that these precompiles are in use and any change to those ABIs might mean a breaking change to unpacking/packing and thus a breaking change to the chain.
>and there is no clear path of how they were generated. 
2- ABI's here are not generated everytime. These were generated once from the precompile's interface and then kept as the defining interface for the precompile. As stated above, if anything changes in those ABIs, there is a risk to break backward compatibility. 

ABIs are not a part of Solidity, they're generic. Even though we used Solidity to generate those, the starting point for a stateful precompile starts with the ABI not the Solidity. So the ABI is really the defining point for the stateful precompiles (thus why we have them in the prod code).

Also see this for how those ABIs are generated in the first place https://build.avax.network/docs/avalanche-l1s/custom-precompiles/create-precompile

ARR4N:
> ABIs are not a part of Solidity, they're generic. Even though we used Solidity to generate those, the starting point for a stateful precompile starts with the ABI not the Solidity. So the ABI is really the defining point for the stateful precompiles (thus why we have them in the prod code).

While this is true, any smart contract that needs to call the precompile needs the Solidity interface, so there must be a source of truth that defines both the precompile and the way it's called.

JonathanOppenheimer:
I have merged @michaelkaplan13's PR into this, and will leave the `contract.abi` rename for the separate cleanup PR as requested.

JonathanOppenheimer:
Superseded by https://github.com/ava-labs/avalanchego/pull/4740

PR #1916, opened 2025-12-10
Reduce ABI Gen diff

michaelkaplan13:
Inlines the ABI generation so that the existing files remain, and keeps the `pretty-json` formatting.

Also removes the metadata hashes from the bytecode of the compiles test contracts, so that it prevents unnecessary diffs in the future.


michaelkaplan13:
The resulting combined diff is best viewed by having this PR be based on the `JonathanOppenheimer/delete-npm` branch.

JonathanOppenheimer:
Sounds good - I can do that in the future PR.

PR #1918, opened 2025-12-11
chore: Rename to avoid lint hits

alarso16:
## Why this should be merged

Removes files named `...test....go`, when it's not a test file. The linter is over-reaching, but this blocks repo-merger

## How this works

Only one constant in each file, with corresponding non-test files that it could use instead.

## How this was tested

Lint

## Need to be documented?

No

No


JonathanOppenheimer (changes requested):
I don't think this will work. The rule `test_import_testing_only_in_tests` ensures that testing-related packages are only imported in actual test files or designated test packages. Both the files you changed the name of still import `testing`. I also think the file names are correct, because they are test files, and there's no need to rename them.

